PR #68, opened 2019-05-02
Update message class template to use the new inheritance pattern

### LINKED ISSUE #63: Update message class templates to use inheritance

AlCalzone:
Fixes: #63


AlCalzone:
Broken/failing test is already removed on master

PR #379, opened 2019-10-31
Move the definition of multilevel sensor types and scales to configuration files

AlCalzone:
in preparation for #318


AlCalzone:
Just needs a manual test, then it's ready!

PR #392, opened 2019-11-05
Hide BasicCC if any other actuator CCs are supported

### LINKED ISSUE #388: Hide BasicCC if another Actuator CC is supported by an endpoint
From SDS14223:
A controlling node MUST NOT use the Basic Command Class for controlling nor showing status
(receiving report) of a node (or endpoint) if the controlling node controls at least one actuator
command class supported by a node (or endpoint). The actuator command classes are defined in the
Application Command Class specification ([2]).

AlCalzone:
Edit: There's no easy way to test this. Should do it manually.

Fixes: #388


AlCalzone:
Tis works

PR #422, opened 2019-11-18
Remove esModuleInterop flag

AlCalzone:


AlCalzone:
no can do... TS node now requires env variables

PR #537, opened 2020-01-04
feat: allow configDir to be passed as an option

ghost:
It's frustrating while developing to keep losing the cache file.  IMO, it should be possible to check the cache file into version control to save having to re-interview.

### DIFF
--- a/src/lib/driver/Driver.ts
+++ b/src/lib/driver/Driver.ts
@@ -79,6 +79,7 @@ export interface ZWaveOptions {
 	 * Set this to true to skip the controller interview. Useful for testing purposes
 	 */
 	skipInterview?: boolean;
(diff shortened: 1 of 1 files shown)


AlCalzone:
LOL, I was just working at the same thing (a bit more complex though)

AlCalzone:
This will be included in PR #538

CliffS:
> LOL, I was just working at the same thing (a bit more complex though)

Thanks.  I don't do *complex* in typescript.  It already took me five goes to get it to compile and three goes to get past your lint. :smile:

AlCalzone:
>  three goes to get past your lint. 😄

:D I have the ESLint and Prettier plugins in VSCode. So just saving the file does most of the work.

PR #539, opened 2020-01-05
Implement removeFailedNode and isFailedNode

### LINKED ISSUE #497: Implement removeFailedNode
Is there any way I can remove nodes 8 & 9 from the controller without the actual devices?

_Originally posted by @CliffS in https://github.com/AlCalzone/node-zwave-js/issues/476#issuecomment-565083650_

AlCalzone:
This PR adds two methods to work with failed nodes:

`isFailedNode(nodeId)` can be used to test if a node is marked as failed.

`removeFailedNode(nodeId)` may be used to remove a failed node. If the process fails, this will throw an exception with the details why.

fixes: #497 
/cc @CliffS here's a goodie for you. Not sure yet which release it will be in.


CliffS:
@AlCalzone Thanks.  I'll test that right now!

CliffS:
Works perfectly.  One suggestion though, now you have `isFailedNode(nodeId)`, you could use that inside the test for `all nodes ready` and ignore failed nodes.

This would mean that the consuming code could rely on `all nodes ready` even if there are failed nodes.  As it stands, `all nodes ready` is not issued, even if `removeFailedNode(nodeId)` has been run as the number is still in the list of nodes.

AlCalzone:
Do you mean that the event is not emitted after the node has been removed? Is it emitted for you after a restart with no more failed nodes?

CliffS:
Correct.  It is not emitted after a node has been removed.  Yes, it works fine after a restart.

AlCalzone:
Ok, I'm tracking that in #546

PR #602, opened 2020-01-24
Upgrade to TypeScript 3.8 and use new features

AlCalzone:
Blocked by:
- [x] https://github.com/babel/babel/issues/10981
- [x] https://github.com/typescript-eslint/typescript-eslint/issues/1436
- [x] https://github.com/Semmle/ql/issues/2691

**TODOs:**
- [x] Upgrade Prettier to an official version (v2.0 / v1.20 ?)
- [x] Upgrade babel to an official version (v7.9)
- [ ] Check if `@babel/core@7.9.x` breaks Node 8 support


lukescott:
@AlCalzone have you considered using `tsc` instead of babel?

At work I've found that using babel wasn't helpful when locked into TypeScript features.  The only place I still use babel is compiling for the web as the env preset is super helpful for targeting browser versions.

The `ts-jest` package lets you run tests using `tsc` instead of going thru babel.

AlCalzone:
I've been using ts-jest earlier but always ran into problems where tests would use some cache I couldn't get rid of. Not very helpful if the test runner is unreliable or shows false-positives...

Not sure if that has changed but I haven't had any similar problems since I made the switch to babel+jest.

PR #681, opened 2020-02-21
Respect mandatory CCs defined for the device class

### LINKED ISSUE #674: Check if mandatory CCs may be omitted from the NIF
Example:
```
20:45:59.852 CNTRLR « [Node 035] received response for protocol info:
                      basic device class:    Static Controller (0x04)
                      generic device class:  Thermostat (0x08)
                      specific device class: Setpoint Thermostat (0x04)
```
and NIF:
```
20:47:07.460 CNTRLR « [Node 035] node info received
                      supported CCs:
                      · Battery
                      · Climate Control Schedule
                      · Clock
                      · Manufacturer Specific
                      · Multi Command
                      · Protection
                      · Thermostat Setpoint
                      · Version
                      · Wake Up
                      controlled CCs:
                      · Climate Control Schedule
                      · Clock
                      · Multi Command
```
According to reports, this device is supposed to support Multilevel Sensor CC.

AlCalzone:
Not sure if this is the way to go... OZW has to remove supported CCs in many config files. Probably better to just rely on the information the devices report us.

We could however use this when a device won't respond to a NIF request.

fixes: #674


AlCalzone:
Some older devices leave out mandatory CCs from the NIF, so we actually need to do this.

Coincidentally, this PR also includes some changes to the build system since it was necessary to fix it for this PR to work.

PR #748, opened 2020-03-28
Color Switch CC

RoboPhred:
Starting to puzzle out how to make CCs.

Current blockers:
[ ]: How do we translate command names to their byte?  Eg: SWITCH_COLOR_SUPPORTED_REPORT
[ ]: How can I test my code?

I tried to `npm link` it against my existing project, but that returns an oddly-typescripted runtime error
```
C:\dev\node-zwave-js\src\lib\commandclass\MultiChannelCC.ts:262
        public constructor(driver: Driver, options: CommandClassOptions) {
                                       ^
ReferenceError: Driver is not defined
    at Object.<anonymous> (C:\dev\node-zwave-js\src\lib\commandclass\MultiChannelCC.ts:262:29)
```

I have disabled source-map-support in my app yet I am still getting a runtime error pointed at typescript here, so I can't tell where in the underlying javascript the error is occurring on.


RoboPhred:
I resolved the `ReferenceError: Driver is not defined` in #751, and fixed the command classes not loading in #750

I made some progress.  By assuming the commands were listed in order of their command ordinal, I was able to successfully implement and test
- SWITCH_COLOR_SUPPORTED_GET (0x01?)
- SWITCH_COLOR_SUPPORTED_REPORT (0x02?)
- SWITCH_COLOR_GET (0x03?)
- SWITCH_COLOR_REPORT (0x04?)

However, SWITCH_COLOR_SET (0x05?) is giving me trouble.  As far as I can tell, the packet structure is correct, but I cannot get my light to change and I never see the new value in place if I issue a GET command afterwards.

On top of this, the spec docs do not indicate any REPORT for a SET.  Despite this, something is coming back down the wire when I send my SET request:
```
>zwave.controller.nodes.get(18).commandClasses["Color Switch"].set({red: 255})
19:33:23.234 DRIVER » [Node 018] [REQ] [SendData]                                        [P: Normal]
                      │ transmitOptions: 0x25
                      │ callbackId:      25
                      └─[ColorSwitchCCSet]
                          payload: 0x0102ff
19:33:23.238 SERIAL » 0x010c0013120533030102ff251907                                      (14 bytes)
19:33:23.241 SERIAL « [ACK]                                                                   (0x06)
19:33:23.247 DRIVER   ACK received from controller for current transaction
19:33:23.250 SERIAL   The receive buffer is empty, waiting for the next chunk...
19:33:23.254 SERIAL « 0x0104011301e8                                                       (6 bytes)
19:33:23.257 SERIAL   Buffer := 0x0104011301e8                                             (6 bytes)
19:33:23.260 SERIAL » [ACK]                                                                   (0x06)
19:33:23.266 DRIVER « [RES] [SendData]                                                [confirmation]
                        wasSent: true
19:33:23.271 SERIAL   The receive buffer is empty, waiting for the next chunk...
19:33:23.277 SERIAL « 0x010700131900                                                       (6 bytes)
19:33:23.283 SERIAL   [incomplete] Buffer := 0x010700131900                                (6 bytes)
19:33:23.287 SERIAL « 0x0002f0                                                             (3 bytes)
19:33:23.291 SERIAL   Buffer := 0x0107001319000002f0                                       (9 bytes)
19:33:23.299 SERIAL » [ACK]                                                                   (0x06)
19:33:23.303 DRIVER « [REQ] [SendData]                                                       [final]
                        callbackId:     25
                        transmitStatus: OK
19:33:23.305 DRIVER   ACK already received, resolving transaction
19:33:23.307 DRIVER   resuming send queue
19:33:23.309 SERIAL   The receive buffer is empty, waiting for the next chunk...
19:33:23.319 DRIVER   The send queue is empty
19:33:23.322 SERIAL « 0x010a000400120433040100d1                                          (12 bytes)
19:33:23.324 SERIAL   Buffer := 0x010a000400120433040100d1                                (12 bytes)
19:33:23.327 SERIAL » [ACK]                                                                   (0x06)
19:33:23.341 SERIAL   The receive buffer is empty, waiting for the next chunk...
```

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+		this._supportsWarmWhite = Boolean(this.payload[0] & 1);
+		this._supportsColdWhite = Boolean(this.payload[0] & 2);
+		this._supportsRed = Boolean(this.payload[0] & 4);
+		this._supportsGreen = Boolean(this.payload[0] & 8);
+		this._supportsBlue = Boolean(this.payload[0] & 16);
+		this._supportsAmber = Boolean(this.payload[0] & 32);
+		this._supportsCyan = Boolean(this.payload[0] & 64);
+		this._supportsPurple = Boolean(this.payload[0] & 128);
+		this._supportsIndex = Boolean(this.payload[2] & 1);
+
```
AlCalzone: You're on the right track, bit masks are Big-Endian.
Variable length usually means that future specs might add additional bytes to the end. The check for at least 2 bytes should suffice, since the command will always contain at least 2.

I have a helper method for parsing and serializing variable-length bit masks. Assuming there won't be any different fields added after it and there won't be any gaps in the bit mask, you can do this:

```ts
const bitMask = parseBitMask(this.payload).map(Boolean);
[
  this._supportsWarmWhite,
  this._supportsColdWhite,
  this._supportsRed,
  this._supportsGreen,
  this._supportsBlue,
  this._supportsAmber,
  this._supportsCyan,
  this._supportsPurple,
  this._supportsIndex
] = bitMask;
```

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+	@ccValue()
+	@ccValueMetadata({
+		...ValueMetadata.ReadOnlyBoolean,
+		label: "Supports Indexed Color",
+	})
+	public get supportsIndex(): boolean {
+		return this._supportsIndex;
+	}
+
+	public toJSON(): JSONObject {
```
AlCalzone, line 500: I've started to deprecate the toJSON method in favor of the `toLogEntry` method. You can find an example in MultiChannelCC, lines 840ff

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+	) {
+		super(driver, options);
+
+		// Docs say 'variable length', but the table shows 2 bytes.
+		validatePayload(this.payload.length >= 2);
+		this._colorComponent = this.payload[0];
+		this._value = this.payload[1];
+
+		// TODO: We probably want to store each color value
+		//	as a different ccValue.  How do we do that when
```
AlCalzone: You'll need to store the values manually in the value DB. `BinarySensorCCReport` has an example. I'll leave it up to you how to translate from the numeric color component to the property name.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+		this._supportsColdWhite = Boolean(this.payload[0] & 2);
+		this._supportsRed = Boolean(this.payload[0] & 4);
+		this._supportsGreen = Boolean(this.payload[0] & 8);
+		this._supportsBlue = Boolean(this.payload[0] & 16);
+		this._supportsAmber = Boolean(this.payload[0] & 32);
+		this._supportsCyan = Boolean(this.payload[0] & 64);
+		this._supportsPurple = Boolean(this.payload[0] & 128);
+		this._supportsIndex = Boolean(this.payload[2] & 1);
+
+		//this.persistValues();
```
AlCalzone: You'll need this line or the values won't be stored in the value DB

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+			this._colorComponent = options.colorComponent;
+		}
+	}
+
+	private _colorComponent: ColorComponent;
+	public get colorComponent(): ColorComponent {
+		return this._colorComponent;
+	}
+	public set colorComponent(value: ColorComponent) {
+		if (!ColorComponent[value]) {
```
AlCalzone, line 618: This will throw if value is 0. The check should probably be `!= undefined`.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+			throw new Error(
+				"colorComponent must be a valid color component index.",
+			);
+		}
+		this._colorComponent = value;
+	}
+
+	public serialize(): Buffer {
+		this.payload = Buffer.allocUnsafe(1);
+		this.payload[0] = this._colorComponent;
```
AlCalzone, line 628: For the sake of consistency, I would use `this.payload = Buffer.from([this._colorComponent])`.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+		return super.serialize();
+	}
+}
+
+export interface ColorSwitchCCSetOptions extends CCCommandOptions, ColorTable {
```
AlCalzone, line 633: Wait... you can do that?!

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+				warmWhite: options.warmWhite,
+				coldWhite: options.coldWhite,
+				red: options.red,
+				green: options.green,
+				blue: options.blue,
+				cyan: options.cyan,
+				purple: options.purple,
+				index: options.index,
+			};
+			this.duration = options.duration || 0;
```
AlCalzone: Since 3.8, I prefer `??` over `||`.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+	}
+	public set duration(value: number) {
+		if (value < 0 || value > 255) {
+			throw new Error("duration out of range.");
+		}
+		this._duration = value;
+	}
+
+	public serialize(): Buffer {
+		let colorComponentKeys = Object.keys(
```
AlCalzone: Ok, now I see why you're doing the stuff at the start. There's probably a way to do a bit simpler, but that seems fine for now.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+		switch (cmd) {
+			case ColorSwitchCommand.SupportedGet:
+			case ColorSwitchCommand.Get:
+			case ColorSwitchCommand.Set:
+				return true; // This is mandatory
+		}
+		return super.supportsCommand(cmd);
+	}
+
+	public async supportedGet() {
```
AlCalzone: I like the API to have somewhat natural names ==> `getSupported`

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+	public serialize(): Buffer {
+		let colorComponentKeys = Object.keys(
+			this._colorTable,
+		) as (keyof ColorTable)[];
+		colorComponentKeys = colorComponentKeys.filter(
+			(x) => !isNaN(this._colorTable[x] as any),
+		);
+		const colorComponentCount = colorComponentKeys.length;
+		this.payload = Buffer.allocUnsafe(1 + colorComponentCount * 2);
+		this.payload[0] = colorComponentCount & 31;
```
AlCalzone: I prefer the bitwise notation `0b11111`.

AlCalzone:
I've added a few remarks in your code. This should solve most of your questions. As I wrote somewhere else, the byte representations can be found in an Excel document in the specs.

Regarding tests: `npm link` is ok to test against a real device. All CCs should have a corresponding test file (`XyzCC.test.ts`). You can fill them with the snippet `zwcctest`.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+		return this._value;
+	}
+}
+
+interface ColorSwitchCCGetOptions extends CCCommandOptions {
+	colorComponent: ColorComponent;
+}
+
+@CCCommand(ColorSwitchCommand.Get)
+@expectedCCResponse(ColorSwitchCCReport)
```
AlCalzone: Ok, I forgot the one comment regarding the response check... You can basically copy the check from `BinarySensorCCGet` and do a few tweaks. You want to make sure that only `ColorSwitchCCReport` with a matching `colorComponent` are accepted as a response to the `Get`.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+};
+
+@API(CommandClasses["Color Switch"])
+export class ColorSwitchCCAPI extends CCAPI {
+	public supportsCommand(cmd: ColorSwitchCommand): Maybe<boolean> {
+		switch (cmd) {
+			case ColorSwitchCommand.SupportedGet:
+			case ColorSwitchCommand.Get:
+			case ColorSwitchCommand.Set:
+				return true; // This is mandatory
```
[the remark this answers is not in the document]
RoboPhred: SupportedGet / Get / Set are in v1, start/stop are added in v2

src/lib/commandclass/ColorSwitchCC.ts
RoboPhred: Thanks, I got that in but I do not know the significance of isPositiveTransmitReport.  What does this represent?  Is the same check needed here?

```js
const testResponseForColorSwitchGet: CCResponsePredicate = (
	sent: ColorSwitchCCGet,
	received,
	isPositiveTransmitReport,
) => {
	return received instanceof ColorSwitchCCReport &&
		sent.colorComponent === received.colorComponent
		? "final"
		: // TODO: What is isPositiveTransmitReport?  Do I need it?
		isPositiveTransmitReport
		? "confirmation"
		: "unexpected";
};
```

src/lib/commandclass/ColorSwitchCC.ts
RoboPhred: Interfaces can take any number of extensions.  Its equivalent to
`type ColorSwitchCCSetOptions = CCComandOptions & ColorTable & { duration: number; }`

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+
+		const cc = new ColorSwitchCCGet(this.driver, {
+			nodeId: this.endpoint.nodeId,
+			endpoint: this.endpoint.index,
+			colorComponent: colorComponent,
+		});
+		const response = (await this.driver.sendCommand<ColorSwitchCCReport>(
+			cc,
+		))!;
+		return {
```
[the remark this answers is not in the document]
RoboPhred: Later versions of Color Switch also return targets and durations, so we should at least keep this in object form.  It will make more sense once I get to color v3

RoboPhred:
Thanks for the feedback.  I think the lack of color response was 2 things:
- The wrong CCCommand, as you pointed out
- The AeoTec LED Light I am using was on an older firmware and I think it was locked into a color index.  I flashed new firmware to it and now it responds correctly.

I still have some concerns:
- I cannot get color index to work.  I am sure my device supported this before as I used it with ozw, but either I am sending it wrong, or the firmware update I just did removed it as an option.  I do have another bulb in storage that should still be on the old firmware I could try.
- My bulb was made before the switch from Color to ColorSwitch, so it does not have the v2 start/stop color change commands.  I am unable to test these at the moment.

How do I get the CC version?  I tried `node.commandClasses["Color Switch"].version` but that returns `0`.  I expected at least `1`.  Or is this an artifact of my bulb having been made before the rename from Color to Color Switch?

AlCalzone:
>     * so it does not have the v2 start/stop color change commands.  I am unable to test these at the moment.

Thats okay but the serialization of those commands should at least be in the CC test file.

> How do I get the CC version? I tried `node.commandClasses["Color Switch"].version` but that returns `0`. I expected at least `1`.

I would need to look at the logfile to know why. Probably you need to remove the cache file to restart the version interview. The interview probably skipped the version check since 1 was the max implemented version and now you're stuck on now knowing the actual version.

RoboPhred:
I implemented v3, got durations worked out, filled out the tests, and added the interview process.

I still need to figure out how to record arbitrary cc values for the get command to complete this.  Can you point me in the right direction for that?

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,446 @@
...
+} from "./CommandClass";
+import { CommandClasses } from "./CommandClasses";
+
+// All the supported commands
+export enum ColorSwitchCommand {
+	SupportedGet = 0x01,
+	SupportedReport = 0x02,
+	Get = 0x03,
+	Report = 0x04,
+	Set = 0x05,
```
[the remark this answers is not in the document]
RoboPhred: v3 has now been implemented, the missing commands have been added.

src/lib/commandclass/ColorSwitchCC.ts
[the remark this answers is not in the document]
RoboPhred: I was wrong, start/stop are in v1 as well.  They have been added.

src/lib/commandclass/ColorSwitchCC.ts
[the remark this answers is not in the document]
RoboPhred: v3 is now implemented, so get now returns targetValue and duration in addition to currentValue and colorComponent

RoboPhred:
I played around and was able to get ccValues being set.  I chose "targetColor", "currentColor" and "supportedColor" for names, and use the color names for propertyKey.  I also implemented setValue in the API.

I am concerned about setting the metadata fresh each time.  Is there a place I should be writing the value metadata at, or is it ok to reset it every time I query data?

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+				description: `Whether the endpoint supports setting the ${key} color.`,
+			});
+			node.valueDB.setValue(
+				valueId,
+				this._supportedColors[supportsColorKey] ?? false,
+			);
+		}
+	}
+
+	private _supportedColors: SupportedColorTable | null;
```
AlCalzone: This is definitely assigned in the constructor, so null should not be necessary.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+			}
+
+			await api.get(key);
+		}
+	}
+}
+
+@CCCommand(ColorSwitchCommand.SupportedReport)
+export class ColorSwitchCCSupportedReport extends ColorSwitchCC
+	implements Partial<SupportedColorTable> {
```
AlCalzone: Is this really Partial? In the API getSupported method you assert all properties to be non-null.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+			const supportsColorKey = supportedColorTableKey(key);
+
+			const valueId: ValueID = {
+				commandClass: CommandClasses["Color Switch"],
+				endpoint: endpoint.index,
+				property: "supportedColor",
+				propertyKey: key,
+			};
+
+			// TODO: Should we be setting this every run?
```
AlCalzone: You could check with `hasMetadata` first.

src/lib/commandclass/ColorSwitchCC.ts
AlCalzone: And you can remove the `| undefined` and `?.` from the properties below.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+@CCCommand(ColorSwitchCommand.Report)
+export class ColorSwitchCCReport extends ColorSwitchCC {
+	public constructor(
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+
+		if (this.version >= 3) {
+			validatePayload(this.payload.length >= 4);
```
AlCalzone: Out of consistency, I would move this check into the if branch below (where you read the payload).
Although this would mean a little additional work, the check is logically closer to where its needed.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+		this._colorComponent = this.payload[0];
+		this._currentValue = this.payload[1];
+
+		if (this.version >= 3) {
+			this._targetValue = this.payload[2];
+			this._duration = Duration.parseReport(this.payload[3]);
+		}
+
+		const node = this.getNode()!;
+		const endpoint = this.getEndpoint()!;
```
AlCalzone: You don't need to access getNode and getEndpoint here. The CommandClass class has a method `getValueDB` and the endpoint index is accessible at `this.endpointIndex`, see also the `IndicatorCCReport` constructor.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+		}
+
+		const valueId: ValueID = {
+			commandClass: CommandClasses["Color Switch"],
+			property: "currentColor",
+			endpoint: endpoint.index,
+			propertyKey: colorName,
+		};
+
+		// TODO: Should we be setting this every run?
```
AlCalzone: You should probably create all the metadata in the interview procedure. See Notification CC line 284ff (and I'm sure I've done this somewhere else too).

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+	) {
+		super(driver, options);
+
+		if (this.version >= 3) {
+			validatePayload(this.payload.length >= 4);
+		} else {
+			validatePayload(this.payload.length >= 2);
+		}
+		this._colorComponent = this.payload[0];
+		this._currentValue = this.payload[1];
```
AlCalzone, line 528: Can this have a "unknown" value? In that case you should use `parseMaybeNumber` like in `BasicCCReport`. In that case, `currentValue` needs the type `Maybe<number>`

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+
+		if (this.version >= 3) {
+			this._targetValue = this.payload[2];
```
AlCalzone, line 532: And please use `parseNumber` for these kinds of values.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+		node.valueDB.setValue(valueId, this._currentValue);
+
+		if (this._targetValue != undefined) {
+			const targetValueId: ValueID = {
+				commandClass: CommandClasses["Color Switch"],
+				property: "targetColor",
+				endpoint: endpoint.index,
+				propertyKey: colorName,
+			};
+			node.valueDB.setMetadata(targetValueId, {
```
AlCalzone: see other comment regarding metadata

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+
+const testResponseForColorSwitchGet: CCResponsePredicate = (
+	sent: ColorSwitchCCGet,
+	received,
+	isPositiveTransmitReport,
+) => {
+	return received instanceof ColorSwitchCCReport &&
+		sent.colorComponent === received.colorComponent
+		? "final"
+		: // TODO: What is isPositiveTransmitReport?  Do I need it?
```
AlCalzone: Yep, you need it - it tells us that the message was received. Just ignore the part after `? "final"` - I'm not happy with this check design, but it is working ;)

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+		if (this.version >= 2) {
+			this.payload[i] = this._duration.serializeSet();
+		}
+		return super.serialize();
+	}
+}
+
+export interface ColorSwitchCCStartLevelChangeOptions extends CCCommandOptions {
+	colorComponent: ColorComponent;
+	down?: boolean;
```
AlCalzone, line 708: Please take a look at `MultilevelSwitchCCStartLevelChange` and implement this in a similar fashion.

AlCalzone: LevelChangeDirection could probably be moved to a shared file then (probably somewhere in `./values`)

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+
+		const cc = new ColorSwitchCCSupportedGet(this.driver, {
+			nodeId: this.endpoint.nodeId,
+			endpoint: this.endpoint.index,
+		});
+		const response = (await this.driver.sendCommand<
+			ColorSwitchCCSupportedReport
+		>(cc))!;
+		return {
+			supportsWarmWhite: response.supportsWarmWhite!,
```
AlCalzone: See comment below regarding the types in `ColorSwitchCCSupportedReport`

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+	) => {
+		if (property !== "targetColor") {
+			throwUnsupportedProperty(this.ccId, property);
+		}
+		if (typeof value !== "number") {
+			throwWrongValueType(this.ccId, property, "number", typeof value);
+		}
+
+		if (!propertyKey) {
+			// No error throw util for this?
```
AlCalzone, line 302: Yep, I've stumbled across this aswell. Should probably be merged with `throwUnsupportedProperty`

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+
+		log.controller.logNode(node.id, {
+			endpoint: this.endpointIndex,
+			message: `${this.constructor.name}: doing a ${
+				complete ? "complete" : "partial"
+			} interview...`,
+			direction: "none",
+		});
+
+		// TODO: Only do this if complete?  Where would we get
```
AlCalzone: Read them from the value DB (where you stored them). See `BinarySensorCC`, line 142ff.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+		super(driver, options);
+		if (gotDeserializationOptions(options)) {
+			// TODO: Deserialize payload
+			throw new ZWaveError(
+				`${this.constructor.name}: deserialization not implemented`,
+				ZWaveErrorCodes.Deserialization_NotImplemented,
+			);
+		} else {
+			// Populate properties from options object
+			this._colorTable = {
```
AlCalzone: I'm doing this so often, I should probably add a `pick` utility method to pick properties from an object.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -0,0 +1,762 @@
...
+	}
+
+	public serialize(): Buffer {
+		this.payload = Buffer.allocUnsafe(1);
+		this.payload[0] = this._colorComponent;
+		return super.serialize();
+	}
+}
+
+// TODO: Move to util library
```
AlCalzone: ==> util/misc.ts

I like it!

AlCalzone:
Hope I didn't miss anything new.

src/lib/commandclass/ColorSwitchCC.ts
RoboPhred: Moved into util/misc

src/lib/commandclass/ColorSwitchCC.ts
RoboPhred: I added a pick() function in util/misc and made use of it here.

src/lib/commandclass/ColorSwitchCC.ts
RoboPhred: For some reason I thought the data was parsed after the constructor call, so I tried to set it up to allow the data to be in an unparsed state.  I will remove the optionality.

RoboPhred:
Do you have an opinion on whether I should be storing supportedColor / targetColor / currentColor by propertyKey, or by storing a ColorTable object as a property without a key?  I am not clear on what use cases propertyKey is made to solve, and I noticed that supportedSensorTypes in BinarySensorCC stores an array, so there is some precedent for complex values stored in a property.

AlCalzone:
For values that are only used internally, I generally use the format that makes the code the easiest (array, map, whatever complicated format [). Every value that will be used outside the library should be stored as a single simple value (number, boolean, string).
To sub-address related properties, I added the property key. Storing all color components as the `currentValue`/`targetValue` property with different property keys seems to make sense. I guess they can all be controlled separately?

If the code is simpler with a numeric propertyKey, you can also do that. In that case, you need to implement the `translatePropertyKey` method on the CC to give it a speaking name. See the IndicatorCC (line 367) for an example.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -93,6 +106,7 @@ const ColorComponentMap: Record<ColorKey, ColorComponent> = {
 	index: ColorComponent.Index,
 };
 const ColorKeys = keysOf(ColorComponentMap);
+const SupportedColorKeys = ColorKeys.map(supportedColorTableKey);
 function isColorKey(key: string): key is ColorKey {
 	// Note: indexing ColorComponentMap would be faster, but
 	//	the linter will not allow it.
```
AlCalzone: Do you have an example what you mean? I'm sure we can make it possible :)

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -293,25 +309,40 @@ export class ColorSwitchCC extends CommandClass {
...
+			const valueDB = this.getValueDB();
+			for (const key of SupportedColorKeys) {
+				const valueId: ValueID = getSupportedColorValueID(
+					this.endpointIndex,
+					key,
+				);
+				const isSupported = valueDB.getValue<boolean>(valueId);
+				supportedColors[key] = isSupported ?? false;
+			}
+		}
```
AlCalzone: Yeah, store it as an object or Map or whatever makes the code the easiest. The Value ID should be marked as internal:
Add a constructor to `ColorSwitchCC`:
```ts
	public constructor(driver: Driver, options: CommandClassOptions) {
		super(driver, options);
		this.registerValue(getSupportedColorValueID(this.endpointIndex).property, true);
	}
```

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -344,68 +375,65 @@ export class ColorSwitchCCSupportedReport extends ColorSwitchCC
...
+					...ValueMetadata.ReadOnly,
+					label: `Supports ${key}`,
+					description: `Whether the endpoint supports setting the ${key} color.`,
+				});
+			}
+			valueDB.setValue(
 				valueId,
 				this._supportedColors[supportsColorKey] ?? false,
 			);
 		}
```
AlCalzone: I guess this makes sense here - I've usually done this kind of stuff in the interview method instead to keep the CC constructors shorter.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -457,35 +480,39 @@ export class ColorSwitchCCReport extends ColorSwitchCC {
...
+		valueDB.setValue(valueId, this._currentValue);
 
 		if (this._targetValue != undefined) {
 			const targetValueId: ValueID = {
 				commandClass: CommandClasses["Color Switch"],
 				property: "targetColor",
-				endpoint: endpoint.index,
+				endpoint: this.endpointIndex,
 				propertyKey: colorName,
 			};
```
AlCalzone: This should probably be in a getXYZValueID method. I'm planning to rework the whole ValueID definition process and having them all at the top makes it easier.

src/lib/commandclass/ColorSwitchCC.test.ts
```diff
@@ -0,0 +1,265 @@
...
+				0b0000_0010, // color: red
+				0b1111_1111, // value: 255
+			]),
+		);
+		const cc = new ColorSwitchCCReport(fakeDriver, {
+      nodeId: 1,
+			data: ccData,
+		});
+		cc.version = 1; // Will not work, as deserialization is performed in the ctor
+
```
AlCalzone, line 97: I forgot to comment here. Deserialization should usually be independent of the version, since you cannot be sure that the version is known before receiving the command. What exactly is the problem here?

AlCalzone (changes requested):
There are some comments from before you haven't addressed yet :)

src/lib/commandclass/ColorSwitchCC.ts
RoboPhred: Previously, the code was something like
```js
function isColorKey(key: string): key is ColorKey {
	return ColorComponentMap[key as ColorKey] != undefined;
}
```
Weirdly, running the linter locally never complained, but the github lint check would.  I no longer remember what error it was giving me though.

src/lib/commandclass/ColorSwitchCC.test.ts
RoboPhred: I was trying to test that the deserialization for older versions works correctly.  I couldn't find a way to pre-set the version so I disabled the entire test.  I am actually confused now as to how the v3 test is working.  Something must be prepopulating the version property to something >= 3 for that test to be passing.

src/lib/commandclass/ColorSwitchCC.ts
RoboPhred: I think now I want to keep them as-is.  It is useful to expose what colors are supported, so the user knows not to try setting colors that will have no effect.

src/lib/commandclass/ColorSwitchCC.ts
RoboPhred: I moved metadata initialization to interview.

It is being done only if the metadata does not already exist, and for a complete interview.  Should it always be done regardless of complete / exists, in case we change the metadata in the future?

src/lib/commandclass/ColorSwitchCC.ts
RoboPhred: I replaced all valueID creation logic with functions.

src/lib/commandclass/ColorSwitchCC.test.ts
AlCalzone: > Something must be prepopulating the version property to something >= 3 for that test to be passing.

Yeah. The mock driver used for tests is setting the version to 100. To test the deserialization for e.g. v1, you should just give the method a buffer that is valid in v1 (usually has less fields).

RoboPhred: I am checking the version field to determine what I should parse though.  In order to make the test work like that, I would have to ignore the version and ,make the cc parser rely only on the length of the payload.  Is this really what we want to do in this case?

AlCalzone: > I would have to ignore the version and ,make the cc parser rely only on the length of the payload. Is this really what we want to do in this case?

I'll need to take a look at the specifics, but this is actually what the specs say about this.

AlCalzone:
Hey @RoboPhred what's the status here? Are you still working on it or should I merge?

RoboPhred:
I think everything is completed here.  Aside from not knowing how to test earlier versions I think it is ready to be merged.

AlCalzone:
Merged as part of #766 
Thanks for your work!

PR #750, opened 2020-03-29
Fix loading command classes.

RoboPhred:
The command class loader looks under the root library directory, which contains no command classes.  It should be looking under `./lib/commandclasses`


AlCalzone:
Oh, yeah I missed that when I refactored things. Good catch! Wonder why I didn't get any reports that stuff broke.

PR #766, opened 2020-04-14
Support ColorSwitchCC

AlCalzone:


AlCalzone:
@RoboPhred I did some changes to your proposal to bring the CC more in line with other CCs:

- [x] Replaced the usage of colorKeys with the actual color component IDs
- [x] Translate value IDs
- [x] Store supported color components as an array in a single value ID
- [x] Reduced the colorTable stuff to a minimum - I do like it for the set command, but I added the numeric enum values and their names as an alternative.

Notice that this might affect your code where you use the CC.

After I've looked at the tests, I'll merge.

src/lib/commandclass/ColorSwitchCC.ts
```diff
@@ -533,53 +380,40 @@ export class ColorSwitchCCReport extends ColorSwitchCC {
...
-		if (this.version >= 3) {
-			validatePayload(this.payload.length >= 4);
-			this._targetValue = this.payload[2];
-			this._duration = Duration.parseReport(this.payload[3]);
-		}
-
-		const colorKey = getColorKeyFromComponent(this._colorComponent);
-		if (!colorKey) {
-			return;
+		if (this.version >= 3 && this.payload.length >= 4) {
```
AlCalzone, line 400: @RoboPhred This is the reason your v1 deserialization test failed. You want to make sure that the CC is at least valid in V1 (validatePayload), but every version above that needs to treat the additional values as optional.

PR #793, opened 2020-05-09
Add compat option to device config to do some node queries on wakeup

### LINKED ISSUE #792: Send nodes quickly back to sleep if we don't have any messages for them
From a log:
```
10:40:48.082 CNTRLR « [Node 033] received wakeup notification
10:40:48.169 CNTRLR   [Node 006] [Meter] value[66049]: metadata updated                 [Endpoint 0]
10:40:48.171 CNTRLR   [Node 006] [~] [Meter] value[66049]: 148 => 71.8                  [Endpoint 0]
10:40:48.174 CNTRLR   [Node 006] [Meter] deltaTime[66049]: metadata updated             [Endpoint 0]
10:40:48.176 CNTRLR   [Node 006] [~] [Meter] deltaTime[66049]: 0 => 0                   [Endpoint 0]
10:40:48.212 CNTRLR   [Node 033] [~] [Battery] level: 0 => 0                            [Endpoint 0]
10:40:48.215 CNTRLR   [Node 033] [~] [Battery] isLow: true => true                      [Endpoint 0]
10:40:48.333 CNTRLR « [Node 033] received wakeup notification
[...]
10:40:57.911 DRIVER   The awake timeout for node 33 has elapsed. Assuming it is asleep.
10:40:57.912 CNTRLR   [Node 033] The node is now asleep.
```

We know that we have no messages, so send the node back to sleep as soon as possible. Maybe after 1s to leave some room for interaction

AlCalzone:
This is required by some devices (like Danfoss LC-13) for normal operation, although they report the relevant data themselves.

This also sends nodes without pending messages back to sleep quickly.
Fixes: #792

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 -->
 
 ## __WORK IN PROGRESS__
+### Features
+* Added compatibility option `queryOnWakeup` to configure which API methods must be called when a device wakes up. Some devices (like the Danfoss thermostats) expect to be queried after wakeup, even if they send the required information themselves.
+
 ### Bugfixes
 * CCs that can split their information into multiple messages now correctly store that information when only a single message is received
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 	## __WORK IN PROGRESS__
 -->
 
+## __WORK IN PROGRESS__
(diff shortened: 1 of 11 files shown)


AlCalzone:
~~This doesn't seem to be necessary anymore.~~
Maybe it still is...

PR #976, opened 2020-09-15
Simplify SendQueue machine logic, handle priority commands during a transaction

### LINKED ISSUE #971: The send thread does not respond to GetNonce requests in some circumstances
If we sent a GetNonce request, but the Node itself requested a nonce, we should answer that as soon as we can (i.e. after the confirmation callback). Otherwise we end up in a deadlock.

AlCalzone:
This PR simplifies the SendQueue machine logic by splitting the part of sending commands, which is currently duplicated between several states, into a separate machine. This `CommandQueueMachine` does no complex logic or validation and simply tries to transmit all the transactions it gets sent. Errors are passed through to the parent machine.

**TODOs:**
- [x] Implement CommandQueueMachine
- [x] Migrate to model based testing
- [x] Refactor SendThreadMachine to use CommandQueueMachine
- [x] Fix SendThreadMachine to always send priority commands (i.e. Handshake responses) by adding them to the CommandQueueMachine queue

fixes: #971


packages/zwave-js/src/lib/driver/SendThreadMachine.test.ts
```diff
@@ -578,6 +578,8 @@ describe("lib/driver/SendThreadMachine", () => {
 					message: testTransactions.NonceResponse.message,
 				});
 				preTransmitHandshakePromise.resolve();
+				// little hack because the event queue seems to work differently on Node.js 10
+				return new Promise((resolve) => setImmediate(resolve));
```
AlCalzone, line 582: grafik

PR #995, opened 2020-09-27
Only change the node status if a node fails to ACK a message

### LINKED ISSUE #993: Don't treat a missing node response after ACK as a dead node

AlCalzone:
fixes: #993


AlCalzone:
manually merged in 98c1e02c

PR #1073, opened 2020-10-23
Alarm Sensor CC: only overwrite the current node id if the source node id is set

### LINKED ISSUE #1072: AlarmSensorCCReport sets Node ID to 0
The current implementation assumes that the first payload byte always contains the correct node id, but apparently this is only the case if the frame is forwarded by another node.

_Originally posted by @AlCalzone in https://github.com/zwave-js/node-zwave-js/issues/1037#issuecomment-715213254_

AlCalzone:
fixes: #1072 

@slangstrom could you test if this resolves your problem with the Everspring device? No new inclusion necessary, just see if the interview completes and the unsolicited reports afterwards are assigned to the correct node.

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/AlarmSensorCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/AlarmSensorCC.ts
@@ -308,7 +308,12 @@ export class AlarmSensorCCReport extends AlarmSensorCC {
 	) {
 		super(driver, options);
 		validatePayload(this.payload.length >= 5, this.payload[1] !== 0xff);
-		this.nodeId = this.payload[0];
+		// Alarm Sensor reports may be forwarded by a different node, in this case
+		// (and only then!) the payload contains the original node ID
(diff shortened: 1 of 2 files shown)


slangstrom:
Yep, looks like it's working now!

```
11:47:38.731 CNTRLR   [Node 002] [~] [Alarm Sensor] state[5]: false => true             [Endpoint 0]
11:47:38.732 CNTRLR   [Node 002] [~] [Alarm Sensor] severity[5]: undefined => undefined [Endpoint 0]
11:47:38.732 CNTRLR   [Node 002] [~] [Alarm Sensor] duration[5]: undefined => undefined [Endpoint 0]
11:47:38.732 SERIAL » [ACK]                                                                   (0x06)
11:47:38.732 DRIVER « [Node 002] [REQ] [ApplicationCommand]
                      └─[AlarmSensorCCReport]
                          sensor type: Water Leak
                          alarm state: true
```

PR #1086, opened 2020-11-02
Use steps output to propagate version and release body

### LINKED ISSUE #1084: [bug] Wrong releases body
Starting from version 5.3.2 the body of releases is wrong. I think the problem is [here](https://github.com/zwave-js/node-zwave-js/blob/master/.github/workflows/test-and-release.yml#L160-L164)

AlCalzone:
fixes: #1084


robertsLando:
Should work based on: https://github.community/t/set-output-truncates-multiline-strings/16852/3

PR #1093, opened 2020-11-04
feat(config): add support for importing the OZW devices database

### LINKED ISSUE #414: Import OpenZWave device config DB
Like #411 but with the files from the OpenZWave repo.

robertsLando:
fixes: #414


robertsLando:
@AlCalzone This is a work in progress, unfortunally xml2json hasn't a type definition and I'm not able to register the type I have created on `maintenance/@types/xml2json`, could you tell me what I'm doing wrong?

AlCalzone:
There you go:
grafik

robertsLando:
@AlCalzone Still not working, had tried that too, same error :confounded:

AlCalzone:
Did you restart the TypeScript editor features?

AlCalzone:
Damn, now it stopped working for me too

robertsLando:
The joy of typescript :rofl: Not a good start for me, actually the download/extract functions work but I'm stucked with this

AlCalzone:
then //@ts-ignore the import for now while I figure out wtf is happening

robertsLando:
@AlCalzone By removing `rootDir: '../'` it works

AlCalzone:
Edit:
- install type definitions: `npm i -D @types/xml2json`
- rootDir needs to be there, or the import of `@zwave-js/config` will complain
- `maintenance/tsconfig.json` should look like this:
```jsonc
{
	"extends": "../tsconfig.json",
	"compileOnSave": true,
	"compilerOptions": {
		"module": "commonjs",
		"moduleResolution": "node",
		// "noImplicitAny": true,
		"target": "es2015",
		"rootDir": "../"
	},
	"include": [
		"*.ts",
		"../packages/*/src/**/*.ts"
	]
}
```

robertsLando:
Yep essentially I just had to install the missing types in that way :cry:  

Now I have fixed it, my question now how do you update `manufacturers.json`? there are lot missing manufacturers

robertsLando:
@AlCalzone Actually I'm able to read eveything, what is missing now is how to edit/create the config files

- `lerna run config -- ozw download`: download ozw archive to tmp dir
- `lerna run config -- ozw manufacturers`: parse manufacturers and products
- `lerna run config -- ozw download manufacturers`: donwload and parse manufacturers and products
- `lerna run config -- clean`: clean temporary directory

robertsLando:
I'm wondering if there are utilities for `Config Database Updates` to update and create both manufacturers/devices files

AlCalzone:
> I'm wondering if there are utilities for `Config Database Updates` to update and create both manufacturers/devices files

Nothing special. The existing code just creates a JS object in the correct format and saves it as JSON.

robertsLando:
What about adding a method to Manufactures.ts to set manufacturer keys and one to write the actual map to json file? And the same for devices?

AlCalzone:
Sounds good

robertsLando:
@AlCalzone I have saw that manufacturers id stored in `manufacturers.json` are upper cases hex, them are lower cases in devices folders and in devices json files. Is this correct?

robertsLando:
@AlCalzone The script has __EDITED__ 504 existing configurations and  imported 257 __NEW__ devices, would you like to test it or would you like me to also add the created files?

robertsLando:
In some cases the default value is `0x09 0x00 0x09 0x00` this happens for values with the label `Temperature Reporting Filter`.

How should I handle them?

__UPDATE__: Seems it's a bug of xml2json, dunno why but that char appears in xml when `value=""`

robertsLando:
There is a REALLLLLY strange bug with the parse of number `2147418112‬`, seems it's impossible to parse it :confused: 

__UPDATE__: Seems the value was `'2147418112‬'` (thought the ' I saw in the debugger was the delimiters, not chars of the string :confounded: ) even this one a xml2json bug I think, used a regex to remove the `'` chars fixed the problem

AlCalzone:
> 
> 
> @AlCalzone I have saw that manufacturers id stored in `manufacturers.json` are upper cases hex, them are lower cases in devices folders and in devices json files. Is this correct?

Huh, I wouldn't mind if it was all the same.

> The script has imported 835 devices, would you like to test it or would you like me to also add the created files?

I'd like to take a look at the files first, but my schedule is quite full today. Maybe tonight.

> In some cases the default value is `0x09 0x00 0x09 0x00` this happens for values with the label `Temperature Reporting Filter`.
> How should I handle them?

I'm not sure, can you show an example?

> There is a REALLLLLY strange bug with the parse of number `2147418112‬`, seems it's impossible to parse it 😕

where?

robertsLando:
@AlCalzone I'm done here, let me know if you want me to upload the parsed configurations or not. I think I have fixed any 'merging' problem, the only thing that ozw db doesn't have is a reference to the devices firmware version so all versions are '0.0' and '255.255'

robertsLando:
Should I do something with [this](https://github.com/OpenZWave/open-zwave/blob/master/config/qubino/ZMNHADx.xml#L166-L180) values?

```xml
<!-- Map endpoints to instances -->
  <CommandClass id="96">
    <Compatibility>
      <MapRootToEndpoint>true</MapRootToEndpoint>
    </Compatibility>
    <Instance index="1" label="Load" />
    <Instance index="2" label="i2" />
    <Instance index="3" label="i3" />
    <Instance index="4" label="Temp Sensor" />
  </CommandClass>
  <!-- COMMAND_CLASS_ALARM.  This class is supported but is missing from the list reported by the ZMNHADx -->
  <CommandClass action="add" id="113">
    <Compatibility>
      <GetSupported>false</GetSupported>
    </Compatibility>
  </CommandClass>
```

Actually the only command classes I'm importing are the configuration (112) associations (133) and multi channel associations (142)

AlCalzone:
> the only thing that ozw db doesn't have is a reference to the devices firmware version so all versions are '0.0' and '255.255'

I think that is okay for now. We can be on the lookout for conflicts when reviewing the config file PRs.

> ```
> <CommandClass id="96">
> ```

I don't have any mechanism to overwrite the reported endpoints, so I would leave that out for now

> ```
> <CommandClass action="add" id="113">
> ```

I also don't have any mechanism to add CCs the device does not report (or remove ones they don't support). Would be interesting to do - maybe add a new key under the (undocumented 😅) `compat` property?

> ```
>     <Compatibility>
>       <GetSupported>false</GetSupported>
> ```

This looks like a compat flag which requires code that uses it - I think this is out of scope of this PR for now.

AlCalzone:
As for the config files - I would like to review this PR in-depth first before having to look at another 800 files :)

robertsLando:
> This looks like a compat flag which requires code that uses it - I think this is out of scope of this PR for now.

Ok Was just for reference :) Will wait for the review so :crossed_fingers:  Be kind :rofl:

robertsLando:
@AlCalzone Fixed, sorry :confounded: I suggest to squash and merge when it's ready for a cleaner master

robertsLando:
Additional questions:

- Should I parse this file too?
    - https://github.com/OpenZWave/open-zwave/blob/master/config/SensorMultiLevelCCTypes.xml
    - https://github.com/OpenZWave/open-zwave/blob/master/config/NotificationCCTypes.xml
    - https://github.com/OpenZWave/open-zwave/blob/master/config/device_classes.xml

- How I should parse bitmask values ?

```xml
<Value bitmask="55" genre="config" index="87" instance="1" label="Sensor is triggered the buzzer will alarm" max="55" min="0" size="4" type="bitset" value="55">
      <Help>What Sensors Trigger the Buzzer</Help>
      <BitSet id="1">
        <Label>Water Leak</Label>
        <Help>If the Water leak is triggered, the buzzer will alarm.</Help>
      </BitSet>
      <BitSet id="2">
        <Label>Vibration</Label>
        <Help>If the vibration is triggered, the buzzer will alarm.</Help>
      </BitSet>
      <BitSet id="3">
        <Label>Tilt Sensor</Label>
        <Help>If the Tilt Sensor is triggered, the buzzer will alarm.</Help>
      </BitSet>
      <BitSet id="5">
        <Label>UnderHeat</Label>
        <Help>If the Under Heat Temperature is triggered, the buzzer will alarm.</Help>
      </BitSet>
      <BitSet id="6">
        <Label>OverHeat</Label>
        <Help>If the Over Heat Temperature is triggered, the buzzer will alarm.</Help>
      </BitSet>
    </Value>
```

- I didn't set the flag `supportsZwavePlus` on products as I didn't find any reference of it on xml files, couldn't you just check if the device supports command class `94` for that (or maybe the device could tell you this in some other ways)?

packages/config/src/Manufacturers.ts
```diff
@@ -93,3 +94,47 @@ export function lookupManufacturer(manufacturerId: number): string | undefined {
...
+		throw new ZWaveError(
+			"The config has not been loaded yet!",
+			ZWaveErrorCodes.Driver_NotReady,
+		);
+	}
+
+	const data: Record<string, string> = {};
+
+	const orderedMap = new Map(
+		[...manufacturers].sort((a, b) => (a[0] > b[0] ? 1 : -1)),
```
AlCalzone, line 136: I think you're supposed to return 0 if both are equal

packages/config/src/Devices.ts
```diff
@@ -84,6 +85,53 @@ export async function loadDeviceIndex(): Promise<void> {
...
+
+	await writeFile(indexPath, JSON.stringify(index, undefined, 4));
+}
+
+/**
+ * Looks up the definition of a given device in the configuration DB
+ * @param manufacturerId The manufacturer id of the device
+ * @param productType The product type of the device
+ * @param productId The product id of the device
+ * @param firmwareVersion If known, configuration for a specific firmware version can be loaded
```
AlCalzone: You don't load here :)

packages/config/src/Devices.ts
```diff
@@ -84,6 +85,53 @@ export async function loadDeviceIndex(): Promise<void> {
...
+		);
+	}
+	// Look up the device in the index
+	const indexEntry: DeviceConfigIndexEntry = {
+		manufacturerId: formatId(manufacturerId),
+		productId: formatId(productId),
+		productType: formatId(productType),
+		firmwareVersion: {
+			min: firmwareVersionMin || "0.0",
+			max: firmwareVersionMax || "255.255",
```
AlCalzone, line 131: ??

packages/config/src/Manufacturers.ts
```diff
@@ -93,3 +94,47 @@ export function lookupManufacturer(manufacturerId: number): string | undefined {
...
+	);
+
+	for (const [id, name] of orderedMap) {
+		const hexId = "0x" + padStart(id.toString(16), 4, "0").toUpperCase();
```
AlCalzone: use `formatId`

maintenance/importConfig.ts
```diff
@@ -93,6 +108,342 @@ async function fetchDevice(id: string): Promise<string> {
...
+			ret.associations = existingDevice.associations;
+		}
+
+		if (existingDevice.paramInformation) {
+			ret.paramInformation = existingDevice.paramInformation;
+		}
+	}
+
+	const parameters = ensureArray(
+		commandClasses.find((c: any) => c.id === 112)?.Value,
```
AlCalzone: 112 -> CommandClasses.Configuration

maintenance/importConfig.ts
```diff
@@ -93,6 +108,342 @@ async function fetchDevice(id: string): Promise<string> {
...
+		.toLocaleUpperCase();
+
+	const productName = product.name;
+
+	// for some reasons some products already have the prefix `0x`, remove it
+	product.id = product.id.replace("0x", "");
+	product.type = product.type.replace("0x", "");
+
+	const productId = "0x" + padStart(product.id, 4);
+	const productType = "0x" + padStart(product.type, 4);
```
AlCalzone: `formatId(...)`

AlCalzone (changes requested):

AlCalzone:
>     * Should I parse this file too?

no, I've got definitions for all of them (straight from the specifications)

About the bitsets: How are the OZW values supposed to be understood?

AlCalzone:
> I didn't set the flag supportsZwavePlus on products as I didn't find any reference of it on xml files

I don't think it is really necessary. I'm not using it anywhere, it is more of a hint to devs who edit the device config files that associations don't need to be defined. Except for devices which support Z-Wave+ but require more lifelines than group 1.

maintenance/importConfig.ts
```diff
@@ -35,6 +43,13 @@ const processedDir = path.join(
 	"../packages/config",
 	"config/devices",
 );
+
+const tmpDir = path.join(__dirname, "../.tmp");
+const ozwTarName = "openzwave.tar.gz";
+const ozwTarUrl =
+	"https://github.com/OpenZWave/open-zwave/archive/master.tar.gz";
```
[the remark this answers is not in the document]
robertsLando: I don't think it's a problem, also with that we should check for latest release in the html table or hardcode the link and change it everytime a new update is out, I would suggest to use master, all devices configs are edited in PRs then also when we will create a github action we will review anyway the changes, if something is wrong we just ignore that change

maintenance/importConfig.ts
```diff
@@ -93,6 +108,342 @@ async function fetchDevice(id: string): Promise<string> {
...
+			reject(err);
+		});
+
+		stream.on("close", () => {
+			if (!hasError) {
+				resolve(fileDest);
+				console.log("ozw archive stored in temporary directory");
+			}
+		});
+	});
```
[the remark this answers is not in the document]
robertsLando: Yep, AFAIK

maintenance/importConfig.ts
```diff
@@ -93,6 +108,342 @@ async function fetchDevice(id: string): Promise<string> {
...
+}
+
+/** Extract `config` folder from ozw archive in `tmpDir` */
+function extractConfigFromTar(): Promise<void> {
+	return new Promise((resolve, reject) => {
+		console.log("extracting config folder from ozw archive...");
+		child.exec(
+			`tar -xzf ${ozwTarName} open-zwave-master/config  --strip-components=1`,
```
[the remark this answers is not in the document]
robertsLando: Yes no problem

maintenance/importConfig.ts
```diff
@@ -93,6 +108,342 @@ async function fetchDevice(id: string): Promise<string> {
...
+	return new Promise((resolve, reject) => {
+		child.exec("rm -rf *", { cwd: tmpDir }, function (error) {
+			if (error) {
+				reject(error);
+			} else {
+				resolve();
+				console.log("temporary directory cleaned");
+			}
+		});
+	});
```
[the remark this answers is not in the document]
robertsLando: Didn't saw you were using fs-extra, I though you was using fs.promises, will edit it :)

maintenance/importConfig.ts
```diff
@@ -93,6 +108,342 @@ async function fetchDevice(id: string): Promise<string> {
...
+	);
+	const manufacturerJson = xml2json.toJson(
+		await fs.readFile(manufacturerFile, "utf8"),
+		{ object: true },
+	);
+
+	await loadManufacturers();
+	await loadDeviceIndex();
+
+	// @ts-ignore
```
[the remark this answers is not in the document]
robertsLando: Sure!

maintenance/importConfig.ts
```diff
@@ -93,6 +108,342 @@ async function fetchDevice(id: string): Promise<string> {
...
+
+/**
+ * When using xml2json some fields expected as array are parsed as objects
+ * when there is only one element. This method ensures that them are arrays
+ *
+ * @param {any} json
+ * @returns {Array<any>}
+ */
+function ensureArray(json: any): Array<any> {
+	json = json || [];
```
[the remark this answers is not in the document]
robertsLando: Does it works even if json is undefined?

maintenance/importConfig.ts
```diff
@@ -93,6 +108,342 @@ async function fetchDevice(id: string): Promise<string> {
...
+	await writeManufacturersToJson();
+	await writeIndexToFile();
+}
+
+/**
+ * When using xml2json some fields expected as array are parsed as objects
+ * when there is only one element. This method ensures that them are arrays
+ *
+ * @param {any} json
+ * @returns {Array<any>}
```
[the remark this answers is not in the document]
robertsLando: Ok will remove types, was auto added by vs-code "document this" function

maintenance/importConfig.ts
```diff
@@ -93,6 +108,342 @@ async function fetchDevice(id: string): Promise<string> {
...
+/**
+ * Read and parse the product xml, add it to index if missing,
+ * create/update device json config and validate the newly added
+ * device
+ *
+ * @param {any} product the parsed product json entry from manufacturer.xml
+ * @param {(string | undefined)} manufacturer manufacturer name string
+ * @param {string} manufacturerId manufacturer hex id `0xXXXX`
+ * @param {number} manufacturerIdInt manufacturer integer id
+ * @returns {Promise<void>} fulfilled once the parsing ends
```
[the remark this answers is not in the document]
robertsLando: You mean manifacturerId and manifacturerIdInt?

maintenance/importConfig.ts
```diff
@@ -93,6 +108,342 @@ async function fetchDevice(id: string): Promise<string> {
...
+ */
+async function parseOzwProduct(
+	product: any,
```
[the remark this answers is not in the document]
robertsLando: I should create an interface for that I think, there is not just the config but also name and id

maintenance/importConfig.ts
```diff
@@ -93,6 +108,342 @@ async function fetchDevice(id: string): Promise<string> {
...
+
+	const productLabel = path
+		.basename(product.config, ".xml")
+		.toLocaleUpperCase();
+
+	const productName = product.name;
+
+	// for some reasons some products already have the prefix `0x`, remove it
+	product.id = product.id.replace("0x", "");
+	product.type = product.type.replace("0x", "");
```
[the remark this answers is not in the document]
robertsLando: It's a typo in the file, it's only present in 2 products, I can make that check but it's almost the same IMO

maintenance/importConfig.ts
```diff
@@ -481,7 +832,18 @@ async function updateManufacturerNames(): Promise<void> {
...
+			await extractConfigFromTar();
+		}
+
+		if (process.argv.includes("manufacturers")) {
+			await parseOzwConfig();
+		}
+	} else if (process.argv.includes("manufacturers")) {
 		await downloadManufacturers();
 	} else if (process.argv.includes("manufacturerNames")) {
 		await updateManufacturerNames();
```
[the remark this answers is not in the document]
robertsLando: We could, yes :)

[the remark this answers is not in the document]
robertsLando: What options names do you suggest?

packages/config/src/Manufacturers.ts
robertsLando: In theory yes, but there couldn't be 2 manufacturers with the same id

robertsLando:
> About the bitsets: How are the OZW values supposed to be understood?

What do you mean? In Z2M I created checkboxes for bitmask values, and then I converted the bits to an integer to write

robertsLando:
I will do the edits tomorrow

robertsLando:
@AlCalzone Once this will be merged I will start working at the github action that will create the PR

__UPDATE__: The script has __EDITED 504__ existing configurations and imported __257 NEW__ devices :rocket:

maintenance/importConfig.ts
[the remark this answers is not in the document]
AlCalzone: For JS it makes sense (and I know it tries to add correct types there). But for TypeScript not really...
Might be worth an issue in the VSCode repo

maintenance/importConfig.ts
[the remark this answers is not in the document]
AlCalzone: ?? takes the right hand side if the left side is exactly `null` or `undefined`. But not if `""`, `0`, or any other falsy value.

AlCalzone:
> What do you mean?

I mean we could use the bitset information to create our partial param values. But I'm not sure which bit `<BitSet id="1">` is targeting. Bit 0 (1<<0) or Bit 1 (1<<1)? Big endian or little endian?

maintenance/importConfig.ts
```diff
@@ -34,7 +35,91 @@ import * as fs from "fs-extra";
...
+import yargs from "yargs";
+
+const execPromise = promisify(child.exec);
+
+yargs
+	.option("source", {
+		description: "source of the import",
+		alias: "s",
+		type: "array",
+		choices: ["oh", "ozw"], // oh: openhub, ozw: openzwave
```
AlCalzone:
```diff
- openhub
- OpenHub
+ OpenHAB
```
;)

maintenance/importConfig.ts
```diff
@@ -34,7 +35,91 @@ import * as fs from "fs-extra";
...
+	})
+	.option("download", {
+		alias: "D",
+		description: "Download devices DB from <source>",
+		type: "boolean",
+		default: false,
+	})
+	.option("clean", {
+		alias: "C",
+		description: "Clean temprorary directory",
```
AlCalzone:
```diff
- temprorary
+ temporary
```

maintenance/importConfig.ts
robertsLando: Fixed :+1:

maintenance/importConfig.ts
```diff
@@ -34,7 +35,91 @@ import * as fs from "fs-extra";
...
+		default: false,
+	})
+	.option("devices", {
+		alias: "d",
+		description: "Parse and update devices configurations",
+		type: "boolean",
+		default: false,
+	})
+	.example(
+		"import -s ozw -Dmid",
```
AlCalzone, line 88: Would it make sense to merge the source and download options into one?

robertsLando:
> But I'm not sure which bit <BitSet id="1"> is targeting. Bit 0 (1<<0) or Bit 1 (1<<1)? Big endian or little endian?

Bit 1 is bit[0], the index starts from +1.  Rappresentation: decimal 3 --> `00000011` bit[0] and bit[1] (bit index 1 and 2 in ozw) are 1. So Bit

Here is how I handled them in z2m: https://github.com/OpenZWave/Zwave2Mqtt/blob/master/lib/ZwaveClient.js#L52
And Here is how I parse them in ozw-shared: https://github.com/OpenZWave/node-openzwave-shared/blob/master/src/utils.cc#L188

>  we could use the bitset information to create our partial param values

Example?

maintenance/importConfig.ts
```diff
@@ -34,7 +35,91 @@ import * as fs from "fs-extra";
...
+class Program {
+	public source: string[];
+	public ids: string[] | undefined;
+	public download: boolean;
+	public clean: boolean;
+	public manufacturers: boolean;
+	public index: boolean;
+	public devices: boolean;
+
+	constructor(json: any) {
```
AlCalzone: You're losing the yargs type definitions this way.
I think you could just `pick` the wanted properties from `yargs.argv` or use `.argv` directly, like here:
https://github.com/ioBroker/create-adapter/blob/master/src/cli.ts#L59-L62

maintenance/importConfig.ts
```diff
@@ -134,35 +225,18 @@ async function downloadOzwConfig(): Promise<string> {
...
-				}
-			},
-		);
-	});
+async function extractConfigFromTar(): Promise<void> {
+	console.log("extracting config folder from ozw archive...");
+	await execPromise(
+		`tar -xzf ${ozwTarName} open-zwave-master/config  --strip-components=1`,
+		{ cwd: tmpDir },
+	);
```
AlCalzone, line 238: cool. I think promisify-child-process did some small extras, but the built-ins should be enough for our purpose.

maintenance/importConfig.ts
```diff
@@ -251,41 +335,35 @@ async function parseOzwProduct(
...
-		(await lookupDevice(
+		!(await lookupDevice(
 			manufacturerIdInt,
 			parseInt(product.type, 16),
 			parseInt(product.id, 16),
-		)) === undefined
+		))
 	) {
+		// let this here, if program.index is false it will not
+		// write the device index to file
```
AlCalzone: couldn't you just add `program.index &&` in front of the if-condition?

maintenance/importConfig.ts
```diff
@@ -348,8 +437,10 @@ async function parseOzwProduct(
 	for (const param of parameters) {
 		if (isNaN(param.index)) continue;
 
-		const parsedParam = ret.paramInformation[param.index] || {};
+		const parsedParam = ret.paramInformation[param.index] ?? {};
 
+		// don't use ?? here, some fields could be empty strings and ?? operator
+		// will not work
```
AlCalzone: agreed

maintenance/importConfig.ts
```diff
@@ -832,33 +924,36 @@ async function updateManufacturerNames(): Promise<void> {
...
+				await updateManufacturerNames();
+			}
+
+			if (program.index) {
+				await generateDeviceIndex();
+			}
+
+			if (program.devices) {
+				await importConfigFiles();
+			}
```
AlCalzone, line 1179: I think this should be done before updating the index

AlCalzone (changes requested):
Just a few more details. Looks pretty good.

Do we need to implement updating for already-existing configs from OpenHAB aswell (like you did for OZW)?

AlCalzone:
> > we could use the bitset information to create our partial param values
> 
> Example?

--> https://zwave-js.github.io/node-zwave-js/#/development/config-files?id=partial-parameters

An example can be found in the config file for `Aeotec Multi Sensor 6`, where I split the report groups from one param into several 1-bit values

maintenance/importConfig.ts
robertsLando: Yes but in case of devices this would make the lookup at the end to fail :)

maintenance/importConfig.ts
robertsLando: I saw that but in this way it's much better typed IMO, if you prefer that I can remove the Program class and use that

maintenance/importConfig.ts
robertsLando: I don't understand, what do you mean?

robertsLando:
> Do we need to implement updating for already-existing configs from OpenHAB aswell

Good question, I think that it would be good yes, but could you do it in another PR as you better know openhab dbs?

maintenance/importConfig.ts
AlCalzone: Ohhh - I didn't make that connection in my head

maintenance/importConfig.ts
AlCalzone: lets keep it like this for now. I thought that the types for argv are pretty good, but I haven't tested it in this case.

maintenance/importConfig.ts
AlCalzone: You have a `download` option `-D` and a `source` option `-s`. 
As far as I understand, `download` requires that you provide the source and `source` without `download` does nothing.
Why not just remove the `download` option and make the script download first when there's a source given?

AlCalzone: Ehhh - nevermind. The import logic is also different for both sources.

AlCalzone:
Ok, just one more thing:
Updating the index must also be done if one manually changes a json config file. This does not depend on any source file format, so I think regenerating the index should be moved out of `parseOzwConfig` and be done after all the imports are done.

maintenance/importConfig.ts
```diff
@@ -437,59 +438,86 @@ async function parseOzwProduct(
...
-		parsedParam.defaultValue = param.value || parsedParam.value || 0;
-
-		if (param.units !== undefined) {
-			parsedParam.units = param.units;
-		}
+		if (param.type === "bitset") {
+			const bitSetIds = ensureArray(param.BitSet);
+			for (const bitSet of bitSetIds) {
+				let id = parseInt(bitSet.id).toString(16);
+				id = `${param.index}[0x${padStart(id, 2, "0")}]`;
```
AlCalzone: This isn't entirely correct, is it?
* Bitset id 1 means Bit 0 of the bit mask is set => 0x01
* Bitset id 2 means Bit 1 of the bit mask is set => 0x02
* Bitset id 3 means Bit 2 of the bit mask is set => 0x04
* ... etc

so you would need to transform `(1 << (bitSet.id - 1))` to hex.

robertsLando:
> Updating the index must also be done if one manually changes a json config file 

I don't understand what you mean

maintenance/importConfig.ts
```diff
@@ -437,59 +438,86 @@ async function parseOzwProduct(
...
+				const label = ensureArray(bitSet.Label)[0];
+				const desc = ensureArray(bitSet.Help)[0];
+
+				// this values are all parsed as switches but could be transformed to
+				// list with two options: `Enable` and `Disable`
+				parsedParam.label = label;
+				parsedParam.description = desc;
+				parsedParam.valueSize = 1;
+				parsedParam.minValue = 0;
+				parsedParam.maxValue = 1;
```
AlCalzone: This is fine, I'm planning to automatically convert these kinds of params to boolean some time in the future.

maintenance/importConfig.ts
robertsLando: Done

maintenance/importConfig.ts
```diff
@@ -437,59 +438,86 @@ async function parseOzwProduct(
...
+				parsedParam.defaultValue = 0;
```
AlCalzone: Shouldn't this be the corresponding bit in `param.value`?

AlCalzone:
> > Updating the index must also be done if one manually changes a json config file
> 
> I don't understand what you mean

I'm talking about this part:
https://github.com/zwave-js/node-zwave-js/blob/5db20fa52117e6f628cf4562d3645d5012649a00/maintenance/importConfig.ts#L958-L985
and this line specifically
https://github.com/zwave-js/node-zwave-js/blob/5db20fa52117e6f628cf4562d3645d5012649a00/maintenance/importConfig.ts#L983

Updating `index.json` must also be done if you manually edit a config file (correct a typo for example). The way this is set up now, it happens as part of the OZW or the OpenHAB import. IMO this should be a separate step after importing the config files - You should be able to run the script to only update the index.

maintenance/importConfig.ts
robertsLando: Unfortunally, yes, you right :confounded:

maintenance/importConfig.ts
```diff
@@ -441,7 +441,7 @@ async function parseOzwProduct(
 		if (param.type === "bitset") {
 			const bitSetIds = ensureArray(param.BitSet);
 			for (const bitSet of bitSetIds) {
-				let id = parseInt(bitSet.id).toString(16);
+				let id = (parseInt(bitSet.id) - 1).toString(16);
```
AlCalzone: No, we need bit shifts! The thing in the square brackets is the bit *mask* itself, e.g. `0xc` to target bits 2 and 3 (zero-based)

robertsLando: Arrggg I feel so stupid. Sorry

robertsLando:
This file is an exampe for bitsets: `/packages/config/config/devices/0x0086/zw100.json`

maintenance/importConfig.ts
robertsLando: @AlCalzone Check it now :)

robertsLando:
> Updating index.json must also be done if you manually edit a config file (correct a typo for example). The way this is set up now, it happens as part of the OZW or the OpenHAB import. IMO this should be a separate step after importing the config files - You should be able to run the script to only update the index.

Ok I understand now, based on this so it's useless for me to call the method to write the devices index to file that I'm using now, I will remove it

maintenance/importConfig.ts
robertsLando: > As far as I understand, download requires that you provide the source and source without download does nothing.
Why not just remove the download option and make the script download first when there's a source given?

Nope because expecially when you are doing testing you don't want to download the source every time, it's a loss of time

AlCalzone:
> This file is an exampe for bitsets: `/packages/config/config/devices/0x0086/zw100.json`

Did you commit that file or should I try to import it?

AlCalzone (approved):

robertsLando:
@AlCalzone Was just a reference to a file to check that has bitsets :)

AlCalzone:
That's good to see the changes, since it was hand edited

maintenance/importConfig.ts
AlCalzone: Actually, I'll push a commit where I changed that. The entire class is unnecessary

AlCalzone:
For reference, the correct command to run the script is `npm run config import -- <options>`. Without `--`, the dashed options seem to be dropped at least partially

robertsLando:
> the dashed options seem to be dropped at least partially

I think it's caused by my `yargs.parse` function, I just saw you have removed it so it sholdn't be a problem now

AlCalzone:
@robertsLando A few points regarding the config files we need to discuss/change:

`0x0001/lfm20.json` (and several others):
```json
{
    "manufacturer": "ACT",
    "manufacturerId": "0x0001",
    "label": "LFM20",
    "description": "LFM-20 Relay Fixture Module",
    "devices": [
        {
            "productType": "0x5246",
            "productId": "0x3133"
        }
    ],
    "firmwareVersion": {
        "min": "0.0",
        "max": "255.255"
    }
}
```
These files just configure the name of the devices. Do you see a point in that or should we leave them out?

---

many, many devices:
```json
    "label": "ZDM230",
    "description": "ZDM230 Wall Dimmer Module",
```
I'd remove the label from the description if the description starts or ends with the label

---

`0x010f/fgrm222.json` and many others:

* Please place `defaultValue` directly under min and max value.
* If `allowManualEntry` is `false`, `minValue` should be equal to the lowest option value (here it is) and `maxValue` should be equal to the highest option value (here it is not)

```json
    "paramInformation": {
        "1": {
            "label": "Local Protection",
            "description": "...",
            "valueSize": 1,
            "minValue": 0,
            "maxValue": 100,
            "readOnly": false,
            "writeOnly": false,
            "allowManualEntry": false,
            "defaultValue": 0,
            "options": [
                {
                    "label": "No protection. Roller Shutter responds to push buttons",
                    "value": 0
                },
                {
                    "label": "Not supported",
                    "value": 1
                },
                {
                    "label": "Local protection active. Roller Shutter does not respond to push buttons",
                    "value": 2
                }
            ]
        },
```

---

`0x0001/zdw103.json` and many others:
* The key `units` should be called `unit` without s
* Please place `unit` under `valueSize` (if unit exists)
* `"units": ""` should be removed (might be due to my proposed `||` to `??` change)
```json
        "7": {
            "label": "On/Off Command Dim Step",
            "description": "Indicates how many levels the dimmer will change for each dimming step.",
            "valueSize": 1,
            "minValue": 1,
            "maxValue": 99,
            "defaultValue": 1,
            "readOnly": false,
            "writeOnly": false,
            "allowManualEntry": true,
            "units": ""
        },
```

---

`0x000c/hs-wd100plus.json` and probably a bunch more:
* Please place the key `"associations"` before `"paramInformation"`

---

A bunch of units need to be automatically normalized or changed during the import (I realize that we need to do that for the existing files too):
* `"Minutes"` -> `"minutes"` (lowercase)
* `"Seconds"` -> `"seconds"` (lowercase)
* `"Decimal"` -> drop entirely
* `"int"` -> drop entirely
* `"cs"` -> any idea WTF that unit is?
* `"fahrenheit"` -> `"°F"`
* `"F"` -> `"°F"`
* `"Degrees Celsius"` -> `"°C"`
* `"Watt"`/`"watts"` -> `"W"`
* `"Volt"`/`"volts"` -> `"V"`
* `"Percent"` -> `"%"`
* `"dimmer level %"` -> `"%"`
* `"degrees"` -> `"°"` (no idea if those devices talk about C or F)

---

When splitting config params, the label of the original param should be included somehow.
Example in `0x0086/zw100.json`, param 101:
```json
        "101[0x01]": {
            "label": "Battery",
            "description": "Send Battery Report",
            "valueSize": 1,
            "minValue": 0,
            "maxValue": 1,
            "defaultValue": 1,
            "readOnly": false,
            "writeOnly": false,
            "allowManualEntry": true
        },
```
This is for group 1, but you can't see that here.

Also, we should sort the params by the param number without the bitmask and after that by bitmask, e.g.: `99, 100[0x01], 100[0x02], 101`

---

There's probably more things, but we'll need to reduce the noise first :)

AlCalzone:
> 
> 
> > the dashed options seem to be dropped at least partially
> 
> I think it's caused by my `yargs.parse` function, I just saw you have removed it so it sholdn't be a problem now

No, these params don't even make it into `process.argv`, so there's nothing to parse them from.

AlCalzone (changes requested):
see my previous loooong comment

robertsLando:
> These files just configure the name of the devices. Do you see a point in that or should we leave them out?

Humm nope, it could be left unchanged if exists IMO

> I'd remove the label from the description if the description starts or ends with the label

I agree, easy to do with a regex with `g` and `i` flags

> If allowManualEntry is false, minValue should be equal to the lowest option value (here it is) and maxValue should be equal to the highest option value (here it is not)

I saw this, even in existing files, I think that we could create a method like `normalizeParams` that check for this thing and the params order

> should be removed (might be due to my proposed || to ?? change)

I think it's caused by a `??` that need to be replaced with `||`

> A bunch of units need to be automatically normalized or changed during the import (I realize that we need to do that for the existing files too):

Here too, I suggest to do such things in the `normalizeParams` method

> When splitting config params, the label of the original param should be included somehow.

Ok it's easy, we already have that

> Also, we should sort the params by the param number without the bitmask and after that by bitmask, e.g.: 99, 100[0x01], 100[0x02], 101

Maybe this too can be added in the `normalizeParams` method?

maintenance/importConfig.ts
```diff
@@ -835,6 +975,45 @@ ${JSON.stringify(parsed, undefined, 4)}`;
...
+	files.sort((a, b) => {
+		const vA = parseFileName(a) ?? "0.0.0";
+		const vB = parseFileName(b) ?? "0.0.0";
+
+		return compare(vA, vB);
+	});
+
+	return files[files.length - 1];
+}
+
```
AlCalzone: I think its better to use the index to look this up instead of parsing file names, similar to:
https://github.com/zwave-js/node-zwave-js/blob/5d5677a0d92ece7db18014cf86e8fafd64d89bb5/packages/config/src/Devices.ts#L94-L115
Then take the highest version from the index entries

maintenance/importConfig.ts
```diff
@@ -308,7 +435,8 @@ async function parseOzwProduct(
 		.basename(product.config, ".xml")
 		.toLocaleUpperCase();
 
-	const productName = product.name;
+	// any products descriptions have productName in it, remove it
+	const productName = product.name.replace(productLabel, "");
```
AlCalzone, line 473: Lets see how this turns out, maybe we need to limit it to the start and end of the string

maintenance/importConfig.ts
```diff
@@ -109,6 +110,8 @@ const processedDir = path.join(
 	"config/devices",
 );
 
+const paramsRegex = /\[0x([0-9]+)\]/;
```
AlCalzone: The params are HEX, so you'll need to include `a-f` in the character class too.

AlCalzone: And probably A-F too just to be sure, see https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/src/Devices.ts#L250

maintenance/importConfig.ts
```diff
@@ -282,6 +285,130 @@ function ensureArray(json: any): any[] {
...
 }
 
+function normalizeUnits(unit: string) {
+	if (!unit) return undefined;
+
+	if (/minutes/i.test(unit)) {
+		return "minutes";
+	} else if (/seconds/i.test(unit)) {
+		return "seconds";
+	} else if (/fahrenheit|\bF\b/i.test(unit)) {
```
AlCalzone: lets make the F lowercase to be consistent ;) I had to think why it was uppercase and the c in the next check was not.

AlCalzone (changes requested):

maintenance/importConfig.ts
robertsLando: I have added the `i` flag to regex so it's case unsensitive

maintenance/importConfig.ts
robertsLando: @AlCalzone I don't understand how to use that to achieve the same result, when I call that method I need to specify also the firmware version

AlCalzone: Sorry, was a bit unclear. I'm suggesting to use
```ts
index.filter(entry => entry.manufacturerId === product.manufacturerId && entry.productType === product.productType && entry.productId === product.productId)
```
to find all matching config files and from that list select the one with the highest firmware version.

The filename-based lookup will break when we decide to rename the files or folders at some point.

AlCalzone:
@robertsLando I think we might need the ability to remove CCs sooner than I expected: https://github.com/zwave-js/node-zwave-js/issues/1107

So while you're at it, we should try to parse this information from the OZW DB. Don't worry about the implementation in the interview.

I suggest we put it under the `"compat"` property (which can be seen in `packages\config\config\devices\0x0002\lc-13.json`) like this:
```jsonc
{
  // ...
  "compat": {
    "cc": {
      "add": [ /* array of CC IDs to be added */ ]
      "remove": [ /* array of CC IDs to be removed */ ]
    }
  }
}
```
I think some OZW configs add CCs aswell - does this information include the CC version?

maintenance/importConfig.ts
robertsLando: I have started using it, but seems there are files with same man, prod type and id but different names :confused:

robertsLando:
@AlCalzone Do you mean: 

```xml
 <CommandClass action="add" id="113">
    <Compatibility>
      <GetSupported>false</GetSupported>
    </Compatibility>
  </CommandClass>
```

This?

> I think some OZW configs add CCs aswell - does this information include the CC version?

If so, nope, there is no version specified

robertsLando:
@AlCalzone Check it now

AlCalzone:
Codewise LGTM. Looking at the config files, there are a couple of small TODOs left:

Partial params should be used instead of normal config params, not in addition to them.
For example (`packages\config\config\devices\0x000c\hs-fc200.json`):
```json
        "31": {
            "label": "LEDs to blink in status mode",
            "description": "Sets LEDs 1-4 to blink in Status mode. Bitmask defines specific LEDs to enable for blinking. Note: this decimal value is derived from a hex code calculation based on the following: \n              Bit 0 = LED 1\n              Bit 1 = LED 2\n              Bit 2 = LED 3\n              Bit 3 = LED 4\n              IE: value of 1 = first LED, 8 = LED 4",
            "valueSize": 1,
            "minValue": 0,
            "maxValue": 15,
            "defaultValue": 0,
            "readOnly": false,
            "writeOnly": false,
            "allowManualEntry": true
        },
        "31[0x01]": {
            "label": "LED 1 Blink Status",
            "description": "LED 1 Blink Status (bottom)",
            "valueSize": 1,
            "minValue": 0,
            "maxValue": 1,
            "defaultValue": 0,
            "readOnly": false,
            "writeOnly": false,
            "allowManualEntry": false,
            "options": [
                {
                    "label": "Solid",
                    "value": 0
                },
                {
                    "label": "Blink",
                    "value": 1
                }
            ]
        },
```

---

It is a bit unfortunate that comments are not preserved when merging the config files, but I see no way of fixing that when parsing and serializing JSON. Maybe an enhancement for a future PR.

---

Do you sort the `devices` array somehow? `packages\config\config\devices\0x0002\lc-13.json` for example. I would prefer sorting by productType, then by productId.

---

Some devices have a space at the start or end of their description, e.g.
`packages\config\config\devices\0x0175\mt2651.json`

robertsLando:
> Partial params should be used instead of normal config params, not in addition to them.

Totally missed this, need to fix it

> It is a bit unfortunate that comments are not preserved when merging the config files, but I see no way of fixing that when parsing and serializing JSON. Maybe an enhancement for a future PR.

I would add a json field for that, like "comments" or "info". I could read the file as a plain text, read all first lines that start with a // and add that to "comments" in the json


> Do you sort the devices array somehow

Nope, TODOs

> Some devices have a space at the start or end of their description

Need to trim it so :)

AlCalzone:
> I would add a json field for that, like "comments" or "info". I could read the file as a plain text, read all first lines that start with a // and add that to "comments" in the json

I meant the specific comments like for the compat options (see the file I linked in https://github.com/zwave-js/node-zwave-js/pull/1093#issuecomment-724098925) for an example. 
As I said, I can just be on the lookout for that when merging PRs. It is not that critical.

robertsLando:
@AlCalzone Should be ready now

robertsLando:
TODOs:

- Override reported endpoints, and `MapRootToEndpoint` flag (??): 

```xml
<CommandClass id="96">
    <Compatibility>
      <MapRootToEndpoint>true</MapRootToEndpoint>
    </Compatibility>
    <Instance index="1" label="Load" />
    <Instance index="2" label="i2" />
    <Instance index="3" label="i3" />
    <Instance index="4" label="Temp Sensor" />
</CommandClass>
```

Compatibility flags: https://github.com/OpenZWave/open-zwave/blob/master/config/device_configuration.xsd#L73

Examples:

- `ForceInstances` flag (wtf is this?) in multi channel associations

```xml
<CommandClass id="142">
		<Compatibility>
			<ForceInstances>true</ForceInstances>
		</Compatibility>
	</CommandClass>
	<!-- Remove COMMAND_CLASS_BASIC -->
	<CommandClass id="32" action="remove" />
```

. `GetSupported` flag

```xml
  <!-- the Device Doesn't support Get for Alarm Class -->
<CommandClass id="113">
    <Compatibility>
      <GetSupported>false</GetSupported>
    </Compatibility>
  </CommandClass>
```

- Override some command classes:

```xml
 <CommandClass id="64">
    <Value genre="user" index="0" instance="1" label="Mode" max="0" min="0" read_only="false" size="1" type="list" units="" value="0" write_only="false">
      <Item label="Off" value="0"/>
      <Item label="Heat (Default)" value="1"/>
      <Item label="Cool" value="2"/>
      <Item label="Energy Heat" value="11"/>
    </Value>
    <SupportedModes>
      <Mode index="0" label="Off"/>
      <Mode index="1" label="Heat"/>
      <Mode index="2" label="Cool"/>
      <Mode index="11" label="Energy Heat"/>
    </SupportedModes>
    <Compatibility/>
  </CommandClass>
  <CommandClass id="67">
    <Instance index="1"/>
    <Value genre="user" index="1" instance="1" label="Heating" max="40" min="5" read_only="false" type="decimal" units="C" value="20" write_only="false"/>
    <Value genre="user" index="2" instance="1" label="Cooling" max="40" min="5" read_only="false" type="decimal" units="C" value="21" write_only="false"/>
    <Value genre="user" index="11" instance="1" label="Energy Heat" max="40" min="5" read_only="false" type="decimal" units="C" value="16" write_only="false"/>
    <Compatibility>
      <Base>0</Base>
      <CreateVars>true</CreateVars>
    </Compatibility>
  </CommandClass>
```

- Eventually parse metadata:

```xml
  <MetaDataItem name="OzwInfoPage">http://www.openzwave.com/device-database/019B:0001:0001</MetaDataItem>
    <MetaDataItem name="ProductPic">images/thermofloor/heatit021.png</MetaDataItem>
 <MetaDataItem name="ExclusionDescription">INCLUSION/EXCLUSION
To include the thermostat to your home automation gateway, press Center (confirm) for 10 seconds.
The display will show OFF. Press Right (down) 4 times till you see Con in the display.
Now start add device in your home automation software. Start inclusion mode by pressing Center (confirm) for approximately
2 seconds. The inclusion/exclusion icon will appear in the display. Confirmation will show Inc/EcL in the display. If inclusion
fails, Err (error) will appear.
Leave programming mode by choosing ESC in menu. Your thermostat is ready for use with default settings.</MetaDataItem>
    <MetaDataItem name="ProductManual">https://Products.Z-WaveAlliance.org/ProductManual/File?folder=&amp;filename=Manuals/1234/Manual_Multireg Z-Wave_Ver2015-A.pdf</MetaDataItem>
    <MetaDataItem id="0001" name="FrequencyName" type="0001">CEPT (Europe)</MetaDataItem>
    <MetaDataItem name="Name">Heatit Z-Wave</MetaDataItem>
    <MetaDataItem id="0001" name="Identifier" type="0001">TF021</MetaDataItem>
    <MetaDataItem name="ResetDescription">FACTORY RESET - RES
By pressing buttons Left and Center (up and confirm) for 20 seconds, the thermostat will perform a complete factory reset.
NB! Please use this procedure only when the primary controller is missing or otherwise inoperable.</MetaDataItem>
    <MetaDataItem name="InclusionDescription">INCLUSION/EXCLUSION
To include the thermostat to your home automation gateway, press Center (confirm) for 10 seconds.
The display will show OFF. Press Right (down) 4 times till you see Con in the display.
Now start add device in your home automation software. Start inclusion mode by pressing Center (confirm) for approximately
2 seconds. The inclusion/exclusion icon will appear in the display. Confirmation will show Inc/EcL in the display. If inclusion
fails, Err (error) will appear.
Leave programming mode by choosing ESC in menu. Your thermostat is ready for use with default settings.</MetaDataItem>
    <MetaDataItem name="WakeupDescription">If Er2 appears in the display, touch any button to restart the display.</MetaDataItem>
```

maintenance/importConfig.ts
```diff
@@ -589,63 +595,64 @@ async function parseOzwProduct(
...
+		// don't use ?? here, some fields could be empty strings and ?? operator
+		// will not work
+		parsedParam.label = ensureArray(param.label)[0] || parsedParam.label;
+		parsedParam.description =
+			ensureArray(param.Help)[0] || parsedParam.description;
+		parsedParam.valueSize = param.size || parsedParam.valueSize || 1;
+		parsedParam.minValue = param.min || parsedParam.min || 0;
+		parsedParam.maxValue = isBitSet
+			? param.bitmask
+			: param.max || parsedParam.max || 100;
```
AlCalzone: In theory it could be possible that a param has a negative min value and a max value of `0`. Shouldn't we use `??` here?

maintenance/importConfig.ts
```diff
@@ -589,63 +595,64 @@ async function parseOzwProduct(
...
-			}
+		if (typeof parsedParam.defaultValue !== "number") {
+			parsedParam.defaultValue = 0;
+		}
 
-			const items = ensureArray(param.Item);
+		// some values have typos, this fixes them
+		if (typeof parsedParam.maxValue === "string") {
+			parsedParam.maxValue = parseInt(
+				parsedParam.maxValue.replace(/[^0-9]/g, ""),
```
AlCalzone: 1. This changes negative values to their absolute value
2. Don't we need to do this for minValue too?

packages/config/src/utils.ts
```diff
@@ -48,3 +48,7 @@ export function formatId(id: number | string): string {
 export function padVersion(version: string): string {
 	return version + ".0";
 }
+
+export function stringify(data: any, beautify: boolean = true): string {
+	return JSON.stringify(data, undefined, beautify ? 4 : 0);
+}
```
AlCalzone: That method already exists here: `packages\shared\src\strings.ts`
You can add the `beautify` param there if you need it.

AlCalzone (changes requested):
Sorry, one more round - should be the last one though 😅

AlCalzone:
You can leave the TODOs for another PR

maintenance/importConfig.ts
robertsLando: What about use `isUndefined` method that checks value is not `null` `undefined` or `''`?

maintenance/importConfig.ts
robertsLando: Will fix this

robertsLando:
> You can leave the TODOs for another PR

TODOs need some thoughts with you to understand where and how to add those informations

robertsLando:
@AlCalzone Hope it's almost ready now :rofl:

robertsLando:
I have noticed some weird things: 

- OZW values (min max and default) are always unsigned
- The param sorting function doesn't always sort in the correct order :thinking:

AlCalzone (approved):

AlCalzone:
I have a couple of small things in mind I want to change (mainly formatting) when merging. I'll check the open points while I'm at it.

AlCalzone:
>     * OZW values (min max and default) are always unsigned

I forgot - zwave-js tries to be smarter here. Nothing to worry about. Do we add `unsigned: true` for the imports?

robertsLando:
> Do we add unsigned: true for the imports

Nope, should I do?

robertsLando:
> The param sorting function doesn't always sort in the correct order

About this, I have found a typo in the code and fixed that, the keys now are sorted in the correct order but the result file still parse the params in a wrong order (sometimes). The zw100.json is an example file that has this problem

AlCalzone:
> Nope, should I do?

Yes, please. Otherwise the config linter might complain.

> the result file still parse the params in a wrong order (sometimes)

I'll take a look at that later, if you don't find the reason.

robertsLando:
> I'll take a look at that later, if you don't find the reason.

Haven't been able to find out wtf is going wrong. As I said the sort algorithm sorts the paramInformations keys correctly, then i iterate them in a `for of` loop to assign the values. Dunno what breaks the order

AlCalzone:
> Dunno what breaks the order

Does this only affect merged config files or also completely new ones?

robertsLando:
> Does this only affect merged config files or also completely new ones?

I have saw this on a merged config, cannot find a new one with the same problem but there could be maybe

AlCalzone:
If it only affects existing config you might have to delete and re-add the Params in the sorted order.

robertsLando:
It's not that the problem as I completely recreate the object from start in normalizeConfig function

AlCalzone:
Seems like there's nothing we can do about the sort order, except maybe represent numeric keys as hex too:

https://stackoverflow.com/a/23202095/10179833

> 1. Integer keys in ascending order (and strings like "1" that parse as ints)
> 2. String keys, in insertion order (ES2015 guarantees this and all browsers comply)
> 3. Symbol names, in insertion order (ES2015 guarantees this and all browsers comply)

robertsLando:
> Seems like there's nothing we can do about the sort order, except maybe represent numeric keys as hex too:

:cry:

AlCalzone:
I'm in the process of merging, please don't add anything else :)

robertsLando:
Sorry just synced the master of the fork with the main one. No more commits now :)

AlCalzone:
I've identified a couple of issues I've fixed. Once the tests pass, I'll merge. Although I'm a bit scared of the config update, that will be quite a bit of work.

robertsLando:
> I've identified a couple of issues I've fixed. Once the tests pass, I'll merge. Although I'm a bit scared of the config update, that will be quite a bit of work.

Let me know if I can help you somehow :) also @chrisns may help with that when he can

PR #1112, opened 2020-11-11
Import OZW config files

AlCalzone:


AlCalzone:
@robertsLando @chrisns I could actually use some help here... I've imported all the new files and fixed the errors, but reviewing the updated files is going to be a heap of work. To help:
1. check out this branch
2. run `npm run config -- import -s ozw -Dmid` (leave out `D` (download) for each subsequent call)
3. review the config file changes one by one
4. run `lerna run lint:config` to check for errors and fix them (the list is gonna be massive). If this shows a file not found error, run `npm run config -- import -i` first
5. create a PR on this branch with your changes

We should probably each choose a directory randomly and delete the others (between step 2 and 3), so we don't do the same work twice.

---

## Regarding 3:

I'd like the changes to be unobtrusive and to make sense. Some examples:

This is fine: ✔
grafik

This is NOT: ❌
grafik

This is fine: ✔
grafik
although `unsigned: true` is not necessary here.

This is not so nice because the original label is more comprehensible: ❌
grafik

This is unnecessary (we don't need the associations key if there is only a single lifeline association and the device supports Z-Wave+): ❌
grafik

Comments are removed sadly, please re-add them if you see that: ❌
grafik

This device label seems wrong: ❌
grafik

... you get the point.

---

## Regarding 4:

When you see an error, ctrl+click on the path to open the config file directly

OZW config files are notoriously bad when it comes to correct valueSizes in combination with the min/max/default values. I often see `-1` used in place of `255`.
This is bullshit:
grafik
**Correct**: min = 0, max = 255, default = 0, the option value should be 255, there should be a second (default) option

Another example:
grafik
**Correct:** valueSize = 4, probably remove `unsigned: true`

And another one:
grafik
**Correct:** valueSize = 2

If you don't know the correct size, make a best-effort guess (only 1, 2 or 4 - 3 is not allowed by the specs, but some devices still use it).

grafik
Some devices send their serial number (or whatever) using a config param with size 16. Remove that, leave a comment.

grafik
Check if these config files really target the same device (or a different variant of it). If so, add the devices array to one of the files (preferably the better one), delete the other.

grafik
One of the labels is wrong or should be more specific.

robertsLando:
@AlCalzone Before anything else, I would suggest to add revision to config files. With the revision number we can check (at least for ozw configs) if a specific config needs to be parsed or not

robertsLando:
ALso wouldn't it be better to allow us to make changes directly on this branch?

AlCalzone:
> Also wouldn't it be better to allow us to make changes directly on this branch?

I don't know how that works 🤷‍♂️

> Before anything else, I would suggest to add revision to config files

Agreed

robertsLando:
@AlCalzone Could you do the import again now with the revision? 

> I don't know how that works

You should add us as maintainers of the library

AlCalzone:
I'll try, it will require a bit of stash/rebase fun

chrisns:
Stupid question is no way to dynamically pull and codify these changes either as rules or patches. rather than having to manually tidy them and repeat when there's ozw changes?

AlCalzone:
The already-added config files have a revision now. @chrisns @robertsLando is planning to create an action that only changes updated files. That should limit the manual work in each PR. 
I'm not sure what burns more time - reviewing the automatic PRs or creating a semi-automatic fix. If you have a good idea, I'm all ears. 
But especially for the in-file comments I'm unsure. I want to preserve them because they are valuable, but I don't know how.

AlCalzone:
I **think** you guys should have push access to this branch now

robertsLando:
@AlCalzone what about adding to the top comment also a link to the original xml file in ozw repo? 

I say this because I saw some strange errors in xml2json parsing and when things are too strange I would like to check the original file to see if there are parsing errors

chrisns:
I don't want to piss on your bonfire here because this is all great work but i feel you may have a license conflict here with a mix of this repo on MIT, openhab on eclipse, and ozw on gplv3.
Has that been thought through?
GPLv3 iirc stipulates we can reuse but only if we keep the same license, not sure about eclipse.

AlCalzone:
> you may have a license conflict

I'm no lawyer or anything, but as far as I understand, copyright licenses protect expression, not ideas or data. And config files are generally not expressive enough to be protected by copyright.
Also consider that the config files (whichever source they are based on), are basically copies from the devices' user manuals.

robertsLando:
@AlCalzone if you didn't know, this kind of things are @chrisns stuffs 🤣 

Jokes aside, I think that it's important to consider license too to prevent any kind of problems in future. @chrisns what would you suggest?

robertsLando:
Question:

Are config files considered part of the code? 🤔 


Could we mix licenses somehow? Like just say that some config are imported from a GPLv3 project? 

We don't really use anything on code side that's imported from others (I mean, the hard part is entirely MIT). Also most of the config files are provided by the community

chrisns:
If we agree it's something we don't want to run the risk of defending later irregardless of if we're right or not:
Either:
- split into separate repos and depend on them
- pull like any other library at build and do the conversion on the fly at runtime

chrisns:
It's not an issue of ownership
Arguably the community submitted their config on the basis of the licence of the repo they made a PR to. So I'd put forward it's not playing nice to take the value of that work and change the licence terms.

chrisns:
> Are config files considered part of the code? 🤔

i'd argue they are intellectual property, so code or otherwise, they can be owned. and often may have been reverse engineered etc, so maybe not just taken from manual.

I'd put forward the device library is more than config, and is code by a dictionary definition

AlCalzone:
> i'd argue they are intellectual property

I don't believe that the majority of config files surpasses the threshold of originality. Sure there might be some exceptions, but where do we draw the line? Most config files are not usable for us without transformation and major modification.

> I'd put forward the device library is more than config, and is code by a dictionary definition

I know GPLv3 is not in discussion here, but going by that definition

> The “Corresponding Source” for a work in object code form means all the source code needed to generate, install, and (for an executable work) run the object code and to modify the work, including scripts to control those activities.

I'd argue that the config files are not part of the code. The library can be built and run perfectly fine without those. Arguably, the experience is not great, but it works.

---

I'm leaning towards not doing anything, but just lets discuss this a bit more.

**1) split into separate repos and depend on them**

How would you structure this? One for OH, one for OZW? Or just a separate one for the configurations? Which license? How do we deal with the additional overhead of synching changes in 2/3 repos?

**2) pull like any other library at build and do the conversion on the fly at runtime**

You mean when installing the library? Where's the difference then? Also, this is going to take minutes on each install and introduce many sources of errors. Some of the changes we need to do require a good understanding of how it all comes together - or quite an extensive ruleset for automatic fixing.
I don't really see that happening.

**3) attribute each imported configuration file with the corresponding license**

Could that be an option? How would we go about the files that are "the best of both worlds"? How about the already existing ones? Is there a point where our modifications are enough that we can say 
grafik

**4) remove information from the config files that does surpass the threshold of originality**

Assuming there is a config param somewhere in the config files that has been found by reverse engineering and a community member does not want us to have that information - I'd be okay with removing it on request.

**5) start from scratch and add config files manually**

For example by a motivated community member who happens to write an import script alone in his room. How could we be sure that it is not taken from another project with a different license? (seriously, no matter what we decide, this is relevant in the future)

robertsLando:
> Most config files are not usable for us without transformation and major modification.

I think that's the main point of the discussion, also ww could place a source link to the original file, sincerly I don't think this could cause any license problem in future, but we should ask to a lawyer or someone expert in this field as those are just suppositions. 

Finally this is OSS project like z2m is and I don't think anyone will cause problems to us for this, just someone who really hate us for some reason could 👀

robertsLando:
I would say one more (last) thing about this. 

If we wouldn't import those files, we should write them one by one, the result would be almost the same as config files are as they are, I mean once you have a config file structure they don't have many different ways of implementation, that is the only one, so even without importing those files the result is the same but will just take some much more (useless) time (and anyway someone could say we have copied them from ozw or elsewhere and who can tell them we have wrote them by our own?)

AlCalzone:
Ok, lets come out ahead and discuss with Justin what he thinks of collaboration.

robertsLando:
@AlCalzone Did you imported only new config files in this pr? I saw only around 200 files here


Could I submit a pr with the source link too? Should I import them all?

robertsLando:
ALso about your comment for config reviews: 

> This is not so nice because the original label is more comprehensible: x
> This is unnecessary (we don't need the associations key if there is only a single lifeline association and the device supports Z-Wave+): x

I think that we could skip this checks and focus more on the min/max/default values as I think them are the most errored

AlCalzone:
Yeah, I only added the new files here because they don't need to be merged. They would need the source link too though.

I didn't want to add all the others in a bulk because this makes it harder to wade through the lint errors than adding them one directory at a time.

But I guess we should wait a bit for a response from Justin.

robertsLando:
> But I guess we should wait a bit for a response from Justin.

Yep not working on it now, was JTK

PR #1117, opened 2020-11-13
feat(controller): implement replaceFailedNode API

### LINKED ISSUE #1082: [feat] Replace failed node api

robertsLando:
Fixes #1082 

@AlCalzone Im't stucked now on the controller function :thinking: 

How do I get the node? How can I specify if I want to replace the node in secure mode or not? Is the inclusion started automatically?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules/
+/packages/*/package-lock.json
 
 # compiled test-files
 *.test.js
@@ -11,11 +12,10 @@ zwave-js-*.tgz
 # ZWave cache
 **/cache
 
+# Temporary (import) config files
+.tmp
+
 # Compiled source files
 **/build
 **/*.tsbuildinfo
 *.map
-
-.tmp
-
-/packages/*/package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,4 @@ zwave-js-*.tgz
 
 .tmp
 
-/**/package-lock.json
+/packages/*/package-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ zwave-js-*.tgz
 *.map
 
 .tmp
+
+/**/package-lock.json
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -948,11 +948,16 @@ export class ZWaveController extends EventEmitter {
 				log.controller.print(
 					`inclusion started, node is ready to be replaced`,
 				);
+				this.emit("inclusion started", !this._includeNonSecure);
 				this._inclusionActive = true;
 				this._replaceFailedPromise?.resolve(true);
-				break;
+
+				// stop here, don't emit incluson failed
+				return true;
 			case ReplaceFailedNodeStatus.FailedNodeReplaceDone:
 				log.controller.print(`node successfully replaced`);
+
+				this.emit("inclusion stopped");
 
 				if (this._nodePendingReplace !== undefined) {
 					this.emit("node removed", this._nodePendingReplace);
@@ -972,6 +977,9 @@ export class ZWaveController extends EventEmitter {
 					this.emit("node added", newNode);
 				}
 
+				// stop here, don't emit incluson failed
+				return true;
+
 			default:
 				this._replaceFailedPromise?.reject(
 					new ZWaveError(
@@ -981,7 +989,9 @@ export class ZWaveController extends EventEmitter {
 				);
 		}
 
-		return true; // Don't invoke any more handlers
+		this.emit("inclusion failed");
+
+		return false; // Don't invoke any more handlers
 	}
 
 	/**
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -908,9 +908,9 @@ export class ZWaveController extends EventEmitter {
 	 * Is called when an ReplaceFailed request is received from the controller.
 	 * Handles and controls the replace process.
 	 */
-	private async handleReplaceNodeRequest(
+	private handleReplaceNodeRequest(
 		msg: ReplaceFailedNodeRequestStatusReport,
-	): Promise<boolean> {
+	): boolean {
 		log.controller.print(
 			`handling replace node request (status = ${
 				ReplaceFailedNodeStatus[msg.replaceStatus]
@@ -926,10 +926,26 @@ export class ZWaveController extends EventEmitter {
 					),
 				);
 				break;
+			case ReplaceFailedNodeStatus.FailedNodeRemoved:
+				this._replaceFailedPromise?.reject(
+					new ZWaveError(
+						`The failed node was removed from the failed nodes list`,
+						ZWaveErrorCodes.ReplaceFailedNode_Failed,
+					),
+				);
+				break;
+			case ReplaceFailedNodeStatus.FailedNodeNotRemoved:
+				this._replaceFailedPromise?.reject(
+					new ZWaveError(
+						`The failed node was not removed from the failing nodes list`,
+						ZWaveErrorCodes.ReplaceFailedNode_Failed,
+					),
+				);
+				break;
 			case ReplaceFailedNodeStatus.FailedNodeReplaceFailed:
 				this._replaceFailedPromise?.reject(
 					new ZWaveError(
-						`The replace process could not be completed`,
+						`The failed node has not been replaced`,
 						ZWaveErrorCodes.ReplaceFailedNode_Failed,
 					),
 				);
@@ -937,12 +953,14 @@ export class ZWaveController extends EventEmitter {
 			case ReplaceFailedNodeStatus.FailedNodeReplace:
 				// failed node is now ready to be replaced and controller is ready to add a new
 				// node with the nodeID of the failed node
-				await this.beginInclusion(this._includeNonSecure);
+				log.controller.print(
+					`inclusion started, node is ready to be replaced`,
+				);
+				this._inclusionActive = true;
+				this._replaceFailedPromise?.resolve(true);
 				break;
 			case ReplaceFailedNodeStatus.FailedNodeReplaceDone:
 				log.controller.print(`node successfully replaced`);
-
-				this._replaceFailedPromise?.resolve(true);
 
 			default:
 				this._replaceFailedPromise?.resolve(false);
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -562,6 +562,7 @@ export class ZWaveController extends EventEmitter {
 	private _includeController: boolean = false;
 	private _nodePendingInclusion: ZWaveNode | undefined;
 	private _nodePendingExclusion: ZWaveNode | undefined;
+	private _nodePendingReplace: ZWaveNode | undefined;
 	// The following variables are to be used for inclusion AND exclusion
 	private _beginInclusionPromise: DeferredPromise<boolean> | undefined;
 	private _stopInclusionPromise: DeferredPromise<boolean> | undefined;
@@ -689,6 +690,62 @@ export class ZWaveController extends EventEmitter {
 			log.controller.print(`the exclusion process was stopped`);
 			this.emit("exclusion stopped");
 		});
+	}
+
+	private async secureBootstrap(node: ZWaveNode): Promise<void> {
+		// If security has been set up and we are allowed to include the node securely, do it
+		if (
+			this.driver.securityManager &&
+			!this._includeNonSecure &&
+			node.supportsCC(CommandClasses.Security)
+		) {
+			// Only try once, otherwise the node stays unsecure
+			try {
+				// SDS13783 - impose a 10s timeout on each message
+				const api = node.commandClasses.Security.withOptions({
+					expire: 10000,
+				});
+				// Request security scheme, because it is required by the specs
+				await api.getSecurityScheme(); // ignore the result
+
+				// Request nonce separately, so we can impose a timeout
+				await api.getNonce({
+					standalone: true,
+					storeAsFreeNonce: true,
+				});
+
+				// send the network key
+				await api.setNetworkKey(this.driver.securityManager.networkKey);
+
+				if (this._includeController) {
+					// Tell the controller which security scheme to use
+					await api.inheritSecurityScheme();
+				}
+
+				// Remember that the node is secure
+				node.isSecure = true;
+			} catch (e: unknown) {
+				let errorMessage = `Security bootstrapping failed, the node is included insecurely`;
+				if (!(e instanceof ZWaveError)) {
+					errorMessage += `: ${e as any}`;
+				} else if (
+					e.code === ZWaveErrorCodes.Controller_MessageExpired
+				) {
+					errorMessage += ": a secure inclusion timer has elapsed.";
+				} else if (
+					e.code !== ZWaveErrorCodes.Controller_MessageDropped &&
+					e.code !== ZWaveErrorCodes.Controller_NodeTimeout
+				) {
+					errorMessage += `: ${e.message}`;
+				}
+				log.controller.logNode(node.id, errorMessage, "warn");
+				// Remember that the node is non-secure
+				node.isSecure = false;
+			}
+		} else {
+			// Remember that the node is non-secure
+			node.isSecure = false;
+		}
 	}
 
 	/**
@@ -824,72 +881,7 @@ export class ZWaveController extends EventEmitter {
 					this._nodes.set(newNode.id, newNode);
 					this._nodePendingInclusion = undefined;
 
-					// If security has been set up and we are allowed to include the node securely, do it
-					if (
-						this.driver.securityManager &&
-						!this._includeNonSecure &&
-						newNode.supportsCC(CommandClasses.Security)
-					) {
-						// Only try once, otherwise the node stays unsecure
-						try {
-							// SDS13783 - impose a 10s timeout on each message
-							const api = newNode.commandClasses.Security.withOptions(
-								{
-									expire: 10000,
-								},
-							);
-							// Request security scheme, because it is required by the specs
-							await api.getSecurityScheme(); // ignore the result
-
-							// Request nonce separately, so we can impose a timeout
-							await api.getNonce({
-								standalone: true,
-								storeAsFreeNonce: true,
-							});
-
-							// send the network key
-							await api.setNetworkKey(
-								this.driver.securityManager.networkKey,
-							);
-
-							if (this._includeController) {
-								// Tell the controller which security scheme to use
-								await api.inheritSecurityScheme();
-							}
-
-							// Remember that the node is secure
-							newNode.isSecure = true;
-						} catch (e: unknown) {
-							let errorMessage = `Security bootstrapping failed, the node is included insecurely`;
-							if (!(e instanceof ZWaveError)) {
-								errorMessage += `: ${e as any}`;
-							} else if (
-								e.code ===
-								ZWaveErrorCodes.Controller_MessageExpired
-							) {
-								errorMessage +=
-									": a secure inclusion timer has elapsed.";
-							} else if (
-								e.code !==
-									ZWaveErrorCodes.Controller_MessageDropped &&
-								e.code !==
-									ZWaveErrorCodes.Controller_NodeTimeout
-							) {
-								errorMessage += `: ${e.message}`;
-							}
-							log.controller.logNode(
-								newNode.id,
-								errorMessage,
-								"warn",
-							);
-							// Remember that the node is non-secure
-							newNode.isSecure = false;
-						}
-					} else {
-						// Remember that the node is non-secure
-						newNode.isSecure = false;
-					}
-
+					await this.secureBootstrap(newNode);
 					this._includeController = false;
 
 					// We're done adding this node, notify listeners
@@ -908,9 +900,9 @@ export class ZWaveController extends EventEmitter {
 	 * Is called when an ReplaceFailed request is received from the controller.
 	 * Handles and controls the replace process.
 	 */
-	private handleReplaceNodeRequest(
+	private async handleReplaceNodeRequest(
 		msg: ReplaceFailedNodeRequestStatusReport,
-	): boolean {
+	): Promise<boolean> {
 		log.controller.print(
 			`handling replace node request (status = ${
 				ReplaceFailedNodeStatus[msg.replaceStatus]
@@ -961,6 +953,24 @@ export class ZWaveController extends EventEmitter {
 				break;
 			case ReplaceFailedNodeStatus.FailedNodeReplaceDone:
 				log.controller.print(`node successfully replaced`);
+
+				if (this._nodePendingReplace !== undefined) {
+					this.emit("node removed", this._nodePendingReplace);
+					this._nodes.delete(this._nodePendingReplace.id);
+
+					const newNode = new ZWaveNode(
+						this._nodePendingReplace.id,
+						this.driver,
+					);
+
+					this._nodes.set(newNode.id, newNode);
+					this._nodePendingReplace = undefined;
+
+					await this.secureBootstrap(newNode);
+
+					// We're done adding this node, notify listeners
+					this.emit("node added", newNode);
+				}
 
 			default:
 				this._replaceFailedPromise?.reject(
@@ -1961,6 +1971,7 @@ ${associatedNodes.join(", ")}`,
 			);
 		} else {
 			this._replaceFailedPromise = createDeferredPromise();
+			this._nodePendingReplace = this.nodes.get(nodeId);
 			return this._replaceFailedPromise;
 		}
 	}
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -565,7 +565,7 @@ export class ZWaveController extends EventEmitter {
 	// The following variables are to be used for inclusion AND exclusion
 	private _beginInclusionPromise: DeferredPromise<boolean> | undefined;
 	private _stopInclusionPromise: DeferredPromise<boolean> | undefined;
-	private _replaceFailedPromise: DeferredPromise<void> | undefined;
+	private _replaceFailedPromise: DeferredPromise<boolean> | undefined;
 
 	/**
 	 * Starts the inclusion process of new nodes.
@@ -957,7 +957,7 @@ export class ZWaveController extends EventEmitter {
 					`inclusion started, node is ready to be replaced`,
 				);
 				this._inclusionActive = true;
-				this._replaceFailedPromise?.resolve();
+				this._replaceFailedPromise?.resolve(true);
 				break;
 			case ReplaceFailedNodeStatus.FailedNodeReplaceDone:
 				log.controller.print(`node successfully replaced`);
@@ -1906,7 +1906,10 @@ ${associatedNodes.join(", ")}`,
 	public async replaceFailedNode(
 		nodeId: number,
 		includeNonSecure: boolean = false,
-	): Promise<void> {
+	): Promise<boolean> {
+		// don't start it twice
+		if (this._inclusionActive || this._exclusionActive) return false;
+
 		log.controller.print(`starting replace failed node process...`);
 
 		this._includeNonSecure = includeNonSecure;
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -986,14 +986,6 @@ export class ZWaveController extends EventEmitter {
 
 				// stop here, don't emit incluson failed
 				return true;
-
-			default:
-				this._replaceFailedPromise?.reject(
-					new ZWaveError(
-						`Unknown error ${msg.replaceStatus}`,
-						ZWaveErrorCodes.ReplaceFailedNode_Failed,
-					),
-				);
 		}
 
 		this.emit("inclusion failed");
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -1888,6 +1888,8 @@ ${associatedNodes.join(", ")}`,
 	/**
 	 * Replace a failed node from the controller's memory. If the process fails, this will throw an exception with the details why.
 	 * @param nodeId The id of the node to replace
+	 * @param includeNonSecure Whether the new node should be included non-securely, even if it supports Security. By default, all nodes will be included securely if possible
+	 *
 	 */
 	public async replaceFailedNode(
 		nodeId: number,
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -691,15 +691,28 @@ export class ZWaveController extends EventEmitter {
 		});
 	}
 
-	private async secureBootstrapS0(node: ZWaveNode): Promise<void> {
-		// If security has been set up and we are allowed to include the node securely, do it
+	private async secureBootstrapS0(
+		node: ZWaveNode,
+		assumeSecure: boolean = false,
+	): Promise<void> {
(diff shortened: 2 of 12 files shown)


AlCalzone:
From the specs I'm assuming that the ZW_FAILED_NODE_REPLACE status means that the controller is in inclusion mode now. I would expect the controller to send us AddNodeToNetwork requests. The inclusion is handled in `packages\zwave-js\src\lib\controller\Controller.ts`, line 747ff. If the new node is added like I am guessin, you can directly reuse that code for replacing the node. You might have to reuse `_beginInclusionPromise` and `_includeNonSecure` for the replacement process too.

robertsLando:
@AlCalzone Something like this?

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1870,23 +1873,24 @@ ${associatedNodes.join(", ")}`,
...
+					// node with the nodeID of the failed node
+					return this.beginInclusion(includeNonSecure);
 				default:
-					// If everything went well, the status is RemoveFailedNodeStatus.NodeRemoved
-
-					// TODO: wtf
-
-					return;
+					// If everything went well, the status is ReplaceFailedNodeStatus.FailedNodeReplaceDone
+					return true;
```
AlCalzone: We need to figure out when which message comes. If responses may be received multiple times (like during the inclusion process) we might need to create a deferred promise to be returned here.

~~I just remembered that I had a log of that from OZW - it might give you some pointers :)
==> https://github.com/zwave-js/node-zwave-js/blob/master/specs/OZW_Log__replace_node.txt~~

AlCalzone: Nevermind, that file doesn't actually include anything useful - I should have taken a look at it when I made it back then 😅

robertsLando: > We need to figure out when which message comes. If responses may be received multiple times (like during the inclusion process) we might need to create a deferred promise to be returned here. 

My zwave understanding at the low level is very low but I think that this could return twice: firstly returns when `beginInclusion` promise is fulfilled and then when we receive a response `ReplaceFailedNodeStatus.FailedNodeReplaceDone`

robertsLando: >  we might need to create a deferred promise to be returned here

You mean the beginInclusionPromise or a new replaceFailed promise?

robertsLando: From what I have understanded It could work like this:

---> ReplaceFailedNodeRequest
<--- ReplaceFailedNodeStatus.FailedNodeReplace
---> AddNodeToNetworkRequest
<--- ReplaceFailedNodeStatus.FailedNodeReplaceDone

So actually in the code the 

```js
const result = await this.driver.sendMessage<
			ReplaceFailedNodeRequestStatusReport | ReplaceFailedNodeResponse
		>(new ReplaceFailedNodeRequest(this.driver, { failedNodeId: nodeId }));
```

Will return twice?

AlCalzone: I think it might work like this:
```
---> ReplaceFailedNodeRequest (start the process)
<--- ReplaceFailedNodeStatus.FailedNodeReplace OR an error (you've handled the error case)
<--- AddNodeToNetworkRequest (MULTIPLE!)
<--- ReplaceFailedNodeStatus.FailedNodeReplaceDone OR Failed
```

The promise returned by the controller method should be rejected in case of an error (either before or during the replacing) or it should be resolved after the node was added (status ReplaceDone)

To be able to handle multiple incoming messages, please take a look at how I've handled the incoming AddNodeToNetworkRequests in https://github.com/zwave-js/node-zwave-js/blob/4dffb8de6796723604da87e26cc8f8bf8ab0b434/packages/zwave-js/src/lib/controller/Controller.ts#L147-L159

AlCalzone: Actually this would be perfect for [state machines](https://github.com/zwave-js/node-zwave-js/issues/884), but this is a whole other beast.

robertsLando:
@AlCalzone Like this? (check latest commit)

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -898,6 +904,64 @@ export class ZWaveController extends EventEmitter {
...
+							`The replace process could not be completed`,
+							ZWaveErrorCodes.ReplaceFailedNode_Failed,
+						),
+					);
+				}
+				break;
+			case ReplaceFailedNodeStatus.FailedNodeReplace:
+				// failed node is now ready to be replaced and controller is ready to add a new
+				// node with the nodeID of the failed node
+				await this.beginInclusion(msg.includeNonSecure);
```
AlCalzone: `_includeNonSecure` is a local variable of the controller, not a property of the received message. You should set it when starting the replacement process.

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -898,6 +904,64 @@ export class ZWaveController extends EventEmitter {
...
+
+				if (this._replaceFailedPromise != null) {
+					this._replaceFailedPromise.resolve(true);
+				}
+
+			default:
+				if (this._replaceFailedPromise != null) {
+					this._replaceFailedPromise.resolve(false);
+				}
+				return false;
```
AlCalzone: remove this line

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -898,6 +904,64 @@ export class ZWaveController extends EventEmitter {
...
+				await this.beginInclusion(msg.includeNonSecure);
+				break;
+			case ReplaceFailedNodeStatus.FailedNodeReplaceDone:
+				log.controller.print(
+					`node ${msg.failedNodeId} successfully replaced`,
+				);
+
+				if (this._replaceFailedPromise != null) {
+					this._replaceFailedPromise.resolve(true);
+				}
```
AlCalzone: This is nicer:
```
this._replaceFailedPromise?.resolve(true);
```

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -898,6 +904,64 @@ export class ZWaveController extends EventEmitter {
...
+
+			default:
+				if (this._replaceFailedPromise != null) {
+					this._replaceFailedPromise.resolve(false);
+				}
```
AlCalzone: dito

packages/zwave-js/src/lib/controller/ReplaceFailedNodeRequest.ts
```diff
@@ -68,6 +68,7 @@ export class ReplaceFailedNodeRequestBase extends Message {
 interface ReplaceFailedNodeRequestOptions extends MessageBaseOptions {
 	// This must not be called nodeId or rejectAllTransactions may reject the request
 	failedNodeId: number;
+	includeNonSecure: boolean;
```
AlCalzone: nope, that doesn't belong here

AlCalzone:
This looks like it could work with my changes. We need to test if the success of the replace promise needs to depend on the inclusion promise or if we always get a success/failure report back.

packages/zwave-js/src/lib/controller/Controller.ts
robertsLando: Humm msg is instance of `ReplaceFailedNodeRequest`, It's a property of the request if you check that, like `failedNodeId`

packages/zwave-js/src/lib/controller/ReplaceFailedNodeRequest.ts
robertsLando: I'm still not sure about this :confused:

robertsLando:
@AlCalzone 

```
z2m:App Zwave api call: replaceFailedNode [ 2, false ] +28s
14:03:11.736 CNTRLR   starting replace failed node process...
14:03:11.739 DRIVER   Send queue:                                                        (1 message)
                      · [REQ] ReplaceFailedNode
14:03:11.741 SERIAL » 0x0105006302019a                                                     (7 bytes)
14:03:11.741 DRIVER » [REQ] [ReplaceFailedNode]
                        payload: 0x0201
14:03:11.742 DRIVER   send thread state: sending.execute
  z2m:Mqtt MQTT client reconnecting +26s
14:03:37.373 SERIAL « [ACK]                                                                   (0x06)
14:03:37.375 DRIVER   send thread state: sending.execute
14:03:37.376 SERIAL « 0x010401630891                                                       (6 bytes)
14:03:37.377 SERIAL » [ACK]                                                                   (0x06)
14:03:37.379 DRIVER « [RES] [ReplaceFailedNode]
                        payload: 0x08
14:03:37.382 CNTRLR   No response from controller after 1/3 attempts. Scheduling next try in 100 ms.
14:03:37.383 DRIVER   send thread state: sending.execute
  z2m:Mqtt Connection refused: Not authorized +14ms
  z2m:Mqtt MQTT client closed +2ms
14:03:37.483 SERIAL » 0x0105006302019a                                                     (7 bytes)
14:03:37.484 DRIVER » [REQ] [ReplaceFailedNode]
                        payload: 0x0201
14:03:37.486 SERIAL « [ACK]                                                                   (0x06)
14:03:37.487 DRIVER   send thread state: sending.execute
14:03:37.488 SERIAL « 0x010401630891                                                       (6 bytes)
14:03:37.489 SERIAL » [ACK]                                                                   (0x06)
14:03:37.490 DRIVER « [RES] [ReplaceFailedNode]
                        payload: 0x08
14:03:37.491 CNTRLR   No response from controller after 2/3 attempts. Scheduling next try in 1100 ms
                      .
14:03:37.492 DRIVER   send thread state: sending.execute
14:03:38.595 SERIAL » 0x0105006302019a                                                     (7 bytes)
14:03:38.597 DRIVER » [REQ] [ReplaceFailedNode]
                        payload: 0x0201
14:03:38.603 SERIAL « [ACK]                                                                   (0x06)
14:03:38.607 DRIVER   send thread state: sending.execute
14:03:38.610 SERIAL « 0x010401630891                                                       (6 bytes)
14:03:38.611 SERIAL » [ACK]                                                                   (0x06)
14:03:38.616 DRIVER « [RES] [ReplaceFailedNode]
                        payload: 0x08
14:03:38.623 DRIVER   send thread state: sending.execute
14:03:38.627 DRIVER   Send queue: (empty)                                               (0 messages)
14:03:38.628 DRIVER   send thread state: idle
```

:cry:

.gitignore
```diff
@@ -17,3 +17,5 @@ zwave-js-*.tgz
 *.map
 
 .tmp
+
+/**/package-lock.json
```
AlCalzone: That does not include the one in the root, correct?
Maybe being explicit would be better: `packages/*/package-lock.json`

robertsLando: > That does not include the one in the root, correct?

Yes correct :) going to edit it to: `packages/**/package-lock.json`

packages/zwave-js/src/lib/controller/Controller.ts
robertsLando: > You should set it when starting the replacement process

It is handled when I call `beginInlcusion`

AlCalzone: The argument (msg) here is the message that is **received** from the controller (the ReplaceFailedNodeRequestStatusReport, `ZW->HOST: REQ | 0x63 | funcID | txStatus` in the specs). 
It won't have the property set.

packages/zwave-js/src/lib/controller/ReplaceFailedNodeRequest.ts
AlCalzone: See other comment above. The message classes do not care about this property, since it is not exchanged with the stick. The controller itself needs to remember that the node it is going to include should be included securely.

packages/zwave-js/src/lib/controller/ReplaceFailedNodeRequest.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+interface ReplaceFailedNodeRequestOptions extends MessageBaseOptions {
+	// This must not be called nodeId or rejectAllTransactions may reject the request
+	failedNodeId: number;
+	includeNonSecure: boolean;
+}
+
+@expectedResponse(FunctionType.ReplaceFailedNode)
+@expectedCallback(FunctionType.ReplaceFailedNode)
```
AlCalzone: If the controller actually sends multiple callbacks, you cannot use `@expectedCallback` to define it here.
See AddNodeToNetworkRequest.ts, line 51
This is okay, since you're handling these messages with the handler function you added to Controller.ts

.gitignore
AlCalzone: one * is enough, don't need to target subdirectories of the packages

AlCalzone:
> 14:03:37.382 CNTRLR   No response from controller after 1/3 attempts.

Weird - it did receive the response. Did you forget the @expectedResponse before creating the log?

robertsLando:
> Did you forget the @expectedResponse before creating the log?

What?

AlCalzone:
> 
> 
> > Did you forget the @expectedResponse before creating the log?
> 
> What?

That:
grafik

Did you have that in place before creating the log?

robertsLando:
@AlCalzone Sure. Actually I have the repo linked in zwavejs2mqtt, when I make some changes I run `npm run build` and relunch zwavejs2mqtt.

AlCalzone:
> when I make some changes I run `npm run build`

You could also keep `npm run watch` running, so your sources are automatically compiled on change.

robertsLando:
> You could also keep npm run watch running, so your sources are automatically compiled on change.

Didn't saw that, nice to know :) What about the error? I'm stucked :(

packages/zwave-js/src/lib/controller/ReplaceFailedNodeRequest.ts
robertsLando: Removed that

AlCalzone:
I'm not sure without debugging myself. For some reason the response is not detected as the expected one.
You could set a conditional breakpoint (`this instanceof ReplaceFailedNodeRequest`) in https://github.com/zwave-js/node-zwave-js/blob/4dffb8de6796723604da87e26cc8f8bf8ab0b434/packages/zwave-js/src/lib/message/Message.ts#L318 and see if you find the problem. If not, I can check later.

robertsLando:
> You could set a conditional breakpoint (this instanceof ReplaceFailedNodeRequest) in

I receive a message `msg instanceof ReplaceFailedNodeResponse`, don't know the flow now but maybe I forgot some reference somewhere?

packages/zwave-js/src/lib/controller/Controller.ts
robertsLando: Fixed

robertsLando:
@AlCalzone I haven't been able to find what's going on, as you said I see the response coming but I dunno why it's not 'catched', will let you debug this if you have time, will be back on monday otherwise

AlCalzone:
Ohhhhhhhh...
The error message is wrong. When retrying, it always logs "no response", even if the reason is a response that is NOK.

robertsLando:
AlCalzone feel free to push your commits here to fix it :)

AlCalzone:
I'll do it later

AlCalzone:
@robertsLando you should be good to go:
```
08:55:31.486 CNTRLR   starting replace failed node process...
08:55:33.289 DRIVER « [REQ] [SendData]
                        callback id:     1
                        transmit status: NoAck
08:55:33.293 DRIVER » [REQ] [ReplaceFailedNode]
                        payload: 0x1102
08:55:33.299 DRIVER « [RES] [ReplaceFailedNode]
                        payload: 0x10
08:55:33.300 CNTRLR   The controller response indicated failure after 1/3 attempts. Scheduling next 
                      try in 100 ms.
08:55:33.416 DRIVER » [REQ] [ReplaceFailedNode]
                        payload: 0x1102
08:55:33.420 DRIVER « [RES] [ReplaceFailedNode]
                        payload: 0x10
08:55:33.421 CNTRLR   The controller response indicated failure after 2/3 attempts. Scheduling next 
                      try in 1100 ms.
08:55:34.534 DRIVER » [REQ] [ReplaceFailedNode]
                        payload: 0x1102
08:55:34.539 DRIVER « [RES] [ReplaceFailedNode]
                        payload: 0x10
08:55:34.545 DRIVER » [Node 017] [REQ] [SendData]
                      │ transmit options: 0x25
                      │ callback id:      3
                      └─[NoOperationCC]
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
ZWaveError: The node replace process could not be started due to the following reasons:
· The node replace process is currently busy
```

robertsLando:
@AlCalzone Will try this on Monday, thanks! 🙏🏻  I'm using z2m to test, I suggest you to give it a try especially for testing it's faster then write examples files :)

AlCalzone:
I have a test job ready I can modify and start with F5 without compiling first.

robertsLando:
@AlCalzone I have tested this, but it has some strange behaviour... Could you give this a try? Actually it's replacing the node but the status report is not correct, I receive 3 (failed) and then if I press the inclusion button on the device it is included (but the inclusion isn't started :confused: )

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -891,6 +904,53 @@ export class ZWaveController extends EventEmitter {
...
+					new ZWaveError(
+						`The replace process could not be completed`,
+						ZWaveErrorCodes.ReplaceFailedNode_Failed,
+					),
+				);
+				break;
+			case ReplaceFailedNodeStatus.FailedNodeReplace:
+				// failed node is now ready to be replaced and controller is ready to add a new
+				// node with the nodeID of the failed node
+				await this.beginInclusion(this._includeNonSecure);
```
robertsLando: I'm not sure this is correct, I think that the controller is automatically set in inclusion mode, the question is how do I specify if the inclusion should be done securely or not?

AlCalzone: > how do I specify if the inclusion should be done securely or not?

You already did by setting the `_includeNonSecure` variable. Remove the call to `beginInclusion` and set the `inclusionActive` variable so a user can't start a second inclusion process in parallel.

I'm thinking if the replaceFailed promise should be resolved here instead of after the replacement. At least that would be in line with `beginInclusion` and the library user would have some kind of notification that the new node can be included now.

robertsLando: > I'm thinking if the replaceFailed promise should be resolved here instead of after the replacement. At least that would be in line with beginInclusion and the library user would have some kind of notification that the new node can be included now.

I think you right!

robertsLando: But anyway the reported status seems wrong

robertsLando: Operation: 

ReplaceFailedNode ---> reported status 3 (replace failed)
Put the device in inclusion mode --->  reported status 4 (unknown)

Restarting the controller results in the node to be added. I'm confused I think that the status report enum is wrong

AlCalzone: > But anyway the reported status seems wrong [...]
> Restarting the controller results in the node to be added

The stick has an internal state machine that easily gets confused, especially if you put it inclusion/exclusion mode multiple times. To be sure that your implementation is correct, I would pull the stick before every test.

packages/zwave-js/src/lib/controller/ReplaceFailedNodeRequest.ts
```diff
@@ -0,0 +1,138 @@
...
+        Meaning that the new node must now
+        emit a nodeinformation frame to be
+        included. */
+	FailedNodeReplace = 1,
+	/** The failed node has been replaced. */
+	FailedNodeReplaceDone = 2,
+
+	/** The replace process has failed */
+	FailedNodeReplaceFailed = 3,
+}
```
AlCalzone: You are right, the enum values are wrong (see the header files in `specs/api_includes`).
```h
/* ZW_ReplaceFailedNode callback status definitions */
#define ZW_FAILED_NODE_REPLACE              3 /* The failed node are ready to be replaced and controller */
                                              /* is ready to add new node with nodeID of the failed node */
#define ZW_FAILED_NODE_REPLACE_DONE         4 /* The failed node has been replaced */
#define ZW_FAILED_NODE_REPLACE_FAILED       5 /* The failed node has not been replaced */
```

robertsLando:
@AlCalzone I think we are almost done here, something is missing: 


```
13:53:33.474 DRIVER   handling request ReplaceFailedNode (99)
13:53:33.475 DRIVER     1 handler registered!
13:53:33.475 DRIVER     invoking handler #0
13:53:33.475 CNTRLR   handling replace node request (status = FailedNodeReplace)
13:53:40.180 CNTRLR   inclusion started, node is ready to be replaced
13:53:44.704 DRIVER       the message was handled
13:53:44.705 DRIVER   send thread state: idle
  z2m:Mqtt MQTT client reconnecting +14s
  z2m:App New connection nsck_h537JlSJb1DAAAB +11s
  z2m:Mqtt Connection refused: Not authorized +6ms
  z2m:App User disconnected fuy-s0owa4BxOL2CAAAA +9ms
  z2m:Mqtt MQTT client closed +4ms
  z2m:Mqtt MQTT client reconnecting +3s
  z2m:Mqtt Connection refused: Not authorized +2ms
  z2m:Mqtt MQTT client closed +1ms
  z2m:Mqtt MQTT client reconnecting +3s
  z2m:Mqtt Connection refused: Not authorized +2ms
  z2m:Mqtt MQTT client closed +1ms
13:53:53.329 SERIAL « 0x0105006303049e                                                     (7 bytes)
13:53:53.330 SERIAL » [ACK]                                                                   (0x06)
13:53:53.331 DRIVER « [REQ] [ReplaceFailedNode]
                        payload: 0x0304
13:53:53.332 DRIVER   Send queue: (empty)                                               (0 messages)
13:53:53.332 DRIVER   send thread state: idle
13:53:53.333 DRIVER   Send queue: (empty)                                               (0 messages)
13:53:53.334 DRIVER   handling request ReplaceFailedNode (99)
13:53:53.334 DRIVER     1 handler registered!
13:53:53.334 DRIVER     invoking handler #0
13:53:53.335 CNTRLR   handling replace node request (status = FailedNodeReplaceDone)
13:54:00.483 CNTRLR   node successfully replaced
13:54:00.484 DRIVER       the message was handled
13:54:00.484 DRIVER   send thread state: idle
  z2m:Mqtt MQTT client reconnecting +10s
  z2m:Mqtt Connection refused: Not authorized +3ms
  z2m:Mqtt MQTT client closed +1ms
13:54:03.306 CNTRLR   stopping inclusion process...
13:54:03.307 DRIVER   Send queue:                                                        (1 message)
                      · [REQ] AddNodeToNetwork
13:54:03.310 SERIAL » 0x0105004ac50471                                                     (7 bytes)
13:54:03.311 DRIVER » [REQ] [AddNodeToNetwork]
                        payload: 0xc504
13:54:03.312 DRIVER   send thread state: sending.execute
13:54:03.314 SERIAL « [ACK]                                                                   (0x06)
13:54:03.315 DRIVER   send thread state: sending.execute
13:54:03.316 DRIVER   Send queue: (empty)                                               (0 messages)
13:54:03.316 DRIVER   send thread state: idle
```

As you can see in the last part a `AddNodeToNetwork` request is sent but the `handleAddNodeRequest` handler isn't called, if I restart the controller the node is correctly added to the replaced id, I think I should send a NIF or something else when the replace is done?

packages/zwave-js/src/lib/controller/ReplaceFailedNodeRequest.ts
robertsLando: Fixed

AlCalzone:
```
13:53:53.334 DRIVER   handling request ReplaceFailedNode (99)
13:53:53.334 DRIVER     1 handler registered!
13:53:53.334 DRIVER     invoking handler #0
13:53:53.335 CNTRLR   handling replace node request (status = FailedNodeReplaceDone)
13:54:00.483 CNTRLR   node successfully replaced
13:54:00.484 DRIVER       the message was handled
13:54:03.306 CNTRLR   stopping inclusion process...
13:54:03.307 DRIVER   Send queue:                                                        (1 message)
                      · [REQ] AddNodeToNetwork
13:54:03.310 SERIAL » 0x0105004ac50471                                                     (7 bytes)
13:54:03.311 DRIVER » [REQ] [AddNodeToNetwork]
                        payload: 0xc504
13:54:03.314 SERIAL « [ACK]                                                                   (0x06)
```
I think we're almost there. The `AddNodeToNetwork` request is for stopping the inclusion process which doesn't seem to be necessary in this case.

Seems we need to react to the `FailedNodeReplaceDone` request and bootstrap the new node, essentially perform the secure inclusion and re-interview it, probably also emit a node removed and node added event.
I find it weird that there is no information about the new node (NIF, is it slave or controller?) sent to the application like it happens on a normal inclusion.

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1901,7 +1906,7 @@ ${associatedNodes.join(", ")}`,
 	public async replaceFailedNode(
 		nodeId: number,
 		includeNonSecure: boolean = false,
-	): Promise<boolean> {
```
AlCalzone: Hm, maybe we should avoid starting this process twice (just like inclusion/exclusion)

robertsLando: I could use the inclusionActive flag?

robertsLando:
> Seems we need to react to the FailedNodeReplaceDone request and bootstrap the new node, essentially perform the secure inclusion and re-interview it, probably also emit a node removed and node added event.

No idea what to do :rofl: The inclusion is automatically started, so what request should I send to start the node interview? 

What I have found on docs is: 

```
ZW_FAILED_NODE_REPLACE

The failed node is ready to be replaced
and controller is ready to add new node
with the nodeID of the failed node.
Meaning that the new node must now
emit a nodeinformation frame to be
included.
```

> I find it weird that there is no information about the new node (NIF, is it slave or controller?) sent to the application like it happens on a normal inclusion.

It is a slave, it's a zwave range extender by aeotec (the only one zwave device I actually have to test :laughing: )

AlCalzone:
> No idea what to do

I would propose:
Emit "node removed", destroy old node instance

https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/controller/Controller.ts#L752-L754
Create the new node instance (we don't receive the NIF, so we don't know the device classes etc. yet, you can leave them out) and remember it.

Request the node's protocol info from the controller (I'm assuming it has it)

Then (again, I'm assuming we can do this - the docs are a bit sparse here) perform the secure bootstrapping:
https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/controller/Controller.ts#L786-L884

Emitting the `node added` event will cause the driver to interview it.

robertsLando:
> Create the new node instance (we don't receive the NIF, so we don't know the device classes etc. yet, you can leave them out) and remember it.

I also don't have the nodeId in the status report context, I should someway add it too...

> Request the node's protocol info from the controller (I'm assuming it has it)

Is there an api on controller for this? Cannot find it

> https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/controller/Controller.ts#L786-L884

Seems this will become a function so, or should I copy and paste it?

AlCalzone:
I've played with this a bit:
```
11:17:33.668 CNTRLR   starting replace failed node process...
11:17:33.671 DRIVER   Send queue:                                                       (2 messages)
                      · [REQ] ReplaceFailedNode
                      · [REQ] SendData (MultiChannelCCEndPointGet) [Node 17, asleep]
11:17:33.673 SERIAL » 0x0105006311028a                                                     (7 bytes)
11:17:33.674 DRIVER » [REQ] [ReplaceFailedNode]
                        payload: 0x1102
11:17:33.676 SERIAL « [ACK]                                                                   (0x06)
11:17:33.684 SERIAL « 0x010401630099                                                       (6 bytes)
11:17:33.684 SERIAL » [ACK]                                                                   (0x06)
11:17:33.685 DRIVER « [RES] [ReplaceFailedNode]
                        payload: 0x00
11:17:33.687 DRIVER   Send queue:                                                        (1 message)
                      · [REQ] SendData (MultiChannelCCEndPointGet) [Node 17, asleep]
11:17:33.802 SERIAL « 0x01050063020398                                                     (7 bytes)
11:17:33.803 SERIAL » [ACK]                                                                   (0x06)
11:17:33.804 DRIVER « [REQ] [ReplaceFailedNode]
                        payload: 0x0203
11:17:33.805 DRIVER   Send queue:                                                        (1 message)
                      · [REQ] SendData (MultiChannelCCEndPointGet) [Node 17, asleep]
11:17:33.806 DRIVER   Send queue:                                                        (1 message)
                      · [REQ] SendData (MultiChannelCCEndPointGet) [Node 17, asleep]
11:17:33.807 DRIVER   handling request ReplaceFailedNode (99)
11:17:33.807 DRIVER     1 handler registered!
11:17:33.808 DRIVER     invoking handler #0
11:17:33.808 CNTRLR   handling replace node request (status = FailedNodeReplace)
11:17:33.809 CNTRLR   inclusion started, node is ready to be replaced
11:17:33.809 DRIVER       the message was handled
11:17:38.693 SERIAL « 0x0105006302049f                                                     (7 bytes)
11:17:38.694 SERIAL » [ACK]                                                                   (0x06)
11:17:38.695 DRIVER « [REQ] [ReplaceFailedNode]
                        payload: 0x0204
11:17:38.696 DRIVER   Send queue:                                                        (1 message)
                      · [REQ] SendData (MultiChannelCCEndPointGet) [Node 17, asleep]
11:17:38.697 DRIVER   Send queue:                                                        (1 message)
                      · [REQ] SendData (MultiChannelCCEndPointGet) [Node 17, asleep]
11:17:38.698 DRIVER   handling request ReplaceFailedNode (99)
11:17:38.698 DRIVER     1 handler registered!
11:17:38.699 DRIVER     invoking handler #0
11:17:38.699 CNTRLR   handling replace node request (status = FailedNodeReplaceDone)
11:17:38.699 CNTRLR   node successfully replaced
11:17:38.700 DRIVER       the message was handled
```
I had to change the check in `replaceFailedNode` to make it not crash on success.
The node was not included securely, but you don't have refactored that part yet, so that is expected.
Also you can see that the pending messages are not discarded because the old node instance is not removed.

What is missing:
- [x] Remove the old node instance on `FailedNodeReplaceDone` 
- [x] Try and see if we can perform secure bootstrapping
- [x] Emit `node added` afterwards to kick off the interview.

robertsLando:
@AlCalzone Seems working :tada:

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -678,6 +692,62 @@ export class ZWaveController extends EventEmitter {
 		});
 	}
 
+	private async secureBootstrap(node: ZWaveNode): Promise<void> {
+		// If security has been set up and we are allowed to include the node securely, do it
+		if (
+			this.driver.securityManager &&
+			!this._includeNonSecure &&
+			node.supportsCC(CommandClasses.Security)
```
robertsLando: The only problelm here is that when replacing a node this check always fails, how should I handle this? I need to fetch command classes before the secure bootstrap (like in add node request)

robertsLando:
@AlCalzone This should be ready now

AlCalzone:
Great, I'll check this out later and try myself

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -880,7 +894,9 @@ export class ZWaveController extends EventEmitter {
 					this._nodes.set(newNode.id, newNode);
 					this._nodePendingInclusion = undefined;
 
-					await this.secureBootstrapS0(newNode);
+					if (!this._includeNonSecure) {
```
robertsLando, line 897: @AlCalzone Wasn't it better to add this check inside the function?

robertsLando: Where do you set secure to false?

AlCalzone: It felt wrong to call a function that does secure bootstrapping if you know you don't want it.
The security CC is removed when one of the commands fails (in the method)

PR #1127, opened 2020-11-17
feat(config): automate index update in PRs

### LINKED ISSUE #1126: Automate index file update in config PRs
Without it, the `lint:config` check won't pick up new files.

For example we could check if a device config file was changed but the index file was not. This action could comment on the PR with a command that updates the index file automatically in the PR.

robertsLando:
Fixes #1126


.github/workflows/updateIndex.yml
```diff
@@ -0,0 +1,53 @@
...
+name: Update Devices Index
+
+on:
+  pull_request:
+    types: [opened, edited]
+  issue_comment:
+    types: [created]
+
+jobs:
+  build:
```
AlCalzone: I think this needs a better name

.github/workflows/updateIndex.yml
```diff
@@ -0,0 +1,53 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
+      - name: Setup repo
+        run: |
+          npm install
+          npm run prerelease
```
AlCalzone: That takes quite a bit of time. We could skip the entire job if no files in `packages/config/config/devices` are changed.

.github/workflows/updateIndex.yml
```diff
@@ -0,0 +1,53 @@
...
+            echo "Index has changed"
+            git add $indexPath
+            echo "::set-output name=PUSH::true"
+          fi
+      - name: Comment a pull_request
+        if: steps.check_index.outputs.PUSH == 'true' && github.event_name == 'pull_request'
+        uses: mb2dev/github-action-comment-pull-request@1.0.0
+        with:
+          message: "The index file isn't updated. Comment 'update index' to automatically update it"
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
AlCalzone: I discussed something similar with @chrisns lately. I'm already doing it with the automerge action, but I'm not really comfortable spreading the `GITHUB_TOKEN` to a bunch of open source actions 🤔
Commenting can be done with octokit, like here https://github.com/zwave-js/node-zwave-js/blob/master/.github/actions/toLogEntry/main.js

.github/workflows/updateIndex.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git commit -m "fix(config): updated devices index.json"
+      - name: Push changes
+        if: steps.check_index.outputs.PUSH == 'true' && github.event_name != 'pull_request' && github.event.comment.body == 'update index'
+        uses: ad-m/github-push-action@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          branch: ${{ github.ref }}
```
AlCalzone: I think we don't need an action if we can just do `git push`:
https://github.community/t/how-does-one-commit-from-an-action/16127/7

AlCalzone (changes requested):

.github/workflows/updateIndex.yml
robertsLando: How do I run the script without building ?

AlCalzone: No I mean looking at the git diff to see if there is even a need to run the script.

AlCalzone: The exit code of `git diff-index --name-only HEAD | grep packages/config/config/devices` should be 0 in that case

.github/workflows/updateIndex.yml
AlCalzone: Or even easier: https://github.com/actions/github-script#comment-on-an-issue

.github/workflows/updateIndex.yml
robertsLando: Check it now

.github/workflows/updateIndex.yml
robertsLando: Done

AlCalzone (approved):

AlCalzone:
Lets try this!

PR #1129, opened 2020-11-17
Test automatic index update in PRs

AlCalzone:


robertsLando:
@AlCalzone You have defined a duplicated `jobs` property in the workflow file

robertsLando:
Now it works but it says `No device changes detected` :thinking:

robertsLando:
@AlCalzone  `github.base_ref` and `head_ref` could be null when the event is not originated from a PR: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions

AlCalzone:
We might need an additional action to detect the changed files in the PR. `git diff` doesn't seem to do it.

AlCalzone:
Or we could do a split approach:
On pull_request events, if there are config changes and the index file is not up to date:
* Delete previous comments by the bot (we can put a HTML comment in there to identify them)
* Add another one at the end

On comment events, if there is a comment by the bot and the comment is `update index` by the PR author:
* Push the changed index file

AlCalzone:
@robertsLando pushing to a branch in a PR is not triggering the action. This comment will

robertsLando:
> pushing to a branch in a PR is not triggering the action. This comment will

I know, you need to add the push event for that

PR #1130, opened 2020-11-17
fix(ci): try fixing the change detection

robertsLando:
Based on: 

https://github.community/t/github-actions-does-not-print-git-diff/17401/2

And: 

https://github.com/actions/checkout/issues/160#issuecomment-589150141

### DIFF
--- a/.github/workflows/update-config-index.yml
+++ b/.github/workflows/update-config-index.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Check if any config files changed
         id: check_changed
         run: |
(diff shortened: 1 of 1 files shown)


robertsLando:
@AlCalzone Could you try with this?

AlCalzone:
I've wanted to avoid hardcoding `master`, but I guess we can't avoid that when we want to support comments too.

AlCalzone:
test comment

AlCalzone (approved):

PR #1131, opened 2020-11-17
Auto-update index file in PRs

AlCalzone:


.github/workflows/update-config-index.yml
```diff
@@ -0,0 +1,100 @@
...
+    needs: check_changed
+    # Skip this job if the config files were not touched
+    if: needs.check_changed.outputs.config_changed
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Install Node.js
+        if: steps.check_changed.outputs.CHANGED == 'true'
```
robertsLando, line 50: Is this if still needed even if the jobs are splitted?

.github/workflows/update-config-index.yml
```diff
@@ -0,0 +1,100 @@
...
+jobs:
+  check_changed:
+    runs-on: ubuntu-latest
+    # Use this job to check whether the next jobs need to run
+    outputs:
+      config_changed: ${{ steps.check_changed.outputs.CHANGED }}
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
```
robertsLando, line 21:
```suggestion
        uses: actions/checkout@v2
        with:
          fetch-depth: 0
```

.github/workflows/update-config-index.yml
AlCalzone: Probably not - we didn't get that far during tests

PR #1141, opened 2020-11-23
Add CC-specific information to value metadata

### LINKED ISSUE #1137: Better identify valueIds types
In OZW I mostly used [this](https://github.com/OpenZWave/open-zwave/blob/master/config/Localization.xml) file to recognize valueIds types paired with [this](https://github.com/OpenZWave/open-zwave/blob/master/config/SensorMultiLevelCCTypes.xml) for sensors types and so on for meters alarms etc... 

I needed this informations because In zwave2mqtt I also support hass discovery, when i discover a new object in hass I need to set it's device_class and other parameters to discover it in the best way. 

So actually what I need is:

How do I retrieve the alarm sensor type from an alarm valueId ?
https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/AlarmSensorCC.ts#L40

The same for sensorType number in sensor multilevel: https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/MultilevelSensorCC.ts#L124

Same for metertype https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/MeterCC.ts#L65

AlCalzone:
fixes: #1137
by adding an optional `ccSpecific` field to the value metadata.

Checklist:

- [x] AlarmSensor (base type `AlarmSensorValueMetadata`)
- [x] BinarySensor (base type `BinarySensorValueMetadata`)
- [x] Indicator
- [x] Meter
- [x] MultilevelSensor (base type `MultilevelSensorValueMetadata`)
- [x] MultilevelSwitch (base type `MultilevelSwitchLevelChangeMetadata`)
- [x] Notification (notification type only for now)
- [x] ThermostatSetpoint (base type `ThermostatSetpointMetadata`) 

The other Thermostat classes don't use any specific type in the value ID


AlCalzone:
@robertsLando let me know if this solves your issues :)

robertsLando:
Tomorrow, thanks so much! 😍

robertsLando:
@AlCalzone What about adding some docs about this? I have added the required changes but unfortunally I haven't any device to test this, will let user to try this in https://github.com/zwave-js/zwavejs2mqtt/pull/20 and let you know :)

AlCalzone:
> What about adding some docs about this?

Yeah... metadata is not documented at all 😅

robertsLando:
I could do it for you :) You have too much work to do here

AlCalzone:
FFS... `lerna`, `npm workspaces` and `npm ci` don't like each other...

robertsLando (dismissed):
LGTM

docs/api/valueid.md
```diff
@@ -29,3 +29,155 @@ interface TranslatedValueID extends ValueID {
...
+	states?: Record<number, string>;
+	unit?: string;
+}
+```
+
+-   `min`: The minumum value that can be assigned to this value
+-   `max`: The maximum value that can be assigned to this value
+-   `steps`: When only certain values between min and max are allowed, this determines the step size
+-   `default`: The default value
+-   `states`: Human-readable names for numeric values
```
robertsLando: Maybe add an example of a state here?

docs/api/valueid.md
```diff
@@ -29,3 +29,155 @@ interface TranslatedValueID extends ValueID {
...
+
+-   `default`: The default value
+
+### string
+
+```ts
+interface ValueMetadataString extends ValueMetadataBase {
+	type: "string";
+	minLength?: number;
+	/** The maximum length this string may have (optional) */
```
robertsLando:
```suggestion
```

docs/api/valueid.md
```diff
@@ -29,3 +29,155 @@ interface TranslatedValueID extends ValueID {
...
+	maxLength?: number;
+	/** The default value */
```
robertsLando:
```suggestion
```

docs/api/valueid.md
```diff
@@ -29,3 +29,155 @@ interface TranslatedValueID extends ValueID {
...
+	};
+};
+```
+
+#### Multilevel Sensor CC
+
+```ts
+type MultilevelSensorValueMetadata = ValueMetadata & {
+	ccSpecific: {
+		sensorType: number;
```
robertsLando, line 172: We could  add a link to sensortypes.json ?

docs/api/valueid.md
```diff
@@ -29,3 +29,155 @@ interface TranslatedValueID extends ValueID {
...
+	};
+};
+```
+
+#### Meter CC
+
+```ts
+type MeterMetadata = ValueMetadata & {
+	ccSpecific: {
+		meterType: number;
```
robertsLando, line 158: Link to meter types json?

robertsLando (changes requested):
Also I would add a link to this section in `getValueMetadata` function

docs/api/valueid.md
robertsLando: https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/sensorTypes.json

docs/api/valueid.md
robertsLando: https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/meters.json

AlCalzone:
> `getValueMetadata`

In the method implementation itself or where do you mean?

robertsLando:
In the method description: 

> Every value in Z-Wave has associated metadata that defines the range of allowed values etc. You can retrieve this metadata using getValueMetadata. Like getValue this takes a single argument of the type ValueMetadata <--- HERE

robertsLando (commented):
Should we reference somewhere in docs also indicators.json notifications.json and scales?

AlCalzone:
I've referenced the scales. Notifications and Indicators might be good to add here too

robertsLando (approved):
LGTM :tada:

PR #1143, opened 2020-11-24
fix: undefined neighbors error

robertsLando:


AlCalzone:
This only solves the symptoms of a bug instead of the disease

robertsLando:
Thought it was something else :) glad you found the source of that bug

PR #1147, opened 2020-11-24
feat: configure logging through driver options

### LINKED ISSUE #1062: Configure logging through driver options
Ideally, this should have preference over env variables. Things to consider:

- [ ] minimum loglevel (currently `LOGLEVEL` env variable) 
- [ ] ability to turn logging off (loglevel: off), should also be supported through the `LOGLEVEL` env variable
- [ ] ability to provide a custom transport for logs (compatible with `winston`)
- [ ] change the logfile location

robertsLando:
Fixes #1062

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: enhancement
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen. Please mention **why** you think the requested feature is useful. Since `zwave-js` is pretty low-level we often have to take a broader approach so many projects can benefit. **Explaining your use-case will go a long way in helping us understand.**
+
+**Describe alternatives you've considered**
+A clear and concise description of any alternative solutions or features you've considered.
+
+**Additional context**
+Add any other context or screenshots about the feature request here. If you want to propose an API, share your thoughts how it could be used (syntax).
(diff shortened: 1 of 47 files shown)


robertsLando:
@AlCalzone What about this?

AlCalzone:
This seems to be going in the right direction. You are not using `setupLogger()` yet, are you? 

I'm thinking if we have to refactor the whole logger stuff to be set up per driver instance. Or the driver could create the base logger instance. All locations with driver access could access it via the driver instance. And we could pass it to all locations without driver access (e.g. serialport).
This way we could run multiple drivers (networks) at once with each driver logging into its own file.

Maybe something for a different PR though.

robertsLando:
> This seems to be going in the right direction. You are not using setupLogger() yet, are you?

Nope I still have to use it in the code, It should be called in driver constructor and logConfig can be an option of Driver options

> I'm thinking if we have to refactor the whole logger stuff to be set up per driver instance. This way we could run multiple drivers (networks) at once with each driver logging into its own file.

I think it's a very good idea but thet needs a complete different approach and a big refactor, could go on with this for now and keep that for 6.0.0 version?

robertsLando:
@AlCalzone I don't understand the lint issues. I have also run `eslint --fix` but it still report lint issues :man_shrugging:

AlCalzone:
I'll check it. Maybe you need to "npm ci" locally, seems like my release fixes also updated Prettier.

AlCalzone (approved):
I've renamed a few methods and updated the documentation a bit. However, this change is breaking since the environment variables are no longer evaluated lazily, but rather when the module is imported.
At least https://github.com/AlCalzone/ioBroker.zwave2 was relying on this behavior.

robertsLando:
Didn't thought about that, I think there is no way to implement this without that breaking, the only one would be to create a Zwave logger instance for each driver like you suggested

AlCalzone:
merged manually

PR #1148, opened 2020-11-24
feat(docs): auto-import interfaces, types, enum declarations in the docs from TS sources

AlCalzone:
With this PR, we have the ability to import the actual definition of types, interfaces and enums into documentation. To do so, I've introduced a comment syntax that is similar to TypeScripts imports:

```md
<!-- #import Firmware from "zwave-js" -->
```
will be converted to
~~~md
<!-- #import Firmware from "zwave-js" -->

```ts
interface Firmware {
    data: Buffer;
    firmwareTarget?: number;
}
```
~~~
for example. There are a few kinks to work out, but the basic functionality is there.

**Next steps:**

- [x] review the generated files if we like the changes
- [x] add the option to remove comments (e.g. if we have better comments in the docs)
- [ ] use this in a github action that checks if the docs can be built
- [ ] use this action to detect changes and create a PR if changes are detected

### DIFF
--- a/docs/api/driver.md
+++ b/docs/api/driver.md
@@ -226,6 +226,7 @@ The `Driver` class inherits from the Node.js [EventEmitter](https://nodejs.org/a
 <!-- #import FileSystem from "zwave-js" -->
 
 ```ts
+/** Defines which methods must be supported by a replacement filesystem */
 interface FileSystem {
 	ensureDir(path: string): Promise<void>;
 	writeFile(
@@ -276,6 +277,7 @@ The message priority must one of the following enum values. Consuming applicatio
 <!-- #import MessagePriority from "zwave-js" -->
 
 ```ts
+/** The priority of messages, sorted from high (0) to low (>0) */
 enum MessagePriority {
 	// Handshake messages have the highest priority because they are part of other transactions
 	// which have already started when the handshakes are needed (e.g. Security Nonce exchange)
@@ -324,6 +326,7 @@ The `onUpdate` has the signature `(status: SupervisionStatus, remainingDuration?
 <!-- #import SupervisionStatus from "zwave-js" -->
 
 ```ts
+/** @publicAPI */
 enum SupervisionStatus {
 	NoSupport = 0x00,
 	Working = 0x01,
@@ -339,6 +342,7 @@ Is used to report the status of a supervised command execution.
 <!-- #import SupervisionResult from "zwave-js" -->
 
 ```ts
+/** @publicAPI */
 interface SupervisionResult {
 	status: SupervisionStatus;
 	remainingDuration?: Duration;
--- a/docs/api/driver.md
+++ b/docs/api/driver.md
@@ -223,10 +223,11 @@ The `Driver` class inherits from the Node.js [EventEmitter](https://nodejs.org/a
 
 ### `FileSystem`
 
+This interface defines which methods must be supported by a replacement filesystem.
+
 <!-- #import FileSystem from "zwave-js" -->
 
 ```ts
-/** Defines which methods must be supported by a replacement filesystem */
 interface FileSystem {
 	ensureDir(path: string): Promise<void>;
 	writeFile(
@@ -268,16 +269,11 @@ interface SendMessageOptions {
 }
 ```
 
-The message priority must one of the following enum values. Consuming applications typically don't need to overwrite the priority.
-
-> [!ATTENTION]
-> DO NOT rely on the numeric values of the enum if you're using it in your application.
-> The ordinal values are likely to change in future updates. Instead, refer to the enum properties directly.
+The message priority must one of the following enum values, which are sorted from high (0) to low (> 0). Consuming applications typically don't need to overwrite the priority.
 
 <!-- #import MessagePriority from "zwave-js" with comments -->
 
 ```ts
-/** The priority of messages, sorted from high (0) to low (>0) */
 enum MessagePriority {
 	// Handshake messages have the highest priority because they are part of other transactions
 	// which have already started when the handshakes are needed (e.g. Security Nonce exchange)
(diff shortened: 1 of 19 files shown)


docs/api/driver.md
```diff
@@ -244,27 +248,32 @@ interface FileSystem {
...
 	supportCheck?: boolean;
-	/** Whether the driver should update the node status to asleep or dead when a transaction is not acknowledged (repeatedly). Setting this to false will cause the simply transaction to be rejected on failure. Default: true */
+	/**
+	 * Whether the driver should update the node status to asleep or dead when a transaction is not acknowledged (repeatedly).
+	 * Setting this to false will cause the simply transaction to be rejected on failure.
+	 * Default: true
+	 */
 	changeNodeStatusOnMissingACK?: boolean;
-	/** Sets the number of milliseconds after which a message expires. When the expiration timer elapses, the promise is rejected with the error code `Controller_MessageExpired`. */
+	/** Sets the number of milliseconds after which a transaction expires. When the expiration timer elapses, the transaction promise is rejected. */
```
AlCalzone: 👎

docs/api/driver.md
```diff
@@ -244,27 +248,32 @@ interface FileSystem {
...
 	expire?: number;
+	/** Internal information used to identify or mark this transaction */
+	tag?: any;
```
AlCalzone: @internal

docs/api/driver.md
```diff
@@ -275,26 +284,30 @@ enum MessagePriority {
 	// Our handshake requests must be prioritized over all other messages
 	PreTransmitHandshake = 1,
 	// Controller commands usually finish quickly and should be preferred over node queries
-	Controller = 2,
+	Controller,
```
AlCalzone, line 285: Not sure, nobody should use the numbers, so there's no need to document them.

docs/api/node.md
```diff
@@ -205,25 +209,30 @@ enum InterviewStage {
 	ProtocolInfo,
 	/** The node has been queried for supported and controlled command classes */
 	NodeInfo,
+
```
AlCalzone, line 212: This comes from removed single-line comments. Not pretty, but I see no simple way around it.

docs/api/valueid.md
```diff
@@ -68,14 +77,22 @@ interface ValueMetadataAny extends ValueMetadataBase {
...
 	max?: number;
+	/** When only certain values between min and max are allowed, this determines the step size */
 	steps?: number;
+	/** The default value */
 	default?: number;
+	/** Speaking names for numeric values */
 	states?: Record<number, string>;
+	/** An optional unit for numeric values */
 	unit?: string;
 }
```
AlCalzone: 👎

docs/api/valueid.md
```diff
@@ -89,9 +106,12 @@ interface ValueMetadataNumeric extends ValueMetadataBase {
 
 #### boolean
 
+<!-- #import ValueMetadataBoolean from "zwave-js" -->
+
 ```ts
 interface ValueMetadataBoolean extends ValueMetadataBase {
 	type: "boolean";
+	/** The default value */
```
AlCalzone: 👎

docs/api/valueid.md
```diff
@@ -100,11 +120,16 @@ interface ValueMetadataBoolean extends ValueMetadataBase {
...
+<!-- #import ValueMetadataString from "zwave-js" -->
+
 ```ts
 interface ValueMetadataString extends ValueMetadataBase {
 	type: "string";
+	/** The minimum length this string must have (optional) */
 	minLength?: number;
+	/** The maximum length this string may have (optional) */
 	maxLength?: number;
+	/** The default value */
```
AlCalzone: :-1:
but the description below should be synchronized with this.

docs/api/valueid.md
```diff
@@ -139,11 +168,13 @@ type BinarySensorValueMetadata = ValueMetadata & {
...
 #### Indicator CC
 
+<!-- #import IndicatorMetadata from "zwave-js" -->
+
 ```ts
 type IndicatorMetadata = ValueMetadata & {
 	ccSpecific: {
 		indicatorId: number;
-		propertyId?: number; // only present on V2+ indicators
+		propertyId?: number;
```
AlCalzone: :-1:

maintenance/generateTypedDocs.ts
```diff
@@ -0,0 +1,207 @@
...
+	module: string;
+	symbol: string;
+	import: string;
+	options: {
+		comments?: boolean;
+		jsdoc?: boolean;
+	};
+}
+
+const importRegex = /(?<import><!-- #import (?<symbol>.*?) from "(?<module>.*?)"(?: with (?<options>[\w\-, ]*?))? -->)(?:[\s\r\n]*```ts[\r\n]*(?<source>(.|\n)*?)```)?/g;
```
robertsLando, line 132: @AlCalzone This regex is pure madness :rofl:

robertsLando (approved):
Everything seems good here. I never used `ts-morph` so cannot say much about the docs generation methods

AlCalzone:
> I never used `ts-morph` so cannot say much about the docs generation methods

It is basically a tool to manipulate the abstract syntax tree (AST) of TypeScript. I just found it today, but it nicely abstracts away the somewhat cumbersome methods the TS API has to offer.
A good example is the one-liner `toggleModifiers` instead of my (incomplete) `updateModifiers` implementation in this commit:
https://github.com/zwave-js/node-zwave-js/pull/1148/commits/c33549169aa27603e2c7876a7f5e5aff1f499be6#

maintenance/generateTypedDocs.ts
AlCalzone: I've seen worse

PR #1153, opened 2020-11-25
feat(cc): refresh current value even if supervision fails or times out

AlCalzone:
We've had a few reports where sending Supervision encapsulated MultilevelSwitchCCSet commands would result in `SupervisionResult.Fail` or no response to the `SupervisionGet` and thus no updated `currentValue`.

This PR drops all the checks and always refreshes the current value to see if that works out better.

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/MultilevelSwitchCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/MultilevelSwitchCC.ts
@@ -17,6 +17,7 @@ import log from "../log";
 import { MessagePriority } from "../message/Constants";
 import {
 	CCAPI,
+	ignoreTimeout,
 	SetValueImplementation,
 	SET_VALUE,
 	throwUnsupportedProperty,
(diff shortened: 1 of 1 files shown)


robertsLando:
It's always nice to see how going against specs could make thing work :rofl:

ODINServ:
Goodmorning,

I've got similar behaviour but with a different request.
Running this branch + zwave2mqtt multicast branch.

Usecase, system was running fine, when i send a open request to all 12 fgr223 nodes at the same time.

```
  z2m:Zwave Error while writing 99 on 8-38-1-targetValue: Message dropped because of an unexpected error: Security CC requires a nonce to be sent! +5s
  z2m:Zwave Unable to write 99 on 8-38-1-targetValue +0ms
```

But the command was received by the node, and the node did respond by opening.
And the value was not correct until i manually requested "open" again which directly updated it to the correct value.

Console output (short since i guess the driver log is more interesting,  and i don't have more since it is running in a "screen" session): [console-output.txt](https://github.com/zwave-js/node-zwave-js/files/5614443/console-output.txt)
Driver log: [zwave-3255.log](https://github.com/zwave-js/node-zwave-js/files/5614444/zwave-3255.log)

Thnx

AlCalzone:
1. I don't think that has to do with this PR.
2. I only see a single `MultilevelSwitchCCSet` which is sent and afterwards the current value is refreshed sucessfully. So the problem you mentioned should be before or after that log snippet.

ODINServ:
Ah yes, that is my retry only: this is (should be) the complete log: 
[zwave-3255-2.log](https://github.com/zwave-js/node-zwave-js/files/5614662/zwave-3255-2.log)

AlCalzone:
Okay, there is something weird going on with node 8...
There are several instances where it doesn't reply. Sometimes it asks us for nonces for secure communication (which we send and which it usually receives) but then doesn't send us a message.

The error above (Security CC requires a nonce to be sent!) happens because the driver tries 3 times to send an encrypted command but never receives the required nonce from node 8.

Judging from that behavior I would say node 8 has a bad connection.

AlCalzone:
So how should we proceed with this?
@ODINServ does it improve the behavior of your devices with Supervision CC?

ODINServ:
Not sure because i'm not able to reproduce it that easily, and i was not that long on this branch.

Not sure how to proceed, i will go to unencrypted nodes and that should make the network more stable aswel.

AlCalzone:
Ok, lets close this PR and reopen if necessary

PR #1160, opened 2020-11-26
fix: get rid of ValueMetadataAny

robertsLando:


AlCalzone:
Hm... maybe it was correct :D

AlCalzone:
I **think** this is technically correct, but it opens a huge can of worms related to assignability issues.

AlCalzone:
`ValueMetadataAny` is now merged with `ValueMetadataBase`. Having it separate didn't really make sense from a type POV. The exported `ValueMetadata.Any, .Readonly and .Writeonly` still have `type: "any"`

AlCalzone (dismissed):

AlCalzone:
merged manually

PR #1176, opened 2020-12-01
Guess firmware format function and docs

### LINKED ISSUE #1174: Add guessFirmwareFormat function
Actually guessFirmwareFormat is not present in zwave-js

robertsLando:
Fixes #1174

### DIFF
--- a/docs/api/node.md
+++ b/docs/api/node.md
@@ -135,14 +135,14 @@ Starts an OTA firmware update process for this node. This method takes two argum
 The library includes 2 helper method (exported from `zwave-js/Utils`) to guess the firmware format and extract the raw firmware image from some common file types:
 
 ```ts
-function guessFirmwareFormat(
+function guessFirmwareFileFormat(
 	filename: string,
-	firmware: Buffer,
+	rawData: Buffer,
 ): FirmwareFileFormat
 ```
 
--   `filename` the name of the firmware file
--   `firmware` the data buffer of the firmware file
+-   `filename`: the name of the firmware file
+-   `rawData`: the file firmware content raw Buffer
 
 ```ts
 function extractFirmware(data: Buffer, format: FirmwareFileFormat): Firmware
@@ -169,7 +169,6 @@ Example usage:
 
 ```ts
 // Extract the firmware from the uploaded file
-const rawData = Buffer.from(firmware);
 let actualFirmware: Firmware;
 try {
 	const format = guessFirmwareFormat(
--- a/docs/api/node.md
+++ b/docs/api/node.md
@@ -130,29 +130,31 @@ beginFirmwareUpdate(data: Buffer, target?: number): Promise<void>
 Starts an OTA firmware update process for this node. This method takes two arguments:
 
 -   `data` - A buffer containing the firmware image in a format supported by the device
--   `target` - _(optional)_ The firmware target (i.e. chip) to upgrade. 0 updates the Z-Wave chip, >=1 updates others if they exist
+-   `target` - _(optional)_ The firmware target (i.e. chip) to upgrade. `0` updates the Z-Wave chip, `>=1` updates others if they exist
 
-The library includes 2 helper method (exported from `zwave-js/Utils`) to guess the firmware format and extract the raw firmware image from some common file types:
-
-```ts
-function guessFirmwareFileFormat(
-	filename: string,
-	rawData: Buffer,
-): FirmwareFileFormat
-```
-
--   `filename`: the name of the firmware file
--   `rawData`: the file firmware content raw Buffer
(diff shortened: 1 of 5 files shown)


docs/api/node.md
```diff
@@ -132,7 +132,17 @@ Starts an OTA firmware update process for this node. This method takes two argum
 -   `data` - A buffer containing the firmware image in a format supported by the device
 -   `target` - _(optional)_ The firmware target (i.e. chip) to upgrade. 0 updates the Z-Wave chip, >=1 updates others if they exist
 
-The library includes a helper method (exported from `zwave-js/Utils`) to extract the raw firmware image from some common file types:
+The library includes 2 helper method (exported from `zwave-js/Utils`) to guess the firmware format and extract the raw firmware image from some common file types:
+
+```ts
+function guessFirmwareFormat(
+	filename: string,
+	firmware: Buffer,
```
AlCalzone: I guess we should call this parameter `rawData` to match the example below

docs/api/node.md
```diff
@@ -155,6 +165,38 @@ interface Firmware {
 
 If no firmware data can be extracted, the method will throw.
 
+Example usage:
+
+```ts
+// Extract the firmware from the uploaded file
+const rawData = Buffer.from(firmware);
```
AlCalzone: Hmm I think this is confusing if you read this the first time. Maybe just state that `rawData` is a Buffer with the file contents and `filename` the corresponding filename?

docs/api/node.md
```diff
@@ -132,7 +132,17 @@ Starts an OTA firmware update process for this node. This method takes two argum
 -   `data` - A buffer containing the firmware image in a format supported by the device
 -   `target` - _(optional)_ The firmware target (i.e. chip) to upgrade. 0 updates the Z-Wave chip, >=1 updates others if they exist
 
-The library includes a helper method (exported from `zwave-js/Utils`) to extract the raw firmware image from some common file types:
+The library includes 2 helper method (exported from `zwave-js/Utils`) to guess the firmware format and extract the raw firmware image from some common file types:
```
AlCalzone: I know I originally wrote this, but we should mention which input files we support.

packages/core/src/util/firmware.ts
```diff
@@ -9,6 +9,28 @@ export interface Firmware {
...
+	firmware: Buffer,
+): FirmwareFileFormat {
+	if (
+		(filename.endsWith(".exe") || filename.endsWith(".ex_")) &&
+		firmware.includes(Buffer.from("Aeon Labs", "utf8"))
+	) {
+		return "aeotec";
+	} else if (/\.(hex|ota|otz)$/.test(filename)) {
+		return filename.slice(-3) as FirmwareFileFormat;
+	} else {
```
AlCalzone, line 34: Should we throw a separate error to let users know that `.hec` is not supported?

packages/core/src/util/firmware.ts
```diff
@@ -9,6 +9,28 @@ export interface Firmware {
...
+		throw new Error("could not detect firmware format");
```
AlCalzone: This should be a `ZWaveError` with a corresponding `ZWaveErrorCode`.

docs/api/node.md
```diff
@@ -132,7 +132,17 @@ Starts an OTA firmware update process for this node. This method takes two argum
...
+
+```ts
+function guessFirmwareFormat(
```
AlCalzone: nit: We should probably call this `guessFirmwareFileFormat` (non-breaking, preferred) or rename `FimwareFileFormat` to `FirmwareFormat` (breaking).

AlCalzone (changes requested):

docs/api/node.md
robertsLando: Yep done

docs/api/node.md
robertsLando: It's mentioned later on extractFirmware function

packages/core/src/util/firmware.ts
robertsLando: Humm nope, it's invalid like the others for now until we don't support it

packages/core/src/util/firmware.ts
robertsLando: fixed

docs/api/node.md
robertsLando: Done

packages/core/src/util/firmware.ts
AlCalzone: I mean saying `.hec` is not (and probably won't be supported) is clearer than "we don't know what this is, maybe we could support it".

AlCalzone (approved):

PR #1177, opened 2020-12-01
feat: support gecko bootloader firmwares

### LINKED ISSUE #1173: Support GBL format for OTA firmware upgrades of Series 700 devices
In order to perform firmware upgrades on SiLabs Series 700 devices,  we need to support the Gecko Bootloader (GBL) file format.

Add the GBL format to the  guessFormat function in zwavejs.

**Additional context**

* Sample GBL firmware files: Update Recessed Door Sensor 7 V1.05 <https://aeotec.freshdesk.com/support/solutions/articles/6000243778-update-recessed-door-sensor-7-v1-05>)"
* Gecko Bootloader User Guide  https://www.wless.ru/files/Silabs/EFR32/ug266-gecko-bootloader-user-guide.pdf

AlCalzone:
fixes: #1173 

@larstobi you might wanna test this :)


larstobi:
@AlCalzone Wow, that was fast! All right, I'll test it, but will have to buy a Series 700 device first. I'll get right on that!

AlCalzone:
LOL I thought you already had one :D

larstobi:
No, haha, but I will soon. I have a "Aeotec Door/ Window Sensor 7" on the way to my doorstep!

packages/core/src/util/firmware.ts
```diff
@@ -2,13 +2,20 @@
 import MemoryMap from "nrf-intel-hex";
 import { ZWaveError, ZWaveErrorCodes } from "../error/ZWaveError";
 
-export type FirmwareFileFormat = "aeotec" | "otz" | "ota" | "hex";
+export type FirmwareFileFormat = "aeotec" | "otz" | "ota" | "hex" | "gecko";
```
robertsLando, line 5: @AlCalzone What about making this an enum? So we prevent to have hardcoded strings

packages/core/src/util/firmware.ts
```diff
@@ -21,7 +28,7 @@ export function guessFirmwareFileFormat(
 ): FirmwareFileFormat {
 	if (
 		(filename.endsWith(".exe") || filename.endsWith(".ex_")) &&
-		rawData.includes(Buffer.from("Aeon Labs", "utf8"))
+		rawData.includes(firmwareIndicators.aeotec)
 	) {
 		return "aeotec";
```
robertsLando, line 33: This could become `FirmwareFileFormat.aeotec`

packages/core/src/util/firmware.ts
```diff
@@ -31,6 +38,11 @@ export function guessFirmwareFileFormat(
 			"Encrypted .hec firmware files are not supported",
 			ZWaveErrorCodes.Unsupported_Firmware_Format,
 		);
+	} else if (
+		filename.endsWith(".gbl") &&
+		rawData.readUInt32BE(0) === firmwareIndicators.gecko
+	) {
+		return "gecko";
```
robertsLando, line 45: `FirmwareFileFormat.gecko`

packages/core/src/util/firmware.ts
```diff
@@ -58,6 +71,10 @@ export function extractFirmware(
 		case "ota":
 		case "hex":
 			return extractFirmwareHEX(rawData);
+		case "gecko":
```
robertsLando, line 74: same

robertsLando (changes requested):

packages/core/src/util/firmware.ts
AlCalzone: I don't really see the benefit of this (breaking!) change when there is auto-completion and an exported type to help the discoverability.

robertsLando: Ok no problem :)

robertsLando (approved):
LGTM if @larstobi test completes correctly :)

larstobi:
@AlCalzone @robertsLando 

Using the latest zj2m (zwave-js 5.7.0) docker image from docker hub, I've tested this on the Aeotec Recessed Door Sensor 7 with the EU gbl file from this article: https://aeotec.freshdesk.com/support/solutions/articles/6000243778-update-recessed-door-sensor-7-v1-05

"Error while calling api beginFirmwareUpdate: Unable to extract firmware from file: could not detect firmware format"

AlCalzone:
:( let me check

AlCalzone:
This seems to be a problem outside of zwave-js. This is the method that throws in your case:
grafik

@robertsLando Any idea?

robertsLando:
@larstobi Could you show me the exact args that are sent to the api? There should be a log for them

larstobi:
@robertsLando I have turned on debug logging to file, as well as web ui console debug logging, but I cannot find any args logged. Where should they appear?

robertsLando:
You should see a log that says you have called an api request `beginFirmwareUpdate` and then the args. Are you using latest from master?

PR #1188, opened 2020-12-03
docs: update typed documentation 🤖

AlCalzone:
The auto-generated documentation has changed. Please review the changes and merge them if desired.

### DIFF
--- a/docs/api/driver.md
+++ b/docs/api/driver.md
@@ -391,7 +391,7 @@ interface ZWaveOptions {
 	/** Allows you to specify a different cache directory */
 	cacheDir: string;
 
-	/** Specify the network key to use for encryption. This must be a Buffer with exactly 16 bytes length */
+	/** Specify the network key to use for encryption. This must be a Buffer of exactly 16 bytes. */
 	networkKey?: Buffer;
 }
 ```


docs/api/driver.md
```diff
@@ -391,7 +391,7 @@ interface ZWaveOptions {
 	/** Allows you to specify a different cache directory */
 	cacheDir: string;
 
-	/** Specify the network key to use for encryption. This must be a Buffer with exactly 16 bytes length */
+	/** Specify the network key to use for encryption. This must be a Buffer of exactly 16 bytes. */
```
robertsLando, line 394: @AlCalzone May an example valid key be useful? Like `"01234267891234567891234567894512"`

robertsLando (approved):
LGTM

docs/api/driver.md
AlCalzone: Well, it must be a Buffer. Doesn't really matter how the applications accept that input.

robertsLando: @AlCalzone You right but I think that 90% of times the buffer will come from a user string input :) `Buffer.from("<32 char string>")`

AlCalzone: Lets put it into your usage example. I would assume that someone who writes an application that supports Z-Wave has enough knowledge to create a buffer ;)

Issue #806, opened 2020-05-17
Color CC: Add RGBW value ID


AlCalzone:

robertsLando:
What do we need for this?

AlCalzone:
We need to which format the hex string should have - which digits, with or without `#`?. OZW concatenates a bunch of HEX digits:
https://github.com/OpenZWave/open-zwave/blob/master/cpp/src/command_classes/Color.cpp#L210

I'm not sure if its worth going past `#RRGGBB` with this, since it is standard and everything else is missing discoverability (maybe `WW`, but that could be warm or cold white...).
We also need to figure out what to do with devices that don't support one of the basic colors. Leave it out, ignore it?

Basically what needs to be done is:
* Probably during the interview, metadata for this value ID should be created
* When a color report is received, the new value of this string must be figured out (from existing values and the new value) and also written
* When this value is set (via the SET_VALUE API) it must be split into its parts to call the set method.

Hope I didn't forget anything.

robertsLando:
> When a color report is received, the new value of this string must be figured out (from existing values and the new value) and also written

And reported to the user by an event :)

AlCalzone:
> And reported to the user by an event :)

That happens when you set a value in the value DB

PR #1195, opened 2020-12-09
feat: implement virtual nodes/endpoints, support broadcast/multicast API

### LINKED ISSUE #1150: Support sending multicast commands through the commandClasses API
The API could look like this:
```ts
const multicastGroup = this.driver.controller.getNodes(2, 3, 4, 5);
// Use the commandClasses API as usual. A group supports a CC if at least one node supports it.
await multicastGroup.commandClasses["Binary Switch"].set(true);
// ^-- This command should only target the nodes that support Binary Switch
```

And there should be a possibility to send a broadcast (node ID 255), for example like this:
```ts
const broadcast = this.driver.controller.getBroadcastNode();
// Use the commandClasses API as usual. The broadcast node supports a CC if at least one node supports it.
await broadcast.commandClasses["Binary Switch"].set(true);
```

AlCalzone:
With this PR we introduce the concept of virtual nodes and endpoints.

Like physical nodes, a virtual node can have multiple (virtual) endpoints. However, it may target multiple nodes (broadcast or multicast) with the following rules:
* A virtual endpoint represents the physical endpoints with the same index, which belong to the the physical nodes targeted by the virtual node
* A virtual endpoint supports a CC if at least one of the physical endpoints it targets has non-secure support for that CC
* The supported CC version is the minimum non-zero version that any of the physical endpoints support. If none support the CC, it is still 0

To use this feature, two new methods were added to `Controller`:
* `getBroadcastNode()` returns a virtual node with the ID 255 (targets ALL nodes)
* `getMulticastGroup(nodeIDs: number[])` returns a virtual node which targets the given nodes.

The functionality of virtual nodes is very limited in comparison with physical nodes. When it comes to CC commands, there are a few limitations:
1. some CCs cannot be used at all (mainly management CCs like Association etc.)
2. some commands (mainly get-type commands) cannot be used
3. Security cannot be used because sending secure communication includes a nonce exchange with each individual node

fixes: #1150 
required for: https://github.com/zwave-js/zwavejs2mqtt/pull/25


AlCalzone:
I think this is ready to be tested

robertsLando (commented):
Missing docs for `getBroadcastNode` and `getMulticastGroup` controller apis

robertsLando:
@AlCalzone Based on this apis I should: 

```
await this.driver.getBroadcastNode().setValue(valueId)

await this.driver.getMulticastGroup(...nodesArray).setValue(valueId)
```

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -273,6 +275,24 @@ export class ZWaveController extends EventEmitter {
...
+	public getBroadcastNode(): VirtualNode {
+		return new VirtualNode(
+			NODE_ID_BROADCAST,
+			this.driver,
+			this.nodes.values(),
+		);
+	}
+
+	/** Creates a virtual node that can be used to send multicast commands to several nodes */
+	public getMulticastGroup(...nodeIDs: number[]): VirtualNode {
```
robertsLando: @AlCalzone why did you use the spread operator? I think that pass an array is good here :)

AlCalzone: I thought being able to do this
```
driver.controller.getMulticastGroup(2, 3, 4)
```
would be nice. Though the case of passing an array is probably more frequent.

AlCalzone:
> Missing docs for `getBroadcastNode` and `getMulticastGroup` controller apis

Didn't get that far yesterday :)

robertsLando:
> Didn't get that far yesterday :)

Was just for reference in case you missed :)

packages/zwave-js/src/lib/controller/Controller.ts
robertsLando: >  Though the case of passing an array is probably more frequent.

Yep it's good to read but I think 99% of times devs will use spread too when calling this

ODINServ:
Hi, I'm willing to test, i saw that https://github.com/zwave-js/zwavejs2mqtt/pull/25 of the mqtt project is merged to master?
Edit: it was merged the other way, so that multicast branch  + this branch

Does that mean when rebuilding that from master and this branch i should be able to send mc / bc messages?

AlCalzone:
I think the zjs2mqtt branch was using the old approach with itering through nodes.

robertsLando:
@AlCalzone I need to know how the setValue works with endpoints. I mean should I specify endpoint? What If I do and what If I don't? Is a broadcast/multicast request sent to all the endpoints if not? 

@ODINServ I also would like to know if you would prefer to set the valueIds from the topic or in the payload. Actually broadcast requests were thinked differently and were using topics but IMO now we should simply call `broadcast/set` and add the valueId CC, property and property key, and the value in the payload

Like

```js
{
commandClass: ...,
property: ...,
propertyKey: ...,
endpoint: ...?
value: ...
}
```

AlCalzone:
> I need to know how the setValue works with endpoints. I mean should I specify endpoint? What If I do and what If I don't? Is a broadcast/multicast request sent to all the endpoints if not?

It should not be different from targeting single nodes. You can use endpoints if desired:
> A virtual endpoint is to a virtual node what a physical endpoint is to a physical node. Since virtual nodes can target multiple physical nodes, virtual endpoints can target multiple physical endpoints by extension.
A virtual endpoint exists if at least one of the virtual node's underlying physical nodes has an endpoint with the same index.

Support of the CC is still tested - meaning a CC is supported if **at least one** of the underlying physical endpoints does. By extension this means if you're trying to send a broad/multicast command to an endpoint no node has, it will tell you that the CC is not supported.

AlCalzone:
Weird... one unrelated test is now failing on node 12 and 14. At least I can reproduce it locally - maybe it has to do with what https://github.com/zwave-js/node-zwave-js/pull/1196 fixes...

ODINServ:
> @ODINServ I also would like to know if you would prefer to set the valueIds from the topic or in the payload. Actually broadcast requests were thinked differently and were using topics but IMO now we should simply call `broadcast/set` and add the valueId CC, property and property key, and the value in the payload
> 
> Like
> 
> ```js
> {
> commandClass: ...,
> property: ...,
> propertyKey: ...,
> endpoint: ...?
> value: ...
> }
> ```

I don't mind I can use both, my preference is the new one I can easily integrate that in openhab not sure how other tools manage that.
Only thing for your proposal is that the docs should contain some examples to use.

An completely different idea is to integrate it on the website.

Rename groups to association, then we could see this multicast feature as groups of endpoints under the groups tab. 

The website can then also make sure that the correct nodes are matched with the command classes, and once a group is created using the website a new topic is created on which we can set values like of it is a normal device.

Groups can than also be detected via the Hass descovery in openhab / Hass , so than all zwave gouping logic is in the zwave2mqtr backend and not in the home automation tool.

The zigbee2mqtt also does this but not sure if the zwave protocol behaves the same.

For me personally I'm already happy with your proposal!

robertsLando:
> The website can then also make sure that the correct nodes are matched with the command classes, and once a group is created using the website a new topic is created on which we can set values like of it is a normal device

This requires much more effort :)

AlCalzone:
FYI: I plan to merge this without support for multiple endpoints. That will require changes in the same order of magnitude and is best left for a followup PR

robertsLando:
@ODINServ zwavejs2mqtt pr is ready for the test. Let me know. Also feel free to open a feature request for groups management, I think it's a good idea but needs some work in the ui

AlCalzone:
I **think** this should be complete now. @robertsLando lets put this to a test - we should see quickly if something is missing.

robertsLando:
@AlCalzone I'm waiting for @ODINServ feedback

ODINServ:
Hey,

**Question**: I'm running the multicast branch of the mqtt project and this branch, is that correct?

I did some testing but without luck :(
It does not matter if i use the optional id in the topic

**Broadcast**
_topic_ ```zwavejs2mqtt/_CLIENTS/ZWAVE_GATEWAY-server/broadcast/set```:
_Payload_:
```
{
  "commandClass": 38,
  "property": "targetValue",
  "value": 99
}
```
Result: 
```08:41:34.823 CNTRLR   [Node 016] [+] [Multilevel Switch] targetValue: undefined         [Endpoint 0]```
```[Node 255] did not respond after 1/3 attempts. Scheduling next try in 500 ms.```
Logs: [zwave-1245-broadcast.log](https://github.com/zwave-js/node-zwave-js/files/5682675/zwave-1245-broadcast.log)


**Multicast**
_topic_ `zwavejs2mqtt/_CLIENTS/ZWAVE_GATEWAY-server/multicast/set`
_payload_:
```
{
  "commandClass": 38,
  "property": "targetValue",
  "value": 99,
  "nodes": [13, 14, 15]
}
```

Result (Not sure if it even was send to the nodes?):
```
08:46:43.473 DRIVER » [MULTICAST] [REQ] [SendDataMulticast]
                      │ transmit options: 0x25
                      │ callback id:      44
                      └─[MultilevelSwitchCCSet]
                          target value: 99
08:46:43.475 SERIAL « [ACK]   
```
Logs: [zwave-1245-multicast.log](https://github.com/zwave-js/node-zwave-js/files/5682680/zwave-1245-multicast.log)

**Comment by AlCalzone** [comment]
> Not sure if it even was send to the nodes?

Looks like it was:
```
08:47:37.294 DRIVER « [REQ] [SendDataMulticast]
                        callback id:     47
                        transmit status: OK
```

> [Node 255] did not respond after 1/3 attempts

That is one thing I need to change. The broadcast "node" cannot acknowledge the receipt. 

Did the commands do something?

**Comment by AlCalzone** [comment]
> That is one thing I need to change. The broadcast "node" cannot acknowledge the receipt.

Oh wait... the transmission is acknowledged. Just the followup command to refresh the current value cannot be done over broadcast.

**Comment by AlCalzone** [comment]
So this might be a problem... Refreshing the value must be done with singlecast followups.
When sending multicast I think it is okay to do that.
However after a broadcast, we might have to (theoretically) send up to 232 get requests to check if the value was actually refreshed, which doesn't seem like a good idea.

**Comment by ODINServ** [comment]
Nothing happens at all, not for the broadcast and not for the multicast

I have two thing in mind:
* Should the nodeId which i send in the mqtt request also be in the logging?
* * According to your example here:  https://github.com/zwave-js/zwavejs2mqtt/pull/25#discussion_r530317689
* * My assumption is that it is just not send because they are also not logged?
* Could it be that it is not working since the nodes are secured included ?

Also i did an additional test by adding the endpoint.

That also seems not to matter, only the ```destination: 1``` seems to be filled now
With endpoint:
```
12:51:25.869 DRIVER » [MULTICAST] [REQ] [SendDataMulticast]
                      │ transmit options: 0x25
                      │ callback id:      23
                        │ source:      0
                        │ destination: 1
                        └─[MultilevelSwitchCCSet]
                            target value: 99
```

Without endpoint:
```
12:53:00.287 DRIVER » [MULTICAST] [REQ] [SendDataMulticast]
                      │ transmit options: 0x25
                      │ callback id:      27
                      └─[MultilevelSwitchCCSet]
                          target value: 99
```
But if i do that for a broadcast:
```12:46:23.451 DRIVER   Virtual node 255, Endpoint 1: Trying to access endpoint instance before Multi Channel interview```

**Comment by ODINServ** [comment]
> Could it be that it is not working since the nodes are secured included ?

I think this seems to be the case

Node 16 is a fgs222 (unsecured) (only unsecure node in the network)
Node 17 is a fgs223 (secured)

Only 16 is responding when i send:
```
{
  "commandClass": 37,
  "property": "targetValue",
  "value": false,
  "nodes": [16,17]
}
```

See logs: [zwave-1245-unsecure.log](https://github.com/zwave-js/node-zwave-js/files/5683022/zwave-1245-unsecure.log)

**Comment by AlCalzone** [comment]
Yep, the security is the problem. It is not possible to send a secure command to multiple nodes, because sending secure commands requires a temporary nonce from the target node.

S2 can be used with multicast, but that isn't implemented yet.

**Comment by ODINServ** [comment]
Ok clear, any idea on a time line for that ?

I will rejoin everything unsecure because I'm also curious if the network becomes more stable than. But that's a bit of a hassle.

But i still I do think having everything encrypted for iot devices is a good thing. What is your opinion? I don't mind that mutch if some one can see state of things (but they could get some interesting information out of that), but controlling them is something else.

But anyhow I'm not controlling locks or something, and not planning to do that either.

**Comment by AlCalzone** [comment]
I can't say for sure - if I find the time, probably at the end of the winter holidays. S2 is pretty complex. I have got a lot of the puzzle pieces together, but putting it all together will take quite a bit of effort.

> But i still I do think having everything encrypted for iot devices is a good thing. What is your opinion?

Yes, it is a good thing, but not really feasible with S0. As long as S0 is the maximum security, I would advise (for maximum stability):
* Only encrypt what is really necessary or irresponsible if not encrypted (e.g. smart locks)
* Think about encrypting sensors (which could leak information if unencrypted)
* Keep everything else unencrypted

**Comment by ODINServ** [comment]
> I can't say for sure - if I find the time, probably at the end of the winter holidays.

Don't start building it only for me, I can live with some not encrypted nodes. I don't have smart locks or zwave sensors so that's fine.

Is saw indeed some flowcharts about S2 m/bcast looks like a lot of work indeed.

I don't have time today to rejoin them, _or you might have a magic command to switch to unencrypted?_
If not, I will take the hassle comming week, I can report back then if you want me to test further before merging.

**Comment by AlCalzone** [comment]
Back to topic: I understand your message above as "it worked for the insecure node", so this feature should be ready.

**Comment by AlCalzone** [comment]
@robertsLando Anything missing from a dashboard POV?

File: docs/api/controller.md [hunk 2]
```diff
@@ -164,6 +164,26 @@ If the target endpoint is not given, the association is a "node association". If
...
+
+```ts
+getMulticastGroup(nodeIDs: number[]): VirtualNode
+```
+
+Creates a virtual node that can be used to send commands to multiple supporting nodes with a single (multicast) message. You can target individual endpoints as usual.
+
+> [!NOTE]
+> Virtual nodes do not support all methods that physical nodes do. Check [`VirtualNode`](api/virtual-node-endpoint.md) for details.  
+> Furthermore, it is not possible to send secure commands using multicast or broadcast.
```
robertsLando: @AlCalzone Isn't more clear to say: Isn't possible to send multicast/broadcast commands to nodes included securely (until S2 support) ?

robertsLando:
LGTM. @AlCalzone will this allow also to add 'virtual valueids' ? I'm thinking about the support for rgb #806

AlCalzone:
> will this allow also to add 'virtual valueids'

I don't think this is necessary. A normal value ID will do just fine for RGB.

PR #1206, opened 2020-12-13
feat: expose node name and location as a writeable property

### LINKED ISSUE #1204: Make `Node Naming And Location CC` valueIDs internal, expose them through properties on the node.
That would be the `Node Naming And Location CC` properties. I'm fine with reusing those properties for our purpose.

Setting them on the node (if the CC is supported) should still happen through the corresponding API.

_Originally posted by @AlCalzone in https://github.com/zwave-js/node-zwave-js/issues/1180#issuecomment-737110291_

AlCalzone:
With this PR, we expose two new properties on the `ZWaveNode` class: `name` and `location`. Both use the corresponding value IDs from `Node Naming and Location CC` to read and locally store the value.
To change it on the node, the `commandClasses` API must be used.

fixes: #1204

### DIFF
--- a/docs/api/node.md
+++ b/docs/api/node.md
@@ -216,6 +216,28 @@ readonly id: number
 
 Returns the ID this node has been assigned by the controller. This is a number between 1 and 232.
 
+### `name`
+
+```ts
+name: string | undefined;
+```
+
+The user-defined name of this node. Uses the value reported by `Node Naming and Location CC` if it exists.
+
+> [!NOTE]
+> Setting this value only updates the name locally. To permanently change the name of the node, use the [`commandClasses` API](api/endpoint.md#commandClasses).
+
+### `location`
+
+```ts
+location: string | undefined;
+```
+
+The user-defined location of this node. Uses the value reported by `Node Naming and Location CC` if it exists.
+
+> [!NOTE]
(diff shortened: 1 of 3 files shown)


AlCalzone:
@robertsLando since you requested this, I'm gonna let you review it

robertsLando (dismissed):
LGTM. @AlCalzone Why does the write doesn't call the command class to also write it to the node?

AlCalzone:
> Why does the write doesn't call the command class to also write it to the node?

Setting properties is synchronous. Updating the value on the node would require an async call (if the CC is supported at all) and the possibility to notify when something went wrong (node dead, node asleep, connectivity issues, etc...).

I mean we could kick off an async process in the background, but that's more or less hoping that setting the property actually has an effect.

robertsLando:
I think that it's better to leave this as you did. Will that property be stored in store?

AlCalzone:
Yep, it will use the (now internal) value IDs from `Node Naming and Location CC` for storage, so changing those through the commandClasses API will have an effect on the property values.

robertsLando:
@AlCalzone Perfect! IMO this is the best solution, it's like the one I have implemented on my side, It's good to have node name and location but unfortunally based on my experience it's not well supported from all devices

PR #1212, opened 2020-12-15
feat(config): support config files without firmware version, load them after Manufacturer Info is known

### LINKED ISSUE #1037: Interview of ID Lock fails
So I tried to include my Remotec ZRC-90, "Scene master", (yes, I do have a lot of Z-Wave devices ;)) but the inclusion fails pretty quickly. Based on the log file attached I don't even think it's being interviewed. I attached my cache-files too, there's nothing in them though.

[remoteczrc90_log.txt](https://github.com/AlCalzone/node-zwave-js/files/5362993/remoteczrc90.txt)
[eabbf325.txt](https://github.com/AlCalzone/node-zwave-js/files/5362996/eabbf325.txt)
[eabbf325.metadata.txt](https://github.com/AlCalzone/node-zwave-js/files/5362997/eabbf325.metadata.txt)
[eabbf325.values.txt](https://github.com/AlCalzone/node-zwave-js/files/5362998/eabbf325.values.txt)

AlCalzone:
With this PR, we add support for config files which explicitly have no firmware version (`firmwareVersion: false`). This allows us to look up those fallback files directly after the `Manufacturer Specific CC` interview, before the `Version CC` is interviewed.

The change is required to support devices like the ID lock, which require compat flags to even complete the Version CC interview.

fixes: #1037

### DIFF
--- a/maintenance/importConfig.ts
+++ b/maintenance/importConfig.ts
@@ -1044,8 +1044,14 @@ function getLatestConfigVersion(
 	configs: DeviceConfigIndexEntry[],
 ): DeviceConfigIndexEntry | undefined {
 	configs.sort((a, b) => {
-		const vA = padVersion(a.firmwareVersion.max);
-		const vB = padVersion(b.firmwareVersion.max);
+		const vA =
+			typeof a.firmwareVersion === "boolean"
+				? "255.255.255"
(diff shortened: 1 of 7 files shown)


robertsLando (dismissed):
Apart from lint issues LGTM :)

Issue #1223, opened 2020-12-18
[bug] Driver logs are not pushed on main STDOUT/STDERR


varet80:
We face an issue with the zwave-js Driver logs. When running the command in our Shell, we can see all logs from zwavejs2mqtt and zwave-js driver in console.
Once we run a Docker container *OR* push stdout/stderr logs to a file, Driver logs missing

Testing this:
Console output:
```
❯ node bin/www
 ______                       _     ___                  _   _
|___  /                      (_)   |__ \                | | | |
   / /_      ____ ___   _____ _ ___   ) |_ __ ___   __ _| |_| |_
  / /\ \ /\ / / _` \ \ / / _ \ / __| / /| '_ ` _ \ / _` | __| __|
 / /__\ V  V / (_| |\ V /  __/ \__ \/ /_| | | | | | (_| | |_| |_
/_____|\_/\_/ \__,_| \_/ \___| |___/____|_| |_| |_|\__, |\__|\__|
                            _/ |                      | |
                           |__/                       |_|
2020-12-17 19:49:14.791 WARN STORE: scenes.json not found
2020-12-17 19:49:14.793 WARN STORE: nodes.json not found
2020-12-17 19:49:15.172 INFO APP: Version: 1.0.0-alpha.0.023fccf
2020-12-17 19:49:15.173 INFO APP: Application path:/home/billias/ongit/zwavejs2mqtt
2020-12-17 19:49:15.218 INFO MQTT: Connecting to mqtt://mqtt.data:1883
2020-12-17 19:49:15.255 INFO ZWAVE: Connecting to /dev/ttyACM0
19:49:15.257 DRIVER   ███████╗ ██╗    ██╗  █████╗  ██╗   ██╗ ███████╗             ██╗ ███████╗
                      ╚══███╔╝ ██║    ██║ ██╔══██╗ ██║   ██║ ██╔════╝             ██║ ██╔════╝
                        ███╔╝  ██║ █╗ ██║ ███████║ ██║   ██║ █████╗   █████╗      ██║ ███████╗
                       ███╔╝   ██║███╗██║ ██╔══██║ ╚██╗ ██╔╝ ██╔══╝   ╚════╝ ██   ██║ ╚════██║
                      ███████╗ ╚███╔███╔╝ ██║  ██║  ╚████╔╝  ███████╗        ╚█████╔╝ ███████║
[...]

PR #1233, opened 2020-12-21
feat: add option to force console logging when not a TTY

AlCalzone:
This PR adds the option `forceConsole` to the logger options, which causes zwave-js to also log to the stdout if it is not connected to a TTY. Setting `logToFile` still has preference though.

fixes: #1223


robertsLando (approved):
LGTM

PR #1248, opened 2020-12-23
feat: change the default retry strategy

### LINKED ISSUE #997: Investigate updating the default retry strategy
Currently, we resend a message up to 3 times when it is either not ACKed by the node or we receive no response. The timer used for those events is the same and is pretty short. It shows that if a device receives a request but does not respond, it is unlikely that it will respond on the next attempt.

A better retry strategy might be the following:
* Attempt sending up to 3 times (configurable count, configurable timeout) if no ACK is received from the node
* By default don't retry sending if an ACK is received and no response, but make this configurable (up to 3 attempts total, including attempts until ACK)
* Increase the node response timeout significantly (10000ms) to give a node time to retry sending the response itself. Currently it is 1600ms.

AlCalzone:
This PR updates the default strategy for retrying SendData commands. For details see #997

fixes: #997


AlCalzone:
Technically not breaking, but can result in worse behavior if node responses often get lost.

PR #1263, opened 2020-12-30
feat: change label for Low battery value to "Low battery level"

varet80:
Make a small label change for low battery.


AlCalzone (approved):
LGTM.

That commit history looks weird though. You might want to hard-reset your master branch to the one of zwave-js. Or branch from `upstream/master`.

PR #1271, opened 2021-01-01
feat(ci): auto-release nightly versions of @zwave-js/config

### LINKED ISSUE #1267: Auto-release nightly versions of @zwave-js/config
if **ONLY** files matching `packages/config/config/**/*.json` were changed since the most recent tag.
**AND** `lerna changed` detects that only `@zwave-js/config` and `zwave-js` (because of the dependency) were changed.

The release should:
* perform the usual checks
* release a new prepatch of `@zwave-js/config`, e.g. `6.0.0` -> `6.0.1-config.yyyymmdd`

By default, lerna would also release a new version of `zwave-js`, but I prefer not to do that.

AlCalzone:
With this PR, we automatically release a new prerelease version of @zwave-js/config every night at 02:00 if the following conditions are met:
1. There are config file changes
2. There are no non-config file changes (except .gitignore)
3. `lerna` tells us that only `@zwave-js/config` and `zwave-js` (because of the dependency) would be published
4. The `lint:config` script succeeds
5. Building `@zwave-js/config` succeeds.

The action does not mirror the version back to git, but chooses a version increase with the least specifity:
`1.0.0 < 1.0.1-20210101 < 1.0.1-alpha.0 < 1.0.1-alpha.0.20200101 < 1.0.1-alpha.1`

This way, versions that are manually released later will always have precedence.

fixes: #1267

### DIFF
--- a/.actrc
+++ b/.actrc
@@ -1,0 +1,1 @@
+-P ubuntu-latest=nektos/act-environments-ubuntu:18.04
--- a/.github/workflows/nightly-config-publish.yml
+++ b/.github/workflows/nightly-config-publish.yml
@@ -1,0 +1,95 @@
+name: Publish nightly configuration updates
+
+on:
+  schedule:
+    - cron: 0 2 * * * # Every day at 02:00
+
+jobs:
+  publish-config:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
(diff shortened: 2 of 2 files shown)


AlCalzone:
Let's just try this...

PR #1282, opened 2021-01-04
feat(config): add checks for partial parameters

AlCalzone:
This PR adds three more checks for partial config parameters:
* Is the value size compatible with the bitmask?
* Is the value size of all related partial params the same?
* Is there a (unnecessary) non-partial config parameter with the same property as a partial one?

### DIFF
--- a/docs/development/config-files.md
+++ b/docs/development/config-files.md
@@ -132,8 +132,8 @@ defines 4 partial parameters that each switch a single bit of parameter #40. Usi
 
 Partial parameters must follow these rules:
 
-1. Each partial parameter must have the same `valueSize`
-1. Each bit mask must fit into the configured `valueSize` of the parameter.
+1. The `valueSize` must be the actual size of the parameter, as defined in the device manual (not just the part of the bitmask). Each partial parameter must have the same `valueSize`.
+1. Each bitmask must fit into the configured `valueSize` of the parameter.
 1. The `minValue`, `maxValue` and `defaultValue` as well as options values are relative to the lowest bit the bit mask. If the bit mask is `0xC` (binary `1100`), these properties must be in the range 0...3 (2 bits). Any required bit shifts are automatically done.
 
 ## `compat`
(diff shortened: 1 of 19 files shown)


AlCalzone:
found a few errors along the way too.

marcus-j-davies:
trying to not add to the list - so saw this, and it may be related?


But setting values (specifically  Configuration Values) via `setValue(ValueID,Value)` is that normal practice?
incorrectly aligns the value to 4 Bytes.

```
17:25:15.253 DRIVER » [Node 002] [REQ] [SendData]
                      │ transmit options: 0x25
                      │ callback id:      17
                      └─[ConfigurationCCSet]
                               parameter #:      31
                               reset to default: false
                               value size:       4
                               value:            6 
```

When in fact, a particular  value might only be 1 byte.
the Configuration Report for a Develo Siren.
[https://products.z-wavealliance.org/ProductManual/File?folder=&filename=Manuals/2308/MT02758_manual.pdf](https://products.z-wavealliance.org/ProductManual/File?folder=&filename=Manuals/2308/MT02758_manual.pdf)

```
17:25:15.320 DRIVER « [Node 002] [REQ] [ApplicationCommand]
                      └─[ConfigurationCCReport]
                               parameter #: 31
                               value size:  1
                               value:       0 
```

Is there a reference to the value size, in the cache anywhere? - it could be used for setValue when targeting the Configuration class?

targeting the CCAPI isn't a problem of course  (and my preferred method) as its asks you for the size.
the value that gets set, is of course ,  incorrect = 0 instead of 6

AlCalzone:
That is configured in the config files. You'll find them in packages/config/config/devices/...
The tricky part will be locating the correct file but you should be able to search for the device label.

AlCalzone:
Oh and feel free to open new issues, this is most likely a bug.

marcus-j-davies:
The config param size states its 4 - manual states 1.
how would you prefer this correction - raise a bug or PR?

> All of the configuration, the data size is 1.  
[MT02758_manual.pdf](https://github.com/zwave-js/node-zwave-js/files/5797942/MT02758_manual.pdf)

config : 0x0175/ph-pse02.json

EDIT
Updating the config correctly aligns the size when using setValue

AlCalzone:
If you have figured it out, pr would be good.

marcus-j-davies:
done - so we'll see ;-)

PR #1295, opened 2021-01-05
feat: export zwave-js package version

### LINKED ISSUE #1288: Expose zwave-js version
Actually would be useful to expose the version of the package

Maybe just export package.json [here](https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/index.ts)

robertsLando:
Fixes #1288

### DIFF
--- a/packages/zwave-js/src/index.ts
+++ b/packages/zwave-js/src/index.ts
@@ -12,6 +12,7 @@ import * as path from "path";
 // locations in error messages
 import "source-map-support/register";
 
+const { version: libVersion } = require("../package.json"); 
 const libraryRootDir = path.join(__dirname, "..");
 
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
Sorry, doesn't work that way. TypeScript wants to import package.json if you do, changing the entire output directory structure.

AlCalzone:
And I prefer if it is called `libVersion` like in Driver.ts, or I'll have two identical exports floating around with different names.

AlCalzone:
Heh, that's tricky :D
Change `../package.json` to `./package.json`. The index file is moved during the build process.

AlCalzone (approved):
Was easier to fix this myself I guess 😅

PR #1296, opened 2021-01-05
feat: implement BatteryCC v3

### LINKED ISSUE #1156: Implement Battery CC V3
Battery CC V3 adds a single field to the Battery Report command:
https://github.com/zwave-js/node-zwave-js/blob/master/specs/SDS13782%20Z-Wave%20Management%20Command%20Class%20Specification.pdf

**TODOs:**
- [x] Parse the new field
- [x] Include it in logging
- [x] update `@implementedVersion` decorator

varet80:
Add support for Battery CC v3 (upgrade from V2

Fixes #1156 

@AlCalzone I will need your review, If I missed something, relatively tiny change!

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/BatteryCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/BatteryCC.ts
@@ -349,13 +349,12 @@ export class BatteryCCReport extends BatteryCC {
 
 	private _lowTemperatureStatus: boolean | undefined;
 	@ccValue({ minVersion: 3 })
-
 	@ccValueMetadata({
 		...ValueMetadata.ReadOnlyBoolean,
-		label: "Battery temperature is low"
+		label: "Battery temperature is low",
 	})
 	public get lowTemperatureStatus(): boolean | undefined {
-		return this._lowTemperatureStatus
+		return this._lowTemperatureStatus;
 	}
 
 	public toLogEntry(): MessageOrCCLogEntry {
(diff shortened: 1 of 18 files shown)


packages/zwave-js/src/lib/commandclass/BatteryCC.ts
```diff
@@ -160,6 +161,7 @@ needs to be replaced or charged: ${
 							batteryStatus.rechargeOrReplace!
 						]
 					}
+is low temperature               ${batteryStatus.lowTemperatureStatus}
```
AlCalzone, line 164: I would call this `temperature is low`

AlCalzone (changes requested):
Just a small change.

And you need to format the file. Either install the recommended VSCode extensions (enabling format-on-save) or run
```
yarn run lint:ts --fix
```

varet80:
there are linting errors coming from other files! which confused. i did run lintfix but still some where not recovereable! and some more changes arrived with linting

Fixed also the change requested

packages/zwave-js/src/lib/commandclass/BatteryCC.ts
varet80: done! added also `:`

AlCalzone:
I can check later. That usually happens when you have different dependency versions than CI.

AlCalzone (approved):

PR #1302, opened 2021-01-06
fix(config): preserve root application CCs for ZHC 5010

### LINKED ISSUE #120: [bug] Endpoint 0 not exposed (Logic-Group ZHC 5010)
**Version**
ZWave2MQTT v1.0.0-alpha.1 (from image zwavejs/zwavejs2mqtt:dev)

Build/Run method
- [X] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build ) 

**Describe the bug**
When including [ZHC5010](http://ww.logichome.dk/index.php/10-support-categories/5-zhc5010-technical-information), only endpoints 1-4 are exposed (in the control panel as well as when inspecting the MQTT-stream and in Home Assistant).

The device consists of an in-built relay (root/endpoint 0) and four buttons (endpoints 1-4).
The relay can be configured to work independently of the four buttons, as decribed in the [manual](http://ww.logichome.dk/plugins/content/cwattachments/cwattachments/helpers/download.php?id=a5bfc9e07964f8dddeb95fc584cd965d&sid=4cf386225c762116a04c1e7a467db8cf):
_"It can be configured so that the relay can be controlled exclusively by sending commands to the root device, which means that the relay isn’t controlled by any of the buttons. In this mode, commands (to Basic Command Class and Binary Switch Command Class) sent to the root device will not be forwarded to device 1, see configuration parameter 15."_

The configuration of paramter 15 is available in the control panel and works fine.
However, the endpoint is not exposed.

**Expected behavior**
The [configuration file](https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x0234/zhc5010_2.0.json) on node-zwave-js seems correct.

The expected behaviour is, that the endpoint is exposed and thereby usable as intended.

**Additional context**
The expected behaviour is seen when including the device in Homey as well as in Fibaro HCL.

AlCalzone:
fixes: https://github.com/zwave-js/zwavejs2mqtt/issues/120

### DIFF
--- a/packages/config/config/devices/0x0234/zhc5010_0.0-1.12.json
+++ b/packages/config/config/devices/0x0234/zhc5010_0.0-1.12.json
@@ -720,5 +720,8 @@
 			"writeOnly": false,
 			"allowManualEntry": true
 		}
+	},
+	"compat": {
+		"preserveRootApplicationCCValueIDs": true
 	}
 }
--- a/packages/config/config/devices/0x0234/zhc5010_2.0.json
+++ b/packages/config/config/devices/0x0234/zhc5010_2.0.json
@@ -840,5 +840,8 @@
 			"writeOnly": false,
 			"allowManualEntry": true
 		}
+	},
+	"compat": {
+		"preserveRootApplicationCCValueIDs": true
 	}
 }


AlCalzone:
@zwave-js-bot pack this

PR #1305, opened 2021-01-06
fix(cc): also guard wakeup interval query against timeout

AlCalzone:
seems to be necessary for FGFS-101


AlCalzone:
@zwave-js-bot pack this

PR #1307, opened 2021-01-06
fix: do not assume that a node falls asleep

AlCalzone:
We have a feature which often causes the communication with battery-powered nodes to seemingly get stuck. After 10s without communication, we assume nodes to be asleep. This is not always correct and causes the node's messages to be moved to the wakeup queue until the next time the node announces that it is awake.

This PR entirely removes this feature.


AlCalzone:
@zwave-js-bot pack this

AlCalzone:
Tests are looking promising

PR #1308, opened 2021-01-06
feat(cc): implement and handle Hail CC

### LINKED ISSUE #1298: [feat] Implement Hail Class (and collect examples for use in doing so)

As discussed in https://github.com/zwave-js/node-zwave-js/issues/1287, the Hail class should be implemented to support older devices and to reduce the need for users to reconfigure their devices. This is a usability issue as devices relying on Hail will not see a state change upon manual control, control via direct association, or when the device intends to send a report.

For users, the Hail class effectively asks the controller to poll devices. **Please collect examples of what values are polled upon a Hail and post them to this issue to assist in implementation.**

Implement new Class/controller actions upon receiving a Hail

AlCalzone:
This PR adds support for the Hail CC. Until we know better what should be done in this case, refresh all Actuator and Sensor values from the hailing node (like we do when a NIF is received without a Lifeline)

fixes: #1298

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/HailCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/HailCC.ts
@@ -1,0 +1,20 @@
+import { CommandClasses } from "@zwave-js/core";
+import {
+	CCCommand,
+	CommandClass,
+	commandClass,
+	implementedVersion,
+} from "./CommandClass";
+
+export enum HailCommand {
+	Hail = 0x01,
+}
+
+// Decorators are applied in the reverse source order, so for @CCCommand to work,
+// it must come before @commandClass
+@CCCommand(HailCommand.Hail)
+@commandClass(CommandClasses.Hail)
+@implementedVersion(1)
+export class HailCC extends CommandClass {
+	declare ccCommand: HailCommand.Hail;
+}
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot pack this

blhoward2:
Ok, today is not my day and I feel like all of the non-technical people I've made fun of in my life. How do I install this? I've been building docker images as I'm running headless and it is easier for me. I thought I could spin a new one, override the command to throw a shell instead of starting, and install this...but it doesn't work and I get errors. Help!?! Is there a ready way to build a generic development docker image that I can use to run these things as we test things?

I tried building with the docker command with the cc/hail branch here, and the fix#23 branch in zwavejs2mqtt, but I still cannot get it to startup that way.

AlCalzone:
@robertsLando?

robertsLando:
@AlCalzone I'm looking at it

blhoward2:
@AlCalzone I cannot access detailed logs because the logger isn't working, but I can confirm that this appears to work on at least one of my devices.  The Aeon Micro Dimmer built into my desk lamp sends a Hail command under the zwavejs2mqtt-dev image and does not update state when I manually change the switch. Running this branch, however, the lamp does change its state correctly in HA when I manually trigger it. Based on that, I think it is working but without logging I cannot confirm that it does. I have other devices I can test it on but need debug logging to do so as the dropping Hail message doesn’t make it to info-level logging which is all I see right now. 

Oddly, I'm not seeing any values changing in the MQTT. But HA is seeing the device go on and off so I’m missing something. I’ll play with this more once logging works again.

blhoward2:
This branch does not presently build due to the recent logging changes in the master branch. PR submitted: #1334 

It also needs to be rebased. #1335

PR #1337, opened 2021-01-10
feat(cc): implement Barrier Operator CC

### LINKED ISSUE #436: [bug] Garage door controller not properly detected
Before submitting a bug please read: <https://zwave-js.github.io/zwavejs2mqtt/#/troubleshooting/bug_report>

### Version

Build/Run method

- [X] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build )

zwavejs2mqtt version: 1.0.3
zwavejs version: 6.1.1

### Describe the bug

Not all attributes are discovered for a garage door controller node after migrating from zwave2mqtt. Specifically, the attributes that allow opening/closing the door, which were properly detected in zwave2mqtt.

### To Reproduce

1. Install and configure zwavejs2mqtt
2. Import nodes.json exported from zwave2mqtt
    - In my case, this file just had names and locations - no other details. Not sure if that's expected or not.
3. Open node details for garage door controller and observe that attributes for opening/closing door are not present

### Expected behavior

When the garage door controller is discovered and attributes populated in zwavejs2mqtt, the attributes for the cover state, and ability to open/close the door, would be discovered, and thus monitorable/controllable over mqtt.

### Additional context

- Controller was originally added securely to zwave2mqtt
- When zwavejs2mqtt loads, it shows that it is still a secure connection
- Screenshots showing configuration details from zwave2mqtt are included, along with logs from zwavejs2mqtt

[logs.zip](https://github.com/zwave-js/zwavejs2mqtt/files/5919291/logs.zip)

#### Node attributes as identified in zwave2mqtt

C6C3100E-A563-4D0A-9AFB-B86F48404E7D
E9312AC9-257A-40A7-A2FF-0963895362EF
6794EFC8-A525-4672-A4BE-A400C1BB5A2C

#### Node attributes as identified in zwavejs2mqtt

D3C3193D-9A19-47F7-B9D2-99B99FBC6B81

varet80:
Add new CC Barrier Operator

Fixes: zwave-js/zwavejs2mqtt#436


packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+
+
+// All possible positions for Report
+export enum BarrierReportState {
+	Closed = 0x00,
+	Closing = 0xFC,
+	Stopped = 0xFD,
+	Openning = 0xFE,
+	Open = 0xFF
+}
```
AlCalzone: This seems to duplicate BarrierPosition. Just use BarrierPosition instead.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * @publicAPI
+ */
+export enum BarrierPosition {
+	Closed = 0x00,
+	Closing = 0xFC,
+	Stopped = 0xFD
+	Openning = 0xFE,
```
AlCalzone: --> `Opening`

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+	Open = 0xFF
+}
+
+/**
+ * @publicAPI
+ */
+export enum BarrierPosition {
+	Closed = 0x00,
+	Closing = 0xFC,
+	Stopped = 0xFD
```
AlCalzone: Missing ,

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// @publicAPI
+export enum SignalType {
+	"Not Supported" = 0x00,
+	Audible = 0x01,
+	Visual = 0x02,
+}
+
+@commandClass(CommandClasses.BarrierOperator)
```
AlCalzone: This is an error, but VSCode offers to correct the spelling for you:
grafik

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+export class BarrierOperatorCC extends CommandClass {
+	declare ccCommand: BarrierOperatorCommand;
+}
+
+// To be reviewed as CC should only accept 0x00 or 0xFF, if more work needed
+@CCCommand(BarrierOperatorCommand.Set)
+export class BarrierOperatorCCSet extends BarrierOperatorCC {
+	public constructor(
+		driver: Driver,
+		options: CommandClassDeserializationOptions | BarrierOperatorCCSetOptions,
```
AlCalzone: BarrierOperatorCCSetOptions is not defined. If you use the `zwcccmd` snippet, there should be an interface definition above the class.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	public targetValue: number;
+
+	public serialize(): Buffer {
+		this.payload = Buffer.from([this.targetValue]);
+		return super.serialize();
+	}
+
+	public toLogEntry(): MessageOrCCLogEntry {
```
AlCalzone, line 410: MessageOrCCLogEntry is not defined like a bunch of other methods. VSCode also offers this quick fix (add all missing imports):
grafik

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+			message: { "target value": this.targetValue },
+		};
+	}
+}
+
+/** Returns the State  */
+interface BarrierOperatorCCGetOptions extends CCCommandOptions {
+	someProperty: number;
+}
+
```
AlCalzone: Remove, its unused

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+@CCCommand(BarrierOperatorCommand.Get)
+@expectedCCResponse(BarrierOperatorCCReport)
```
AlCalzone: Because of this line, Get-type commands should appear after the corresponding Report-type commands in the source file. ESLint will complain about this :)

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+		// return raw value if cannot decode
+		this._barrierPosition = this.payload[0];
+		this.persistValues();
+	}
+
+	private _barrierPosition: BarrierPosition | undefined;
+	@ccValue()
+	@ccValueMetadata({
+		...ValueMetadata.ReadOnlyUInt8,
+		label: "Barrier Operator State",
```
AlCalzone: Shouldn't this rather be `"Barrier Position"`?
I'm thinking whether there's a better name, since the value doesn't just reflect the position, but also the status.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+		states: enumValuesToMetadataStates(BarrierPosition)
+	})
+
```
AlCalzone, line 456: I'd remove empty lines between decorators and methods (to make it clear they belong together)

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+	public get barrierPosition(): BarrierPosition | undefined {
+		return this._barrierPosition;
+	}
+}
+
+@CCCommand(BarrierOperatorCommand.CapabilitiesGet)
+@expectedResponse(BarrierOperatorCCCapabilitiesReport)
+export class BarrierOperatorCCCapabilitiesGet extends BarrierOperatorCC {}
```
AlCalzone: In order to avoid confusion in case there's another Capabilities command added in the future, these two commands (and their `BarrierOperatorCommand` enum values should be called `...SignalingCapabilities...` instead of just `...Capabilities...`.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+	) {
+		super(driver, options);
+		
+		validatePayload(this.payload.length >=1);
+		const bitMaskLength = this.payload.length;
+		validatePayload(this.payload.length >= bitMaskLength);
+		this._supportedSignalTypes = parseBitMask(
+			this.payload.slice(0, bitMaskLength),
+			SignalType['Not Supported'],
+		);
```
AlCalzone: Not quite correct:
The payload length must be `>= 1 + bitMaskLength`, since the bitmask starts at the second byte.
The bitmask starts with value 1 at bit 0, so the 2nd argument to `parseBitMask` must be `SignalType.Audible`.

I'd probably remove the signal type `Not Supported` altogether if it is not used elsewhere.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+
+		this this.persistValues();
```
AlCalzone: typo

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	private _supportedSignalTypes: SignalType[];
+	@@ccValue({ internal: true})
```
AlCalzone: The `@` is duplicated

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
AlCalzone: I think it would make sense to split this into `position` and `state`.
If the value is either 0..99 or 0xff, `position` would be 0..99 or 100 (instead of 0xff) and `state` would be closed, stopped or open.
If the value is one of the other special values, `position` would be `undefined` and `state` would be that special value.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
AlCalzone: Should be called `BarrierState` instead IMO.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,157 @@
...
+			this.targetValue = options.targetValue;
+		}
+	}
+
+	public targetValue: number;
+
+	public serialize(): Buffer {
+		this.payload = Buffer.from([this.targetValue]);
+		return super.serialize();
+	}
```
AlCalzone: In order to be able to set this through the value that reports the state, I think we should do what I proposed below and call the property `state`. In the `BarrierOperatorCCSetOptions` you'd only allow `BarrierState.Open` and `BarrierState.Closed`.
```
node.commandClasses["Barrier Operator"].set(true); // or false, there's no inbetween
```

AlCalzone (changes requested):
See other comments. Remaining TODOs:

- [x] Interview procedure
- [x] `SET_VALUE` API

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: now is renamed to BarrierState! as makes more sense

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: thnx! agree

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: yeah it is smarter than me :) thanks

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: I already have found this and some more, defined some missing ones.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: still scanning these! and adding them

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: already been removed

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: good one

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: :)

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: outdated lets see if the new update is what should be. Thank you for the detailed explanation

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: Good one!

AlCalzone:
closing and reopening to test our bot

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,267 @@
...
+			throw new ZWaveError(
+				`${this.constructor.name}: deserialization not implemented`,
+				ZWaveErrorCodes.Deserialization_NotImplemented,
+			);
+		} else {
+			this.targetValue = options.targetValue;
+		}
+	}
+
+	public targetValue: boolean;
```
AlCalzone: Thinking about it again, this property should have the same name and type as the corresponding one in the report (`barrierState`). You can specify the type `BarrierState.Open | BarrierState.Closed` on `BarrierOperatorCCSetOptions` to ensure only those get used.

AlCalzone: I would however rename `barrierState` to `state` and `barrierPosition` to `position`.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,267 @@
...
+			this._barrierPosition = payloadPosition;
+			if (payloadValue > 0 && payloadValue <= 99) {
+				this._barrierState = undefined;
+			} else {
+				this._barrierState = payloadValue;
+			}
+		} else {
+			this._barrierPosition = undefined;
+			this._barrierState = payloadValue;
+		}
```
AlCalzone: This could probably be simplified logic-wise but let's keep it like this for now.

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: changed all:
BarrierState -> State
(_)barrierPosition -> (_)position
(_)barrierState -> (_)state

shorter names

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -1,70 +1,79 @@
...
+}
+
+@commandClass(CommandClasses["Barrier Operator"])
 @implementedVersion(1)
 export class BarrierOperatorCC extends CommandClass {
 	declare ccCommand: BarrierOperatorCommand;
 }
 
+interface BarrierOperatorCCSetOptions extends CCCommandOptions {
+	targetValue: number;
```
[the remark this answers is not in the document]
varet80: will make it easier for users

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: I simplified the logic on my next commit

robertsLando:
@billiaz Remove yarn-error.log file from the PR

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -36,7 +36,7 @@ export enum BarrierOperatorCommand {
 /**
  * @publicAPI
  */
-export enum BarrierState {
+export enum State {
```
AlCalzone: The enum name was fine

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -81,22 +81,22 @@ export class BarrierOperatorCCSet extends BarrierOperatorCC {
...
+			this.state = options.state;
 		}
 	}
 
-	public targetValue: boolean;
+	public state: State.Open | State.Closed;
 
 	public serialize(): Buffer {
-		const payload: number[] = [this.targetValue ? 0xff : 0x00];
+		const payload: number[] = [this.state ? 0xff : 0x00];
```
AlCalzone: `this.state` is already either `0xff` or `0x00`, so you can just do `this.payload = Buffer.from([this.state]);`

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -81,22 +81,22 @@ export class BarrierOperatorCCSet extends BarrierOperatorCC {
...
 		this.payload = Buffer.from(payload);
 		return super.serialize();
 	}
 
 	public toLogEntry(): MessageOrCCLogEntry {
 		return {
 			...super.toLogEntry(),
-			message: { "target value": this.targetValue },
+			message: { "state value": this.state },
```
AlCalzone: `"state value"` sounds weird.

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: reverted this

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: `"target value"` or `"target state"` ?  
I prefer the later

AlCalzone: target state 👍🏻 
(or just state)

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -216,25 +211,66 @@ export class BarrierOperatorCCEventSet extends BarrierOperatorCC {
...
 		}
 	}
+	public type: SignalType;
+	public state: SignalState;
 
 	public serialize(): Buffer {
 		this.payload = Buffer.from([
-			/* TODO: serialize */
+			(this.payload[0] = this.type),
+			(this.payload[1] = this.state),
```
AlCalzone:
```suggestion
			this.type,
			this.state,
```

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -216,25 +211,66 @@ export class BarrierOperatorCCEventSet extends BarrierOperatorCC {
...
+		super(driver, options);
+
+		validatePayload(this.payload.length >= 2);
+		this._type = this.payload[0];
+		this._state = this.payload[1];
+
+		this.persistValues();
+	}
+
+	private _type: SignalType | undefined;
```
AlCalzone: You're definitely setting a value in the constructor, so this should not include `| undefined`.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -216,25 +211,66 @@ export class BarrierOperatorCCEventSet extends BarrierOperatorCC {
...
+	@ccValue()
+	@ccValueMetadata({
+		...ValueMetadata.ReadOnlyUInt8,
+		label: "Event Signal Type",
+		states: enumValuesToMetadataStates(SignalType),
+	})
+	public get type(): SignalType | undefined {
```
AlCalzone: dito

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -216,25 +211,66 @@ export class BarrierOperatorCCEventSet extends BarrierOperatorCC {
...
+		return this._type;
+	}
+
+	private _state: SignalState | undefined;
```
AlCalzone: dito

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -216,25 +211,66 @@ export class BarrierOperatorCCEventSet extends BarrierOperatorCC {
...
+	@ccValue()
+	@ccValueMetadata({
+		...ValueMetadata.ReadOnlyUInt8,
+		label: "Event Signal State",
+		states: enumValuesToMetadataStates(SignalState),
+	})
+	public get state(): SignalState | undefined {
```
AlCalzone: dito

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -216,25 +211,66 @@ export class BarrierOperatorCCEventSet extends BarrierOperatorCC {
...
+}
+
 interface BarrierOperatorCCEventGetOptions extends CCCommandOptions {
 	signalType: SignalType;
 }
 
 @CCCommand(BarrierOperatorCommand.EventGet)
-@expectedCCResponse(BarrierOperatorCCReport)
+@expectedCCResponse(BarrierOperatorCCEventReport)
 export class BarrierOperatorCCEventGet extends BarrierOperatorCC {
```
AlCalzone: You forgot the the `/* TODO: serialize */` in `serialize` :)

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -105,8 +112,7 @@ export class BarrierOperatorCCAPI extends CCAPI {
 			this.commandOptions,
 		))!;
 		return {
-			// interpret unknown values as false
-			currentValue: response.currentValue,
+			state: response.state,
```
AlCalzone: `BarrierOperatorCCReport` has two properties that work hand in hand - so this should return both.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -141,6 +147,72 @@ export class BarrierOperatorCCAPI extends CCAPI {
...
+		const cc = new BarrierOperatorCCEventSet(this.driver, {
+			nodeId: this.endpoint.nodeId,
+			endpoint: this.endpoint.index,
+			signalType,
+			signalState,
+		});
+		if (this.isSinglecast()) {
+			// remember the value in case the device does not respond with a target value
+			await this.getEvent(signalType);
+		}
```
AlCalzone: This doesn't make sense here.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -261,6 +332,16 @@ export class BarrierOperatorCCReport extends BarrierOperatorCC {
 	public get position(): number | undefined {
 		return this._position;
 	}
+
+	public toLogEntry(): MessageOrCCLogEntry {
+		return {
+			...super.toLogEntry(),
+			message: {
+				"barrier position": this.position,
+				"barrier state": this.state, // I think i need to enumerate this
```
AlCalzone: yep: `getEnumMemberName(BarrierState, this.state)`

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -291,6 +372,15 @@ export class BarrierOperatorCCCapabilitiesReport extends BarrierOperatorCC {
 	public get supportedSignalTypes(): readonly SignalType[] {
 		return this._supportedSignalTypes;
 	}
+
+	public toLogEntry(): MessageOrCCLogEntry {
+		return {
+			...super.toLogEntry(),
+			message: {
+				"supported types": this.supportedSignalTypes,
```
AlCalzone: Needs to be enumerated too (and printed as a list). Look at MultilevelSensor, this has something similar.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -329,6 +419,16 @@ export class BarrierOperatorCCEventSet extends BarrierOperatorCC {
 		this.payload = Buffer.from([this.type, this.state]);
 		return super.serialize();
 	}
+
+	public toLogEntry(): MessageOrCCLogEntry {
+		return {
+			...super.toLogEntry(),
+			message: {
+				"signal type": this.type,
+				"signal state": this.state,
```
AlCalzone: use `getEnumMemberName` (for both I think)

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -367,6 +467,16 @@ export class BarrierOperatorCCEventReport extends BarrierOperatorCC {
...
 		return this._state;
 	}
+
+	public toLogEntry(): MessageOrCCLogEntry {
+		return {
+			...super.toLogEntry(),
+			message: {
+				"signal type": this.type,
+				"signal state": this.state,
+			},
```
AlCalzone: dito

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -402,4 +512,13 @@ export class BarrierOperatorCCEventGet extends BarrierOperatorCC {
 		]);
 		return super.serialize();
 	}
+
+	public toLogEntry(): MessageOrCCLogEntry {
+		return {
+			...super.toLogEntry(),
+			message: {
+				"signal type": this.signalType,
```
AlCalzone: dito

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -402,4 +512,13 @@ export class BarrierOperatorCCEventGet extends BarrierOperatorCC {
 		]);
 		return super.serialize();
 	}
+
```
AlCalzone, line 660: above this line, the `serialize` method is not completed yet.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -367,6 +467,16 @@ export class BarrierOperatorCCEventReport extends BarrierOperatorCC {
 	public get state(): SignalState {
 		return this._state;
 	}
+
```
AlCalzone, line 608: Question: Does BarrierOperatorCCEventSet allow configuring a different signal state for every type? Then the implementation is not correct.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -141,6 +147,72 @@ export class BarrierOperatorCCAPI extends CCAPI {
...
+			signalType,
+		});
+		const response = (await this.driver.sendCommand<BarrierOperatorCCEventReport>(
+			cc,
+			this.commandOptions,
+		))!;
+		return {
+			signalType: response.type,
+			signalState: response.state,
+		};
```
AlCalzone: I think you're retrieving the signalState for a signalType. Thus the method should only return the signalState.

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: you right!

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: map and join... i think the next commit should be fixing that! thanks!

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: done my best! might be aweful or no sense!

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: you right! enumerated it also

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: I confused myself
I think i need to move this after the sendCommand. AFter sending an event do i need to get the Event value?

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
varet80: every type has on and off state

2 types: Audible/Visual and Common states: On/Off
Screenshot 2021-01-16 170024

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
AlCalzone: Uhh, yes. This after the sendCommand and remove the other thing.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -141,6 +147,72 @@ export class BarrierOperatorCCAPI extends CCAPI {
...
+		if (this.isSinglecast()) {
+			// remember the value in case the device does not respond with a target value
+			await this.getEvent(signalType);
+		}
+		await this.driver.sendCommand(cc, this.commandOptions);
+
+		if (this.isSinglecast()) {
+			// Refresh the current value
+			await this.get();
+		}
```
AlCalzone:
```suggestion
```

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
AlCalzone: So, in that case, state and type are not two separate properties. Instead you have a map where `type` is the key and `state` the value for that key. This also makes the SET_VALUE API more complicated.

An example would be the `Color Switch CC` (`targetColor` and `currentColor`).
==> https://github.com/zwave-js/node-zwave-js/blob/e1c251f5430f5637f1948d9443068d8c479e586f/packages/zwave-js/src/lib/commandclass/ColorSwitchCC.ts#L265-L287
and
==> https://github.com/zwave-js/node-zwave-js/blob/e1c251f5430f5637f1948d9443068d8c479e586f/packages/zwave-js/src/lib/commandclass/ColorSwitchCC.ts#L494-L500

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -132,6 +139,86 @@ export class BarrierOperatorCCAPI extends CCAPI {
...
+		} else {
+			switch (property) {
+				case "state":
+				case "signalType":
+				case "signalState":
+					await this.set(value);
+					break;
+				default:
+					throwUnsupportedProperty(this.ccId, property);
+			}
```
AlCalzone: Yeah, that won't work. You have to use differnet commands for different properties. See the linked examples from the other comment.

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,514 @@
...
+		if (this.isSinglecast()) {
+			// remember the value in case the device does not respond with a target value
+			this.endpoint
+				.getNodeUnsafe()
+				?.valueDB.setValue(
+					getStateValueId(this.endpoint.index),
+					state,
+					{ noEvent: true },
+				);
+		}
```
AlCalzone:
```suggestion
```

robertsLando:
Fixes https://github.com/zwave-js/zwavejs2mqtt/issues/436

AlCalzone:
Finished the missing things and rebased onto master, so we're up to date.

AlCalzone:
@zwave-js-bot pack this

AlCalzone (dismissed):

kpine:
I tested this PR, here are my observations.

First, deviceClasses.json is missing many specific types for Entry Control device class, including type 0x7 `"Secure Barrier Addon"` ([SPECIFIC_TYPE_SECURE _BARRIER_ADDON](https://github.com/Z-WavePublic/libzwaveip/blob/ade735d6b8f364716334c89f4593b22cc5220d92/config/ZWave_custom_cmd_classes.xml#L26)) reported by my device (shows up as `"UNKNOWN (0x07)"`). HA uses this to discover that it's a garage door opener.

Second, there are Barrier values populated now, but they are all read-only so I can't actually control anything.

image

Lastly, I did manually open and close the garage door, and it looks like the barrier state is being determined correctly.

Attached are log and cache files, including an HA dumpfile. Threw in the ozw cache file just in case it's useful. The device is node number 12.

[logs.zip](https://github.com/zwave-js/node-zwave-js/files/5923578/logs.zip)

varet80:
@kpine home assisfNt is the next phase.
Now the only part missong is setting a state on the device.
@AlCalzone i suspect i did something, in my first commits, wrong with the targetValue.

kpine:
> @kpine home assisfNt is the next phase.

Sorry, I meant a zwave-js-server dump file (which was downloaded from HA). I am not attempting to use it with HA yet (obviously it can't be supported without this PR).

AlCalzone:
Barrier State should be writable. If not, we missed a flag somewhere.

pittbull:
I am trying to test this build but I am not fluent in docker/github yet. Could I get some guidance as to how my docker-compose file needs to be setup for it to pull this build?

kpine:
> 
> 
> I am trying to test this build but I am not fluent in docker/github yet. Could I get some guidance as to how my docker-compose file needs to be setup for it to pull this build?

Make a custom docker build. I'm using the `local-copy-src` option and just cloned both repos and cherry-picked this PR.

https://zwave-js.github.io/zwavejs2mqtt/#/troubleshooting/docker_custom_builds

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,641 @@
...
+			this.position = 100;
+			this.state = payloadValue;
+		}
+
+		this.persistValues();
+	}
+
+	@ccValue()
+	@ccValueMetadata({
+		...ValueMetadata.ReadOnlyUInt8,
```
kpine, line 451: Not familiar with typescript or this code much yet so... does this ReadOnlyUInt8 mean the value id is not writeable? If so, is this value supposed to be used for reporting and setting the state?

AlCalzone: good catch. That implies `writeable: false`.

AlCalzone:
@zwave-js-bot pack this again for the npm users

edgauthier:
With the latest changes, I am able to open/close the garage door.

kpine:
It's working for me as well, but I have a few comments...

First, I now understand why the barrier state was originally read-only (maybe it came up in the review comments). The "state" property represented the "Barrier Report::State" (4.12.4 Barrier Operator Report Command), which has Closed, Closing, Stopped, Opening, Open values.

The "Barrier Set::Target Value" has only two valid states. CLOSE and OPEN. So the "state" property is now a combination of both states. In order to open the door I can use "Open" value, but to close I have to use "Closed". I can then attempt to set the state to "Opening" or any of the others, which apparently gets sent to the device, which it ignores. Technically I think the spec forbids the controller from doing this (CC:0066.01.01.11.003 All other values are reserved and MUST NOT be used by a sending node.)

It feels to me like there should be write-only targetValue that can be set only with "Close" and "Open", and a read-only "state" value that reflects the Barrier Report::State? But, an application like HA would just use the "Closed" and "Open" and hide the invalid states. 🤷‍♂️ 

Log attached in case you're interested. [zwave.log](https://github.com/zwave-js/node-zwave-js/files/5930711/zwave.log)

kpine:
> It feels to me like there should be write-only targetValue that can be set only with "Close" and "Open", and a read-only "state" value that reflects the Barrier Report::State? But, an application like HA would just use the "Closed" and "Open" and hide the invalid states.

As a counter argument, a single value for set/report is the same way OZW is implemented and it just [protects](https://github.com/OpenZWave/open-zwave/blob/f42e4f58572f38f6e4c2b95a74ab445af444f286/cpp/src/command_classes/BarrierOperator.cpp#L314-L316) the Set from setting an invalid value (it also allows any value except Closed to mean open 😄 ). So maybe just prevent bad target values and call it a day? (I should have looked closer at the code, there's a `// TODO: When written, this should only accept OPEN and CLOSE`).

edgauthier:
HA is my front end, and considering that the previous system worked the way it did (using the single state attribute, and protections/handling for the received states), my vote would be to follow that model. I would assume other front ends are likely to expect that behavior. 

I'm still new to this though, just wanted to provide my input.

kpine:
My only concern is that there could be a short time that the value transitions in the wrong order:

1. State is Closed
2. Set State to Open (to open door).
3. For a short time, garage door appears to have a state of Open. A UI might display or application send events that the garage door is open, depending on how it uses the values internally.
4. Barrier sends report, it's now in Opening state
5. Barrier sends report, it's now in Open state

So the value went from Closed -> Open -> Opening -> Open, or Open -> Closed -> Closing -> Closed. Maybe some devices skip reporting the transitional stage, in that case it doesn't matter. I also don't see any problem with this behavior in HA and ozw. Perhaps HA only records state changes when it gets some value update from zwave-js vs. an internal change. So maybe I'm overthinking it.

AlCalzone:
I would be fine with separating it into `current/target`. I'm just cautious with that, since it has sparked a lot of confusion for the switch classes.

ratsputin:
Using this PR version of zwavejs, I'm seeing the following error:
```
    0: ExitFrame [pc: 0x43066be8]
Security context: 0x4d89e09d <JSObject>
    1: byteLength(aka byteLength) [0x3c6fc4dd] [buffer.js:530] [bytecode=0x39ff8b3d offset=204](this=0x222042f1 <undefined>,string=0x396cd935 <Very long string[174306121]>,encoding=0x4d8c6481 <String[4]: utf8>)
    2: arguments adaptor frame: 3->2
    3: fromString(aka fromString) [0x2ad794dd] [buffer.js:341] [bytecode=0x39ff5725 offset=74](this=0x222042f1 <undefine...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
Aborted
```

jaburges:
> I would be fine with separating it into `current/target`. I'm just cautious with that, since it has sparked a lot of confusion for the switch classes.

I'd prefer write states = Opened / Closed
Read states = Opened / Opening / Closed / Closing (and stopped i think is the additional)

Reason being is I have routines that run on 'Opening' as it takes some time for the garage door to actually open and change state to "Opened".

AlCalzone:
With the most recent commit, we now distinguish between the `targetState` for commanding and `currentState` and reading the status.

kpine:
Just an observation, with the latest commit, if I set the target state to Open while the door is already in the Open state, it transitions to Opening, and remains there until I really close it. Not sure what to think of that.

kpine:
Would you be able to add the specific device classes for barrier operators into this PR?

AlCalzone:
> Just an observation, with the latest commit, if I set the target state to Open while the door is already in the Open state, it transitions to Opening, and remains there until I really close it. Not sure what to think of that.

Yeah, that's not good. I think I'll just remove the manual update of `currentState` in favor of a poll.

> Would you be able to add the specific device classes for barrier operators into this PR?

Got a link or something?

pittbull:
It is still showing up as a switch and nothing more when using the latest build. Is this correct?

kpine:
> 
> 
> It is still showing up as a switch and nothing more when using the latest build. Is this correct?

If you are talking about Home Assistant, garage door covers are not implemented yet (it requires this PR), so yes you only will see a switch.

pittbull:
> > It is still showing up as a switch and nothing more when using the latest build. Is this correct?
> 
> If you are talking about Home Assistant, garage door covers are not implemented yet (it requires this PR), so yes you only will see a switch.

I was under the impression I was using this PR - as per earlier comment about building it locally?

kpine:
> Got a link or something?

I've got a few, but [SDS13740](https://raw.githubusercontent.com/zwave-js/specs/master/SDS13740-1%20-%20Z-Wave%20Plus%20Device%20and%20Command%20Class%20Types%20and%20Defines%20Specification.pdf) seems to be the most authoritative.

```
/* Device class Entry Control */
#define GENERIC_TYPE_ENTRY_CONTROL                      0x40 /*Entry Control*/
#define SPECIFIC_TYPE_NOT_USED                          0x00 /*Specific Device Class Not Used*/
#define SPECIFIC_TYPE_DOOR_LOCK                         0x01 /*Door Lock*/
#define SPECIFIC_TYPE_ADVANCED_DOOR_LOCK                0x02 /*Advanced Door Lock*/
#define SPECIFIC_TYPE_SECURE_KEYPAD_DOOR_LOCK           0x03 /*Door Lock (keypad –lever) Device Type*/
#define SPECIFIC_TYPE_SECURE_KEYPAD_DOOR_LOCK_DEADBOLT  0x04 /*Door Lock (keypad –deadbolt) Device Type*/
#define SPECIFIC_TYPE_SECURE_DOOR                       0x05 /*Barrier Operator Specific Device Class*/
#define SPECIFIC_TYPE_SECURE_GATE                       0x06 /*Barrier Operator Specific Device Class*/
#define SPECIFIC_TYPE_SECURE_BARRIER_ADDON              0x07 /*Barrier Operator Specific Device Class*/
#define SPECIFIC_TYPE_SECURE_BARRIER_OPEN_ONLY          0x08 /*Barrier Operator Specific Device Class*/
#define SPECIFIC_TYPE_SECURE_BARRIER_CLOSE_ONLY         0x09 /*Barrier Operator Specific Device Class*/
#define SPECIFIC_TYPE_SECURE_LOCKBOX                    0x0A /*SDS12724*/
#define SPECIFIC_TYPE_SECURE_KEYPAD                     0x0B
```

There are also textual descriptions of those specific types in SDS11847. I've also found them (with better descriptions) publicly in [libzwaveip](https://github.com/Z-WavePublic/libzwaveip/blob/ade735d6b8f364716334c89f4593b22cc5220d92/config/ZWave_custom_cmd_classes.xml#L18-L31) and [OpenZWave](https://github.com/OpenZWave/open-zwave/blob/f42e4f58572f38f6e4c2b95a74ab445af444f286/config/device_classes.xml#L127-L142).
```
  <gen_dev key="0x40" name="GENERIC_TYPE_ENTRY_CONTROL" help="Entry Control" comment="Entry Control">
    <spec_dev key="0x00" name="SPECIFIC_TYPE_NOT_USED" help="Not Used" comment="Specific Device Class Not Used" />
    <spec_dev key="0x01" name="SPECIFIC_TYPE_DOOR_LOCK" help="Door Lock" comment="Door Lock" />
    <spec_dev key="0x02" name="SPECIFIC_TYPE_ADVANCED_DOOR_LOCK" help="Advanced Door Lock" comment="Advanced Door Lock" />
    <spec_dev key="0x03" name="SPECIFIC_TYPE_SECURE_KEYPAD_DOOR_LOCK" help="Secure Keypad Door Lock" comment="Door Lock (keypad –lever) Device Type" />
    <spec_dev key="0x04" name="SPECIFIC_TYPE_SECURE_KEYPAD_DOOR_LOCK_DEADBOLT" help="Door Lock Keypad Deadbolt" comment="Door Lock (keypad – deadbolt) Device Type" />
    <spec_dev key="0x05" name="SPECIFIC_TYPE_SECURE_DOOR" help="Secure Door" comment="Barrier Operator Specific Device Class" />
    <spec_dev key="0x06" name="SPECIFIC_TYPE_SECURE_GATE" help="Secure Gate" comment="Barrier Operator Specific Device Class" />
    <spec_dev key="0x07" name="SPECIFIC_TYPE_SECURE_BARRIER_ADDON" help="Secure Barrier Addon" comment="Barrier Operator Specific Device Class" />
    <spec_dev key="0x08" name="SPECIFIC_TYPE_SECURE_BARRIER_OPEN_ONLY" help="Secure Barrier Open Only" comment="Barrier Operator Specific Device Class" />
    <spec_dev key="0x09" name="SPECIFIC_TYPE_SECURE_BARRIER_CLOSE_ONLY" help="Secure Barrier Close Only" comment="Barrier Operator Specific Device Class" />
    <spec_dev key="0x0A" name="SPECIFIC_TYPE_SECURE_LOCKBOX" help="Secure Lockbox" comment="SDS12724" />
    <spec_dev key="0x0B" name="SPECIFIC_TYPE_SECURE_KEYPAD" help="Secure Keypad" />
  </gen_dev>
```

My device currently shows up as `"UNKNOWN (0x07)"` which means it's a `"Secure Barrier Addon"`. HA uses this string to match the device to applicable entities.

kpine:
> I was under the impression I was using this PR - as per earlier comment about building it locally?

HA is separate software. If you are using this PR, then you need to implement the discovery schema and Cover class in HA. As mentioned, this is not done yet as we need this PR.

pittbull:
> > I was under the impression I was using this PR - as per earlier comment about building it locally?
> 
> HA is separate software. If you are using this PR, then you need to implement the discovery schema and Cover class in HA. As mentioned, this is not done yet as we need this PR.

Understood. Sorry for the confusion.

kpine:
> Understood. Sorry for the confusion.

No worries. I have a basic implementation on my [fork](https://github.com/kpine/core/commits/zwave-js-barrier) if you're eager to try.

AlCalzone:
@kpine gotcha. I think there are some others of the new ZW+ types missing. I might do that in a separate PR.

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,672 @@
...
+			}
+			if (typeof value !== "number") {
+				throwWrongValueType(
+					this.ccId,
+					property,
+					"number",
+					typeof value,
+				);
+			}
+			await this.setEventSignaling(propertyKey, value);
```
kpine, line 261: As mentioned in the other comment, this works fine using a `signalingState` property that is marked as read-only.

I hate to be the one to suggest it, but should this be refreshed? Or should the application assume it is set, or request its own refresh?

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
```diff
@@ -0,0 +1,672 @@
...
+		property: "signalingState",
+		propertyKey: subsystemType,
+	};
+}
+
+function getSignalingStateMetadata(
+	subsystemType: SubsystemType,
+): ValueMetadata {
+	return {
+		...ValueMetadata.ReadOnlyUInt8,
```
kpine: What is the significance of this being read only? In zwavejs2mqtt it's not editable. However, I can call the set API and modify it successfully. Should it also be writable?

Here's what zwave-js-server shows me, for reference:
```json
    {
      "endpoint": 0,
      "commandClass": 102,
      "commandClassName": "Barrier Operator",
      "property": "signalingState",
      "propertyKey": 1,
      "propertyName": "signalingState",
      "propertyKeyName": "1",
      "ccVersion": 0,
      "metadata": {
        "type": "number",
        "readable": true,
        "writeable": false,
        "min": 0,
        "max": 255,
        "label": "Signaling State (Audible)",
        "states": {
          "0": "Off",
          "255": "On"
        }
      },
      "value": 255
    },
```

AlCalzone: 🙈 yes that should be writable too

packages/zwave-js/src/lib/commandclass/BarrierOperatorCC.ts
AlCalzone: signalingState is now writable. As with many other values, I've decided not to verify with a poll. If setting the value does not fail, applications should assume that it worked. If this is **really, really** critical somehow, there's always the option to poll on demand.

PR #1338, opened 2021-01-10
feat(config): add Fibaro Smart Switch FGS214

LaplaceII:
@AlCalzone : Based on your config file for Fibaro Double Smart Switch FGS224 I adopted it to Single Smart Switch FGS224. Please check....I'm not an Expert!


AlCalzone:
Thanks, looks good!

PR #1351, opened 2021-01-11
fix(config): correction to config value size

marcus-j-davies:
As discussed.
config value size and some defaults were incorrect according the manual.

### DIFF
--- a/packages/config/config/devices/0x0175/ph-pse02.json
+++ b/packages/config/config/devices/0x0175/ph-pse02.json
@@ -22,7 +22,7 @@
 			"description": "NotificationReport/BinaryReport",
 			"valueSize": 1,
 			"minValue": 0,
-			"maxValue": 65535,
+			"maxValue": 255,
 			"defaultValue": 8,
 			"readOnly": false,
 			"writeOnly": false,
--- a/packages/config/config/devices/0x0175/ph-pse02.json
+++ b/packages/config/config/devices/0x0175/ph-pse02.json
@@ -1,55 +1,85 @@
 // Devolo ph-pse02
 // Multisound indoor siren
 {
-	"manufacturer": "Devolo",
-	"manufacturerId": "0x0175",
-	"label": "ph-pse02",
-	"description": "Multisound indoor siren",
-	"devices": [
-		{
-			"productType": "0x0004",
-			"productId": "0x000a"
-		}
-	],
-	"firmwareVersion": {
-		"min": "0.0",
-		"max": "255.255"
-	},
-	"supportsZWavePlus": true,
-	"paramInformation": {
-		"7": {
-			"label": "Costumer  Function",
-			"description": "NotificationReport/BinaryReport",
-			"valueSize": 1,
-			"minValue": 0,
-			"maxValue": 255,
-			"defaultValue": 8,
-			"readOnly": false,
-			"writeOnly": false,
+    "manufacturer": "Devolo",
+    "manufacturerId": "0x0175",
+    "label": "ph-pse02",
+    "description": "Multisound indoor siren",
+    "devices": [
+        {
+            "productType": "0x0004",
+            "productId": "0x000a"
+        }
+    ],
+    "firmwareVersion": {
+        "min": "0.0",
+        "max": "255.255"
+    },
+    "supportsZWavePlus": true,
+    "paramInformation": {
+        "7[0x04]": {
+            "label": "Disable Sound",
+            "description": "Prevents audio from the siren - leaving only the strobe function.",
+            "valueSize": 1,
+            "minValue": 0,
+            "maxValue": 1,
+            "defaultValue": 1,
+            "readOnly": false,
+            "writeOnly": false,
 			"allowManualEntry": true,
-			"unsigned": true
-		},
-		"29": {
-			"label": "Disable Alarm",
-			"description": "Disable the alarm  function.",
-			"valueSize": 1,
-			"minValue": 0,
-			"maxValue": 1,
-			"defaultValue": 0,
-			"readOnly": false,
-			"writeOnly": false,
-			"allowManualEntry": true
-		},
-		"31": {
-			"label": "Alarm Duration",
-			"description": "Play alarm sound duration.",
-			"valueSize": 1,
-			"minValue": 0,
-			"maxValue": 127,
-			"defaultValue": 6,
-			"readOnly": false,
-			"writeOnly": false,
-			"allowManualEntry": true
-		}
-	}
+            "options": [
+                {
+                    "label": "Enables Sound",
+                    "value": 0
+                },
+                {
+                    "label": "Disables Sound",
+                    "value": 1
+                }
+            ]
+        },
+        "7[0x08]": {
+            "label": "Notification Type",
+            "description": "Notification event type to use for trigger events",
+            "valueSize": 1,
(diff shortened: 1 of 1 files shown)


AlCalzone:
To verify, these are the allowed values for this parameter?
https://github.com/OpenZWave/open-zwave/blob/370bc03c6e849883a0576e1f0b6c324e969c745f/config/philio/pse02.xml#L34-L37

If so, then I'm leaning towards splitting this into two partial params.

marcus-j-davies:
> To verify, these are the allowed values for this parameter?
> https://github.com/OpenZWave/open-zwave/blob/370bc03c6e849883a0576e1f0b6c324e969c745f/config/philio/pse02.xml#L34-L37
> 
> If so, then I'm leaning towards splitting this into two partial params.

param 7 is actually a bit mask.

bit 0-2    : Reserved (not sure what for)
bit 3        : Disable Sound
bit 4.       : Notification type (0: Using Notification Report, 1: Using Sensor Binary Report)
bit 5-7    : Reserved (not sure what for) 

bits 0-2, 5-7 I would think have no affect.

I think it's safer to keep the param as it is, and just go with the length fix.
theoretically the max value is 255 - depending on the combination?


Screenshot 2021-01-11 at 22 19 09

AlCalzone:
The length fix is fine, you just need to make the config parameter `unsigned`. I am opposed to the practice of stuffing multiple unrelated options into one config parameter. By splitting these into two partials (each targeting the single bit) it makes more sense for the user in the end.

Do you want to take a stab at this or should I?

marcus-j-davies:
I'll give it a go, need to get more into this project.
any documentation/examples of how to add 'quirks' to the configs? 

EDIT:
2nd thought - my Typescript game, isn't up to scratch yet.
JS, I Got
.Net, I eat for breakfast
Typescript - what in the world!!(Evan though I know its JS)

I'll add the unsigned flag now, but I think its safer (and probably less time consuming) to leave in the hands of someone who knows what they are doing at this point.

AlCalzone:
> how to add 'quirks' to the configs?

Sure, you'll find it in the docs (they are linked from the repo's main page btw.)
https://zwave-js.github.io/node-zwave-js/#/development/config-files

marcus-j-davies:
> > how to add 'quirks' to the configs?
> 
> Sure, you'll find it in the docs (they are linked from the repo's main page btw.)
> https://zwave-js.github.io/node-zwave-js/#/development/config-files

Ahh - ok, I thought it was going to be a little more involved, meaning typescript stuff - which I'm still getting to grips with.
I will indeed take a stab at it.

marcus-j-davies:
@AlCalzone,

Not sure if I am doing anything wrong with the config here (still learning the internals of ZWJS), or its a bug.
taking the below to split param 7 into 2 partials

bit 3 : Disables Sound
bit 4 : Sets Notification type.

This comes through  the ValueID interface fine.
The other bits are non user configurable 

```
"paramInformation": {
        "7[0x03]": {
            "label": "Disable Sound",
            "description": "Prevents audio from the siren - leaving only the strobe function.",
            "valueSize": 1,
            "minValue": 0,
            "maxValue": 1,
            "defaultValue": 1,
            "readOnly": false,
            "writeOnly": false,
            "allowManualEntry": true,
            "options": [
                {
                    "label": "Enables Sound",
                    "value": 0
                },
                {
                    "label": "Disables Sound",
                    "value": 1
                }
            ]
        },
        "7[0x04]": {
            "label": "Notification Type",
            "description": "Notification event type to use for trigger events",
            "valueSize": 1,
            "minValue": 0,
            "maxValue": 1,
            "defaultValue": 0,
            "readOnly": false,
            "writeOnly": false,
            "allowManualEntry": true,
            "options": [
                {
                    "label": "Notification Report",
                    "value": 0
                },
                {
                    "label": "Sensor Binary Report",
                    "value": 1
                }
            ]
        },
        "29": {
            "label": "Disable Alarm",
            "description": "Disables the alarm function entierly, including any tamper detection.",
            "valueSize": 1,
            "minValue": 0,
            "maxValue": 1,
            "defaultValue": 0,
            "readOnly": false,
            "writeOnly": false,
            "allowManualEntry": true
        },
        "31": {
            "label": "Alarm Duration",
            "description": "Play alarm sound duration.",
            "valueSize": 1,
            "minValue": 0,
            "maxValue": 127,
            "defaultValue": 6,
            "readOnly": false,
            "writeOnly": false,
            "allowManualEntry": true
        }
    }

```

However, after setting bit 3 via
```
let ValueID = {
  "commandClassName": "Configuration",
  "commandClass": 112,
  "endpoint": 0,
  "property": 7,
  "propertyKey": 3,
  "propertyName": "Disable Sound"
}
 
```

it does not actually have any affect on the device.
checking param 7 it is coming back as 1 - instead of 8 as expected when setting bit 3 to 1.

Setting param 7 to 8 manually via the CCAPI,  correctly makes the alarm silence - but then, the partials do not report the correct value.

Am I missing something?

AlCalzone:
The thing in square brackets is a bit mask. You need to put 0x04 there to target bit 3 and 0x08 for bit 4.

marcus-j-davies:
Ok, thats done, sadly it hasn't helped.

 when I set a value of 1 to
```
let ValueID = {
  "commandClassName": "Configuration",
  "commandClass": 112,
  "endpoint": 0,
  "property": 7,
  "propertyKey": 4,
  "propertyName": "Disable Sound"
}

```
the value written to the device is now 4, instead of the previous 1, but still not 8 as expected.
the value returned from getValue however, is showing 1:
```
{
  "commandClassName": "Configuration",
  "commandClass": 112,
  "property": 7,
  "propertyKey": 4,
  "newValue": 1,
  "prevValue": 1,
  "propertyName": "Disable Sound"
}
```

but the value of 4 is written to the device, it should be 8.
could this be an incorrect endianness order on the document, or internally?





it could Evan be my lack of understanding  with bit masking.

AlCalzone:
Can you push the changes so I can verify?

marcus-j-davies:
> Can you push the changes so I can verify?

Thats done - I don't think its an error in the document as setting 8 manually, yields the correct result.

AlCalzone:
Ohh... I think we were off by one bit.
Bit 3 is (binary) 0000_1000, so (hex) 0x08
Bit 4 is (binary) 0001_0000, so (hex) 0x10.

marcus-j-davies:
Hold the line.....

AlCalzone:
On a side note: Your lint is failing because the JSON formatting is not okay.
Please don't attempt to fix the formatting. I would expect our bot to comment but it doesn't. So this serves as a good base for debugging the bot :)

marcus-j-davies:
> Ohh... I think we were off by one bit.
> Bit 3 is (binary) 0000_1000, so (hex) 0x08
> Bit 4 is (binary) 0001_0000, so (hex) 0x10.

ok - that seems better.

setting bit 3 only : yields 8
setting bit 4 only : yields 16
setting both : yields 24.

I'll push the change  - without pretty.

AlCalzone:
closing and reopening to trigger bot

AlCalzone:
@zwave-js-bot fix lint

marcus-j-davies:
> @zwave-js-bot fix lint

 Does he have relations with skynet?

AlCalzone:
Can work on that if you need it 🤖

PR #1358, opened 2021-01-12
fix(config): Q-Light Puck - Fix param 11

geirra:
Simple fix, cross checked with the norwegian manual.
https://qlight.no/media/nvllykma/instructions_z-wave_puck_norsk_a4.pdf

Should get rid of the lint errors (duplicate)


AlCalzone:
perfect, thanks!

PR #1374, opened 2021-01-13
fix: destroy loggers when destroying driver instance

robertsLando:


packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -1252,6 +1252,9 @@ export class Driver extends EventEmitter {
 		this._wasDestroyed = true;
 		this.driverLog.print("destroying driver instance...");
 
+		// destroy loggers
+		this.logContainer.destroy();
```
AlCalzone: that comes too early, we're using the loggers below.

AlCalzone (changes requested):

packages/zwave-js/src/lib/driver/Driver.ts
robertsLando: arggg sorry you right

AlCalzone (approved):

PR #1379, opened 2021-01-13
feat(config): add compat option treatBasicSetAsEvent and use it for GE devices

### LINKED ISSUE #1277: GE Switches and Dimmers do not reflect double-up or double-down because they use the BasicCC for those events
**Describe the bug**

GE dimmer/switches are capable of double tap events, but those are not captured here and result in the current value going out of sync with the device. 

These are the steps to make it work under qt-openzwave: https://community.home-assistant.io/t/howto-add-double-tap-support-for-ge-switches-and-dimmers-in-ozw-beta/228871/16.

The original qt-openzwave issue which has more detailed info is at: OpenZWave/qt-openzwave#60.

This appears to result from this because the devices use BasicCC for these events despite supporting the Multi-Level Command Class: #388 

Non-compliant or not, it is how they work... Is there a way to IgnoreMapping here like in ozw 1.6?

**Note: If possible, these devices should also auto-associate the controller into group 3 so that it receives the double-tap events.**

**To Reproduce**
1. Pair switches in zwavejs2mqtt. 
2. Double tap up or down

**Device information**
Which device(s) is/are affected?  Various GE switches and dimmers (confirmed on 14294, 14291)
What are the node IDs?

**Logfile:** 

Note: I have a script that resets the light to the default levels upon turning on. That accounts for the 50 --> 50 messages.

Double-tap up (from a prior double-tap up state, hence 255-->255):

Screen Shot 2021-01-02 at 10 13 38 PM

Double-tap up from off, just in case being in a former double-tap state affects it (note this one is out of order on the console log):

Screen Shot 2021-01-02 at 10 16 16 PM

(The current value goes to 255 and HA thinks the device is now at 100%, but the physical dim has not changed. **They are now out of sync. The currentValue is 255, the target Value is 50**.)

Double-tap down:

Screen Shot 2021-01-02 at 10 14 02 PM

For comparison, off to normal on:

Screen Shot 2021-01-02 at 10 15 11 PM

For comparison, on to normal off:

Screen Shot 2021-01-02 at 10 15 43 PM

Console log:

Screen Shot 2021-01-02 at 10 17 08 PM

AlCalzone:
fixes: #1277

### DIFF
--- a/docs/development/config-files.md
+++ b/docs/development/config-files.md
@@ -190,6 +190,13 @@ The Z-Wave+ specs mandate that the root endpoint must **mirror** the application
 
 Some devices spam the network with hundreds of invalid `ConfigurationCC::InfoReport`s when one is requested. Set this flag to `true` to skip this query for affected devices.
 
+### `treatBasicSetAsEvent`
+
+By default, `Basic CC::Set` commands are interpreted as status updates. This flag causes the driver to emit a `value event` for the `"value"` property instead. Note that this property is exclusively used in this case in order to avoid conflicts with regular value IDs.
+
+> [!NOTE]
+> If this option is `true`, it has precedence over `disableBasicMapping`.
+
 ## Contributing configuration files
 
 In order to get your configuration file included in this library, this is the way:
--- a/maintenance/schemas/device-config.json
+++ b/maintenance/schemas/device-config.json
@@ -228,6 +228,9 @@
 				},
 				"skipConfigurationInfoQuery": {
 					"const": true
+				},
+				"treatBasicSetAsEvent": {
+					"const": true
 				}
 			},
 			"additionalProperties": false
--- a/packages/config/config/devices/0x0063/46201.json
+++ b/packages/config/config/devices/0x0063/46201.json
@@ -62,6 +62,6 @@
 		}
 	},
 	"compat": {
-		"disableBasicMapping": true
+		"treatBasicSetAsEvent": true
 	}
 }
--- a/packages/config/config/devices/0x0063/46202.json
+++ b/packages/config/config/devices/0x0063/46202.json
@@ -57,6 +57,6 @@
 		}
 	},
 	"compat": {
-		"disableBasicMapping": true
+		"treatBasicSetAsEvent": true
 	}
 }
--- a/packages/config/config/devices/0x0063/46203.json
+++ b/packages/config/config/devices/0x0063/46203.json
@@ -57,6 +57,6 @@
 		}
 	},
 	"compat": {
-		"disableBasicMapping": true
+		"treatBasicSetAsEvent": true
 	}
 }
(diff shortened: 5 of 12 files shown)


blhoward2:
It now shows in the logs:

```
==> zwavejs_1.log <==
21:38:08.788 DRIVER « [Node 003] [REQ] [ApplicationCommand]
                      └─[BasicCCSet]
                          target value: 255

==> zwavejs2mqtt.log <==
2021-01-13 21:38:08.791 INFO ZWAVE: Node 3: value notification: 32-0-value undefined => 255

==> zwavejs_1.log <==
21:38:08.790 CNTRLR   [Node 003] treating BasicCC Set as a value event
```

From what I can tell (with @jcam help), a value for command class 32 is never created, so when updateValue is called it does nothing. I get lost when I try to figure out what would trigger the value to be called in node-zwave-js, when the new compat flag is set. (You probably don't want to create a value when BasicCC is mapped).

AlCalzone:
What are the implications of not having Metadata for a value @robertsLando ?
Do I need to create that too?

AlCalzone:
@blhoward2 try now please. The metadata is now created before emitting the value event and I suspect zwavejs2mqtt requires that to be present.

AlCalzone:
This might go into 6.0.0 if you can verify it works before the release ;)

blhoward2:
```
2021-01-14 17:17:00.904 INFO ZWAVE: Node 3: metadata updated: 32-0-value
2021-01-14 17:17:00.907 INFO ZWAVE: Node 3: value notification: 32-0-value undefined => 255
```

```
==> zwavejs_1.log <==
17:17:00.902 DRIVER « [Node 003] [REQ] [ApplicationCommand]
                      └─[BasicCCSet]
                          target value: 255

==> zwavejs2mqtt.log <==
2021-01-14 17:17:00.905 INFO ZWAVE: Node 3: metadata updated: 32-0-value

==> zwavejs_1.log <==
17:17:00.903 CNTRLR   [Node 003] treating BasicCC Set as a value event

==> zwavejs2mqtt.log <==
2021-01-14 17:17:00.907 INFO ZWAVE: Node 3: value notification: 32-0-value undefined => 255
```
I now get the node value in mqtt. It goes to 255 for a second, then one second later sends again without a value included in the payload. We still need to define a sensor at the zwavejs2mqtt level, but this all seems working to me. @robertsLando

AlCalzone:
> then one second later sends again without a value included in the payload

That is behavior by zwavejs2mqtt. Not sure if it is necessary when the value is not retained, but I'll let you be the judge of that.

blhoward2:
> > then one second later sends again without a value included in the payload
> 
> That is behavior by zwavejs2mqtt. Not sure if it is necessary without persistence, but I'll let you be the judge of that.

It's hard to test without it being in HA...

blhoward2:
I manually added a sensor and it works great in HA! Whether it needs to clear it or not, I'm not sure but I think they determined in the context of scene events that it was or HA ignores the repeat value.

AlCalzone:
Okay, that's a go for me.

blhoward2:
Thanks!

PR #1384, opened 2021-01-14
fix(config): change manufacturer Jasco Products to GE/Jasco

blhoward2:
We currently label GE devices as Jasco, as Jasco makes them for GE. This could be confusing for users as Jasco sells very few devices under its own branding and all other major projects (and the Zwave Alliance) list these solely under GE's name.

### DIFF
--- a/packages/config/config/devices/0x0063/12718.json
+++ b/packages/config/config/devices/0x0063/12718.json
@@ -1,7 +1,7 @@
-// Jasco Products 12718
+// GE/Jasco 12718
 // Smart Dimmer
 {
-	"manufacturer": "Jasco Products",
+	"manufacturer": "GE/Jasco",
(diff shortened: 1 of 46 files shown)


AlCalzone:
You might also need to edit the config files since they also contain a reference to the manufacturer.

AlCalzone:
@blhoward2 I went ahead and did it - please cross-check if that makes sense.

blhoward2:
Thanks, I was going to do that later. These look right. Out of curiosity why do we have manufacturers.json and it’s defined in the file?

AlCalzone:
If I remember correctly, OpenHAB imports needed this for lookup. 

And there are cases where multiple manufacturers share an ID or an old one gets replaced.
By having that info in the device files, at least the manufacturer for each device matches.

PR #1394, opened 2021-01-14
fix(build): export ZWaveNodeValueNotificationArgs

AlCalzone:
@robertsLando this means you can replace
```
 * @param {import('zwave-js/build/lib/node/Types').ZWaveNodeValueNotificationArgs} args
```
with
```
 * @param {import('zwave-js').ZWaveNodeValueNotificationArgs} args
```
since you shouldn't count on the internal structure to stay like it is.


robertsLando:
@AlCalzone thanks!

PR #1397, opened 2021-01-14
fix: correctly change node status after successful ping

### LINKED ISSUE #1387: After a successful ping, devices are not detected as awake/alive
https://github.com/zwave-js/node-zwave-js/issues/1364#issuecomment-760016477

AlCalzone:
fixes: #1387

### DIFF
--- a/packages/zwave-js/src/lib/driver/Driver.ts
+++ b/packages/zwave-js/src/lib/driver/Driver.ts
@@ -1937,12 +1937,9 @@ ${handlers.length} left`,
 		let node: ZWaveNode | undefined;
 
 		// Don't send messages to dead nodes
-		if (
-			isNodeQuery(msg) ||
-			(isCommandClassContainer(msg) && !messageIsPing(msg))
-		) {
+		if (isNodeQuery(msg) || isCommandClassContainer(msg)) {
 			node = msg.getNodeUnsafe();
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot pack this

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -2047,6 +2044,8 @@ ${handlers.length} left`,
 					if (!node.keepAwake) {
 						this.debounceSendNodeToSleep(node);
 					}
+					// The node must be awake because it answered
+					node.markAsAwake();
```
jcam, line 2048: This looks odd, the previous line would be sending the node to sleep, ergo it would not be awake anymore... right? should this be markAsAlive()?

AlCalzone: Not entirely correct. The previous line will actively send the node to sleep after 1 second if there are no messages for it in the queue. Until then, it is awake.

jcam: right! because this is fast :)

AlCalzone: And depending on the queue length, the "send to sleep" might take pretty long until it actually happens.

PR #1403, opened 2021-01-15
feat(node): expose refreshValues method

blhoward2:
To support exposing the refreshValues() function upstream in mqtt. I think this is necessary...


AlCalzone:
This should **never** be used to work around issues with single device values. Calling `refreshValues` will update all actuator and sensor values of all endpoints of a node, which may add up to a lot of traffic.

I think for what you are trying to achieve, we need a read-version of the `setValue` API.

robertsLando:
>  we need a read-version of the setValue API.

Absolutely. That could also allow to create some kind of manual polling?

blhoward2:
> This should **never** be used to work around issues with single device values. Calling `refreshValues` will update all actuator and sensor values of all endpoints of a node, which may add up to a lot of traffic.
> 
> I think for what you are trying to achieve, we need a read-version of the `setValue` API.

@AlCalzone That may be necessary though, as you may need to update multiple values and may not know which ones? It’s like triggering a manual Hail. What I’m trying to duplicate is ozw has a service called RefreshDynamic that refreshes all dynamic values at once. Perhaps we need both?

blhoward2:
> > we need a read-version of the setValue API.
> 
> Absolutely. That could also allow to create some kind of manual polling?

This is what I think we need. As an example when I use my Leviton scene controllers they send scene events by direct association and some devices don’t update their status via BasicCC or Hail. So what I used to do was refresh their values after detecting a scene event.

AlCalzone:
@blhoward2 thats a valid use case. I'm just seeing several reports that some devices don't report the final status of a transition and I want to avoid that users try to work around that by polling all values.

IMO I agree we need both.

AlCalzone:
Added an explicit warning in the docs, now I'm okay with it.

Issue #1347, opened 2021-01-11
Handle partial params when bulk-setting config params


AlCalzone:
as discussed in https://github.com/zwave-js/zwavejs2mqtt/issues/196

PR #1411, opened 2021-01-15
Adding config parameter 8 to Inovelli Red On_Off switch

StarFleetCPTN:
Adding config parameter 8 will allow sending notifications to the switch with one publish.

### DIFF
--- a/packages/config/config/devices/0x031e/lzw30-sn.json
+++ b/packages/config/config/devices/0x031e/lzw30-sn.json
@@ -339,6 +339,17 @@
 			"writeOnly": false,
 			"allowManualEntry": true
 		},
+		"8": {
+			"label": "LED Strip Effect",
+			"description": "LED Strip Effect",
+			"valueSize": 4,
+			"minValue": 0,
+			"maxValue": 83823359,
+			"defaultValue": 0,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true
+		},
 		"8[0xff]": {
 			"label": "LED Strip Effect (Color)",
 			"description": "LED Strip Effect (Color)",


AlCalzone:
Config parameter #8 is split into multiple partial parameters, each of which targets a specific function, which makes it much easier to work with for humans without a calculator. In general we want to avoid duplication here.

Is there a reason you need to update them all in one go? And if so, would having a way to [bulk-set](https://github.com/zwave-js/node-zwave-js/issues/1347) the partial params help you?

StarFleetCPTN:
Coming from ozw all my current automations are setup to use a single conf parm for notifications. I'm lazy and not having make changes to split what has been single call into 4 would be great. Having a way to bulk-set params could be useful. 

Currently the Inovelli Red Dimmer is setup to accept a single value for notification which works with my current setup out of the box. https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x031e/lzw31-sn.json

AlCalzone:
That config param is something we're likely also going to change.

I'll take your use case into account when designing the bulk set API. How are you currently controlling the values? MQTT?

StarFleetCPTN:
I'm using MQTT to send the values, and I just noticed that sending all 4 config values at the same will not work. All 4 values are needed to in-order to display the correct LED notification on the switch, and zwavejs2mqtt is only writing the last config value sent to the switch.

2021-01-17 23:55:04.535 INFO MQTT: Message received on zwave/nodeID_11/configuration/endpoint_0/LED_Strip_Effect_Color/255/set, '25'
2021-01-17 23:55:04.536 INFO ZWAVE: Writing 25 to 112-0-8-255
2021-01-17 23:55:04.551 INFO MQTT: Message received on zwave/nodeID_11/configuration/endpoint_0/LED_Strip_Effect_Duration/16711680/set, '255'
2021-01-17 23:55:04.552 INFO ZWAVE: Writing 255 to 112-0-8-16711680
2021-01-17 23:55:04.670 INFO ZWAVE: Node 11: value updated: 112-0-8-255 0 => 0
2021-01-17 23:55:04.672 INFO ZWAVE: Node 11: value updated: 112-0-8-65280 10 => 10
2021-01-17 23:55:04.674 INFO ZWAVE: Node 11: value updated: 112-0-8-16711680 10 => 255
2021-01-17 23:55:04.675 INFO ZWAVE: Node 11: value updated: 112-0-8-2130706432 2 => 2
2021-01-17 23:55:04.737 INFO ZWAVE: Node 11: value updated: 112-0-8-255 0 => 0
2021-01-17 23:55:04.740 INFO ZWAVE: Node 11: value updated: 112-0-8-65280 10 => 10
2021-01-17 23:55:04.742 INFO ZWAVE: Node 11: value updated: 112-0-8-16711680 255 => 255
2021-01-17 23:55:04.745 INFO ZWAVE: Node 11: value updated: 112-0-8-2130706432 2 => 2

AlCalzone:
You'd probably have to wait until the previous value is confirmed, as zwave-js merges the new value with the known current value behind the scenes.

AlCalzone:
@StarFleetCPTN I'm not sure if there's a way to set `112-0-8` (without the bitmask suffix) through MQTT. If it does, it should work if you just specify the compound value there.

If not, I'll have to ask for a bit of patience. I'll get this sorted out together with #1436

crkochan:
Just to add my voice to this, I was also angling to submit a PR to add the parameter after the recent release of Home Assistant 2021.3.0 introduced the `zwave_js.set_config_parameter` service.

My typical invocation is via said Home Assistant service, from various flows in the Node Red add-on.

AlCalzone:
@crkochan HA should be able to set the compound value AFAIK, even if there's no value shown for it.

crkochan:
It's early enough in the game that I'm not sure what that looks like.

Under OZW, my flows would compose the payload to the service call with a static value for the parameter based on what I wanted the dimmer/switch to do for the given notification (flash red for two minutes at 100% brightness = 71174657). Yes, I'd have have to use an external tool to figure out what the notification value should be, but it would typically never change, so it didn't feel like a burden.

With the current zwave_js service, even while providing a bitmask, it doesn't seem as though it's possible to set four different values (color, duration, effect, brightness) in a single invocation of `set_config_parameter`.

Like OP, the switch from a single service call to up to four seems like a regression in terms of DRY-ness, but I do concede that the approach of breaking out partial parameters is a lot more user friendly to new users. It's just a shame that we can't have both options available by allowing both partial params as well as the "some assembly required" old school param.

AlCalzone:
You can set the old school compound parameter. I'm not sure what's the correct way to do this in HomeAssistant, but if you have an API to set arbitrary values, you just need to use the value ID without propertyKey/bitmask to set the compound param.
@marcelveldt @kpine any hints?

crkochan:
Some operational details first.
* Home Assistant version: 2021.3.0 (on Home Assistant OS)
* Controller interface: zwavejs2mqtt on remote host (using websocket server, mqtt gateway disabled)
*  Zwavejs-server Version: 1.1.0
* Zwavejs Version: 6.6.0

Now with that out of the way, I have an [Inovelli LZ30-SN Red Series On/Off switch](https://devices.zwave-js.io/?jumpTo=0x031e:0x0002:0x0001:1.20) I want to have perform a notification. 
[Per Inovelli](https://support.inovelli.com/portal/en/kb/articles/inovelli-device-manuals), the correct parameter for this is param 8. BTW, the firmware for said device is 1.2, but I have another in a more recent beta version of 1.19. 

So I flip over to the services tab of the developer tools in HA, and I compose a service call.
```yaml
service: zwave_js.set_config_parameter
data:
  parameter: '8'
  value: 37554733
target:
  entity_id: switch.living_room_plug
```

The response I get back in the UI and logs is `Configuration parameter with value ID 69-112-0-8-00-00 could not be found`. And indeed, there is no such parameter shown in the zwavejs2mqtt control panel for the device.

What I do have is (among the others)...
* [69-112-0-8-255] LED Strip Effect (Color)
* [69-112-0-8-65280] LED Strip Effect (Brightness)
* [69-112-0-8-16711680] LED Strip Effect (Duration)
* [69-112-0-8-2130706432] LED Strip Effect (Effect)

AlCalzone:
Yes, it is true that zwave-js does not expose the full param currently. That requires some rework. But nonetheless, this works on the zwave-js side.

> Configuration parameter with value ID 69-112-0-8-00-00 could not be found

I don't know which additional values HA encodes into the value ID, but it looks like it shouldn't add `propertyKey` 0 by default after the `8`. I'll ask the devs.

crkochan:
Ya, I traced it back to their [zwavejs-js-server-python](https://github.com/home-assistant-libs/zwave-js-server-python/tree/0.20.1) lib.

Their [async_set_config_parameter function](https://github.com/home-assistant-libs/zwave-js-server-python/blob/a5094e822ba82ba6768b690b0268179505f661c0/zwave_js_server/util/node.py#L11) makes [a call to get_value_id](https://github.com/home-assistant-libs/zwave-js-server-python/blob/a5094e822ba82ba6768b690b0268179505f661c0/zwave_js_server/util/node.py#L40) when setting a parameter by id number. 

The [get_value_id function](https://github.com/home-assistant-libs/zwave-js-server-python/blob/a5094e822ba82ba6768b690b0268179505f661c0/zwave_js_server/model/value.py#L56) sets `property_key` and `property_key_name` to `00` if they are undef or None, and spits back an f-string.

Then they try to find the string in [a dictionary of config values](https://github.com/home-assistant-libs/zwave-js-server-python/blob/a5094e822ba82ba6768b690b0268179505f661c0/zwave_js_server/util/node.py#L23) put together by [get_configuration_values](https://github.com/home-assistant-libs/zwave-js-server-python/blob/a5094e822ba82ba6768b690b0268179505f661c0/zwave_js_server/model/node.py#L299) which I think, by way of [get_command_class_values](https://github.com/home-assistant-libs/zwave-js-server-python/blob/a5094e822ba82ba6768b690b0268179505f661c0/zwave_js_server/model/node.py#L288) sources its information based on the contents of the device configuration files, which for the [lz30-sn](https://github.com/zwave-js/node-zwave-js/blob/v6.6.0/packages/config/config/devices/0x031e/lzw30-sn.json) does not have parameter 8  by itself where it isn't split into multiple partial parameters. 

It's also entirely possible that it's rather late, and I'm chasing the wrong whale.

AlCalzone:
> It's also entirely possible that it's rather later, and I'm chasing the wrong whale.

Looks like the right 🐳 to me :)

marcelveldt:
The value_id in the python lib is just internal how we store and look it up.
We do not yet support setting a config param's raw value. Only the partials (so for each propertyKey)

AlCalzone:
Is there any workaround? Like setting the value for a given value ID without internal checks (and letting zwave-js complain if that's not a valid ID)?

crkochan:
I think, for the time being, my approach will be to adapt [existing examples of HA scripting yaml](https://github.com/brianhanifin/Home-Assistant-Config/blob/93589cc2a747074813081c3a4eb9f4c802160336/scripts/notifications/inovelli_led/inovelli_led.yaml) with additional functionality for setting four parameters instead of one.

Hopefully the "yet" in that sentence means setting raw config values may someday be possible, and I'll be able to re-simplify my automations.

Not that that resolves OP's issue. Hopefully I didn't take the PR convo off on too far of a tangent.

marcelveldt:
> Is there any workaround? Like setting the value for a given value ID without internal checks?

Not yet, we'll work on this later. @raman325 has this on his TODO list.
So, yes, this will be added later. For now just use the partials when setting config values. 
You can even set them by name, which is way more easier then calculating values yourself.

You can set multiple values at once by providing an object of key:value in the value parameter of the service call

raman325:
> > Is there any workaround? Like setting the value for a given value ID without internal checks?
> 
> Not yet, we'll work on this later. @raman325 has this on his TODO list.
> So, yes, this will be added later. For now just use the partials when setting config values.
> You can even set them by name, which is way more easier then calculating values yourself.
> 
> You can set multiple values at once by providing an object of key:value in the value parameter of the service call

You can’t use the key value object yet. We need a way to bulk set parameters in node-zwave-ja which isn’t available yet

PR #1412, opened 2021-01-15
fix(controller): return boolean on stop inclusion/exclusion

Michsior14:
This fixes the response of `stopInclusion`/`stopExclusion` commands to fulfill the defined interface.


AlCalzone:
Huh, this is caused by a type definition error in https://github.com/AlCalzone/shared-utils/
I'll update that and edit your PR accordingly.

AlCalzone (approved):

PR #1423, opened 2021-01-16
fix: add parameter validation to setManufacturer function

blhoward2:
Fixes a bug noted during revisions to importConfig.ts. Passing an undefined manufacturerId causes silent corruption in manufacturers.json. Rather than rely on upstream error checking, this skips such calls and logs an error to console (since the function is only called from importConfig.ts and that file is typically run in that manner).


AlCalzone (changes requested):
IMO that's not the correct fix. The argument type for `manufacturerId` is `number`, so if an `undefined` value sneaks in here, there's most likely an `any` floating around when it shouldn't be.

If you show me some of the calling code, I can check that.

blhoward2:
No, what happens is some of the files (like 8 of 3,700) have no ManufacturerId. It is null. If you inadvertently parse one like that it causes it to be undef. This is just a failsafe in case that happens and it hasn't been accounted for by the calling function.

AlCalzone:
Still you can't check in every typescript function that the arguments match the function signature.

blhoward2:
Yeah, it just struck me that this one fails silently but appears to succeed. If you didn look at the manufacturers.json file you’d never know you’d wiped out part of the file. If you don’t want it, it’s not a huge deal.

AlCalzone:
I'd rather use the unknown type when parsing XML to force checks at the calling site

blhoward2:
ok

PR #1438, opened 2021-01-17
feat: export controller event callback types

balloob:
This exports the controller event callback types so we can use the types in Z-Wave JS Server here https://github.com/zwave-js/zwave-js-server/blob/master/src/lib/forward.ts#L6

### DIFF
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -123,7 +123,7 @@ interface ControllerEventCallbacks {
 	"heal network done": (result: ReadonlyMap<number, HealNodeStatus>) => void;
 }
 
-type ControllerEvents = Extract<keyof ControllerEventCallbacks, string>;
+export type ControllerEvents = Extract<keyof ControllerEventCallbacks, string>;
 
 export interface ZWaveController {
 	on<TEvent extends ControllerEvents>(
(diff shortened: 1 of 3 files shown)


AlCalzone (changes requested):
They should also be re-exported from `src/Controller.ts`, so I'm free to change the internal structure without breaking depending libs.

balloob:
@zwave-js-bot fix lint

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -123,7 +123,7 @@ interface ControllerEventCallbacks {
 	"heal network done": (result: ReadonlyMap<number, HealNodeStatus>) => void;
 }
 
-type ControllerEvents = Extract<keyof ControllerEventCallbacks, string>;
+export type ControllerEvents = Extract<keyof ControllerEventCallbacks, string>;
```
balloob, line 126: @AlCalzone any reason you prefer `Extract<keyof ControllerEventCallbacks, string>` over just `keyof ControllerEventCallbacks` ?

AlCalzone: I'm sure there was a reason for this 🤷🏻‍♂️
Or I copied it from somewhere where it was necessary and didn't remove it.

AlCalzone:
@balloob can you write the fix-lint bot command again? I'm trying to debug why it is not working for other users.

balloob:
@zwave-js-bot fix lint

balloob:
(right now there is nothing to fix though)

AlCalzone:
Okay, the bot thinks the comment is in an issue, not a PR, so it doesn't add the PR author to the list of authorized users. 👍🏻

AlCalzone (approved):

PR #1440, opened 2021-01-17
feat(config): add ThermoFloor z-temp2 thermostat

ivdimitro:
Adding HeatIt Z-Temp2 battery operated thermostat


ivdimitro:
I messed up something with the commits. Will try again.

PR #1444, opened 2021-01-18
feat: add controller method markNodeAsFailed

### LINKED ISSUE #1392: Add a method to trigger isFailed for a node
From https://github.com/zwave-js/zwavejs2mqtt/issues/233:
> These nodes are not available anymore since they had the same problems as above but had been re-included to make them work again. Unfortunately these are left as zombies which cannot be removed since "Is node failed" does not set them to failed:

AlCalzone:
fixes: #1392

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -85,6 +85,15 @@ isFailedNode(nodeId: number): Promise<boolean>
 
 Checks if a node was marked as failed in the controller. If it is, it can be removed from the network with [`removeFailedNode`](#removeFailedNode).
 
+### `markNodeAsFailed`
+
+```ts
+markNodeAsFailed(nodeId: number): Promise<boolean>
+```
+
+Sometimes failed nodes are not detected by the controller hardware, so [`removeFailedNode`](#removeFailedNode) and [`replaceFailedNode`](#replaceFailedNode) won't work. This method can be used to ping a node a couple of times, which should reliably mark the node as failed.
+It returns `true` when the node was successfully detected as failed.
+
 ### `removeFailedNode`
 
 ```ts
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -1870,6 +1870,46 @@ ${associatedNodes.join(", ")}`,
 	}
 
 	/**
+	 * Mark a node as failed in the controller's memory, so it can be removed with @see removeFailedNode.
+	 * The return value indicates whether marking the node as failed succeeded.
+	 * @param nodeId The id of the node in question
+	 */
+	public async markNodeAsFailed(nodeId: number): Promise<boolean> {
+		const node = this.nodes.get(nodeId);
+		if (!node) {
+			throw new ZWaveError(
+				`Node ${nodeId} was not found!`,
+				ZWaveErrorCodes.Controller_NodeNotFound,
+			);
+		}
+
+		if (await this.isFailedNode(nodeId)) return true;
+
+		const attempts = 5;
+		for (let i = 1; i <= attempts; i++) {
+			this.driver.controllerLog.logNode(
+				nodeId,
+				`Trying to mark node as failed, attempt ${i}/${attempts}...`,
+			);
+			if (await node.ping()) {
+				this.driver.controllerLog.logNode(
+					nodeId,
+					`responded, it is not a failed node.`,
+				);
+				return false;
+			} else if (await this.isFailedNode(nodeId)) {
+				this.driver.controllerLog.logNode(
+					nodeId,
+					`was marked as failed.`,
+				);
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	/**
 	 * Tests if a node is marked as failed in the controller's memory
 	 * @param nodeId The id of the node in question
 	 */


robertsLando:
Hummm, IMO this creates confusion. This shopuld be what isFailed node does

I would not create a separete api for this. Is failed node should check if the node is failed and return true if it is, false if not

ahochsteger:
@robertsLando but how would it then be possible to replace a failed node that is still able to communicate but otherwise not able to serve its purpose (e.g. a mechanically broken button)?
Using "is failed node" would always return false in that case, leaving no way to replace this broken device with the same node id.
IMHO there should be a way to override the automatic failure detection to be able to replace these broken devices.

robertsLando:
@ahochsteger did you try to simply disconnect the device and trigger again the request? Until it's connected and aswering to the controller it will never be marked as failed

ahochsteger:
@robertsLando yes, I've done that and it almost always works.

In some cases though (e.g. when re-including the same device again and keep the node id) it causes the failed flag to be reset again as soon as I bring the device online again after marking it as failed and "replace failed node" fails. Here the only solution is to fully exclude and include the device, causing a lot of re-configuration for all affected smart home automation which I'd like to prevent.
I'm not sure though, markNodeAsFailed will help in that case.

FYI: I've been using the Silicon Labs Z-Wave PC Controller v5 for these kind of maintenance since it did not work reliably using zwavejs2mqtt after some time (sorry, no issue yet, since it's much harder to reproduce and I'm not yet fully sure if it has to do with the Z-Wave.Me UZB stick).

robertsLando:
>  causing a lot of re-configuration for all affected smart home automation which I'd like to prevent.

Why? Couldn't you exclude the node, mark it as failed and replace it?

ahochsteger:
No, this did not work - at least not for me.

It always gives me the following error:
image

Even though the node is shown as failed in the table:
image

robertsLando:
@ahochsteger Needs a log to understand what's going on

ahochsteger:
> @ahochsteger Needs a log to understand what's going on

@robertsLando  here are the relevant logs for zwavejs2mqtt:
```bash
# First try to replace failed node:
2021-01-18 16:47:35.435 INFO APP: Zwave api call: replaceFailedNode [ 4, false, [length]: 2 ]
2021-01-18 16:47:37.095 INFO ZWAVE: The node replace process could not be started due to the following reasons:
· Node 4 is not in the list of failed nodes replaceFailedNode 

# Second try to replace failed node:
2021-01-18 16:47:47.614 INFO APP: Zwave api call: replaceFailedNode [ 4, false, [length]: 2 ]
2021-01-18 16:47:48.860 INFO ZWAVE: The node replace process could not be started due to the following reasons:
· Node 4 is not in the list of failed nodes replaceFailedNode 

# Try isFailedNode before:
2021-01-18 16:49:49.680 INFO APP: Zwave api call: isFailedNode [ 4, [length]: 1 ]
2021-01-18 16:49:49.707 INFO ZWAVE: Success zwave api call isFailedNode 

# Third try to replace failed node:
2021-01-18 16:50:12.157 INFO APP: Zwave api call: replaceFailedNode [ 4, false, [length]: 2 ]
2021-01-18 16:50:13.413 INFO ZWAVE: The node replace process could not be started due to the following reasons:
· Node 4 is not in the list of failed nodes replaceFailedNode 
```

ahochsteger:
And here are the zwavejs logs - in case they are helpful:
[zwavejs.log](https://github.com/zwave-js/node-zwave-js/files/5831438/zwavejs.log)

AlCalzone:
> In some cases though (e.g. when re-including the same device again and keep the node id) it causes the failed flag to be reset again as soon as I bring the device online again after marking it as failed and "replace failed node" fails.

I don't understand. If the device is not failed, why would you replace it with replaceFailedNode? That leads to situations where two devices share a node ID and causes confusion for the driver.

If a device is in the controller's failed nodes list is out of our reach. But you can only replace it if it is in there.
Hence the method I added to force some communication with it - which should fail and make the controller put it in the list.

> 2021-01-18 16:49:49.680 INFO APP: Zwave api call: isFailedNode [ 4, [length]: 1 ]
> 2021-01-18 16:49:49.707 INFO ZWAVE: Success zwave api call isFailedNode 
> 2021-01-18 16:50:12.157 INFO APP: Zwave api call: replaceFailedNode [ 4, false, [length]: 2 ]
> 2021-01-18 16:50:13.413 INFO ZWAVE: The node replace process could not be started due to the following reasons:
> · Node 4 is not in the list of failed nodes replaceFailedNode 

`isFailedNode` just **checks** if the node is in the failed list. It does not change the status.

ahochsteger:
> I don't understand. If the device is not failed, why would you replace it with replaceFailedNode? That leads to situations where two devices share a node ID and causes confusion for the driver.

@AlCalzone the reason was that node initialization was not reliable (see [zwave-js/zwavejs2mqtt#233]) and marking them as failed and then replacing them with the same device caused to re-initialize the node which almost always helped as a workaround.
Hopefully this won't be required anymore, when the root cause is solved ;-).

> `isFailedNode` just **checks** if the node is in the failed list. It does not change the status.

Ah, I see - I assumed that the communication is check as well and the status updated.
Maybe you're interested in the fact that the Silicon Labs Z-Wave PC Controller (which always worked reliably for me) seems to do the check as well.

Here is a screenshot of the resulting logs when executing the "Is Failed Node" action:
image

But then we are talking about 3 different functionality here:
1. Check, if a node is on the failed nodes list (no state change): -> isFailedNode()
2. Ping the device and update the failed nodes list, if it is not reachable: -> testFailedNode()
3. Force a device to be placed on the failed nodes list. -> markNodeAsFailed()

If I understand correctly, the 3rd option is not possible, since the failed nodes list cannot be modified directly, right?

AlCalzone:
> 3\. Force a device to be placed on the failed nodes list. -> markNodeAsFailed()
> If I understand correctly, the 3rd option is not possible, since the failed nodes list cannot be modified directly, right?

That is exactly what this method is trying to do, although we cannot influence if it will actually work.
In the case that a node is really failed (it should not respond then), the method pings it until it appears on the failed nodes list (up to 5 times max).

Regarding your original problem: An InterviewStatus of `NodeInfo` means that querying the node info is the last completed step. At this point all the necessary information to query all Command Classes (CC) is known. If a re-inclusion fixes anything here, it is a happy coincidence. Instead, being stuck at `NodeInfo` implies that the CC interview breaks somewhere.
If you really want to reset all information we have about a node, `refreshInfo` will do the trick.

AlCalzone:
superseded by #1456

robertsLando:
@AlCalzone So this method is useless right?

AlCalzone:
Yep

PR #1453, opened 2021-01-19
fix(config): delay refreshing the current value after a set command

AlCalzone:
This PR implements a short-term solution for the incorrect `currentValue` after a set command, which is due to too early polls on devices with a slow transition.

It uses the specified command duration to delay the poll and falls back to a default of 5 seconds if none was given.

I'm going to try and make this a bit smarter and eventually move it out of the driver - but until then this will be the workaround.


AlCalzone:
@zwave-js-bot pack this

PR #1456, opened 2021-01-20
fix: ping a node before attempting to remove or replace

AlCalzone:
Based on https://github.com/zwave-js/node-zwave-js/discussions/1447 this is how we should handle the node removal/replacement process.

closes: #1444 

@robertsLando you can remove the isFailed check if you do that before calling these methods.

### DIFF
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -1908,6 +1908,21 @@ ${associatedNodes.join(", ")}`,
 	 * @param nodeId The id of the node to remove
 	 */
 	public async removeFailedNode(nodeId: number): Promise<void> {
+		const node = this.nodes.get(nodeId);
+		if (!node) {
+			throw new ZWaveError(
+				`Node ${nodeId} was not found!`,
+				ZWaveErrorCodes.Controller_NodeNotFound,
+			);
+		}
+
+		if (await node.ping()) {
+			throw new ZWaveError(
+				`The node removal process could not be started because the node responded to a ping.`,
+				ZWaveErrorCodes.ReplaceFailedNode_Failed,
+			);
+		}
+
 		const result = await this.driver.sendMessage<
 			RemoveFailedNodeRequestStatusReport | RemoveFailedNodeResponse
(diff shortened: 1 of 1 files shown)


packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1885,6 +1885,21 @@ ${associatedNodes.join(", ")}`,
...
+				ZWaveErrorCodes.Controller_NodeNotFound,
+			);
+		}
+
+		if (await node.ping()) {
+			throw new ZWaveError(
+				`The node removal process could not be started because the node responded to a ping.`,
+				ZWaveErrorCodes.ReplaceFailedNode_Failed,
+			);
+		}
```
robertsLando: This is repeated later, couldn't this be a new private function like `pingNode` taht returns a node?

AlCalzone: I don't follow? The second part is in another method.
I could extract a helper getNodeOrThrow but that should be separate from the ping

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1970,6 +1985,21 @@ ${associatedNodes.join(", ")}`,
...
+				ZWaveErrorCodes.Controller_NodeNotFound,
+			);
+		}
+
+		if (await node.ping()) {
+			throw new ZWaveError(
+				`The node replace process could not be started because the node responded to a ping.`,
+				ZWaveErrorCodes.ReplaceFailedNode_Failed,
+			);
+		}
```
robertsLando: Here you use the same code you used previously, could be a function so

AlCalzone: Different error message. I don't see the benefit in having a method like getNodeOrThrowAndThrowIfPingSuceeded.
The first part, okay.

AlCalzone:
Refactored based on #1457 to address review comments

robertsLando (dismissed):
Yep much cleaner now! LGTM :)

AlCalzone:
@zwave-js-bot rebase

AlCalzone:
try again, @zwave-js-bot rebase

AlCalzone:
try again, @zwave-js-bot rebase

Issue #1349, opened 2021-01-11
[feat] Import all certified devices from the zWave Alliance


blhoward2:
**Is your feature request related to a problem? Please describe.**
The zWave Alliance has an xml file on every certified device, which includes the device id, firmware versions, configuration parameters, etc. The plan would be to import all of those provided device xmls into the project to ensure the largest number of potential working devices, amending existing device files where necessary and adding new ones. There are 3,466 device xmls available as of 1/10/2021, though there is duplication among devices due to firmware revisions so there are in practice fewer "devices" as most users would consider them. This covers all frequency spectrums/countries.

**Note: I've already downloaded all xmls over a period of time so as to not abuse their server.** Let's please not all go pull down all 3,466 xmls from the zwave alliance as I don't want them to decide the database isn't worth their server being pounded.

Here is an example file: https://products.z-wavealliance.org/Products/4024/XML

For curiosity sake, here is the year-by-year breakdown of devices included, by certification date:

2014: 71
2015: 234
2016: 425
2017: 513
2018: 398
2019: 486
2020: 280
2021: 10

These xmls also include inclusion/exclusion instructions, reset directions, etc. This information can be used to auto-generate some of the supported devices documentation: https://github.com/zwave-js/zwavejs2mqtt/issues/148

**Describe the solution you'd like**
Amend the current import configuration routines to process these xmls and to generate/amend project device files.

AlCalzone:
> These xmls also include inclusion/exclusion instructions, reset directions, etc.

I'm open to adding that (which we can always do later though). Question is how much use the inclusion instructions are, as we can only show them **after** the device was included.

blhoward2:
> I'm open to adding that (which we can always do later though). Question is how much use the inclusion instructions are, as we can only show them **after** the device was included.

They'd be in the documentation, which is what I meant, so if the user knows what device they have they could look. This was added in response to someone wanting to start hand jamming documentation.

cgarwood:
Having the wakeup/exclusion/etc metadata would be useful for the Home Assistant integration as well

AlCalzone:
@blhoward2 anything left to do here?

PR #1473, opened 2021-01-22
feat(config): import device configs from Z-Wave Alliance (part 1: no params)

blhoward2:
This is the first push of zWave Alliance files, which have no parameters included. Only minor naming changes to present files are included alongside new devices. In support of #1349


blhoward2:
@zwave-js-bot fix lint

packages/config/config/devices/0x0131/zt_zweuzbee.json
```diff
@@ -0,0 +1,22 @@
...
+// Zipatile
+{
+	"manufacturer": "Zipato",
+	"manufacturerId": "0x0131",
+	"label": "ZT.ZWEUZBEE",
+	"description": "Zipatile",
+	"devices": [
+		{
+			"productType": "0x0x03",
+			"productId": "0x0x01",
```
AlCalzone: That is not a valid HEX ID

AlCalzone (changes requested):
Please remove `\r` and `\t` from the descriptions

AlCalzone:
@blhoward2

Do we need to check before merging whether the label of a device is the start of another one? Example: `packages\config\config\devices\0x001d\vrc0p_v1.json` vs. `packages\config\config\devices\0x001d\vrc0p.json`. Or are these two different devices?

How did you determine the file labels? Or didn't you change them? Example: `packages\config\config\devices\0x0157\gr-105n.json` vs its label

blhoward2:
I may have missed that one. I’m having to do these ones more manually since they have no parameters to use to distinguish from each other. Where the label doesn’t already match I do the best I can by dropping some last characters/suffixes and comparing the files to each other. If they match it combines them and keeps the shortened label (e.g. XXXXXUS and XXXXXEU become XXXXX) and then re-adding the suffix if that didn’t result in them being merged. There are limits though as to how aggressive I can be or we add different models together. That said, I can probably add v1 to that suffix list without an issue. I’ve been using my review of this set to train it a little better if you will. 

There are a few with very similar file names but which are actually different devices according to the descriptions. 

If a device existed the file name stays the same unless I manually change it. If it’s new it uses the label as it did before.

blhoward2:
> @blhoward2
> 
> Do we need to check before merging whether the label of a device is the start of another one? Example: `packages\config\config\devices\0x001d\vrc0p_v1.json` vs. `packages\config\config\devices\0x001d\vrc0p.json`. Or are these two different devices?
> 
> How did you determine the file labels? Or didn't you change them? Example: `packages\config\config\devices\0x0157\gr-105n.json` vs its label

These are two different devices. One is an in-wall outlet, one is an appliance module.

packages/config/config/devices/0x0131/zt_zweuzbee.json
blhoward2: Fixed

blhoward2:
@zwave-js-bot fix lint

blhoward2:
@AlCalzone Unless you see other errors, these should be ready to merge. I fixed everything you noted, plus some. These should be all of the no-parameter devices.

Note: There are some places where there are two spaces and it should be one. Rather than rebuild these, that is a simple grep/sed on the command line. Doing so actually fixes a bunch of our current files too, though, so it was going to pull those into this pile making it hard for you to review. **I'll submit a second PR fixing the 2-->1 space across all current files and these ones once these are merged.** I'll of course fix importConfig.ts too, but I'd already made a bunch of manual changes to these so I don't want to have to redo all of that.

packages/config/config/devices/0x0002/hrvccm.json
```diff
@@ -1,4 +1,4 @@
-// Danfoss HRVCCM
+// Danfoss H
```
AlCalzone: wonder what happened here 🤷🏻‍♂️

packages/config/config/devices/0x007a/506219.json
```diff
@@ -39,7 +39,7 @@
 	},
 	"paramInformation": {
 		"0": {
-			"label": "Switching/dimming/shutters dual-surface:",
+			"label": "Switching/dimming/shutters dual-surface: ",
```
AlCalzone: The old label had the same flaw, but we should make sure that there are no trailing punctuation or whitespaces in labels

packages/config/config/devices/0x008a/molite.json
```diff
@@ -88,7 +88,7 @@
 					"value": 1
 				},
 				{
-					"label": "Mode 2: normal operation mode:",
+					"label": "Mode 2: normal operation mode: ",
```
AlCalzone: dito

packages/config/config/devices/0x014f/ps15emz51.json
```diff
@@ -40,7 +40,7 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "LED ON when the load is ON,",
+					"label": "LED ON when the load is ON, ",
```
AlCalzone: dito

packages/config/config/devices/0x015d/zen22.json
```diff
@@ -29,7 +29,7 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Upper paddle turns the light on,",
+					"label": "Upper paddle turns the light on, ",
```
AlCalzone: dito

packages/config/config/devices/0x016a/ft132.json
```diff
@@ -160,8 +160,8 @@
 			]
 		},
 		"90": {
-			"label": "Enables/disables parameter 91 and 92 below:",
-			"description": "Enables/disables parameter 91 and 92 below:",
+			"label": "Enables/disables parameter 91 and 92 below: ",
+			"description": "Enables/disables parameter 91 and 92 below: ",
```
AlCalzone: dito

packages/config/config/devices/0x031e/lzw36.json
```diff
@@ -47,7 +47,7 @@
 	"paramInformation": {
 		"1": {
 			"label": "Light Dimming Speed",
-			"description": "This changes the speed in which the attached light dims up or down. A setting of 0 should turn the light immediately on or off (almost like an on/off switch). Increasing the value should slow down the transition speed.\n      Range:0-98\n      Default: 4",
+			"description": "This changes the speed in which the attached light dims up or down. A setting of 0 should turn the light immediately on or off (almost like an on/off switch). Increasing the value should slow down the transition speed. Range:0-98 Default: 4",
```
AlCalzone, line 50: 👍🏻 These are nice!

packages/config/config/devices/0x0060/ad142.json
```diff
@@ -1,18 +0,0 @@
-// Everspring AD142
```
AlCalzone: What happened here? Merged with another one?

packages/config/config/devices/0x0060/an157.json
```diff
@@ -1,22 +0,0 @@
-// Everspring AN157
```
AlCalzone, line 1: And here?

AlCalzone (changes requested):
I have a few nits regarding label sanitization, but mostly looks good.

packages/config/config/devices/0x0002/hrvccm.json
blhoward2: Looks like I just missed updating the comment when either merging or making the identifier more descriptive. Fixed.

packages/config/config/devices/0x008a/molite.json
blhoward2: Fixed across all files

packages/config/config/devices/0x015d/zen22.json
blhoward2: Fixed across all files

packages/config/config/devices/0x016a/ft132.json
blhoward2: Fixed across all files

packages/config/config/devices/0x014f/ps15emz51.json
blhoward2: Fixed globally

packages/config/config/devices/0x0060/ad142.json
blhoward2: Not sure but I re-added it. Had to have been my manual mistake as this runs non-destructively (at least for now). I think I meant to delete an142 when I was merging devices.

packages/config/config/devices/0x0060/an157.json
blhoward2: This one is correct. It was merged with an145 (and an148/an142). They are actually slightly different devices (screw in lamp holder and plug-in lamp modules), but the manufacturer used the same product type/id for all.

AlCalzone (approved):

PR #1483, opened 2021-01-24
feat(config): import device configs from Z-Wave Alliance (Part 2: cleanup values and labels)

blhoward2:
Push 2: Minor config file changes from the zWave Alliance Files (primarily adding zwa file numbers and min/max values. In support of #1349


blhoward2:
@zwave-js-bot fix lint

blhoward2:
@AlCalzone These should be ready to merge. They are all minor changes to current device files. Where there is a deletion/addition of a whole file it is because I renamed the file to match a corrected identifier.

Min/max changes were all verified eyes on. These are primarily changes from 255 to 1 for binary 0/1 values.

packages/config/config/devices/0x0086/dsb09.json
```diff
@@ -36,7 +37,6 @@
 		"1": {
 			"label": "Voltage",
 			"description": "Voltage to calculate wattage from amperage.",
-			"unit": "V",
```
AlCalzone, line 39: That doesn't look correct. (same for the other removed units in this file)

packages/config/config/devices/0x0002/014g0210.json
```diff
@@ -69,7 +69,6 @@
 		"3": {
 			"label": "PWM period",
 			"description": "Pulse-width modulation period",
-			"unit": "min",
```
AlCalzone, line 72: Is this correct?

packages/config/config/devices/0x001a/rf9542.json
```diff
@@ -29,8 +30,7 @@
 	"paramInformation": {
 		"1": {
 			"label": "Delayed OFF time",
-			"description": "delay time in seconds",
-			"unit": "s",
```
AlCalzone, line 33: this param should have `"unit": "seconds"`

packages/config/config/devices/0x001a/rf9542.json
```diff
@@ -42,8 +42,7 @@
 		},
 		"2": {
 			"label": "Panic ON time",
-			"description": "time in seconds for panic mode",
-			"unit": "s",
```
AlCalzone, line 46: this param should have `"unit": "seconds"`

packages/config/config/devices/0x001a/rf9542.json
```diff
@@ -55,8 +54,7 @@
 		},
 		"3": {
 			"label": "Panic OFF time",
-			"description": "off time in seconds for panic mode",
-			"unit": "s",
```
AlCalzone, line 59: this param should have `"unit": "seconds"`

packages/config/config/devices/0x0059/sir-321.json
```diff
@@ -61,7 +62,6 @@
 		"3": {
 			"label": "Temperature reporting intervals",
 			"description": "Define the time interval for the temperature report that is sent.",
-			"unit": "sec",
```
AlCalzone, line 64: this param should have `"unit": "seconds"` (I think)

packages/config/config/devices/0x0060/hac01.json
```diff
@@ -38,7 +39,6 @@
 		"2": {
 			"label": "Amount Of Delay",
 			"description": "Amount Of Delay (seconds) to send the Basic Set Off command.",
-			"unit": "s",
```
AlCalzone, line 41: this param should have `"unit": "seconds"`

packages/config/config/devices/0x0115/zme_06433.json
```diff
@@ -65,10 +66,9 @@
 		"2": {
 			"label": "Automatically switch off after",
 			"description": "If not zero, automatically switch off after a user defined time",
-			"unit": "Sec",
```
AlCalzone, line 68: this param should have `"unit": "seconds"`

packages/config/config/devices/0x0115/zme_06433.json
```diff
@@ -65,10 +66,9 @@
...
 			"valueSize": 2,
 			"minValue": 0,
-			"maxValue": 65535,
```
AlCalzone, line 71: And this doesn't seem correct either. The param is used to set the interval. The options are what is wrong.
--> remove options, set "allowManualEntry" to true.

packages/config/config/devices/0x0115/zme_06433.json
```diff
@@ -138,10 +138,9 @@
 		"5": {
 			"label": "Dimming speed",
 			"description": "Time to dim on button presses",
-			"unit": "10ms",
```
AlCalzone, line 141: I'm sure this was here for a reason

packages/config/config/devices/0x0115/zme_06433.json
```diff
@@ -138,10 +138,9 @@
...
 			"valueSize": 1,
 			"minValue": 0,
-			"maxValue": 255,
+			"maxValue": 30,
```
AlCalzone: no

AlCalzone: Get rid of the options instead

AlCalzone (changes requested):
I don't think this is entirely correct. In cases where `allowManualEntry` is true, you cannot just take the min/max of the options. The user is free to choose any value between min and max in this case and the options are just "presets".

Besides from this, I've got a couple of comments

marcus-j-davies:
@blhoward2 @AlCalzone ,

Not sure if this is too late, but https://github.com/zwave-js/node-zwave-js/pull/1473 mentions to remove `\r` and / or `\n`
is it too late to keep them? I can swop these new line characters for `<br />` at render time.

Some of the metadata's are very "busy" hence the line breaks make it easier on the eye.

Screenshot 2021-01-24 at 13 14 07

AlCalzone:
@marcus-j-davies I don't want to remove it entirely, but make it consistent. Only \n, no \r, no \t, etc.

Hence why I proposed markdown so you could also render some formatting easily.

marcus-j-davies:
As long as there are still `\n` I can use them to break the lines up.

I still think markdown is risky to use here, the layout is very organised, given that markdown can render tables, whole HTML elements etc etc, it could introduce artifacts that affects the non-markdown code.

I have only ever used markdown for read me documents, GitHub type stuff.

marcus-j-davies:
@blhoward2 

This XML [2509](https://products.z-wavealliance.org/Products/2509/XML) 
Has spaces where I would expect them i.e **the following warning. CAUTION: This will factory**
And for the JSON Equivalent **the following warning.CAUTION: This will factory** 

Apologies, but wasn't sure if that is still a working progress, or is a glitch in the conversion routine?

packages/config/config/devices/0x0086/dsb09.json
blhoward2: This is being done by your normalize routine, so I assumed you wanted it.

packages/config/config/devices/0x0115/zme_06433.json
blhoward2: These maxes come from the file. Are you sure it can handle 255?

blhoward2:
> @blhoward2
> 
> This XML [2509](https://products.z-wavealliance.org/Products/2509/XML)
> Has spaces where I would expect them i.e **the following warning. CAUTION: This will factory**
> And for the JSON Equivalent **the following warning.CAUTION: This will factory**
> 
> Apologies, but wasn't sure if that is still a working progress, or is a glitch in the conversion routine?

I messed up at one point removing \n instead of replacing them with a space. I’ve had to chase them manually in that pile but I’ll take a look. It’s fixed in the others.

packages/config/config/devices/0x0115/zme_06433.json
AlCalzone: I don't see any config parameters in the file https://products.z-wavealliance.org/products/19/json
Are you sure you didn't infer that from the option values?

marcus-j-davies:
No worries,

If I understand it correctly, we are replacing `\r` with... well... nothing "" so leaving `\n` in tact?
so any`\r\n` will change to `\n`

Re:
>Please remove \r and \t from the descriptions

blhoward2:
> @blhoward2 @AlCalzone ,
> 
> Not sure if this is too late, but #1473 mentions to remove `\r` and / or `\n`
> is it too late to keep them? I can swop these new line characters for `<br />` at render time.
> 
> Some of the metadata's are very "busy" hence the line breaks make it easier on the eye.
> 
> Screenshot 2021-01-24 at 13 14 07

I'm going to redo these in a new PR.

marcus-j-davies:
> I'm going to redo these in a new PR.

I think it will look pretty slick once formatted (replacing `\r\n`with `\n`)

Screenshot 2021-01-24 at 16 29 04

packages/config/config/devices/0x0115/zme_06433.json
blhoward2: I've changed this back to 65535, verified in the manual. I also added a few missing parameters.

packages/config/config/devices/0x0115/zme_06433.json
blhoward2: Confirmed 10 ms in manual

blhoward2:
@AlCalzone All ready.

AlCalzone (approved):

PR #1484, opened 2021-01-24
feat(config): document zwaveAllianceId and metadata fields

AlCalzone:
This PR documents the fields added in #1473 and updates the JSON schema to support the new properties

### DIFF
--- a/docs/development/config-files.md
+++ b/docs/development/config-files.md
@@ -14,6 +14,7 @@ The following properties are defined and should always be present in the same or
 | `description`       | A longer description of the device, usually the full name                                                             |
 | `devices`           | An array of product type and product ID combinations, [see below](#devices) for details.                              |
 | `firmwareVersion`   | The firmware version range this config file is valid for, [see below](#firmwareVersion) for details.                  |
+| `metadata`          | Metadata that is intended to help the user, like inclusion instructions etc. [See below](#metadata) for details.      |
 | `associations`      | The association groups the device supports, [see below](#associations) for details.                                   |
 | `supportsZWavePlus` | If set to `true`, the device complies with the Z-Wave+ standard. In this case, omit the `associations` property.      |
 | `proprietary`       | A dictionary of settings for the proprietary CC. The settings depend on each proprietary CC implementation.           |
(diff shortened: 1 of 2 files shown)


robertsLando:
@AlCalzone i would also support the image link

AlCalzone:
It is not in any files yet (and not parsed by zwave-js yet). Will add that as soon as @blhoward2 gets to it and we've found consensus on a name.

blhoward2:
@AlCalzone @robertsLando @marcus-j-davies The issue with the image links is, we have to decide where they'll be hosted. Thoughts on that? The links in the zwa files are not absolute and we shouldn't link to them, so we'd need to download and host them.

marcus-j-davies:
There isn't a lot of options here.

 1) Link to them - is their small print to stop this? we could credit them in the Browser?
 2) CDN/Host - the obvious costs in doing so.
 3) Package them (either as a file, or as a data URI in the config) - the size of the package in doing so.
 
 Point 3 can be improved, i.e reduce the resolution of the image to a few kb, IF that is where we go.

AlCalzone:
The images seem to be quite large (around 500-1000 KB per file), so I don't like packaging them.
If we can verify (@blhoward2) that the links are always like `https://products.z-wavealliance.org/ProductImages/ProductImage?prod=<id>&which=21`, I can set up a reverse-proxy (under the existing domain) with auto-resize and cache that handles the images. So best case, the alliance server is only ever queried once for each image.

marcus-j-davies:
One note, whilst it doesn't make a difference to browser (I will just use the URI value from the config).
This is how I envision  the image being used (and to give you some indication on resolution)

Thoughts?

Untitled

AlCalzone:
Another use case would be UIs in automation systems which might need a higher resolution.
So maybe just include the desired size in the self-hosted URL.

marcus-j-davies:
So would nginx resize in real time?
or are you thinking the frontend should be responsible (but a with a base resolution being offered)

I'm thinking of resources needed to do Real time resolution changes, with the potential of high volume traffic.

Again - I might not be giving enough credit to performance these days

AlCalzone:
nginx has plugins that can handle it. The resized images will be stored, so it only ever needs to happen once per file.

I have a server set up that is solely responsible to be an SSL-enabled (letsencrypt) reverse proxy for a couple of hosted websites that would charge extra for SSL.
It is currently very bored, so a bit of extra work won't hurt. If traffic becomes an issue, I can set up another one.

PR #1492, opened 2021-01-25
feat(config): track file hash in device index and regenerate on demand

AlCalzone:
With this PR, we hash the existing config files and store that hash in the index to detect if it needs to be regenerated. This should make it easier to add, lint and test new configurations because the index.json no longer needs to be regenerated manually.


AlCalzone:
Making this a draft for now - I have an alternate approach without hashing in mind.

PR #1495, opened 2021-01-25
feat(config): support EcoDim dimmers

pascalwinters:
Adds support for EcoDim 0.7 Basic variant and optimized EcoDim 0.7 configuration.


blhoward2:
@pascalwinters From where did you get these parameters? The zwave alliance shows two devices, the DIM.10 and the DIM.7. You're calling this the DIM.07 but have the product ID for the DIM.10 (0x0002) instead of the DIM.07 (0x0001). You're also including the parameters for the DIM.10, which include many more options than the DIM.07. From where are you getting your ProductID and that it is the DIM.07? What is your firmware version?

What you're calling the DIM.07 Basic is actually the DIM.10 in the database. This may be a DIM.07 with new firmware that now supports more parameters. They may sell the DIM.07 (some variant) and the DIM.10 that report the same info, but if so the device file should reflect that. And the original one (0x0001) should be called just the DIM.07.

Edit: I suspect from googling that this is a firmware split or multiple devices using the same Product ID. The DIM.07 has the shorter parameter list. The Dim.07 Pro does show the parameters from the DIM.10. But unless you can verify it I'm not sure this is the correct Product ID. If it is, the shorter one should be the DIM.07 and this one should be the DIM.07 Pro / Dim.10. If it isn't, it may be that the two DIM.07s have the same product ID but at a specific firmware version they pickup the extra options. Make sense? In any event this is wrong as what we’re calling the DIM.07 Pro isn’t that device and that device had many more parameters available. 

DIM.07: https://products.z-wavealliance.org/products/3796

DIM.10: https://products.z-wavealliance.org/products/3932
https://www.ecodim.nl/eco-dim10-zwave.html

@AlCalzone Hold off on merging this one as I'm not sure that it is correct.

pascalwinters:
@blhoward2 I have all the info from the ZWaveJS2MQTT dashboard:
image
Here you can see the ProductId 0x002, and I have an ECO-DIM.07 Basic variant.
I had contact with the manufacturer yesterday, the basic variant isn't yet online on zwavealliance, it's a matter of time.

I copied the parameters from the 07 Pro version, because it's a copy of that device. But maybe it's better to contact the supplier and ask for all information and process it in de config files. Maybe I can then also add the 10 variant, and I can also the mystery x0202 productid.

I will contact the supplier to clear up everything.

AlCalzone:
@pascalwinters some manufacturers reuse the same device IDs over multiple firmwares. It might be that we have to distinguish the devices by firmware version. Any chance you see your version anywhere? If not, you could send me your cache files and I'll check.

pascalwinters:
@AlCalzone I have firmware version 1.12.
Is there an central place to see all zwave firmware versions?

AlCalzone:
You mean in zwavejs2mqtt? I don't think so. @robertsLando ?

robertsLando:
> Is there an central place to see all zwave firmware versions?

Could be good to add the fw version in the table? @pascalwinters

pascalwinters:
@AlCalzone No not especially in zwavejs2mqtt, but how and where can you see if there is an new firmware available?

pascalwinters:
> > Is there an central place to see all zwave firmware versions?
> 
> Could be good to add the fw version in the table? @pascalwinters

Which table do you mean? @robertsLando 
It's already available in the zwavejstmqtt:
image

robertsLando:
@pascalwinters I mean the main table, where you see all your nodes

pascalwinters:
> @pascalwinters I mean the main table, where you see all your nodes

@robertsLando Maybe it can, but it's not part of this PR

AlCalzone:
> where can you see if there is an new firmware available?

That's not possible

pascalwinters:
I have requested the info by the manufacturer. I will edit this PR when I received the info.

pascalwinters:
The manufacturer indicates that the chips used in the 3 dimmers are all the same and can therefore all have the same product ID. 
@AlCalzone How do we handle this?

I have replied with the question if we can detect it with the firmware versions.

AlCalzone:
That is a weird explanation. Let's wait for info on the firmware.

pascalwinters:
Ok, we’ll wait. Otherwise I will call them tomorrow.

packages/config/config/devices/0x0431/ecodim.json
```diff
@@ -101,29 +110,29 @@
 			"description": "Enable or Disable external switch to pair network. Default value: 0.",
 			"valueSize": 1,
 			"minValue": 0,
-			"maxValue": 255,
-			"defaultValue": 0,
+			"maxValue": 1,
+			"defaultValue": 1,
```
AlCalzone: Description says "default 0", line 114 says "default 1".

packages/config/config/devices/0x0431/ecodim.json
```diff
@@ -101,29 +110,29 @@
 			"description": "Enable or Disable external switch to pair network. Default value: 0.",
```
AlCalzone:
```suggestion
			"description": "Enable or Disable external switch to pair network.",
```
I prefer not having default values or value ranges in the description.

packages/config/config/devices/0x0431/ecodim.json
```diff
@@ -78,13 +83,17 @@
...
+				{
+					"label": "Send Switch Multilevel Report only when Load status is not changed by Z-WAVE Command",
+					"value": 2
 				}
 			]
 		},
 		"3": {
-			"label": "Dimming speed",
+			"label": "Default dimming speed",
 			"description": "Default dimming speed，unit is second. Valid values: 0-100. Default value: 5 0 - Instanly. 1 - Fast. ... 100 - Slow.",
```
AlCalzone:
```suggestion
			"description": "Default dimming speed",
```
but that is rather pointless

AlCalzone (changes requested):

packages/config/config/devices/0x0431/ecodim.json
pascalwinters: Yes agree, but it's from the manual

packages/config/config/devices/0x0431/ecodim.json
pascalwinters: Changed it

AlCalzone (approved):

PR #1498, opened 2021-01-26
fix(perf): improve ValueDB lookup performance

### LINKED ISSUE #1493: Performance bottleneck in getDefinedValueIDs
grafik

Not sure what that `(bytecode) ~Abort.ExtraWide` is all about but we can try iterating without the `forEach` instead.

/cc @marcelveldt

AlCalzone:
This PR attempts to fix the performance issues found in #1493. The Value DB is deliberately huge (550000 values), so we can judge the impact better.

## Step 1:
`dbKeyToValueIdFast` is optimized, but still too slow to be in the hot path. We now filter values using string comparison instead of parsing JSON first and likely discarding the result afterwards.

**Before:**
```
db contains 550000 values
calling getValues 100 times
took 550.62 ms / call
```

**After:**
```
db contains 550000 values
calling getValues 100 times
took 44.97 ms / call
```

**total speedup:** 12x

## Step 2:
The `clear()` methods were parsing the DB keys although the result is only used conditionally. That no longer happens now.
**Before:**
```
calling clear for each node (100)
took 106.55 ms / call
```

**After:**
```
calling clear for each node (100)
took 97.88 ms / call
```

## Step 3:
`indexOf` turns out to be faster, but only if we include the leading `{` and `,`.
**After:**
```
calling getValues 100 times
took 39.94 ms / call
calling clear for each node (100)
took 95.81 ms / call
```

**total speedup:** 13.8x

## Step 4:
Keeping track of the own IDs completely avoids looping altogether. However this adds an upfront cost for the index creation:
**Before:**
```
index value DB: 0.162ms (NO-OP)

calling getValues 100 times
took 40.04 ms / call

calling clear for each node (100)
took 91.89 ms / call
```

**After:**
```
index value DB: 6770.065ms

calling getValues 100 times
took 1.23 ms / call

calling clear for each node (100)
took 12.35 ms / call
```

**total speedup:** 🔥 447x (ignoring the upfront cost)

## Step 5: TODO: Speed up index generation
**After:**
```
index value DB: 122.762ms

calling getValues 100 times
took 1.23 ms / call

calling clear for each node (100)
took 12.31 ms / call
```

**total speedup:** 🔥 447x (with negligible upfront cost 🎉)

---

/cc @marcelveldt
fixes: #1493

### DIFF
--- a/packages/core/src/values/ValueDB.ts
+++ b/packages/core/src/values/ValueDB.ts
@@ -310,9 +310,9 @@ export class ValueDB extends EventEmitter {
 		});
 
 		oldValues.forEach(([key, prevValue]) => {
-			const { nodeId, ...valueId } = this.dbKeyToValueId(key);
 			this._db.delete(key);
 			if (options.noEvent !== true) {
+				const { nodeId, ...valueId } = this.dbKeyToValueId(key);
 				const cbArg: ValueRemovedArgs = {
 					...valueId,
 					prevValue,
@@ -321,10 +321,10 @@ export class ValueDB extends EventEmitter {
 			}
 		});
 		oldMetadataKeys.forEach((key) => {
-			const { nodeId, ...valueId } = this.dbKeyToValueId(key);
 			this._metadata.delete(key);
 
 			if (options.noEvent !== true) {
+				const { nodeId, ...valueId } = this.dbKeyToValueId(key);
 				const cbArg: MetadataUpdatedArgs = {
 					...valueId,
 					metadata: undefined,
--- a/packages/core/src/values/ValueDB.ts
+++ b/packages/core/src/values/ValueDB.ts
@@ -160,11 +160,26 @@ export class ValueDB extends EventEmitter {
 		this.nodeId = nodeId;
 		this._db = valueDB;
 		this._metadata = metadataDB;
+
+		this._index = this.buildIndex();
 	}
 
 	private nodeId: number;
 	private _db: JsonlDB<unknown>;
 	private _metadata: JsonlDB<ValueMetadata>;
+	private _index: Set<string>;
+
+	private buildIndex(): Set<string> {
+		const ret = new Set<string>();
+		for (const key of this._db.keys()) {
+			if (compareDBKeyFast(key, this.nodeId)) ret.add(key);
+		}
+		for (const key of this._metadata.keys()) {
+			if (!ret.has(key) && compareDBKeyFast(key, this.nodeId))
+				ret.add(key);
+		}
+		return ret;
+	}
 
 	private valueIdToDBKey(valueID: ValueID): string {
 		return JSON.stringify({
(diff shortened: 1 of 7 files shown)


robertsLando:
@AlCalzone indexOf is even faster: https://jsben.ch/om6Xb (around 20%)

Anyway I think that this could be even faster with a cache

test/valuedb-perf.ts
```diff
@@ -0,0 +1,50 @@
...
+
+	console.log(`db contains ${values.size} values`);
+	const numRounds = 100;
+	console.log(`calling getValues ${numRounds} times`);
+
+	const start = highResTimestamp();
+	for (let i = 1; i <= numRounds; i++) {
+		valueDBs.get(1)!.getValues(5);
+	}
+	const duration = highResTimestamp() - start;
```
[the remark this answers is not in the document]
AlCalzone: Not if I want to average per run

AlCalzone:
> indexOf is even faster

Gave it a try 👎🏻:
```
calling getValues 10 times
took 1362.05 ms / call
```

robertsLando:
> Gave it a try 👎🏻:

Lol maybe that's only for browsers :man_shrugging:

AlCalzone:
Nevermind, I messed up.
```
db contains 550000 values
calling getValues 100 times
took 41.28 ms / call
```

robertsLando:
@AlCalzone Personally I think that the build index time is really too high, I think that the step 3 is enought

packages/core/src/values/ValueDB.ts
```diff
@@ -160,11 +160,26 @@ export class ValueDB extends EventEmitter {
...
+
+	private buildIndex(): Set<string> {
+		const ret = new Set<string>();
+		for (const key of this._db.keys()) {
+			if (compareDBKeyFast(key, this.nodeId)) ret.add(key);
+		}
+		for (const key of this._metadata.keys()) {
+			if (!ret.has(key) && compareDBKeyFast(key, this.nodeId))
+				ret.add(key);
+		}
```
robertsLando, line 188: Use normal for loops with index, is much faster then for-of

marcelveldt: Is the index build only once at startup ? In that case it is much more preferred to have fast lookups at runtime and accept the few seconds at startup.

AlCalzone: Not possible. `Map.keys()` returns an iterator.

marcelveldt: Great job btw so far hunting this down!

AlCalzone: Somehow I missed your comment @marcelveldt. Yes the index is built only once at startup but it currently does (# of nodes)x the work that is necessary.
So I'm going to look into that as a last step.

AlCalzone:
Index generation for the 550k values DB is down to ~123 ms, so all concerns should be addressed.

@marcelveldt I'll ask the bot to create a test build so you can check this out.
@zwave-js-bot pack this.

robertsLando:
@AlCalzone Awesome!

marcelveldt:
I gave this version a testdrive and it's REALLY amazing. That nasty dumpState call in the server took 7+ seconds on my main mesh yesterday (without this fix) and this now takes about 50ms :-)

Also can confirm that it didn't break anything else I could discover.

PR #1503, opened 2021-01-26
feat(config): use mtime to detect if index needs to be regenerated

AlCalzone:
With this PR, we compare the `mtime` of the index.json with the maximum `mtime` of the device config directory (recursively) to detect if regenerating the index is necessary.

This means that adding and testing new configuration files is easier because the index is updated automatically.

closes #1492

### DIFF
--- a/packages/config/src/Devices.ts
+++ b/packages/config/src/Devices.ts
@@ -1,4 +1,3 @@
-import { ZWaveError } from "@zwave-js/core";
 import {
 	JSONObject,
 	ObjectKeyMap,
@@ -7,6 +6,7 @@ import {
 } from "@zwave-js/shared";
 import { entries } from "alcalzone-shared/objects";
 import { isArray, isObject } from "alcalzone-shared/typeguards";
+import * as fs from "fs-extra";
 import { pathExists, readFile, writeFile } from "fs-extra";
 import JSON5 from "json5";
 import path from "path";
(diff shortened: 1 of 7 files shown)


packages/config/src/Devices.ts
```diff
@@ -41,66 +41,107 @@ export const devicesDir = path.join(configDir, "devices");
...
+		const stat = await fs.stat(fullPath);
+		if (
+			(dir !== devicesDir || f !== "index.json") &&
+			(stat.isFile() || stat.isDirectory()) &&
+			stat.mtime > lastChange
+		) {
+			return true;
+		} else if (stat.isDirectory()) {
+			// we need to go deeper!
+			if (await hasChangedDeviceFiles(fullPath, lastChange)) return true;
```
robertsLando, line 64:
```suggestion
			return hasChangedDeviceFiles(fullPath, lastChange);
```

robertsLando: Also is mtime and lastChange in timestamp format? Otherwise I suggest to use getTime to comprare the dates

packages/config/src/Devices.ts
```diff
@@ -41,66 +41,107 @@ export const devicesDir = path.join(configDir, "devices");
...
+	let index: DeviceConfigIndex | undefined;
+	let mtimeIndex: Date | undefined;
+	// ...or if cannot be parsed
+	if (!needsUpdate) {
+		try {
+			const fileContents = await readFile(indexPath, "utf8");
+			index = JSON5.parse(fileContents);
+			mtimeIndex = (await fs.stat(indexPath)).mtime;
+		} catch {
+			// console.error("Error while parsing index file - regenerating...");
```
robertsLando: Remove console.error or replace them with logger errors

packages/config/src/Devices.ts
```diff
@@ -41,66 +41,107 @@ export const devicesDir = path.join(configDir, "devices");
...
+		if (forceWrite || process.env.NODE_ENV !== "test") {
+			// Save the index to disk (but not during unit tests)
+			await writeFile(
+				path.join(devicesDir, "index.json"),
+				`// This file is auto-generated. DO NOT edit it by hand if you don't know what you're doing!"
+${stringify(index, "\t")}
+`,
+				"utf8",
+			);
+			console.log(`Device index regenerated`);
```
robertsLando: Use logger here

robertsLando (changes requested):

packages/config/src/Devices.ts
AlCalzone: That proposed change is not correct. We loop through the sub-directories (depth first) and if one directory has (deep) changes, we bail. If not, we need to continue with the next sub-directory.
Your change makes it so that we dive into the first directory and ignore all others.

`mtime` is a Date object, so is `lastChange` (see line 46):
```
lastChange: Date,
```

packages/config/src/Devices.ts
AlCalzone: that's commented out, but I can remove it ;)

packages/config/src/Devices.ts
AlCalzone: It is not guaranteed that we have a logger instance available. The code is also used in unit and CI tests without a driver instance.

robertsLando: So log if logger is present and use console if not?

AlCalzone: I'm not so invested in that log line, I'll rather remove it altogether 😅
Too much hassle passing the driver instance through ;)

packages/config/src/Devices.ts
robertsLando: Yep sorry, by looking at it again it's wrong :)

robertsLando: About dates, I see them are `Date` objects, usually when dealing with dates I prefer to convert them to millis (Date.getTime()) as for example the equality operator would not work between two dates. But it's ok you can keep it like it is :)

packages/config/src/Devices.ts
robertsLando: I mean it's useful so for example if log to file is enabled you will see that

packages/config/src/Devices.ts
robertsLando: I would use the logger here too (if present), fallback to console if not

packages/config/src/Devices.ts
AlCalzone: > the equality operator would not work between two dates

I've stumbled across that, but `>` and `>=` work :)

robertsLando (approved):
LGTM

PR #1515, opened 2021-01-27
fix(cc): improve value labels for Meter CC

### LINKED ISSUE #1485: MeterCC.ts creates value labels that cannot be distinguished
**Describe the bug**

When selecting a meter value in zwavejs2mqtt (e.g. from a Shenzehen Neo NAS-WR01ZE Wall Plug) the selection box shows identical labels for all meter values which cannot be distinguished for which unit it is (kWh, W, V, A).

This is a screenshot from the value selection:
image

This is a screenshot that shows that multiple values have identical labels and would not be able to distinguish, if the UI would not output the unit as well:
image

Maybe the value labels could be improved a bit as well (shorter with more meaningful names).

**Device information**

Which device(s) is/are affected (make/model)?  Shenzehen Neo NAS-WR01ZE Wall Plug
What are the node IDs? Node 080 (see logs). Device IDs: 600-4135-512 (0x0258-0x1027-0x0200)

The bug affects not only this device but all devices using the meter CC.

**Installation information**
How did you install `node-zwave-js`? Zwavejs2mqtt (dev) docker image

**To Reproduce**
1. Go to zwavejs2mqtt settings
2. Click on 'New Value' on the bottom left
3. Select a device supporting the meter CC
4. Open the value select - it shows multiple identically named values, unable to distinguish

**Logfile:**
[zwavejs-log.zip](https://github.com/zwave-js/node-zwave-js/files/5862252/zwavejs-log.zip)

AlCalzone:
fixes: #1485

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/MeterCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/MeterCC.ts
@@ -139,6 +139,24 @@ export function getResetValueId(endpoint: number, type?: number): ValueID {
 	};
 }
 
+function getValueLabel(
+	configManager: ConfigManager,
+	meterType: number,
+	scale: MeterScale,
+	rateType: RateType,
+	suffix?: string,
+): string {
+	let ret = getMeterTypeName(configManager, meterType);
+	if (rateType !== RateType.Unspecified) {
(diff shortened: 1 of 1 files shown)


AlCalzone:
@ahochsteger you might wanna try this :)

@zwave-js-bot pack this

PR #1516, opened 2021-01-27
fix: correctly implement Z-Wave+ Role Type Specs for interview procedure

### LINKED ISSUE #1450: Change interview procedure to match Z-Wave+ Role Type Specs
> From https://github.com/zwave-js/node-zwave-js/issues/1037#issuecomment-762753312
After the protocol inclusion is completed, and security is negotiated, the first thing a controller should do with a Z-Wave Plus device is setup its Lifeline association.

AlCalzone:
With this PR, we match the Z-Wave+ Role Type Specs in that we now set up the Z-Wave+ Lifeline, WakeUp destination and return route to controller as soon as possible. Not doing it caused issues with the ID Lock 150 which made it look like the device reused nonces.

Since the non-spec-compliant compat option `keepS0NonceUntilNext` isn't necessary anymore, we drop that too, along with the ability for fallback confing files without firmware versions.

fixes: #1450


AlCalzone:
@zwave-js-bot pack this

roflmao:
[zwave_node_39.log](https://github.com/zwave-js/node-zwave-js/files/5889585/zwave_node_39.log)

Interview failed after exclude, zwave reset, include

AlCalzone:
@roflmao I think you need to try again:
```
20:19:11.411 DRIVER » [Node 039] [REQ] [SendData]
                      │ transmit options: 0x25
                      │ callback id:      48
                      └─[SecurityCCSchemeGet]
...
20:19:14.938 DRIVER « [REQ] [SendData]
                        callback id:     48
                        transmit status: NoAck
...
20:19:21.415 CNTRLR   [Node 039] Security bootstrapping failed, the node is included insecurely: a s
                      ecure inclusion timer has elapsed.
```
The node could not be included securely because it didn't respond, so it won't communicate correctly. Guess we could a better job displaying that.

roflmao:
[zwave_node_40.log](https://github.com/zwave-js/node-zwave-js/files/5893368/zwave_node_40.log)

Interview finished, without config it seems (unknown manufacter/produt)

AlCalzone:
Hmm, this time it didn't tell us the securely supported commands:
```
11:50:52.848 CNTRLR   [Node 040] Querying securely supported commands timed out, skipping interview.
                      ..
```
Is that device far from the controller? Could you move it closer?

This time, re-interviewing instead of re-including should be enough.

AlCalzone:
@roflmao I've changed one more thing that happens while including, so if you don't mind, re-including it would be appreciated.

roflmao:
The lock is a bit far away, inclusion is done via neighbors. I would need some time to move the stick closer and then ill test.

AlCalzone:
Ok maybe the last commit improves that behavior since the return route is now configured as the very first thing we do after inclusion.

roflmao:
[zwave_node_40.log](https://github.com/zwave-js/node-zwave-js/files/5894454/zwave_node_40.log)

Seems the interview got completed after some time (no re-include, only restart with last zwave-js rev)
Though it did not get all 109 User Codes.

PR #1522, opened 2021-01-28
fix: treat no error as successful command, update currentValue immediately

### LINKED ISSUE #1519: [bug] recent changes exacerbate issue of switch turning back on in HA during transitions
**Describe the bug**

Changes made in response to #452 exacerbate the switch flipping back on in HA on my setup. Coincidentally, I inadvertently started up an old image from two weeks ago that doesn't have the flipping issue at all, suggesting interim changes are the cause. Regardless, #1453 makes it worse than it was before that PR. The issue is non-existent in a build off 1.0.0-alpha.2+9f99728.9f99728.

**Device information**

This is a GE 14294, with a step of 20 and a timing of 1 (10 ms).

- [ ] New device
- [ ] Previously working device (node-zwave-js)
    - Which library version/docker image/adapter version?

    - Which platform?
    
**Installation information**
- [ ] ioBroker.zwave2 adapter
- [ ] Pkg
- [X ] Manual Docker build
    - `node-zwave-js` branch: master
    - `zwavejs2mqtt` branch: master
- [ ] Other: 

**To Reproduce**
2. Click on '...'
3. Scroll down to '...'

**Additional context**

**Logfile:**

### LINKED ISSUE #1521: Handle all refresh-after-set calls similar

AlCalzone:
With this PR, we no longer refresh or update values in the lower level `commandClasses` APIs. Instead this now happens in the `SET_VALUE` API if the low-level command succeeded. A few non-critical CCs (like name/location) no longer automatically poll and instead just assume it worked.

Also, we now update `currentValue` immediately where applicable and we update the newly set value in the value DB if the command was successful, so UIs should no longer flicker between different states.

Fixes: #1519
Fixes: #1521

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/BinarySwitchCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/BinarySwitchCC.ts
@@ -135,16 +135,7 @@ export class BinarySwitchCCAPI extends CCAPI {
 			// Verify the current value after a delay
 			// TODO: #1321
 			const duration = undefined as Duration | undefined;
-
-			if (this.refreshTimeout) clearTimeout(this.refreshTimeout);
-			setTimeout(async () => {
-				this.refreshTimeout = undefined;
-				try {
-					await this.get();
-				} catch {
-					/* ignore */
-				}
-			}, duration?.toMilliseconds() ?? 1000).unref();
+			this.schedulePoll({ property }, duration?.toMilliseconds() ?? 1000);
 		}
 	};
 
--- a/packages/zwave-js/src/lib/commandclass/BinarySwitchCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/BinarySwitchCC.ts
@@ -37,11 +37,11 @@ import {
 	implementedVersion,
 } from "./CommandClass";
 
-function getTargetValueValueId(endpoint?: number): ValueID {
+function getCurrentValueValueId(endpoint?: number): ValueID {
 	return {
 		commandClass: CommandClasses["Binary Switch"],
 		endpoint,
-		property: "targetValue",
+		property: "currentValue",
 	};
 }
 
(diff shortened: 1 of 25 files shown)


packages/zwave-js/src/lib/commandclass/BinarySwitchCC.ts
```diff
@@ -145,6 +122,21 @@ export class BinarySwitchCCAPI extends CCAPI {
...
+			valueDB?.setValue(
+				getCurrentValueValueId(this.endpoint.index),
+				value,
+			);
+
+			// Verify the current value after a delay
+			// TODO: #1321
+			const duration = undefined as Duration | undefined;
+			this.schedulePoll({ property }, duration?.toMilliseconds() ?? 1000);
+		}
```
nevir, line 139: Given the frequency of this pattern, I wonder if abstracting into a `setAndVerify` abstraction might be a good idea? Something like…

```suggestion
		// Verify the current value after a delay
		// TODO: #1321
		const duration = undefined as Duration | undefined;
		await this.setAndVerify(value, duration?.toMilliseconds() ?? 1000);
```

packages/zwave-js/src/lib/commandclass/ThermostatModeCC.ts
```diff
@@ -94,6 +94,11 @@ export class ThermostatModeCCAPI extends CCAPI {
 			throwWrongValueType(this.ccId, property, "number", typeof value);
 		}
 		await this.set(value);
+
+		if (this.isSinglecast()) {
+			// Verify the current value after a delay
+			this.schedulePoll({ property });
+		}
```
nevir, line 101: Dumb question since I'm not that familiar w/ the code at all yet - should this also be setting it in the `valueDB`?

nevir (commented):
Hopefully it's ok for a rando to leave some comments - hoping to slowly ramp up on this repo a bit; but I have very little context, so take everything with a grain of salt!

packages/zwave-js/src/lib/commandclass/BinarySwitchCC.ts
AlCalzone: The problem here is that the value we're immediately updating is different for each CC.
I plan on reworking the value definitions, but until then that's not doable

packages/zwave-js/src/lib/commandclass/ThermostatModeCC.ts
AlCalzone: Thermostat mode only has a single value. That gets stored in node.setValue()

PR #1542, opened 2021-01-31
chore(config): add check for similar descriptions

AlCalzone:
With this PR, the config linter computes the normalized Levenshtein distance between param label and description. If that is less than 0.5 (i.e. very similar), the parameter gets flagged.

@blhoward2 I think we need your script:
```
@zwave-js/config: Found 0 errors and 2855 warnings!
```

Most of thes are identical param label and description. Others are actually similar descriptions:
```
@zwave-js/config: packages/config/config/devices/0x5254/zxt-600_1.3.json:
@zwave-js/config: [WARN] Parameter #34 has a very similar label and description (normalized distance 0.17). Consider removing the description if it does not add any information:
@zwave-js/config: label:       Set Auto Report Cond by Time Interv
@zwave-js/config: description: Set Auto Report Condition by Time Interval
```


AlCalzone:
Down to 711 warnings by removing the exact duplicates

blhoward2:
We can try it, but I tried the Levenshtein and eventually gave up. The parameters do similar things and have short values so there will be a huge number of false hits. There is also no way to programmatically rewrite a similar description so every one will need to be verified and corrected by hand. 

Similarity also isn't a bad thing for most. Dim Step (Manual) and Dim Step (Z-Wave) should be similar. As should Dim Steps and Dim Timing.

blhoward2:
I've looked at every description for every file I've pushed as well. As have you. Can we exclude those somehow?

blhoward2:
Also, are we ok with having no description if the label makes it apparent? I've been making them up when they're similar or missing.

blhoward2:
What do you want me to do? Just give me a list. You want me to parse and remove exact matches to start?

I'd wait to worry about the similar ones. We look at those anyway as we process the zwa files and the zwa files overwrite the prior descriptions. What's left are the most complex devices with lots of parameters so I think we'll hit much of that as we go. 

(Also ignore my deleted comment if you get them by email. I was thinking labels for which many are and should be similar.)

AlCalzone:
I don't think there's anything that needs to be done right now if we can live with the warnings for now. The identical ones were easy to remove with a regex, the others need a script that is a little more elaborate (or we just fix them while we add/import new files).
Except if you have a clever trick up your sleeve, that is.

Would merging this now mess up your current work?

blhoward2:
No, everything I'm working on at the moment is a new device. I'll rebase before I do the next run.

PR #1556, opened 2021-02-02
fix(config): remove double tap support from small number of early ge devices

### LINKED ISSUE #1555: GE 14291 switch failing to discover double tap error in logs
**Describe the bug**

When using zwavejs2mqtt 1.0.2 and zwavjs 6.1.1 the GE 14291 switch gives an error on discovery.  This is due to firmware version 5.20 not support double tap, so there is a failure getting the discovery_payload.   
This impacts firmware 5.20 on the GE/Jasco 14291/14292 

**Device information**

Which device(s) is/are affected (make/model)?  GE/Jasco 14291/14292 
What are the node IDs? 13

- [ ] New device
- [X] Previously working device (node-zwave-js)
    - Which library version/docker image/adapter version? 6.0
   just upgraded to 6.1.1

    - Which platform?
    
**Installation information**
- [X ] `zwavejs2mqtt` (latest) docker image
- [X ] `zwavejs2mqtt` (dev) docker image
- [ ] ioBroker.zwave2 adapter
- [ ] Pkg
- [ ] Manual Docker build
    - `node-zwave-js` branch: 
    - `zwavejs2mqtt` branch: 
- [ ] Other: 

**To Reproduce**
1. Discover switch
2. watch logs

**Additional context**
Discussion in Slack:  
https://zwave2mqtt.slack.com/archives/C01DQ24SDKL/p1612220970387700

**Logfile:** 
[zwavejs_1.log](https://github.com/zwave-js/node-zwave-js/files/5907524/zwavejs_1.log)

blhoward2:
Closes #1555. A small number of early run 14291/14292 do not support double tap. They run firmware version 5.20. I'm also splitting 14294 as a precaution.


blhoward2:
@AlCalzone Don't merge these yet. There is a larger issue at play.

blhoward2:
@AlCalzone Please wait to merge until https://github.com/zwave-js/zwavejs2mqtt/pull/422 is tested. I do think the 14291/14292 change is necessary based on forum posts I found, but I think the 14294 change is unnecessary and the result of a now-fixed bug at the zwavejs2mqtt level.

AlCalzone:
Aight, let me know if this should be merged - or closed

blhoward2:
@AlCalzone Please merge. I've removed the 14294 changes as I believe that device does work. There is anecdotal evidence that older 14291 switches probably don't support it.

AlCalzone (approved):

cmamatey:
Hey @AlCalzone @blhoward2 I'm just setting up zwave-js now and found that double tap is not working on my one GE 14291 switch, but it worked previously on other z-wave setups.
I was able to get double tap to work with my GE 14294 dimmer on zwave-js, so I began looking for what's different between the two. I noticed zwave-js debugging for double taps was different between the two and that they identified groups differently, so I came to github to figure out why and found this change...
My 14291 reports firmware version 5.20 but has supported double tap with other platforms, so I'm wondering what could be done to re-enable double tap for it on zwave-js. Commenting here first to ask before opening a new issue for my problem.

blhoward2:
I suspected this might come up. @AlCalzone that leaves two options. Either we revert the commit and include it for all 14291 switches and handle it better if the switch doesn't actually support it, or we leave it as is and some people can't use it that should be able to do so. Given the bug fixed at the zwavejs2mqtt level I'm not sure whether they'd even get an error message anymore. How does your forced discovery process work that you implemented so the sensors showed up right away?

AlCalzone:
What happens if we revert? Do users just have a useless value?

blhoward2:
I suspect that now they'd get a sensor that does nothing. Before they were getting a discovery error from zwavejs2mqtt but that was a different bug. I assume the switches still have the Basic Set command class, so you're forcing the endpoint creation (without a real interview that would throw an error?) and then zwavejs2mqtt is going to see that and create the sensor. But a double tap would just send two on events via the basic command class (and act upon that through the regular command class) instead of a double tap decoupled from the actual relay.

Given the small number of likely affected devices I would err to the side of a useless sensor. 

Keep in mind there is a much bigger bug here if any are dimmers if we don't revert. If the device supports double tap and it isn't enabled and they double tap the state will become decoupled in HA. HA will think it is at 255 due to the mapping even when it isn't. Switches could become decoupled on/off but that is less likely.

AlCalzone:
Okay, let's revert then. The entire PR?

blhoward2:
That's fine. The Honeywell one at the top was missing double tap initially I think but I can clean that up later. I need to check the others anyway.

PR #1561, opened 2021-02-02
fix(cc): verify that targetValue is a valid currentValue

### LINKED ISSUE #1558: Multilevel switch currentValue reports level 255 when setting targetLevel to last (255)
**Describe the bug**

If you turn on a light using the last level (targetValue 255) zwave-js immediately returns the currentValue as 255, which is a "deprecated" value. I assume this is some new behavior from #1522.

Should the application handle the value 255, or should zwave-js not send it? Unfortunately, if nothing is sent, the problem of flickering UIs returns.

As of now, HA only expects values from 1-99 so the 255 results in HA thinking the light is as bright as the sun (or close to it!). 

**Device information**

Which device(s) is/are affected (make/model)? All dimmers (state exists until a new value is reported, that time various by device). If an application prefers to report targetValue for V4+ switches, it will not be observable.
What are the node IDs? 2

- [ ] New device
- [X] Previously working device (node-zwave-js)
    - Which library version/docker image/adapter version? v6.1.0
    - Have you made any recent configuration changes to the device? Describe. No

    - Which platform?
    
**Installation information**
- [ ] ioBroker.zwave2 adapter
- [ ] Pkg
- [X] Manual Docker build
    - `node-zwave-js` branch:  b1b4f74ef95da72f87e68b474e513867d7e33c68
    - `zwavejs2mqtt` branch:  v1.0.1
- [ ] Other: 

**To Reproduce**
1. Set the multilevel switch targetValue to 255 (turn on to last level)
2. Watch the currentValue go to 255
3. Wait a few seconds for a refresh
4. Watch the currentValue show the actual value

**Additional context**

**Logfile:** [zwavejs_1.log](https://github.com/zwave-js/node-zwave-js/files/5908984/zwavejs_1.log)

AlCalzone:
fixes: #1558

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/BasicCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/BasicCC.ts
@@ -93,11 +93,14 @@ export class BasicCCAPI extends CCAPI {
 		// If the command did not fail, assume that it succeeded and update the currentValue accordingly
 		// so UIs have immediate feedback
 		if (this.isSinglecast()) {
-			const valueDB = this.endpoint.getNodeUnsafe()?.valueDB;
-			valueDB?.setValue(
-				getCurrentValueValueId(this.endpoint.index),
-				value,
-			);
+			// Only update currentValue for valid target values
+			if (value >= 0 && value <= 99) {
+				const valueDB = this.endpoint.getNodeUnsafe()?.valueDB;
+				valueDB?.setValue(
+					getCurrentValueValueId(this.endpoint.index),
+					value,
+				);
+			}
 
 			// and verify the current value after a delay
 			this.schedulePoll({ property });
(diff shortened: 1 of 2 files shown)


jcam:
Tested with HA 2021.2.0b5 and zwavejs2mqtt 1.0.3 in server mode.

This is working well, the brightness indicated does not jump up to max and back.
For the V4 dimmers, the UI toggle glitch does not return, it is visible with the V3 and V2 I have, though somehow seems less pronounced... perhaps the default refresh window in my build is less than the 5s i had before.

AlCalzone:
That confirms what I mentioned in the issue. Essentially HA needs to be a bit more optimistic so the flickering is avoided.

blhoward2:
So does this re-break the toggle flickering for everyone? Because if it does...the icon momentarily going to full bright is less of a ui annoyance than the switch during back on for five seconds...

AlCalzone:
It shouldn't

jcam:
On mine, on older dimmers, it changes it so that the switch flips back on for a half a second or less.

blhoward2:
Always or only when setting 255?

AlCalzone:
I checked with my dimmers - they are pretty quick though, but it should work as intended:
targetValue: 0...99 updates the currentValue immediately
targetValue: 255 does not update currentValue immediately, but the following report from the device does

jcam:
> Always or only when setting 255?

only when setting 255.
and either way it is significantly faster than it was before.

PR #1569, opened 2021-02-03
fix(cc): assume that node is secure when it sends or requests a nonce

### LINKED ISSUE #1051: If a node sends or responds to NonceGet, assume it is secure
Check docs first if that is valid.

AlCalzone:
based on #1565
fixes: #1051


AlCalzone:
@zwave-js-bot rebase

Issue #1571, opened 2021-02-03
Additional Association groups from device config files are not respected if a device claims to support less groups


AlCalzone:
An example is Aeotec ZW100. It has configuration for groups 2 and 3, but the Association CC only reports 1 group.
Not sure what should have priority here and if the device would even respond.

AlCalzone:
If something doesn't show up, it probably shouldn't be in the config files anyways.

PR #1572, opened 2021-02-03
fix(config): define additional Association groups for ZW100

AlCalzone:
Depends on the outcome of #1571 whether this should be merged or not.

The manual and the device only report a single group, but the configuration allows setting the reports for groups 2 and 3 🤷🏻‍♂️


AlCalzone:
So this is not correct. Groups don't refer to associations but to "groups" of reports.

PR #1591, opened 2021-02-04
feat(config): add additional product type and id number for Shenzen Neo PD03Z

wauswaus:
I own a Shenzen Neo PD03Z motion sensor whose product type and id where not supported yet however the sensor itself is.
Therefore, I added the product type and id number to the already existing config file.


AlCalzone:
@zwave-js-bot rename commit

AlCalzone (approved):

AlCalzone:
Thanks!

PR #1607, opened 2021-02-05
feat(config): add alternative device id for Heatit Z-Smoke 230V

roflmao:
SSIA

### DIFF
--- a/packages/config/config/devices/0x019b/4512689.json
+++ b/packages/config/config/devices/0x019b/4512689.json
@@ -13,8 +13,7 @@
 		},
 		{
 			"productType": "0x0003",
-			"productId": "0x000d",
-			"zwaveAllianceId": 3933
+			"productId": "0x000d"
 		}
 	],
 	"firmwareVersion": {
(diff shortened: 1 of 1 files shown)


roflmao:
@zwave-js-bot fix lint

packages/config/config/devices/0x019b/4512689.json
```diff
@@ -10,7 +10,12 @@
 			"productType": "0x0003",
 			"productId": "0x3601",
 			"zwaveAllianceId": 3933
-		}
+		},
+    {
+      "productType": "0x0003",
+      "productId": "0x000d",
+      "zwaveAllianceId": 3933
```
AlCalzone: @blhoward2 should we remove that line if the device ID does not match the one on the alliance webite?

AlCalzone:
@zwave-js-bot fix lint for @roflmao 🤣

AlCalzone:
Will wait for @blhoward2's response

blhoward2:
Please remove the zwa id from the file where the device id isn't included in the database. There is no benefit to having it included twice in the same device file and I like to keep track of which are and which aren't in the database.

AlCalzone (approved):

Issue #1600, opened 2021-02-04
📢 Announcement: About supporting new device configuration files


AlCalzone:
# This is outdated. Accepting PRs for all devices for now, as long as they conform to the [style guide](https://zwave-js.github.io/node-zwave-js/#/config-files/contributing-files)

We are in the process of importing all remaining devices from the Z-Wave Alliance Database (https://products.z-wavealliance.org/products/). If your device is listed below, we're already working on it and should have a file within the next few weeks. In that case, please **don't** open a new request. As you can see, it is an extensive list and having to track and close duplicate device requests will slow our progress.

We will begin pushing these out in groups as we go. We will do our best to prioritize common brands so that more obscure devices come later. 

### If you are willing to help manually convert partial parameters or review batches of simpler files, please message @blhoward2 on [Slack](https://join.slack.com/t/zwave2mqtt/shared_invite/enQtNjc4NjgyNjc3NDI2LTc3OGQzYmJlZDIzZTJhMzUzZWQ3M2Q3NThmMjY5MGY1MTc4NjFiOWZhZWE5YjNmNGE0OWRjZjJiMjliZGQyYmU)
We auto-generate 95% of these files so most of the work is in error checking and cleaning up descriptions.

#### The following devices will be either updated or imported (yes, we're that busy!):

A.KEEMPLECOM, Ltd. - KP-SW-07 - Smart Touch Light Switch (Single)
A.KEEMPLECOM, Ltd. - KP-SW-08 - Smart Touch Light Switch (2-way)
A.KEEMPLECOM, Ltd. - KP-SW-09 - Smart Touch Light Switch (Double)
ABUS August Bremicker Söhne KG - HOMETEC PRO CFA3010 - Z-Wave Door Lock
ABUS Security-Center GmbH & Co. KG - PLBE10000 - Keypad
ABUS Security-Center GmbH & Co. KG - PLBW10000 - Z-Wave Dual Pet PIR
ABUS Security-Center GmbH & Co. KG - PLBW10100 - PIR Sensor Lux
ABUS Security-Center GmbH & Co. KG - PLHA10000 - Z-Wave 230V relay module
ABUS Security-Center GmbH & Co. KG - PLMK10000 - ABUS Z-Wave Door-/Window Sensor
ABUS Security-Center GmbH & Co. KG - PLMK10100 - Door-/Window Sensor Bypass
ABUS Security-Center GmbH & Co. KG - PLSG10000 - Z-Wave indoor siren
[...]

PR #1648, opened 2021-02-08
feat(config): adds support for Dawon DNS Plugs

nukusinji:
Adds support for Dawon DNS Plugs.
Config was imported from z-wavealliance.
https://products.z-wavealliance.org/products/2661
https://products.z-wavealliance.org/products/2532
https://products.z-wavealliance.org/products/2531
https://products.z-wavealliance.org/products/1607


blhoward2:
You're missing all of the configuration parameters for both. And we already have these files built and ready for review. #1600. Please be patient and we'll get these added shortly from our end.

PR #1660, opened 2021-02-08
fix(config): swap product type and id for ZEN30

aneisch:


AlCalzone:
@blhoward2 any objections?

AlCalzone (approved):
assuming blhoward2 is okay with this

blhoward2:
Does this match ozw? We haven't touched those yet so I don't know how zwa has them, if at all.

aneisch:
> Does this match ozw? We haven't touched those yet so I don't know how zwa has them, if at all.

Yes, matches OZW. https://github.com/OpenZWave/open-zwave/blob/master/config/zooz/zen30.xml#L38

Unfortunately I overlooked this mistake in my initial ZEN30 conversion.

AlCalzone:
What, they have `manufacturerId:productId:productType` in their files? Or was that a mistake in the xml?
If it is that way round, we might have a couple of other files with that error.

aneisch:
> What, they have `manufacturerId:productId:productType` in their files? Or was that a mistake in the xml?
> If it is that way round, we might have a couple of other files with that error.

If you use the line `<MetaDataItem name="OzwInfoPage">http://www.openzwave.com/device-database/027a:a008:a000</MetaDataItem>` it is `manufacturerId:productId:productType`, yes.

blhoward2:
You really shouldn't pull the device info from that line as it is non-operative. Their index lives in manufacturer_specific.xml.

aneisch:
> You really shouldn't pull the device info from that line as it is non-operative. Their index lives in manufacturer_specific.xml.

That's great to know, thank you. I will reference that for future contributions.

blhoward2:
Most of the Zooz devices only have one set of type/id but you may check that you didn't miss any. In the index they'll often have multiple sets of values pointing to one device file, but the device file only has the metadata for one of them.

AlCalzone:
> Their index lives in manufacturer_specific.xml

Yeah, right... our import script respects that.

blhoward2 (approved):

PR #1716, opened 2021-02-11
feat(config): add compat option manualValueRefreshDelayMs and use it for Leviton DZMX1

### LINKED ISSUE #1695: Leviton DZMX1 dimmer needs a delay before automagic refresh after being manually tapped
**Describe the bug**

When the physical button on the Levition DZMX1 dimmer switch is tapped, the device emits an ApplicationUpdateRequest.  zwave-js correctly interprets this as a state-change notification (logging "`Node does not send unsolicited updates, refreshing actuator and sensor values...`") and immediately performs a `refreshValues()` on the node, sending a `MultilevelSwitchCCGet` command.  But, the device immediately responds with an "Application Status" command indicating "busy, try again later", which is dropped as unimplemented by zwave-js.  Ten seconds later, the outstanding Get command times out.

The consequences of all this is:
 * The state of the device is not automatically updated after a physical button tap.
 * Attempts to communicate with the device after a physical button tap (e.g., to manually poll the value, or change the value via zwave) will be delayed for 10 seconds, waiting for the automatic Get command to timeout.

I figure the "busy, try again later" response from the device was a technique to avoid violating the old Lutron patent.  A `refreshValues()` on the node __does__ succeed if it is delayed at least 950ms after receiving the ApplicationUpdateRequest.  So, it looks like Leviton implemented "Well, let's just require a 1 second delay, so that no one can possibly call this an 'instant update'."

To handle this, I propose inserting a device-specific delay before `refreshValues()` is called in this block https://github.com/zwave-js/node-zwave-js/blob/6cc6466e2ce050b95364b1740cc6d123af3cdba1/packages/zwave-js/src/lib/node/Node.ts#L1505-L1512  The delay is trivial, but what about the device-specific aspect?  Would it be reasonable to just add a new parameter to the schema for device configuration files?

(Another possibility would be to implement handling the "Application Status, busy, try again later" message, and automatically retrying the request.  But... even that would require knowing how long to wait before the retry, hence a device-specific parameter.)

**Device information**

Which device(s) is/are affected (make/model)?  Leviton DZMX1
What are the node IDs?  Node 2

- [x] New device
- [ ] Previously working device (node-zwave-js)
    
**Installation information**
- [ ] ioBroker.zwave2 adapter
- [x] Pkg (`npm install @zwave-js/server`)
- [ ] Manual Docker build
    - `node-zwave-js` branch: 
    - `zwavejs2mqtt` branch: 
- [x] Manually built (to experiment with delays before `refreshValues()`) 
- [ ] Other: 

**To Reproduce**
1. Have DZMX1 in network.
2. Tap its button.
3. Within the next 10 seconds, try to do something with it over RF.
4. Look at logs.

**Additional context**
_none_

**Logfile:**

This logfile is from stock zwave-js (without any delays hacked in).  The timeline is:
 * Light is off, zwave-js-server started up.
 * Human taps physical button.
 * Device calls for help, zwave-js responds, device says "Naw, I am busy", zwave-js ignores that.
 * 3-4 seconds later, human runs script asking zwave-js to poll the current level value and then set the value to zero.
 * 10 seconds after the button tap, when zwave-js's original automatic Get request times out, the requested poll/set finally occur.

[zwave-175924.log](https://github.com/zwave-js/node-zwave-js/files/5960034/zwave-175924.log)

mdoggydog:
Fixes #1695.

### DIFF
--- a/docs/development/config-files.md
+++ b/docs/development/config-files.md
@@ -233,7 +233,7 @@ By default, `Basic CC::Set` commands are interpreted as status updates. This fla
 
 ### `manualValueRefreshDelayMs`
 
-Some legacy devices emit an NIF when a local event occurs (e.g., button press) to signal that the controller should request a status update.  Some of these devices, however, require a delay before receiving the update request (or they ignore it or respond with a "busy" message).  `manualValueRefreshDelayMs` specifies that delay, expressed in milliseconds.  If unset, the default delay is 0ms.
+Some legacy devices emit an NIF when a local event occurs (e.g. a button press) to signal that the controller should request a status update. However, some of these devices require a delay before they are ready to respond to this request. `manualValueRefreshDelayMs` specifies that delay, expressed in milliseconds. If unset, there will be no delay.
 
 ## Contributing configuration files
 
--- a/docs/development/config-files.md
+++ b/docs/development/config-files.md
@@ -231,6 +231,10 @@ By default, `Basic CC::Set` commands are interpreted as status updates. This fla
 > [!NOTE]
 > If this option is `true`, it has precedence over `disableBasicMapping`.
 
+### `manualValueRefreshDelayMs`
+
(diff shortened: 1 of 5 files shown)


AlCalzone:
Thanks - just had some nits I fixed myself. Will merge when the checks are green

mdoggydog:
Great --- I'm glad I got to contribute.  

(And now, if anyone asks me "Have you ever written any Typescript?", I can say, "Yes, I have.")

PR #1719, opened 2021-02-11
fix(config): add LED always on to GE 46201

crxporter:
These switches respond to a value of 3 to keep the LED always on.

### DIFF
--- a/packages/config/config/devices/0x0063/46201.json
+++ b/packages/config/config/devices/0x0063/46201.json
@@ -56,6 +56,10 @@
 				{
 					"label": "LED always OFF",
 					"value": 2
+				},
+				{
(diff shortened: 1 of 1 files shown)


blhoward2:
Have you tested on the switch itself? Some don't include this in the manual but I've suspected it may be there.

crxporter:
This one yes, it accepts 3 and keeps LED on.

blhoward2 (dismissed):

blhoward2:
Ok, thanks.

AlCalzone (dismissed):

AlCalzone:
@zwave-js-bot rename commit

AlCalzone (approved):

PR #1720, opened 2021-02-11
Adding LED always on to GE 14288

crxporter:
Found another one


blhoward2:
As with the other one, have you tested this on the switch itself? We omitted the option on these to match the manual. There are a number that lack the option but we need testing to add it contrary to the manual.

crxporter:
You're right about this one. It does not respond to 3.

PR #1741, opened 2021-02-12
feat(cc): implement Thermostat Fan Mode CC

Ikcelaks:
I've created a Draft implementation of the Thermostat Fan Mode CC and an incomplete test suite.

The implementation should be complete, but I haven't yet tested it on actual hardware.  I'm submitting this PR as a draft to get early feedback on whether you're interested in me developing it further.

(I also intend to add the Thermostat Fan State CC and expand the tests)

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
@@ -67,6 +67,22 @@ export enum ThermostatFanMode {
 	"External circulation" = 0x0b,
 }
 
+export function getOffStateValueID(endpoint: number): ValueID {
+	return {
+		commandClass: CommandClasses["Thermostat Fan Mode"],
+		endpoint,
+		property: "off",
+	};
+}
+
+export function getModeStateValueID(endpoint: number): ValueID {
+	return {
+		commandClass: CommandClasses["Thermostat Fan Mode"],
+		endpoint,
+		property: "mode",
+	};
+}
+
 @API(CommandClasses["Thermostat Fan Mode"])
 export class ThermostatFanModeCCAPI extends CCAPI {
 	public supportsCommand(cmd: ThermostatFanModeCommand): Maybe<boolean> {
@@ -84,17 +100,47 @@ export class ThermostatFanModeCCAPI extends CCAPI {
 		{ property },
 		value,
 	): Promise<void> => {
-		if (property !== "mode") {
+		const valueDB = this.endpoint.getNodeUnsafe()!.valueDB;
+		if (property === "mode") {
+			if (typeof value !== "number") {
+				throwWrongValueType(
+					this.ccId,
+					property,
+					"number",
+					typeof value,
+				);
+			}
+			const off = valueDB.getValue<boolean>(
+				getOffStateValueID(this.endpoint.index),
+			);
+			await this.set(value, off);
+			if (this.isSinglecast()) {
+				// Verify the current value after a delay
+				this.schedulePoll({ property });
+			}
+		} else if (property === "off") {
+			if (typeof value !== "boolean") {
+				throwWrongValueType(
+					this.ccId,
+					property,
+					"boolean",
+					typeof value,
+				);
+			}
+			const mode = valueDB.getValue<ThermostatFanMode>(
+				getModeStateValueID(this.endpoint.index),
+			);
+			// if (!value && mode == undefined) {
+			// 	// This should NOT be supported, but I don't know which
+			// 	// throw function to call.
+			// }
+			await this.set(mode ?? ThermostatFanMode["Auto low"], value);
+			if (this.isSinglecast()) {
+				// Verify the current value after a delay
+				this.schedulePoll({ property });
+			}
+		} else {
 			throwUnsupportedProperty(this.ccId, property);
-		}
-		if (typeof value !== "number") {
-			throwWrongValueType(this.ccId, property, "number", typeof value);
-		}
-		await this.set(value);
-
-		if (this.isSinglecast()) {
-			// Verify the current value after a delay
-			this.schedulePoll({ property });
 		}
 	};
 
@@ -103,6 +149,7 @@ export class ThermostatFanModeCCAPI extends CCAPI {
 	}): Promise<unknown> => {
 		switch (property) {
 			case "mode":
+			case "off":
 				return (await this.get())?.[property];
 
 			default:
@@ -130,7 +177,7 @@ export class ThermostatFanModeCCAPI extends CCAPI {
 		}
 	}
 
-	public async set(mode: ThermostatFanMode): Promise<void> {
+	public async set(mode: ThermostatFanMode, off?: boolean): Promise<void> {
 		this.assertSupportsCommand(
 			ThermostatFanModeCommand,
 			ThermostatFanModeCommand.Set,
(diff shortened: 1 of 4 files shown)


packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
```diff
@@ -0,0 +1,390 @@
...
+
+		this.persistValues();
+	}
+
+	private _mode: ThermostatFanMode;
+	@ccValue()
+	@ccValueMetadata({
+		...ValueMetadata.UInt8,
+		min: 0,
+		max: 15,
```
AlCalzone: This should be replaced with `states: enumValuesToMetadataStates(ThermostatFanMode)`, by default don't limit it to any modes. You're doing that in the `SupportedReport`.

packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
```diff
@@ -0,0 +1,390 @@
...
+		label: "Thermostat fan mode",
+	})
+	public get mode(): ThermostatFanMode {
+		return this._mode;
+	}
+
+	private _off: boolean | undefined;
+	@ccValue({ minVersion: 2 })
+	public get off(): boolean | undefined {
```
AlCalzone, line 375: missing metadata

packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
```diff
@@ -0,0 +1,390 @@
...
+			this.mode = options.mode;
+		}
+	}
+
+	public mode: ThermostatFanMode;
+
+	public serialize(): Buffer {
+		this.payload = Buffer.from([
+			// 0 R R R M M M M
+			this.mode & 0b1111,
```
AlCalzone: you're missing the `off` flag

packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
```diff
@@ -0,0 +1,390 @@
...
+			endpoint: this.endpointIndex,
+			property: "mode",
+		};
+		// Only update the dynamic part
+		this.getValueDB().setMetadata(valueId, ({
+			states: enumValuesToMetadataStates(
+				ThermostatFanMode,
+				this._supportedModes,
+			),
+		} as unknown) as ValueMetadata);
```
AlCalzone: It is not ideal, but you can get around the double cast by doing this:
```ts
		this.getValueDB().setMetadata(valueId, {
			...ValueMetadata.UInt8, // <== HERE!
			states: enumValuesToMetadataStates(
				ThermostatFanMode,
				this._supportedModes,
			),
		});
```

packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
```diff
@@ -0,0 +1,390 @@
...
+@CCCommand(ThermostatFanModeCommand.SupportedReport)
+export class ThermostatFanModeCCSupportedReport extends ThermostatFanModeCC {
+	public constructor(
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+		this._supportedModes = parseBitMask(
+			this.payload,
+			ThermostatFanMode["Auto high"],
```
AlCalzone:
```suggestion
			ThermostatFanMode["Auto low"],
```

packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
```diff
@@ -0,0 +1,390 @@
...
+			// Verify the current value after a delay
+			this.schedulePoll({ property });
+		}
+	};
+
+	protected [POLL_VALUE]: PollValueImplementation = async ({
+		property,
+	}): Promise<unknown> => {
+		switch (property) {
+			case "mode":
```
AlCalzone, line 151:
```suggestion
			case "mode":
			case "off":
```

packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
```diff
@@ -0,0 +1,390 @@
...
+		}
+		if (typeof value !== "number") {
+			throwWrongValueType(this.ccId, property, "number", typeof value);
+		}
+		await this.set(value);
+
+		if (this.isSinglecast()) {
+			// Verify the current value after a delay
+			this.schedulePoll({ property });
+		}
```
AlCalzone: This should also allow setting the `"off"` property. In that case, I'd load the current mode from the value DB and use it together with `off`.

packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
```diff
@@ -0,0 +1,390 @@
...
+			message: "querying current thermostat fan mode...",
+			direction: "outbound",
+		});
+		const currentStatus = await api.get();
+		if (currentStatus) {
+			this.driver.controllerLog.logNode(node.id, {
+				endpoint: this.endpointIndex,
+				message:
+					"received current thermostat fan mode: " +
+					getEnumMemberName(ThermostatFanMode, currentStatus.mode),
```
AlCalzone: I'd log `off` too.

AlCalzone (changes requested):
Looks pretty good - a few minor things need to be addressed

packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
Ikcelaks: I just reread the specs on the Set command, and you're of course correct. I do think they really could have written that paragraph better.

packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
Ikcelaks: I essentially copied the metadata info from the ThermostatModeCC.ts file (which has min 0 and max 31), so I thought it was a guard on the size of the bitmask.  Might want to look at that CC too.

packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
Ikcelaks: I also copied this bit from ThermostatModeCC.ts. This looks much saner.

packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
AlCalzone: Sounds familiar 😅

Ikcelaks:
I'm very unsure that this is handling the set and report correctly with respect to the Off bit.  In particular, I think that setting the Off bit to 0 without specifying the Mode is nonsensical.

My device doesn't have an "off" mode, so I will not be able to test if this is correct with any hardware.

packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts
AlCalzone: Good point!

AlCalzone:
>  In particular, I think that setting the Off bit to 0 without specifying the Mode is nonsensical.

I'm not sure about this. Doesn't it make sense to turn the device back on without changing the mode?

AlCalzone (dismissed):

AlCalzone (approved):

PR #1743, opened 2021-02-13
feat(cc): implement Thermostat Fan State CC

Ikcelaks:
This is a draft implementation of the Thermostat Fan State CC and a basic set of tests.  Not yet tested on real hardware.

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/ThermostatFanStateCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/ThermostatFanStateCC.ts
@@ -1,0 +1,174 @@
+import {
+	CommandClasses,
+	Maybe,
+	MessageOrCCLogEntry,
+	MessageRecord,
+	validatePayload,
+	ValueMetadata,
+} from "@zwave-js/core";
+import { getEnumMemberName, pick } from "@zwave-js/shared";
+import type { Driver } from "../driver/Driver";
+import { MessagePriority } from "../message/Constants";
+import {
+	CCAPI,
+	PollValueImplementation,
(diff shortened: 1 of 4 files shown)


packages/zwave-js/src/lib/commandclass/ThermostatFanStateCC.ts
```diff
@@ -0,0 +1,174 @@
...
+
+		this.persistValues();
+	}
+
+	private _state: ThermostatFanState;
+	@ccValue()
+	@ccValueMetadata({
+		...ValueMetadata.UInt8,
+		min: 0,
+		max: 15,
```
AlCalzone: This should be replaced with `states`.

packages/zwave-js/src/lib/commandclass/ThermostatFanStateCC.ts
```diff
@@ -0,0 +1,174 @@
...
+			nodeId: this.endpoint.nodeId,
+			endpoint: this.endpoint.index,
+		});
+		const response = await this.driver.sendCommand<ThermostatFanStateCCReport>(
+			cc,
+			this.commandOptions,
+		);
+		if (response) {
+			return pick(response, ["state"]);
+		}
```
AlCalzone: I think I'd replace this with `return response?.state`.
If the standard ever adds a new property, we can handle it by returning an object for higher versions.

AlCalzone (changes requested):

AlCalzone (approved):

AlCalzone:
Thank you! We should see if this works soon :)

PR #1755, opened 2021-02-14
docs: explain the basics

### LINKED ISSUE #1313: docs: Explain the basics like yarn install etc.
I’ll try that later, thanks. The documentation does a good job explaining some of these things but I feel like in some areas (like here) it assumes a level of knowledge with JavaScript/typescript that not everyone will have as the project grows. Just a thought. I guess ultimately it’s a question of how accessible you want it to be for people to add devices and such.

_Originally posted by @blhoward2 in https://github.com/zwave-js/node-zwave-js/issues/1312#issuecomment-756214327_

AlCalzone:
fixes: #1313

@blhoward2 since you originally raised the issue as a comment, please check if you think this is better now.


blhoward2:
I think this is much better. Looks good to me.

blhoward2 (approved):

PR #1758, opened 2021-02-14
fix(cc): convert CommandClass event params to JSON

AlCalzone:
fixes: https://github.com/AlCalzone/ioBroker.zwave2/issues/535


robertsLando:
@AlCalzone Could you explain this better? Where is this used?

AlCalzone:
In the `"notification"` event. Previously, the eventParameter could be a CC instance (besides Buffer and Record<string, number>), now that CC is converted to a Record, so you don't have to handle it yourself.

Issue #1321, opened 2021-01-08
allow setting of duration/transition


marcelveldt:
**Is your feature request related to a problem? Please describe.**
Looks like setting the light transition duration is not implemented yet.

**Describe the solution you'd like**
Be able to set transition duration value. 

**Explaining your use-case will go a long way in helping us understand.**
Home Assistant integration

<img width="318" alt="Schermafbeelding 2021-01-08 om 23 17 02" src="https://user-images.githubusercontent.com/6389780/104070615-9ef46e80-5207-11eb-9196-b25e9a241921.png">

<img width="952" alt="Schermafbeelding 2021-01-08 om 23 17 49" src="https://user-images.githubusercontent.com/6389780/104070665-baf81000-5207-11eb-9cf7-cff7ba517b2e.png">

Thanks!

AlCalzone:
What is the expected behavior here?

1. Let's say, a user sets a duration, then changes the target value.
2. The device later gets triggered physically and reports a different duration.
3. The user changes the target value again.

Which duration (if any) gets used for the set command?

robertsLando:
> The device later gets triggered physically and reports a different duration

Why should this happen? Does a trigger also report a DUration?

AlCalzone:
Yes, some newer devices report current and target value as well as the remaining duration.

robertsLando:
I think that the device should always use the user selected duration

AlCalzone:
Currently, there's only one `duration` value ID, so that would get overwritten when we receive a report.
And looking at the confusion around current/target value I'm really hesitant to add another of these value pairs.

marcelveldt:
I think it would be nice to be able to retrieve the current configured transitionDuration and must-have to set a transitionDuration. maybe adjust it the same so you'll have currentDuration and targetDuration ?

AlCalzone:
I'm wondering if it is desirable **not** to have an extra value to set the transition duration and instead pass it somehow along with the set command?
I'm not very familiar with the MQTT API to determine if that is possible and makes sense.

robertsLando:
> I'm wondering if it is desirable not to have an extra value to set the transition duration and instead pass it somehow along with the set command?

This could be good but also setting a default value should be possible

AlCalzone:
@marcelveldt we've been thinking - what's your take on adding another argument to `setValue` that specifies options for the set call? E.g.
```ts
[...]

PR #1772, opened 2021-02-15
feat(cc): implement Scene Controller Configuration CC

Ikcelaks:
Here's a draft for the Scene Controller Configuration CC.

I copied what UserCodeCC did for persisting values and setting metadata, but I'm not certain that they're completely alike.  Scene Controller Configuration CC is a little wonky, because it encodes a disabled group / scene assignment with `sceneId = 0`.  I believe this means that `dimming duration` is effectively cleared by setting `sceneId = 0`, but the documentation is 100% clear to me.

I haven't implemented SetValue for `dimmingDuration`, because I don't know how to handle the Duration type here.  If it can be done, this seems like a valuable feature that users would expect.

The interview potentially makes an API call with the Association CC, because that's how to determine the number of supported groups.  I believe that the AssociationCC interview should always happen before the SceneControllerConfigurationCC interview, but I did allow the interview function to do  the actual query if the cached `groupCount` is missing.

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
@@ -300,11 +300,6 @@ export class SceneControllerConfigurationCC extends CommandClass {
 			return;
 		}
 
-		this.driver.controllerLog.logNode(node.id, {
-			endpoint: this.endpointIndex,
-			message: `supports ${groupCount} association groups`,
-			direction: "none",
-		});
 
 		// Always query scene configuration for each association group
 		for (let groupId = 1; groupId <= groupCount; groupId++) {
--- a/packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
@@ -23,7 +23,7 @@ import {
 	throwUnsupportedPropertyKey,
 	throwWrongValueType,
 } from "./API";
-import { getGroupCountValueId } from "./AssociationCC";
+import type { AssociationCC } from "./AssociationCC";
 import {
 	API,
 	CCCommand,
@@ -82,9 +82,7 @@ function persistSceneConfig(
 
 	if (!valueDB.hasMetadata(sceneIdValueId)) {
 		valueDB.setMetadata(sceneIdValueId, {
-			...ValueMetadata.Number,
-			min: 0,
-			max: 255,
+			...ValueMetadata.UInt8,
 			label: `Associated Scene ID (${groupId})`,
 		});
 	}
@@ -308,9 +306,8 @@ export class SceneControllerConfigurationCC extends CommandClass {
 			direction: "none",
 		});
 
-		// Always query scene and dimmer duration for each association group
-		// skipping group #1, which is reserved for Lifeline
-		for (let groupId = 2; groupId <= groupCount; groupId++) {
+		// Always query scene configuration for each association group
+		for (let groupId = 1; groupId <= groupCount; groupId++) {
 			this.driver.controllerLog.logNode(node.id, {
 				endpoint: this.endpointIndex,
 				message: `querying scene configuration for association group #${groupId}...`,
@@ -338,8 +335,14 @@ dimming duration: ${group.dimmingDuration.toString()}`;
 	 * This only works AFTER the node has been interviewed by this CC
 	 * or the AssociationCC.
 	 */
-	public getGroupCountCached(): number {
-		return this.getValueDB().getValue(getGroupCountValueId()) ?? 0;
+	protected getGroupCountCached(): number {
+		return (
+			this.getEndpoint()
+				?.createCCInstanceUnsafe<AssociationCC>(
+					CommandClasses.Association,
+				)
+				?.getGroupCountCached() ?? 0
+		);
 	}
 }
 
@@ -370,9 +373,12 @@ export class SceneControllerConfigurationCCSet extends SceneControllerConfigurat
 			this.sceneId = options.sceneId;
 			this.dimmingDuration = options.dimmingDuration;
 
-			if (this.groupId < 2 || this.groupId > groupCount) {
+			// The client SHOULD NOT specify group 1 (the life-line group).
+			// We don't block it here, because the specs don't forbid it,
+			// and it may be needed for some devices.
+			if (this.groupId < 1 || this.groupId > groupCount) {
 				throw new ZWaveError(
-					`${this.constructor.name}: The user ID must be between 2 and the number of supported groups ${groupCount}.`,
+					`${this.constructor.name}: The group ID must be between 1 and the number of supported groups ${groupCount}.`,
 					ZWaveErrorCodes.Argument_Invalid,
 				);
 			}
@@ -448,12 +454,24 @@ export class SceneControllerConfigurationCCReport extends SceneControllerConfigu
 	}
 }
 
+function testResponseForSceneControllerConfigurationGet(
+	sent: SceneControllerConfigurationCCGet,
+	received: SceneControllerConfigurationCCReport,
+) {
+	// We expect a Scene Controller Configuration Report that matches
+	// the requested groupId, unless groupId 0 was requested
+	return sent.groupId === 0 || received.groupId === sent.groupId;
+}
+
 interface SceneControllerConfigurationCCGetOptions extends CCCommandOptions {
 	groupId: number;
 }
 
 @CCCommand(SceneControllerConfigurationCommand.Get)
-@expectedCCResponse(SceneControllerConfigurationCCReport)
+@expectedCCResponse(
+	SceneControllerConfigurationCCReport,
+	testResponseForSceneControllerConfigurationGet,
+)
 export class SceneControllerConfigurationCCGet extends SceneControllerConfigurationCC {
 	public constructor(
 		driver: Driver,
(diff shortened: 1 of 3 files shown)


AlCalzone:
I'll do an in-depth review tomorrow. Some thoughts regarding your comments:

> I haven't implemented SetValue for dimmingDuration, because I don't know how to handle the Duration type here. If it can be done, this seems like a valuable feature that users would expect.

That likely depends on #1321, so don't worry about it for now.

> The interview potentially makes an API call with the Association CC, because that's how to determine the number of supported groups. I believe that the AssociationCC interview should always happen before the SceneControllerConfigurationCC interview, but I did allow the interview function to do the actual query if the cached groupCount is missing.

I'm not a fan of that. CC implementations shouldn't repeat the work that others. I think the best approach here is to fail softly, i.e. assume a group count of 0 and do nothing. In that case, you could skip setting the `interviewCompleted` property to true. Some CCs do that if a required query times out (e.g. support information).

You should also force that `Association CC` is interviewed first by implementing `determineRequiredCCInterviews` (see `Association CC` for an example).
Also: Do we need to handle `Multi Channel Assocation CC` alternatively? Is `SceneControllerConfigurationCC` supported by endpoints or only the node itself?

Ikcelaks:
The Z-Wave Application Command Class Specification PDF specifically identifies the Association CC v1 Supported Groupings Get command, so I don't think we need the Multi Channel Association CC here.

How do I determine if a CC can be supported by endpoints vs the node itself?

AlCalzone:
> How do I determine if a CC can be supported by endpoints vs the node itself?

That should be somewhere in the specs. I think the Command Class Control specification.

Ikcelaks:
I add `AssociationCC` to `determineRequiredCCInterviews()`.  I had a feeling there was a more elegant solution, and this is definitely more elegant.

I found the list of CCs that must only be supported and advertised by the Root Device (3.2.1 of SDS12783).  `Scene Controller Configuration CC` is NOT in that list.  However, I do not believe we need to worry about `Multi Channel Association CC`, because we are only using the `Association CC` to determine the group count (which should be the same, since they're all shared).  We do not actually care about the groups themselves.

Finally, it occurs to me that I do NOT really understand the distinction between the `CCAPI` and `PhysicalCCAPI`.  I can see in the source code that `PhysicalCCAPI` is simply a `CCAPI` that asserts it is not a `VirtualEndpoint`, but I don't really understand what effect that has.  In this CC, I used `PhysicalCCAPI`.  In the Thermostat Fan CCs, I used `CCAPI`.  In each case I was simply matching the CC I was using as a template.

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -0,0 +1,490 @@
...
+				`${this.constructor.name}: deserialization not implemented`,
+				ZWaveErrorCodes.Deserialization_NotImplemented,
+			);
+		} else {
+			const groupCount = this.getGroupCountCached();
+			this.groupId = options.groupId;
+			this.sceneId = options.sceneId;
+			this.dimmingDuration = options.dimmingDuration;
+
+			if (this.groupId < 2 || this.groupId > groupCount) {
```
AlCalzone: From the specs:
> A sending node SHOULD NOT specify Group ID 1

I think since it is not forbidden, we should allow it, because one might want to configure the command that is sent to the lifeline

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -0,0 +1,490 @@
...
+		};
+	}
+}
+
+interface SceneControllerConfigurationCCGetOptions extends CCCommandOptions {
+	groupId: number;
+}
+
+@CCCommand(SceneControllerConfigurationCommand.Get)
+@expectedCCResponse(SceneControllerConfigurationCCReport)
```
AlCalzone: This should have a predicate whether the correct report is received, see `testResponseForAlarmSensorGet` as an example. You can special-case the id 0 there to allow any report.

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -0,0 +1,490 @@
...
+		this.endpointIndex,
+		groupId,
+	);
+	const valueDB = this.getValueDB();
+
+	if (!valueDB.hasMetadata(sceneIdValueId)) {
+		valueDB.setMetadata(sceneIdValueId, {
+			...ValueMetadata.Number,
+			min: 0,
+			max: 255,
```
AlCalzone: Can be replaced with `...ValueMetadata.UInt8`

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -0,0 +1,490 @@
...
+		});
+
+		this.driver.controllerLog.logNode(node.id, {
+			message: `${this.constructor.name}: doing a ${
+				complete ? "complete" : "partial"
+			} interview...`,
+			direction: "none",
+		});
+
+		const groupCount = this.getGroupCountCached();
```
AlCalzone, line 293: This should just use the `Association CC` instance
```suggestion
		const groupCount = endpoint
			.createCCInstanceUnsafe<AssociationCC>(
				CommandClasses.Association,
			)
			?.getGroupCountCached() ?? 0;
```

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -0,0 +1,490 @@
...
+		if (groupCount === 0) {
+			this.driver.controllerLog.logNode(node.id, {
+				endpoint: this.endpointIndex,
+				message: `soft failing interview (missing association group count)`,
+				direction: "none",
+			});
+
+			this.interviewComplete = false;
+			return;
```
AlCalzone:
```suggestion
			this.driver.controllerLog.logNode(node.id, {
				endpoint: this.endpointIndex,
				message: `skipping Scene Controller Configuration interview because Association group count is unknown`,
				direction: "none",
				level: "warn",
			});
			return;
```

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -0,0 +1,490 @@
...
+		}
+
+		this.driver.controllerLog.logNode(node.id, {
+			endpoint: this.endpointIndex,
+			message: `supports ${groupCount} association groups`,
+			direction: "none",
+		});
```
AlCalzone:
```suggestion
```

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -0,0 +1,490 @@
...
+
+		// Always query scene and dimmer duration for each association group
+		// skipping group #1, which is reserved for Lifeline
+		for (let groupId = 2; groupId <= groupCount; groupId++) {
```
AlCalzone: See other comment regarding group 1, I think it would be good to have this information.

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -0,0 +1,490 @@
...
+			this.driver.controllerLog.logNode(node.id, {
+				endpoint: this.endpointIndex,
+				message: `querying association group #${groupId}...`,
```
AlCalzone:
```suggestion
				message: `querying scene configuration for association group #${groupId}...`,
```

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -0,0 +1,490 @@
...
+				direction: "outbound",
+			});
+			const group = await api.get(groupId);
+			if (group != undefined) {
+				const logMessage = `received information for association group #${groupId}:
+scene Id: ${group.sceneId}
+dimming duration: ${group.dimmingDuration.toString()}`;
```
AlCalzone:
```suggestion
				const logMessage = `received scene configuration for association group #${groupId}:
scene ID:         ${group.sceneId}
dimming duration: ${group.dimmingDuration.toString()}`;
```

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -0,0 +1,490 @@
...
+					"number",
+					typeof value,
+				);
+			}
+
+			if (value === 0) {
+				// Disable Group ID / Scene ID
+				await this.disable(propertyKey);
+			} else {
+				// We need to set the dimming duration along with
```
AlCalzone:
```suggestion
				// We need to set the dimming duration along with the scene ID
```

AlCalzone:
Please re-request a review once you're done. I'm planning to do a release tonight, so maybe it could already go in there.

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
Ikcelaks: Since `getGroupCountCached()` is called three times (`interview`, `SceneControllerConfigurationCCSet`, and `SceneControllerConfigurationCCGet`), I changed `getGroupCountCached` to a `protected` function that forwards the call to the `Association CC` instance.

AlCalzone (dismissed):
I have one more small thing but I'll fix it myself

AlCalzone (approved):

PR #1777, opened 2021-02-16
feat(config): add templates to kwikset locks

blhoward2:
First application of templates, to Kwikset locks


packages/config/config/devices/0x0090/templates/kwikset_template.json
```diff
@@ -0,0 +1,887 @@
...
+			{
+				"label": "Guest",
+				"value": 3
+			},
+			{
+				"label": "Worker",
+				"value": 4
+			}
+		]
+	},
```
AlCalzone: I'd define this once and then import it from the `usercode_x` keys, overwriting the label.

AlCalzone (dismissed):
We can cut down more repetition :)

AlCalzone:
@blhoward2 I've switched the imports and labels, just to avoid any confusion in the future, in case anyone adds a property above the import and that gets overwritten.

AlCalzone (approved):

Issue #1761, opened 2021-02-14
Inovelli LZW31-SN firmware update does not complete due to 'waiting for activation' status


prwood80:
**Describe the bug**

Updating firmware on Inovelli LZW31-SN results in result 0xFD (OK_WaitingForActivation).  The activation never occurs, the node is refreshed and the update does not appear to succeed.

**Device information**

Inovelli LZW31-SN, current firmware version is 1.47.  Updating to v1.52

**Last Known Working Configuration**
- [ ] New device
- [ ] Previously working device (node-zwave-js)
    - Which library version/docker image/adapter version?
    - Have you made any recent configuration changes to the device? Describe.

- [x] Previously working device (other platform)
    - Which platform? Z-Wave PC Controller 5
    - Have you made any recent configuration changes to the device? No
    
**Installation information**
How did you install `node-zwave-js`?
- [ ] `zwavejs2mqtt` (latest) docker image
- [ ] `zwavejs2mqtt` (dev) docker image
- [ ] ioBroker.zwave2 adapter
- [ ] Pkg
- [x] Manual Docker build
    - `node-zwave-js` branch: master
    - `zwavejs2mqtt` branch: master
- [ ] Manually built (as described in the [docs](https://zwave-js.github.io/node-zwave-js/#/development/installing-from-github)) 
- [ ] Other: 

**To Reproduce**
1. Initiate 'Begin Firmware Update' for node on the internal Control Panel web page.
2. Select Target 0 & otz file
3. Wait for update to complete
4. Observe Controller Status reports completion with status code 0xFD
5. Refresh node
6. Observe node firmware is not updated

**Additional context**

**Logfile:**
Could not successfully start a firmware update with the logging enabled.  Will continue to try.  Even after restarting the docker container and waiting for all nodes to complete the interview stage, I received a number of errors (firmware update in progress, node does not support firmware update [though it seemed to be identified and healthy])

AlCalzone:
AFAIK that means you have to manually push a button on the device to continue.
It is basically a safety measure so you can't just update the firmware without physical access to the device.

prwood80:
I'll try that.  I didn't see any documentation from Inovelli in those regards (and the Z-wave PC Controller 5 doesn't require that).

AlCalzone:
> the Z-wave PC Controller 5 doesn't require that

I might be mistaken then.

AlCalzone:
Nevermind, I found it:

[...]

PR #1791, opened 2021-02-17
fix(cc): use alternate firmware ID when updating target > 0

prwood80:
Set firmware id by target

Partially addresses: [(#1761)](https://github.com/zwave-js/node-zwave-js/issues/1761).


AlCalzone:
@zwave-js-bot rebase

AlCalzone (approved):

PR #1799, opened 2021-02-17
fix(docs): avoid search collisions

AlCalzone:
@robertsLando I think this is why search isn't working in your docs. You might want to apply this to your repo too


robertsLando (approved):
Didn't know about this, will add a fix on my side too! Thanks

PR #1805, opened 2021-02-17
feat(cc): implement Scene Actuator Configuration CC

Ikcelaks:
This implements the Scene Actuator Configuration CC and a very basic set of tests.

I have NOT implemented an interview, because I don't think it's appropriate.  Supporting devices are required to support 255 scenes, and there is no method I see for determining which scenes are relevant.  The user will know which scenes they are using and can request the current configuration selectively.  Blindly querying all 255 scenes is excessive, I think.

Setting the `level` is optional in the Set command.  The Zwave interface adds a 1 bit `override` buffer field.  I have chosen to make the `level` field optional and setting the `override` bit automatically on a `options.level != undefined` test.  I'm not sure if that's the preferred interface.

I'm still am not certain about the distinction between PhysicalCCAPI and CCAPI.

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
@@ -106,7 +106,6 @@ export class SceneActuatorConfigurationCCAPI extends CCAPI {
 			case SceneActuatorConfigurationCommand.Get:
 				return this.isSinglecast();
 			case SceneActuatorConfigurationCommand.Set:
-			case SceneActuatorConfigurationCommand.Report:
 				return true;
 		}
 		return super.supportsCommand(cmd);
@@ -193,14 +192,37 @@ export class SceneActuatorConfigurationCCAPI extends CCAPI {
 		await this.driver.sendCommand(cc, this.commandOptions);
 	}
 
+	public async getActive(): Promise<
+		| Pick<
+				SceneActuatorConfigurationCCReport,
+				"sceneId" | "level" | "dimmingDuration"
+		  >
+		| undefined
+	> {
+		this.assertSupportsCommand(
+			SceneActuatorConfigurationCommand,
+			SceneActuatorConfigurationCommand.Get,
+		);
+
+		const cc = new SceneActuatorConfigurationCCGet(this.driver, {
+			nodeId: this.endpoint.nodeId,
+			endpoint: this.endpoint.index,
+			sceneId: 0,
+		});
+		const response = await this.driver.sendCommand<SceneActuatorConfigurationCCReport>(
+			cc,
+			this.commandOptions,
+		);
+
+		if (response) {
+			return pick(response, ["sceneId", "level", "dimmingDuration"]);
+		}
+	}
+
 	public async get(
 		sceneId: number,
 	): Promise<
-		| {
-				sceneId: number;
-				level: number;
-				dimmingDuration: Duration;
-		  }
+		| Pick<SceneActuatorConfigurationCCReport, "level" | "dimmingDuration">
 		| undefined
 	> {
 		this.assertSupportsCommand(
@@ -218,8 +240,13 @@ export class SceneActuatorConfigurationCCAPI extends CCAPI {
 			this.commandOptions,
 		);
 
-		if (response) {
-			return pick(response, ["sceneId", "level", "dimmingDuration"]);
+		// A response to a
+		if (
+			response &&
+			response.level != undefined &&
+			response.dimmingDuration != undefined
+		) {
+			return pick(response, ["level", "dimmingDuration"]);
 		}
 	}
 }
@@ -264,33 +291,20 @@ export class SceneActuatorConfigurationCCSet extends SceneActuatorConfigurationC
 			}
 			this.sceneId = options.sceneId;
 			this.dimmingDuration = options.dimmingDuration;
-			if (options.level != undefined) {
-				if (options.level < 0 || options.level > 255) {
-					throw new ZWaveError(
-						`level ${options.level} out of range [1..255]`,
-						ZWaveErrorCodes.Argument_Invalid,
-					);
-				}
-				this.override = true;
-				this.level = options.level;
-			} else {
-				this.override = false;
-				this.level = 0xff; // will be ignored
-			}
+			this.level = options.level;
 		}
 	}
 
 	public sceneId: number;
 	public dimmingDuration: Duration;
-	public override: boolean;
-	public level: number;
+	public level?: number;
 
 	public serialize(): Buffer {
 		this.payload = Buffer.from([
 			this.sceneId,
 			this.dimmingDuration.serializeSet(),
-			this.override ? 0b1000_0000 : 0x00,
-			this.level,
+			this.level != undefined ? 0b1000_0000 : 0,
+			this.level ?? 0xff,
 		]);
 		return super.serialize();
 	}
@@ -305,25 +319,35 @@ export class SceneActuatorConfigurationCCReport extends SceneActuatorConfigurati
 		super(driver, options);
 		validatePayload(this.payload.length >= 3);
 		this.sceneId = this.payload[0];
-		this.level = this.payload[1];
-		this.dimmingDuration =
-			Duration.parseReport(this.payload[2]) ?? new Duration(0, "unknown");
 
-		this.persistValues();
+		if (this.sceneId != 0) {
+			this.level = this.payload[1];
+			this.dimmingDuration =
+				Duration.parseReport(this.payload[2]) ??
+				new Duration(0, "unknown");
+
+			this.persistValues();
+		}
 	}
 
-	public sceneId: number;
-	public level: number;
-	public dimmingDuration: Duration;
+	public readonly sceneId: number;
+	public readonly level?: number;
+	public readonly dimmingDuration?: Duration;
 
 	public persistValues(): boolean {
-		persistSceneActuatorConfig.call(
+		if (
+			this.sceneId === 0 ||
+			this.level == undefined ||
+			this.dimmingDuration == undefined
+		) {
+			return false;
+		}
+		return persistSceneActuatorConfig.call(
 			this,
 			this.sceneId,
 			this.level,
 			this.dimmingDuration,
 		);
-		return true;
 	}
 
 	public toLogEntry(): MessageOrCCLogEntry {
@@ -332,7 +356,7 @@ export class SceneActuatorConfigurationCCReport extends SceneActuatorConfigurati
 			message: {
 				sceneId: this.sceneId,
 				level: this.level,
-				dimmingDuration: this.dimmingDuration.toString(),
+				dimmingDuration: this.dimmingDuration?.toString(),
 			},
 		};
 	}
@@ -342,7 +366,9 @@ function testResponseForSceneActuatorConfigurationGet(
 	sent: SceneActuatorConfigurationCCGet,
 	received: SceneActuatorConfigurationCCReport,
 ) {
-	return received.sceneId === sent.sceneId;
+	// We expect a Scene Actuator Configuration Report that matches
+	// the requested sceneId, unless groupId 0 was requested
+	return sent.sceneId === 0 || received.sceneId === sent.sceneId;
 }
 
 interface SceneActuatorConfigurationCCGetOptions extends CCCommandOptions {
@@ -369,12 +395,6 @@ export class SceneActuatorConfigurationCCGet extends SceneActuatorConfigurationC
 				ZWaveErrorCodes.Deserialization_NotImplemented,
 			);
 		} else {
-			if (options.sceneId < 1 || options.sceneId > 255) {
-				throw new ZWaveError(
-					`scene id ${options.sceneId} out of range [1..255]`,
-					ZWaveErrorCodes.Argument_Invalid,
-				);
-			}
 			this.sceneId = options.sceneId;
 		}
 	}
--- a/packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
@@ -230,6 +230,14 @@ export class SceneActuatorConfigurationCCAPI extends CCAPI {
 			SceneActuatorConfigurationCommand.Get,
 		);
 
+		if (sceneId === 0) {
+			// We need to prevent this, since it will cause the Report sceneId
+			// to be different from the requested sceneId, and the client won't
+			// be able to determine the actual sceneId, because it's hidden in
+			// the return value
+			return;
+		}
+
 		const cc = new SceneActuatorConfigurationCCGet(this.driver, {
 			nodeId: this.endpoint.nodeId,
 			endpoint: this.endpoint.index,
@@ -240,13 +248,12 @@ export class SceneActuatorConfigurationCCAPI extends CCAPI {
 			this.commandOptions,
 		);
 
-		// A response to a
-		if (
-			response &&
-			response.level != undefined &&
-			response.dimmingDuration != undefined
-		) {
+		if (!response) return;
+		if (response.sceneId === sceneId) {
 			return pick(response, ["level", "dimmingDuration"]);
+		} else {
+			// I don't think this can happen, if we block sceneId = 0
+			return;
 		}
 	}
 }
--- a/packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
@@ -130,7 +130,7 @@ export class SceneActuatorConfigurationCCAPI extends CCAPI {
 				);
 			}
 
-			// We need to set the dimming duration along with the level
+			// We need to set the dimming duration along with the level.
 			// If no duration is set, we default to 0 seconds
 			const node = this.endpoint.getNodeUnsafe()!;
 			const dimmingDuration =
@@ -249,14 +249,12 @@ export class SceneActuatorConfigurationCCAPI extends CCAPI {
 	}
 }
 
+// @noInterview - We don't want to query 255 scenes
+
 @commandClass(CommandClasses["Scene Actuator Configuration"])
 @implementedVersion(1)
 export class SceneActuatorConfigurationCC extends CommandClass {
 	declare ccCommand: SceneActuatorConfigurationCommand;
-
-	// interview is omitted, because our only option would be
-	// to query all 255 possible sceneIds, which is a lot of traffic
-	// for little benefit
 }
 
 interface SceneActuatorConfigurationCCSetOptions extends CCCommandOptions {
@@ -283,7 +281,7 @@ export class SceneActuatorConfigurationCCSet extends SceneActuatorConfigurationC
 		} else {
 			if (options.sceneId < 1 || options.sceneId > 255) {
 				throw new ZWaveError(
-					`scene id ${options.sceneId} out of range [1..255]`,
+					`The scene id ${options.sceneId} must be between 1 and 255!`,
 					ZWaveErrorCodes.Argument_Invalid,
 				);
 			}
@@ -333,6 +331,7 @@ export class SceneActuatorConfigurationCCReport extends SceneActuatorConfigurati
 	public readonly dimmingDuration?: Duration;
 
 	public persistValues(): boolean {
+		// Do not persist values for an inactive scene
 		if (
 			this.sceneId === 0 ||
 			this.level == undefined ||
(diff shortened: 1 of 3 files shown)


AlCalzone:
> I'm still am not certain about the distinction between PhysicalCCAPI and CCAPI.

Should the commands be supported in multicast or broadcast? If not, then PhysicalCCAPI, else CCAPI.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,395 @@
...
+			}
+			this.sceneId = options.sceneId;
+			this.dimmingDuration = options.dimmingDuration;
+			if (options.level != undefined) {
+				if (options.level < 0 || options.level > 255) {
+					throw new ZWaveError(
+						`level ${options.level} out of range [1..255]`,
+						ZWaveErrorCodes.Argument_Invalid,
+					);
+				}
```
AlCalzone: This is unnecessary (and the error message is wrong). If you pass a value into a buffer, it is automatically `&`-ed with `0xff`.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,395 @@
...
+				this.override = true;
+				this.level = options.level;
+			} else {
+				this.override = false;
+				this.level = 0xff; // will be ignored
+			}
```
AlCalzone: I'd get rid of the override variable and store level as `?: number`

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,395 @@
...
+	public dimmingDuration: Duration;
+	public override: boolean;
+	public level: number;
+
+	public serialize(): Buffer {
+		this.payload = Buffer.from([
+			this.sceneId,
+			this.dimmingDuration.serializeSet(),
+			this.override ? 0b1000_0000 : 0x00,
+			this.level,
```
AlCalzone:
```suggestion
			this.level != undefined ? 0b1000_0000 : 0,
			this.level ?? 0xff,
```

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,395 @@
...
+	) {
+		super(driver, options);
+		if (gotDeserializationOptions(options)) {
+			// TODO: Deserialize payload
+			throw new ZWaveError(
+				`${this.constructor.name}: deserialization not implemented`,
+				ZWaveErrorCodes.Deserialization_NotImplemented,
+			);
+		} else {
+			if (options.sceneId < 1 || options.sceneId > 255) {
```
AlCalzone: > The value 0 MUST indicate that the current active scene (if any).

So 0 is also okay. I think this is actually just the same as in `Scene Controller Configuration CC` including the predicate function.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,395 @@
...
+	public constructor(
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+		validatePayload(this.payload.length >= 3);
+		this.sceneId = this.payload[0];
+		this.level = this.payload[1];
+		this.dimmingDuration =
+			Duration.parseReport(this.payload[2]) ?? new Duration(0, "unknown");
```
AlCalzone: > The value 0 MUST indicate that no scene is currently active at the sending node and no scene setting is
advertised in this command. In this case, the Level and Dimming Duration fields SHOULD be ignored.

I think it would make sense to keep level and dimming duration as optional. What do you think?

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,395 @@
...
+			endpoint: this.endpoint.index,
+			sceneId: sceneId,
+		});
+		const response = await this.driver.sendCommand<SceneActuatorConfigurationCCReport>(
+			cc,
+			this.commandOptions,
+		);
+
+		if (response) {
+			return pick(response, ["sceneId", "level", "dimmingDuration"]);
```
AlCalzone: If 0 wasn't used to retrieve the current scene, I'd remove `sceneId` from the return value.
I already didn't like this in Scene Controller Configuration CC, maybe we should split the API command into two:
`getCurrentScene()` and `get(sceneId)` with `sceneId > 0`?

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,395 @@
...
+					);
+				}
+				return (await this.get(propertyKey))?.[property];
+			}
+			default:
+				throwUnsupportedProperty(this.ccId, property);
+		}
+	};
+
+	public async set(
```
AlCalzone, line 170: Maybe we should call this `enableScene`

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,395 @@
...
+			case SceneActuatorConfigurationCommand.Get:
+				return this.isSinglecast();
+			case SceneActuatorConfigurationCommand.Set:
+			case SceneActuatorConfigurationCommand.Report:
+				return true;
+		}
+		return super.supportsCommand(cmd);
+	}
+
+	protected [SET_VALUE]: SetValueImplementation = async (
```
AlCalzone, line 114: The setValue API seems awkward for this CC. It is meant to enable a scene, optionally with a target level. But to enable the scene, one would have to set the `level` property for that scene. Maybe this should be left out.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,395 @@
...
+@API(CommandClasses["Scene Actuator Configuration"])
+export class SceneActuatorConfigurationCCAPI extends CCAPI {
+	public supportsCommand(
+		cmd: SceneActuatorConfigurationCommand,
+	): Maybe<boolean> {
+		switch (cmd) {
+			case SceneActuatorConfigurationCommand.Get:
+				return this.isSinglecast();
+			case SceneActuatorConfigurationCommand.Set:
+			case SceneActuatorConfigurationCommand.Report:
```
AlCalzone:
```suggestion
```
We don't support sending Reports

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
Ikcelaks: I think of "enable" as being a toggle.  Something is either enabled or disabled.  Enabling something that is already enabled shouldn't change anything

In this case, the user is also adjusting duration and (optionally) level settings.

Maybe `setupScene` or `setScene` ?

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
Ikcelaks: Yeah, I rushed this one and didn't read the specs carefully.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
Ikcelaks: Yes, the `level` and `dimmingDuration` fields should be undefined in a report for `sceneId = 0`.  That also means we should not persist values if `sceneId = 0`.

I'm changing the constructor to only call `persistValues()` if `sceneId > 0`, but `persistValues` is a `public` function, so I assume I have to consider the possibility that the function is called from outside the constructor.  Is it correct to have `persistValues` abort and `return false` if it's called when `sceneId = 0`?

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
Ikcelaks: I thought this class was for configuring how the actuator should react when a scene is activated.  When we call a `Set`, we're telling the endpoint that when scene `sceneId` is activated, that actuator should take `dimmingDuration` amount of time to go to a target value of `level` (or optionally the `current level` of the device as of when we're calling the `SceneActuatorConfigurationCC.Set`).  In some sense setting the `level` of scene `sceneId` is the most fundamental part of the `Set` command, because this `level` is always used when the scene is activated (while the `dimmingDuration` assigned here can be overridden by a value from the node that sends the `SceneActivation.Set` command.

So, I think being able to use the setValue API to update the `level` assigned to the scene does make some sense.  However, I do think it's a bit awkward to allow setting of `level` but not allow setting of `dimmingDuration`.  Also, you have a much better understanding of the whole API.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
Ikcelaks: We definitely should have `getCurrentScene()` in the API.

If we want to require `sceneId > 0` for `get(sceneId)` (and omit `sceneId` from the return value), we should update Scene Controller Configuration CC to behave the same way.  That would be a breaking change, but I doubt it would mess anyone up, since it was just added.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
Ikcelaks: The more I think about this, the more I prefer just leaving it as `set`.  The `set` function is configuring the endpoint for how the actuator should behave when a scene is activating.  It's basically setting the scene's actuator configuration.

Ikcelaks:
I made the requested changes, except for changing the name of `set`.

The checks don't pass, because I don't call `persistValues()` when the `Report` sets `sceneId = 0`.  I haven't yet looked into what I need to do in that case.

Let me know if you like how I handled the new `getActive()` and `get(sceneId)` api functions.  The `get()` function is lacking logging, etc., because I want to make sure I have the API that you envisioned.  If we omit `sceneId` from the regular `get()` call, we must prevent `get(0)` from being processed, because we'd have no way of reporting the actual `sceneId` to the caller.  I assume there is a way you want to log and / or return an error.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -193,21 +192,52 @@ export class SceneActuatorConfigurationCCAPI extends CCAPI {
...
 			SceneActuatorConfigurationCommand,
 			SceneActuatorConfigurationCommand.Get,
 		);
 
+		if (sceneId === 0) {
+			// We need to prevent this, since it will cause the Report sceneId
+			// to be different from the requested sceneId, and the client won't
+			// be able to determine the actual sceneId, because it's hidden in
+			// the return value
+			return;
```
AlCalzone: Just throw here with error code `Argument_Invalid` and tell the dev to use `getActive`.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -218,8 +248,12 @@ export class SceneActuatorConfigurationCCAPI extends CCAPI {
...
 
-		if (response) {
-			return pick(response, ["sceneId", "level", "dimmingDuration"]);
+		if (!response) return;
+		if (response.sceneId === sceneId) {
+			return pick(response, ["level", "dimmingDuration"]);
+		} else {
+			// I don't think this can happen, if we block sceneId = 0
+			return;
 		}
```
AlCalzone:
```suggestion
		if (response) {
			return pick(response, ["level", "dimmingDuration"]);
		}
```

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -305,25 +326,35 @@ export class SceneActuatorConfigurationCCReport extends SceneActuatorConfigurati
...
 
 	public persistValues(): boolean {
-		persistSceneActuatorConfig.call(
+		if (
+			this.sceneId === 0 ||
+			this.level == undefined ||
+			this.dimmingDuration == undefined
+		) {
+			return false;
+		}
```
AlCalzone, line 341:
```suggestion
```
No need, you're not calling this if sceneId is 0.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -305,25 +326,35 @@ export class SceneActuatorConfigurationCCReport extends SceneActuatorConfigurati
 		super(driver, options);
 		validatePayload(this.payload.length >= 3);
 		this.sceneId = this.payload[0];
-		this.level = this.payload[1];
-		this.dimmingDuration =
-			Duration.parseReport(this.payload[2]) ?? new Duration(0, "unknown");
 
-		this.persistValues();
+		if (this.sceneId != 0) {
```
AlCalzone:
```suggestion
		if (this.sceneId !== 0) {
```

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
Ikcelaks: Okay, I added that because I wasn't sure if there was a possibility for `persistValues()` to be called from outside of my code.

If it can't be called from outside the class, why is it `public`?

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,422 @@
...
+
+		if (this.sceneId !== 0) {
+			this.level = this.payload[1];
+			this.dimmingDuration =
+				Duration.parseReport(this.payload[2]) ??
+				new Duration(0, "unknown");
+
+			this.persistValues();
+		}
+	}
```
Ikcelaks, line 327: Lint is unhappy with potentially skipping the call to `persistValues()`.

Error: The application CC report class SceneActuatorConfigurationCCReport needs a call to persistValues() in the constructor or an implementation for mergePartialCCs()

`mergePartialCCs` is not appropriate.  I'm going to move `persistValues()` out of the `if`, so it's always called, and leave the check for `sceneId === 0` in the `persistValues` function (which additionally protects against the possibility that something outside the class calls `persistValues` while `sceneId === 0`).

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
Ikcelaks: I ended up leaving in this code and instead moving the call to `persistValues()` out of the `sceneId === 0` check in the constructor.

This was mostly done to please the Lint gods, but I'm also wary of removing necessary checks from a public function (even if I know it should never be called with parameters that fail the check).

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,417 @@
...
+		this.assertSupportsCommand(
+			SceneActuatorConfigurationCommand,
+			SceneActuatorConfigurationCommand.Get,
+		);
+
+		if (sceneId === 0) {
+			throw new ZWaveError(
+				`sceneId 0 is not valid for get. Use getActive() instead.`,
+				ZWaveErrorCodes.Argument_Invalid,
+			);
```
Ikcelaks, line 233: I'm not sure about the wording of this error.  It feels a bit cryptic, but I didn't want to make it too long.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
AlCalzone: > Is it correct to have persistValues abort and return false if it's called when sceneId = 0?

The return type of the subclassed method is currently not evaluated anywhere, but for consistency: **yes**!

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
AlCalzone: Agreed.

>  That would be a breaking change, but I doubt it would mess anyone up, since it was just added.

Also agreed!

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
AlCalzone: > I thought this class was for configuring how the actuator should react when a scene is activated.

You're right - I mixed it up while reading yesterday. Setting level is okay, setting the duration can also be done if you accept Duration objects. I was thinking of those values where the duration is a kind of meta value for another one, e.g. "go to level X with duration Y now!"
Same probably also goes for the SceneControllerConfigurationCC then?

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
AlCalzone: Yeah, see next comment. I got confused

AlCalzone:
Seems like I missed a few comments yesterday - sorry!

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
AlCalzone: Maybe `Invalid scene ID 0. To get the currently active scene, use getActive() instead.`

Ikcelaks:
Do you have any thoughts on the Interview (or lack thereof)?

Currently, not putting anything into the valueDB until a client makes a `get` request.  Is that going to cause issues?  Should we initialize the valueDB with values set to `undefined`?

I'm wondering how zwavejs2mqtt (and other clients) will handle this CC.

AlCalzone:
The Z-Wave CC Control specifications do not include this CC and I agree that it is overkill to query all possible 255 values.
That should probably be done on demand by the applications - and only for the relevant scenes, either using the API methods or `pollValue`.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,445 @@
...
+			const dimmingDuration = node.getValue<Duration>(
+				getDimmingDurationValueID(this.endpoint.index, propertyKey),
+			);
+			// If dimmingDuration is missing, we must abort
+			if (dimmingDuration == undefined) {
+				throw new ZWaveError(
+					`The "dimmingDuration" property cannot be changed before the level is known!`,
+					ZWaveErrorCodes.Argument_Invalid,
+				);
+			}
```
AlCalzone: We're stuck in an infinite loop here, because you also cannot set the dimmingDuration without the level.

I feel like I'm making you run in circles, but it seems like we do need to treat dimmingDuration as a meta property that needs to be set along with the level (#1321)

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
```diff
@@ -0,0 +1,445 @@
...
+			await this.set(propertyKey, value, level);
+		} else {
+			// setting dimmingDuration value alone not supported,
+			// to be added when setting Duration values is supported
+			throwUnsupportedProperty(this.ccId, property);
+		}
+
+		// Verify the current value after a delay
+		this.schedulePoll({ property, propertyKey });
```
AlCalzone: I don't think it is mission critical to verify the change. I only want to do this when it is important, e.g. lights, thermostats and config params (which might not get accepted).

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
Ikcelaks: I'm removing the option to set the `dimmingDuration` alone and reallowing a default of `0 seconds` to be set for the `dimmingDuration` if the user attempts to do a SET_VALUE on `level` before a `dimmingDuration` has been set.  That seems like a reasonable default.

packages/zwave-js/src/lib/commandclass/SceneActuatorConfigurationCC.ts
Ikcelaks: Done.

AlCalzone:
Sounds good - I'll do a couple of smaller changes, then merge.
I think some of these changes need to be applied to the other CC too, right?

AlCalzone (approved):

PR #1808, opened 2021-02-17
feat(config): add Zooz zen72, update zen71 description

bepsoccer:
- add zen72
- update zen71 name to reflect the non use of S2 in the naming by Zooz for the Zen21/22/71/72 product lines


AlCalzone (dismissed):

AlCalzone:
@blhoward2 for a second look, it is late :)

packages/config/config/devices/0x027a/zen72.json
```diff
@@ -0,0 +1,429 @@
...
+					"label": "30%",
+					"value": 2
+				}
+			]
+		},
+		"3": {
+			"label": "Auto Turn-Off Timer",
+			"description": "Set the number of minutes to automatically turn off switch once it has been turned on (0 to disable).",
+			"valueSize": 4,
+			"minValue": 0,
```
blhoward2, line 143: Please use the unit to below valueSize throughout. We like to keep these uniform.

packages/config/config/devices/0x027a/zen72.json
```diff
@@ -0,0 +1,429 @@
...
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Previous brightness",
+					"value": 0
+				},
+				{
+					"label": "Configured custom brightness level",
```
blhoward2: Where and how do you define the custom brightness level?

blhoward2 (changes requested):
Minor changes for consistency.

packages/config/config/devices/0x027a/zen72.json
bepsoccer: I don't understand the comment here, can you clarify?  Also, shall I update the Zen77 as well since this is almost an identical copy of the existing Zen77 with simply the addition of the 3-way momentary switch option.

packages/config/config/devices/0x027a/zen72.json
bepsoccer: Param 18, but once again this is simply a copy of the Zen77, I can change this to reference the other param, but then the Zen77 will need updating as well to be consistent.

packages/config/config/devices/0x027a/zen72.json
blhoward2: Typo. We have a specific order for these fields. The unit should be directly under the valueSize.

packages/config/config/devices/0x027a/zen72.json
blhoward2: Just do this one. We're in the process of templating these so we'll fix it during that process.

packages/config/config/devices/0x027a/zen72.json
bepsoccer: gotcha, I'll remember that.  Fix coming shortly.

blhoward2 (approved):
Approved with changes I directly committed.

PR #1810, opened 2021-02-18
feat(config): support conditional settings with "$if"

AlCalzone:
This PR introduces conditional config settings to cut down on repetition for devices that only slighly change between firmware versions or hardware revisions. The `"$if"` property accepts a string with logic inside, e.g.

```jsonc
{
	"$if": "firmwareVersion >= 1.0 && productType === 0x1234"
	// ... other settings
}
```

The logic supports the JavaScript operators `<`, `<=`, `>`, `>=` and `===`, as well as `&&`, `||` and `(...)`. Available variables are `manufacturerId`, `productType`, `productId` and `firmwareVersion`.

You can use `"$if"` in the following locations:

-   Inside association groups
-   Inside config parameters
-   Inside config parameter options

It is also possible to select one of multiple variants of a config parameter by specifying an array. In this case `"$if"` is mandatory in each parameter, except the last one. The first matching definition will be used, the others don't apply.


docs/development/config-files.md
```diff
@@ -399,3 +399,49 @@ Properties listed before the `$import` statement may get overwritten by the impo
 	"template": true
 }
 ```
+
+### Conditional configuration settings
+
+For devices that only slighly change between firmware versions or hardware revisions, you can enable some settings conditionally with the `"$if"` property. It accepts a string with logic inside, e.g.
+
+```json
```
robertsLando, line 407: I suggest use to use js instead of json so the syntax highlight doesn't highlight comments

robertsLando:
Awesome!

docs/development/config-files.md
AlCalzone: That's just here in the diff. Github knows the type `jsonc` where it does not complain, but Docsify doesn't. However it renders json with comments just fine:

grafik

blhoward2:
Looks great to me. I'm not sure when we'd differentiate by more than firmware version but it could come up, so makes sense to include it. 

Does this (or could this) also apply to association groups? I noticed recently that the Kwikset locks seems to gain a group over time but I'm not actually sure it's even definable as they use the same everything else. It may come up again though.

packages/config/src/Devices.ts
```diff
@@ -455,18 +543,19 @@ noEndpoint in association ${groupId} must be either true or left out`,
...
+		deviceId?: DeviceID,
 	) {
 		this.parameterNumber = parameterNumber;
 		this.valueBitMask = valueBitMask;
 
 		if (typeof definition.label !== "string") {
 			throwInvalidConfig(
 				"devices",
-				`packages/config/config/devices/${filename}:
+				`packages/config/config/devices/${parent.filename}:
```
robertsLando, line 558: there are plenty of `parent.filename` here. Maybe a new var could be useful

AlCalzone:
Yeah you can use it for the individual association groups just not with the array notation of parameters.

blhoward2 (approved):

robertsLando (approved):

PR #1834, opened 2021-02-20
fix(cc): update SceneControllerConfigurationCC API

Ikcelaks:
This PR updates the SceneControllerConfigurationCC API to follow the decisions we made while working on the SceneActuatorConfigurationCC API.

I also applied a few of the other refinements:

1. Set no longer schedules a Poll.
2. `PhysicalCCAPI` was changed to CCAPI, because the Set command is allowed to be part of a Multicast. (and related changes)
3. Prevented Reports with `groupId = 0` from being persisted.


AlCalzone:
>     1. Set no longer schedules a Poll.

Forgot to push that?
grafik

AlCalzone (changes requested):
Other than the refresh thing this looks good

Ikcelaks:
> > ```
> > 1. Set no longer schedules a Poll.
> > ```
> 
> Forgot to push that?
> grafik

I think I just missed that I was polling in both the `SET_VALUE` and `set()` functions.  I did an actual search this time.

AlCalzone (approved):

PR #1836, opened 2021-02-20
feat(config): template and cleanup HomeSeer device files

blhoward2:
Add template, cleanup values.


packages/config/config/devices/0x000c/hs-pa100.json
```diff
@@ -14,32 +14,5 @@
 	"firmwareVersion": {
 		"min": "0.0",
 		"max": "255.255"
-	},
```
blhoward2, line 17: Note: This param does not exist in the manual so I'm removing it.

packages/config/config/devices/0x000c/hs-wd200.json
```diff
@@ -0,0 +1,125 @@
+// HomeSeer Technologies HS-WD200+
```
blhoward2, line 1: Note: These files were split by firmware version, but the files were identical. I recombined them.

AlCalzone (approved):
I love this

PR #1838, opened 2021-02-20
feat(cc): add Entry Control command class

WizKid:
My attempt to add support for Entry Control command class.

Tested it with my Ring Keypad and everything I tested works.

In this pull request there is still nothing that handle the Notification message but I figured it is better to submit this first and then work on that.

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
@@ -80,26 +80,42 @@ export enum EntryControlEventTypes {
 }
 
 export enum EntryControlDataTypes {
-	NA = 0x00,
+	None = 0x00,
 	Raw = 0x01,
-	Ascii = 0x02,
-	Md5 = 0x03,
+	ASCII = 0x02,
+	MD5 = 0x03,
 }
 
-export function getKeyCachedSizeStateValueID(endpoint: number): ValueID {
+function getValueID(property: string, endpoint: number): ValueID {
 	return {
 		commandClass: CommandClasses["Entry Control"],
 		endpoint,
-		property: "keyCachedSize",
+		property: property,
 	};
 }
 
-export function getKeyCachedTimeoutStateValueID(endpoint: number): ValueID {
-	return {
-		commandClass: CommandClasses["Entry Control"],
-		endpoint,
-		property: "keyCachedTimeout",
-	};
+export function getKeyCacheSizeStateValueID(endpoint: number): ValueID {
+	return getValueID("keyCacheSize", endpoint);
+}
+
+export function getKeyCacheTimeoutStateValueID(endpoint: number): ValueID {
+	return getValueID("keyCacheTimeout", endpoint);
+}
+
+export function getMinKeyCacheSizeStateValueID(endpoint: number): ValueID {
+	return getValueID("minKeyCacheSize", endpoint);
+}
+
+export function getMaxKeyCacheSizeStateValueID(endpoint: number): ValueID {
+	return getValueID("maxKeyCacheSize", endpoint);
+}
+
+export function getMinKeyCacheTimeoutStateValueID(endpoint: number): ValueID {
+	return getValueID("minKeyCacheTimeout", endpoint);
+}
+
+export function getMaxKeyCacheTimeoutStateValueID(endpoint: number): ValueID {
+	return getValueID("maxKeyCacheTimeout", endpoint);
 }
 
 @API(CommandClasses["Entry Control"])
@@ -117,7 +133,7 @@ export class EntryControlCCAPI extends CCAPI {
 	}
 
 	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-	public async getKeySupported() {
+	public async getSupportedKeys() {
 		this.assertSupportsCommand(
 			EntryControlCommand,
 			EntryControlCommand.KeySupportedGet,
@@ -131,13 +147,11 @@ export class EntryControlCCAPI extends CCAPI {
 			cc,
 			this.commandOptions,
 		);
-		if (response) {
-			return pick(response, ["keySupported"]);
-		}
+		return response?.supportedKeys;
 	}
 
 	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-	public async getEventSupported() {
+	public async getEventCapabilities() {
 		this.assertSupportsCommand(
 			EntryControlCommand,
 			EntryControlCommand.EventSupportedGet,
@@ -153,12 +167,12 @@ export class EntryControlCCAPI extends CCAPI {
 		);
 		if (response) {
 			return pick(response, [
-				"dataTypeSupported",
-				"eventTypeSupported",
-				"keyCachedSizeMin",
-				"keyCachedSizeMax",
-				"keyCachedTimeoutMin",
-				"keyCachedTimeoutMax",
+				"supportedDataTypes",
+				"supportedEventTypes",
+				"minKeyCacheSize",
+				"maxKeyCacheSize",
+				"minKeyCacheTimeout",
+				"maxKeyCacheTimeout",
 			]);
 		}
 	}
@@ -179,7 +193,7 @@ export class EntryControlCCAPI extends CCAPI {
 			this.commandOptions,
 		);
 		if (response) {
-			return pick(response, ["keyCachedSize", "keyCachedTimeout"]);
+			return pick(response, ["keyCacheSize", "keyCacheTimeout"]);
 		}
 	}
 
@@ -193,6 +207,56 @@ export class EntryControlCCAPI extends CCAPI {
 			EntryControlCommand.ConfigurationGet,
 		);
 
+		const valueDB = this.endpoint.getNodeUnsafe()!.valueDB;
+
+		// Validate key cache size
+		const minKeyCacheSize = valueDB.getValue<number>(
+			getMinKeyCacheSizeStateValueID(this.endpoint.index),
+		);
+		if (minKeyCacheSize !== undefined && keyCacheSize < minKeyCacheSize) {
+			throw new ZWaveError(
+				`The "keyCacheSize" (${keyCacheSize}) must be >= than "minKeyCacheSize" (${minKeyCacheSize})!`,
+				ZWaveErrorCodes.Argument_Invalid,
+			);
+		}
+
+		const maxKeyCacheSize = valueDB.getValue<number>(
+			getMaxKeyCacheSizeStateValueID(this.endpoint.index),
+		);
+		if (maxKeyCacheSize !== undefined && keyCacheSize > maxKeyCacheSize) {
+			throw new ZWaveError(
+				`The "keyCacheSize" (${keyCacheSize}) must be <= than "maxKeyCacheSize" (${maxKeyCacheSize})!`,
+				ZWaveErrorCodes.Argument_Invalid,
+			);
+		}
+
+		// Validate key cache timeout
+		const minKeyCacheTimeout = valueDB.getValue<number>(
+			getMinKeyCacheTimeoutStateValueID(this.endpoint.index),
+		);
+		if (
+			minKeyCacheTimeout !== undefined &&
+			keyCacheTimeout < minKeyCacheTimeout
+		) {
+			throw new ZWaveError(
+				`The "keyCacheTimeout" (${keyCacheTimeout}) must be >= than "minKeyCacheTimeout" (${minKeyCacheTimeout})!`,
+				ZWaveErrorCodes.Argument_Invalid,
+			);
+		}
+
+		const maxKeyCacheTimeout = valueDB.getValue<number>(
+			getMaxKeyCacheTimeoutStateValueID(this.endpoint.index),
+		);
+		if (
+			maxKeyCacheTimeout !== undefined &&
+			keyCacheTimeout > maxKeyCacheTimeout
+		) {
+			throw new ZWaveError(
+				`The "keyCacheTimeout" (${keyCacheTimeout}) must be <= than "maxKeyCacheTimeout" (${maxKeyCacheTimeout})!`,
+				ZWaveErrorCodes.Argument_Invalid,
+			);
+		}
+
 		const cc = new EntryControlCCConfigurationSet(this.driver, {
 			nodeId: this.endpoint.nodeId,
 			endpoint: this.endpoint.index,
@@ -204,7 +268,7 @@ export class EntryControlCCAPI extends CCAPI {
 			this.commandOptions,
 		);
 		if (response) {
-			return pick(response, ["keyCachedSize", "keyCachedTimeout"]);
+			return pick(response, ["keyCacheSize", "keyCacheTimeout"]);
 		}
 	}
 
@@ -212,7 +276,7 @@ export class EntryControlCCAPI extends CCAPI {
 		{ property },
 		value,
 	): Promise<void> => {
-		if (property !== "keyCachedSize" && property !== "keyCachedTimeout") {
+		if (property !== "keyCacheSize" && property !== "keyCacheTimeout") {
 			throwUnsupportedProperty(this.ccId, property);
 		}
 		if (typeof value !== "number") {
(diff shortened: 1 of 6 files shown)


AlCalzone:
Cool - I'll review tomorrow

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+	Unlock = 0x17,
+	Test = 0x18,
+	Cancel = 0x19,
+}
+
+export enum EntryControlDataTypes {
+	NA = 0x00,
+	Raw = 0x01,
+	Ascii = 0x02,
+	Md5 = 0x03,
```
AlCalzone:
```suggestion
	None = 0x00,
	Raw = 0x01,
	ASCII = 0x02,
	MD5 = 0x03,
```

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+			}
+		}
+
+		// Remember that the interview is complete
+		this.interviewComplete = true;
+	}
+}
+
+@CCCommand(EntryControlCommand.Notification)
+export class EntryControlCCNotication extends EntryControlCC {
```
AlCalzone:
```suggestion
export class EntryControlCCNotification extends EntryControlCC {
```

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+
+		validatePayload(this.payload.length >= 4);
+		this._sequenceNumber = this.payload[0];
+		this._dataType = this.payload[1] & 0b11;
+		this._eventType = this.payload[2];
+		const eventDataLength = this.payload[3];
```
AlCalzone, line 346: > The value MUST be in the range 0..32.

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+
+		validatePayload(this.payload.length >= 4 + eventDataLength);
+		this._eventData = Buffer.from(
```
AlCalzone: I think we should distinguish between the types here.
(None)/Raw/MD5 -> keep as buffer
ASCII -> Convert to ASCII string

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+	@ccValue()
+	@ccValueMetadata({
+		...ValueMetadata.ReadOnly,
+		type: "number[]",
+		label: "Supported keys",
+		description: "A list of supported keys",
+	})
+	public get keySupported(): number[] {
+		return this._keySupported;
+	}
```
AlCalzone:
```suggestion
    @ccValue({internal: true})
	public readonly supportedKeys: readonly number[];
```
The support information is internal to zwave-js. UIs don't want to display this without context.

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+export class EntryControlCCEventSupportedReport extends EntryControlCC {
+	public constructor(
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+
+		validatePayload(this.payload.length >= 6);
+
+		const dataTypeLength = this.payload[0] & 0b0000_0011;
```
AlCalzone:
```suggestion
		const dataTypeLength = this.payload[0] & 0b11;
```

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+		validatePayload(this.payload.length >= 6 + dataTypeLength);
```
AlCalzone:
```suggestion
		validatePayload(this.payload.length >= 1 + dataTypeLength);
```

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+		this._dataTypeSupported = parseBitMask(
+			this.payload.slice(1, 1 + dataTypeLength),
+			0,
```
AlCalzone: Please use the enum member here

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+		);
+
+		const eventTypeLength = this.payload[1 + dataTypeLength] & 0b0001_1111;
```
AlCalzone:
```suggestion
		const eventTypeLength = this.payload[1 + dataTypeLength] & 0b11111;
```

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+		validatePayload(
+			this.payload.length >= 6 + dataTypeLength + eventTypeLength,
```
AlCalzone:
```suggestion
			this.payload.length >= 1 + dataTypeLength + eventTypeLength,
```
I prefer to work with an offset variable in cases where multiple variable-length fields are included in a command.
Otherwise you're going to repeat yourself a lot.

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+		);
+		this._eventTypeSupported = parseBitMask(
+			this.payload.slice(
+				2 + dataTypeLength,
+				2 + dataTypeLength + eventTypeLength,
+			),
+			0,
```
AlCalzone: Also here, enum member please.

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+	@ccValue()
+	@ccValueMetadata({
+		...ValueMetadata.ReadOnly,
+		type: "number[]",
+		label: "Supported event types",
+		description: "A list of supported event types",
+	})
+	public get eventTypeSupported(): EntryControlEventTypes[] {
+		return this._eventTypeSupported;
+	}
```
AlCalzone: see previous command

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+	private _keyCachedTimeoutMax: number;
+
+	@ccValue()
+	@ccValueMetadata({
+		...ValueMetadata.ReadOnlyUInt8,
+		label: "Maximum supported value of key cache timeout",
+	})
+	public get keyCachedTimeoutMax(): number {
+		return this._keyCachedTimeoutMax;
+	}
```
AlCalzone: These shouldn't be values, but rather change the min/max of the key cache size/timeout value metadata.

AlCalzone: Also regarding the naming:
`minKeyCacheSize` etc sounds more natural.

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+
+	@ccValue()
+	@ccValueMetadata({
+		...ValueMetadata.UInt8,
+		label: "Key cache size",
+		description: "Number of character that must be stored before sending",
+	})
+	public get keyCachedSize(): number {
+		return this._keyCachedSize;
+	}
```
AlCalzone:
```suggestion
	@ccValue()
	@ccValueMetadata({
		...ValueMetadata.UInt8,
		label: "Key cache size",
		description: "Number of character that must be stored before sending",
	})
	public readonly keyCacheSize: number;
```
same for the timeout. The default metadata should contain min/max values, see specs.

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+		super(driver, options);
+		if (gotDeserializationOptions(options)) {
+			// TODO: Deserialize payload
+			throw new ZWaveError(
+				`${this.constructor.name}: deserialization not implemented`,
+				ZWaveErrorCodes.Deserialization_NotImplemented,
+			);
+		} else {
+			this.keyCacheSize = options.keyCacheSize;
+			this.keyCacheTimeout = options.keyCacheTimeout;
```
AlCalzone, line 585: Validate the options against the range defined in the specs

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+			case EntryControlCommand.ConfigurationGet:
+				return this.isSinglecast();
+			case EntryControlCommand.ConfigurationSet:
+				return true;
+		}
+		return super.supportsCommand(cmd);
+	}
+
+	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+	public async getKeySupported() {
```
AlCalzone: I'd call this `getSupportedKeys`

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+			nodeId: this.endpoint.nodeId,
+			endpoint: this.endpoint.index,
+		});
+		const response = await this.driver.sendCommand<EntryControlCCKeySupportedReport>(
+			cc,
+			this.commandOptions,
+		);
+		if (response) {
+			return pick(response, ["keySupported"]);
+		}
```
AlCalzone:
```suggestion
		return response?.supportedKeys;
```

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+	}
+
+	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+	public async getEventSupported() {
```
AlCalzone: I'd call this `getEventCapabilities`.

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+		if (property === "keyCachedSize") {
+			const oldKeyCachedTimeout = valueDB.getValue<number>(
+				getKeyCachedTimeoutStateValueID(this.endpoint.index),
+			);
+			if (oldKeyCachedTimeout == undefined) {
+				throw new ZWaveError(
+					`The "keyCachedSize" property cannot be changed before the key cached timeout is known!`,
+					ZWaveErrorCodes.Argument_Invalid,
+				);
+			}
```
AlCalzone: Can we assume/use a default timeout here? Otherwise the API user gets stuck in a loop

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -0,0 +1,683 @@
...
+			direction: "none",
+		});
+
+		this.driver.controllerLog.logNode(node.id, {
+			endpoint: this.endpointIndex,
+			message: "requesting entry control supported keys...",
+			direction: "outbound",
+		});
+
+		{
```
AlCalzone: Interesting approach with the blocks. I think I still prefer better names for `resp` over this trick :)

AlCalzone (changes requested):
The notification command needs to be surfaced for the user somehow. I think a separate event like for the Notification CC (although less complicated) would make sense.

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
WizKid: How do I make them value metadata instead of value?

WizKid: How do I make them value metadata instead of value?

AlCalzone: An example can be found in `ThermostatFanModeCCSupportedReport` where the states are changed to match what the device reports.
You can use the same approach to change min/max of the value's corresponding metadata.

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -193,6 +207,56 @@ export class EntryControlCCAPI extends CCAPI {
...
+		);
+		if (
+			maxKeyCacheTimeout !== undefined &&
+			keyCacheTimeout > maxKeyCacheTimeout
+		) {
+			throw new ZWaveError(
+				`The "keyCacheTimeout" (${keyCacheTimeout}) must be <= than "maxKeyCacheTimeout" (${maxKeyCacheTimeout})!`,
+				ZWaveErrorCodes.Argument_Invalid,
+			);
+		}
```
AlCalzone: I would only check if a value is passed that lies outside what the specs allow. If the user provides a value the device does not report as supported, that is on him. The UIs already give enough guidance when you limit min/max with metadata.
That check should be inside the CC implementation, see https://github.com/zwave-js/node-zwave-js/pull/1838/files/efaee1bc81b088fdb8fbd7791cf23a718a1de1f9#diff-5e7c890d091030ea844ddf9bc6acf697c2883b79e41c86e3861e177b5bd49729

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -204,65 +268,56 @@ export class EntryControlCCAPI extends CCAPI {
...
-			case "keyCachedTimeout":
+			case "supportedDataTypes":
+			case "supportedEventTypes":
+			case "minKeyCacheSize":
+			case "maxKeyCacheSize":
+			case "minKeyCacheTimeout":
+			case "maxKeyCacheTimeout":
+				return (await this.getEventCapabilities())?.[property];
+			case "supportedKeys":
+				return this.getSupportedKeys();
```
AlCalzone: Sorry, it was late last night when I reviewed. Most of these shouldn't be public values, so they shouldn't be pollable with this API. `minKeyCacheSize` etc should be included in metadata, `supportedDataTypes, supportedEventTypes, supportedKeys` should be known after the interview and not change.

AlCalzone (changes requested):
Looking better. In addition to the comments below, the min/max timeout/size should be part of the metadata.

packages/zwave-js/src/lib/commandclass/EntryControlCC.ts
```diff
@@ -513,9 +438,35 @@ export class EntryControlCCEventSupportedReport extends EntryControlCC {
...
+
+		const keyCacheTimeoutValueId = getKeyCacheTimeoutStateValueID(
+			this.endpointIndex,
+		);
+		this.getValueDB().setMetadata(keyCacheTimeoutValueId, {
+			...ValueMetadata.UInt8,
+			min: this.minKeyCacheTimeout,
+			max: this.maxKeyCacheTimeout,
+		});
+
```
AlCalzone, line 469: Looks cleaner now - please remove the `@ccValue({ internal: true })` annotations below this line for the properties that are now part of metadata.

AlCalzone (changes requested):
Almost... One comment below and we need to surface the notification through an event.

AlCalzone:
Just a heads up: We've discovered yesterday that the `"notification"` event needs some rework. I think the way forward is to use it as a base for all CCs that send more complex notifications like this one.

I'm going to hold off on merging until we've decided how the revamped event should look like, in order not to introduce additional breaks when we ultimately change the event signature.

WizKid:
Do you want me to remove the event sending part and land the rest?

AlCalzone:
How useful is the CC is without that?

WizKid:
I forgot that I develop this on the master branch. I think this restores it.

marcus-j-davies:
@WizKid,

More on the 1st Gen Ring keypad.

I'm coming across reports that state, their is a proprietary pairing process and some mention that it enforces S2.
Mine is being delivered today (EU).

Have you been able to pair the device in the normal manner, are their certain steps you have to take?

I'm confused with your success in using it without the official hub, and the articles I'm reading.
I will be testing my device with your branch (which has all the inner workings I believe?)

UPDATE:
Wow - it does indeed work (input, arm buttons) 
I note it supports the indicator CC - have you had any luck with that?

Naturally, it can be used to indicate, the current security status. 
I'm using PC Controller to interrogate/set its abilities without the official hub

WizKid:
@marcus-j-davies Yes Indicator works for me. These are the values I been able to find https://gist.github.com/WizKid/baa467cb0a3424b401f411832c4d0cee

We should probably start a discussion thread or continue on Slack/Discord regarding general thing with the Keypad that has nothing to do with this PR.

AlCalzone:
I've rebased this PR onto the dev-branch for 7.0.0 and made it compatible. The changes are now in https://github.com/zwave-js/node-zwave-js/pull/2022 - thanks for your work so far!

PR #1840, opened 2021-02-21
feat(cc): implement Powerlevel CC

Ikcelaks:
Implements Powerlevel CC.

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
@@ -1,11 +1,9 @@
 import {
 	CommandClasses,
-	enumValuesToMetadataStates,
 	Maybe,
 	MessageOrCCLogEntry,
 	MessageRecord,
 	validatePayload,
-	ValueMetadata,
 	ZWaveError,
 	ZWaveErrorCodes,
 } from "@zwave-js/core";
@@ -17,8 +15,6 @@ import {
 	API,
 	CCCommand,
 	CCCommandOptions,
-	ccValue,
-	ccValueMetadata,
 	CommandClass,
 	commandClass,
 	CommandClassDeserializationOptions,
@@ -38,16 +34,16 @@ export enum PowerlevelCommand {
 }
 
 export enum Powerlevel {
-	NormalPower = 0x00,
-	Minus1dBm = 0x01,
-	Minus2dBm = 0x02,
-	Minus3dBm = 0x03,
-	Minus4dBm = 0x04,
-	Minus5dBm = 0x05,
-	Minus6dBm = 0x06,
-	Minus7dBm = 0x07,
-	Minus8dBm = 0x08,
-	Minus9dBm = 0x09,
+	"Normal Power" = 0x00,
+	"-1 dBm" = 0x01,
+	"-2 dBm" = 0x02,
+	"-3 dBm" = 0x03,
+	"-4 dBm" = 0x04,
+	"-5 dBm" = 0x05,
+	"-6 dBm" = 0x06,
+	"-7 dBm" = 0x07,
+	"-8 dBm" = 0x08,
+	"-9 dBm" = 0x09,
 }
 
 export enum PowerlevelTestStatus {
@@ -76,7 +72,7 @@ export class PowerlevelCCAPI extends CCAPI {
 		const cc = new PowerlevelCCSet(this.driver, {
 			nodeId: this.endpoint.nodeId,
 			endpoint: this.endpoint.index,
-			powerlevel: Powerlevel.NormalPower,
+			powerlevel: Powerlevel["Normal Power"],
 		});
 		await this.driver.sendCommand(cc, this.commandOptions);
 	}
@@ -114,7 +110,7 @@ export class PowerlevelCCAPI extends CCAPI {
 		}
 	}
 
-	public async startTestNode(
+	public async startNodeTest(
 		testNodeId: number,
 		powerlevel: Powerlevel,
 		testFrameCount: number,
@@ -134,10 +130,10 @@ export class PowerlevelCCAPI extends CCAPI {
 		await this.driver.sendCommand(cc, this.commandOptions);
 	}
 
-	public async getTestNodeResult(): Promise<
+	public async getNodeTestStatus(): Promise<
 		| Pick<
 				PowerlevelCCTestNodeReport,
-				"testNodeId" | "testStatus" | "testFrameAcknowlegedCount"
+				"testNodeId" | "status" | "acknowledgedFrames"
 		  >
 		| undefined
 	> {
@@ -150,15 +146,16 @@ export class PowerlevelCCAPI extends CCAPI {
 			nodeId: this.endpoint.nodeId,
 			endpoint: this.endpoint.index,
 		});
-		const response = await this.driver.sendCommand<PowerlevelCCTestNodeReport>(
-			cc,
-			this.commandOptions,
-		);
+		const response =
+			await this.driver.sendCommand<PowerlevelCCTestNodeReport>(
+				cc,
+				this.commandOptions,
+			);
 		if (response) {
 			return pick(response, [
 				"testNodeId",
-				"testStatus",
-				"testFrameAcknowlegedCount",
+				"status",
+				"acknowledgedFrames",
 			]);
 		}
 	}
@@ -177,7 +174,8 @@ type PowerlevelCCSetOptions = CCCommandOptions &
 				timeout: number;
 		  }
 		| {
-				powerlevel: typeof Powerlevel.NormalPower;
+				powerlevel: typeof Powerlevel["Normal Power"];
+				timeout?: undefined;
 		  }
 	);
 
@@ -195,16 +193,16 @@ export class PowerlevelCCSet extends PowerlevelCC {
 				ZWaveErrorCodes.Deserialization_NotImplemented,
 			);
 		} else {
-			if ("timeout" in options) {
+			this.powerlevel = options.powerlevel;
+			if (options.powerlevel !== Powerlevel["Normal Power"]) {
 				if (options.timeout < 1 || options.timeout > 255) {
 					throw new ZWaveError(
-						`${this.constructor.name}: The timeout parameter must be between 1 and 255.`,
+						`The timeout parameter must be between 1 and 255.`,
 						ZWaveErrorCodes.Argument_Invalid,
 					);
 				}
 				this.timeout = options.timeout;
 			}
-			this.powerlevel = options.powerlevel;
 		}
 	}
 
@@ -218,7 +216,7 @@ export class PowerlevelCCSet extends PowerlevelCC {
 
 	public toLogEntry(): MessageOrCCLogEntry {
 		const message: MessageRecord = {
-			powerlevel: getEnumMemberName(Powerlevel, this.powerlevel),
+			"power level": getEnumMemberName(Powerlevel, this.powerlevel),
 		};
 		if (this.timeout != undefined) {
 			message.timeout = this.timeout;
@@ -239,36 +237,26 @@ export class PowerlevelCCReport extends PowerlevelCC {
 		super(driver, options);
 
 		this.powerlevel = this.payload[0];
-		if (this.powerlevel !== Powerlevel.NormalPower) {
+		if (this.powerlevel !== Powerlevel["Normal Power"]) {
 			this.timeout = this.payload[1];
 		}
 
 		this.persistValues();
 	}
 
-	@ccValue({ stateful: false })
-	@ccValueMetadata({
-		...ValueMetadata.UInt8,
-		states: enumValuesToMetadataStates(Powerlevel),
-		label: "Power level",
-		description: "The current power level  in effect on the node.",
-	})
 	public readonly powerlevel: Powerlevel;
-
-	@ccValue({ stateful: false })
-	@ccValueMetadata({
-		...ValueMetadata.UInt8,
-		min: 1,
-		max: 255,
-		label: "Timeout",
-		description: "Time in seconds before node returns to Normal power.",
-	})
 	public readonly timeout?: number;
 
 	public toLogEntry(): MessageOrCCLogEntry {
+		const message: MessageRecord = {
+			"power level": getEnumMemberName(Powerlevel, this.powerlevel),
+		};
+		if (this.timeout != undefined) {
+			message.timeout = `${this.timeout} s`;
+		}
 		return {
 			...super.toLogEntry(),
-			message: { "Power level": this.powerlevel, Timeout: this.timeout },
+			message,
 		};
 	}
 }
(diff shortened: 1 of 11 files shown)


Ikcelaks:
I removed the `@expectedCCResponse(PowerlevelCCTestNodeReport)` from the `PowerlevelCCTestNodeSet` command.  The `TestNodeSet` command is supposed to cause the receiver to send a `TestNodeReport` when the test is complete, but I realized it wasn't appropriate to block waiting for that to return.

All of the report values are non-stateful.  Hopefully I handled those correctly.

The `TestNodeSet` command requires a `TestNodeId` for a node that is (according to the specs) "Not asleep" and not "FLiRS".  I took that to mean that it should specifically be "Awake" (and not "FLiRS"), because it doesn't make much sense to me to target a node in either of the other states either, but maybe I'm wrong.

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+
+type PowerlevelCCSetOptions = CCCommandOptions &
+	(
+		| {
+				powerlevel: Powerlevel;
+				timeout: number;
+		  }
+		| {
+				powerlevel: typeof Powerlevel.NormalPower;
+		  }
```
AlCalzone: This might need `timeout?: undefined` in order to make some checks less awkward in TypeScript. For example the one in the constructor below.

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+			// TODO: Deserialize payload
+			throw new ZWaveError(
+				`${this.constructor.name}: deserialization not implemented`,
+				ZWaveErrorCodes.Deserialization_NotImplemented,
+			);
+		} else {
+			if ("timeout" in options) {
+				if (options.timeout < 1 || options.timeout > 255) {
+					throw new ZWaveError(
+						`${this.constructor.name}: The timeout parameter must be between 1 and 255.`,
```
AlCalzone:
```suggestion
						`The timeout parameter must be between 1 and 255.`,
```

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+			}
+			this.powerlevel = options.powerlevel;
+		}
+	}
+
+	public powerlevel: Powerlevel;
+	public timeout?: number;
+
+	public serialize(): Buffer {
+		this.payload = Buffer.from([this.powerlevel, this.timeout ?? 0x00]);
```
AlCalzone, line 236: Just thinking out of the box, since I've been burned a few times.
If valid values for the timeout are ranging from 1 to 255 and the timeout must be ignored for "normal power", it could theoretically be that some weird device chokes on a timeout value of 0.
So to be sure, lets make 1 the default.

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+
+		this.persistValues();
+	}
+
+	@ccValue({ stateful: false })
+	@ccValueMetadata({
+		...ValueMetadata.UInt8,
+		states: enumValuesToMetadataStates(Powerlevel),
+		label: "Power level",
+		description: "The current power level  in effect on the node.",
```
AlCalzone:
```suggestion
		description: "The current power level in effect on the node.",
```

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+	})
+	public readonly powerlevel: Powerlevel;
+
+	@ccValue({ stateful: false })
+	@ccValueMetadata({
+		...ValueMetadata.UInt8,
+		min: 1,
+		max: 255,
+		label: "Timeout",
+		description: "Time in seconds before node returns to Normal power.",
```
AlCalzone:
```suggestion
		description: "Time before node returns to Normal power.",
```

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+		description: "The current power level  in effect on the node.",
+	})
+	public readonly powerlevel: Powerlevel;
+
+	@ccValue({ stateful: false })
+	@ccValueMetadata({
+		...ValueMetadata.UInt8,
+		min: 1,
+		max: 255,
+		label: "Timeout",
```
AlCalzone:
```suggestion
		unit: "seconds",
		label: "Timeout",
```

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+		description: "Time in seconds before node returns to Normal power.",
+	})
+	public readonly timeout?: number;
+
+	public toLogEntry(): MessageOrCCLogEntry {
+		return {
+			...super.toLogEntry(),
+			message: { "Power level": this.powerlevel, Timeout: this.timeout },
```
AlCalzone: I think all other of these messages begin lowercase - lets keep it this way.
Also, timeout should only be logged if not undefined.

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+			// TODO: Deserialize payload
+			throw new ZWaveError(
+				`${this.constructor.name}: deserialization not implemented`,
+				ZWaveErrorCodes.Deserialization_NotImplemented,
+			);
+		} else {
+			const testNode = driver.controller.nodes.get(options.testNodeId);
+			if (!testNode) {
+				throw new ZWaveError(
+					`${this.constructor.name}: Node Id ${options.testNodeId} doesn't exist.`,
```
AlCalzone: Please get rid of all the `${this.constructor.name}: `. The ones in `if (gotDeserializationOptions)` may stay due to consistency.

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+					ZWaveErrorCodes.Argument_Invalid,
+				);
+			}
+			if (testNode.isFrequentListening) {
+				throw new ZWaveError(
+					`${this.constructor.name}: Node Id ${options.testNodeId} is FLiRS, which is not allowed for the Test Node.`,
+					ZWaveErrorCodes.Argument_Invalid,
+				);
+			}
+			if (testNode.status !== NodeStatus.Awake) {
```
AlCalzone: This check should only apply if `testNode.canSleep` is true.

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+		return super.serialize();
+	}
+
+	public toLogEntry(): MessageOrCCLogEntry {
+		return {
+			...super.toLogEntry(),
+			message: {
+				testNodeId: this.testNodeId,
+				powerlevel: this.powerlevel,
+				testFrameCount: this.testFrameCount,
```
AlCalzone: the property names should be a bit nicer in a log: `test node ID`, `powerlevel`, `test frame count`.

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+
+		validatePayload(this.payload.length >= 4);
+		this.testNodeId = this.payload[0];
+		this.testStatus = this.payload[1];
+		this.testFrameAcknowlegedCount =
+			(this.payload[2] << 8) | this.payload[3];
```
AlCalzone: Use `readUInt16BE` instead

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+		states: enumValuesToMetadataStates(PowerlevelTestStatus),
+		label: "Test Status",
+	})
+	public readonly testStatus: PowerlevelTestStatus;
+
+	@ccValue({ stateful: false })
+	@ccValueMetadata({
+		...ValueMetadata.ReadOnlyUInt16,
+		min: 1,
+		max: 65535,
```
AlCalzone: Not necessary, that is implied by `ReadOnlyUInt16`

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+		label: "Power level",
+		description: "The current power level  in effect on the node.",
+	})
+	public readonly powerlevel: Powerlevel;
+
+	@ccValue({ stateful: false })
+	@ccValueMetadata({
+		...ValueMetadata.UInt8,
+		min: 1,
+		max: 255,
```
AlCalzone: Not necessary, that is implied by UInt8

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+		this.testFrameAcknowlegedCount =
+			(this.payload[2] << 8) | this.payload[3];
+
+		this.persistValues();
+	}
+
+	@ccValue({ stateful: false })
+	@ccValueMetadata({
+		...ValueMetadata.ReadOnlyUInt8,
+		label: "Test Node Id",
```
AlCalzone: If we're strict, this should have `min: 1, max: 232`

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
```diff
@@ -0,0 +1,409 @@
...
+	Minus1dBm = 0x01,
+	Minus2dBm = 0x02,
+	Minus3dBm = 0x03,
+	Minus4dBm = 0x04,
+	Minus5dBm = 0x05,
+	Minus6dBm = 0x06,
+	Minus7dBm = 0x07,
+	Minus8dBm = 0x08,
+	Minus9dBm = 0x09,
+}
```
AlCalzone: Since these will end up as labels somewhere through the metadata, I'm thinking they should be formatted like this:
```suggestion
export enum Powerlevel {
	"Normal Power" = 0x00,
	"-1 dBm" = 0x01,
	"-2 dBm" = 0x02,
	"-3 dBm" = 0x03,
	"-4 dBm" = 0x04,
	"-5 dBm" = 0x05,
	"-6 dBm" = 0x06,
	"-7 dBm" = 0x07,
	"-8 dBm" = 0x08,
	"-9 dBm" = 0x09,
}
```

AlCalzone (changes requested):
> All of the report values are non-stateful. Hopefully I handled those correctly.

I just read this after the review. In this case, remove all the @ccValue and @ccValueMetadata stuff to make sure the values aren't stored automatically.

A few pointers in terms of the API:
I'd rename `startTestNode` -> `startNodeTest` and `getTestNodeResult` to `getNodeTestStatus` (since it can be ongoing).
I agree that the `TestNodeSet` shouldn't wait for the unsolicited report. This functionality would best be offered through a method on `ZWaveNode`, that returns a Promise which resolves when the report comes in.

packages/zwave-js/src/lib/commandclass/PowerlevelCC.ts
Ikcelaks: > > All of the report values are non-stateful. Hopefully I handled those correctly.
> 
> I just read this after the review. In this case, remove all the @ccValue and @ccValueMetadata stuff to make sure the values aren't stored automatically.

Doesn't `@ccValue({ stateful: false })` cause the value to be emitted as a "value notification" and not stored in the DB?  I do see that having all of the separate properties will cause multiple events, each with just one part of the result.

What I really want to do is have `PowerlevelCCReport` produce one "value notification" with `timeout` as the value and `powerlevel` in the label.

`PowerlevelCCTestNodeReport` is a bit less obvious to implement.  It should produce one "value notification" with `testFrameAcknowlegedCount` as the value and `testStatus`, `testFrameCount`, `powerlevel`, and `testNodeId` imbedded as metadata.  The `testStatus` value would be indexed by the `testNodeId` (like Central Scene indexes values by `sceneNumber`). I believe that I can accomplish that with `ccSpecific` metadata.

> A few pointers in terms of the API:
> I'd rename `startTestNode` -> `startNodeTest` and `getTestNodeResult` to `getNodeTestStatus` (since it can be ongoing).
> I agree that the `TestNodeSet` shouldn't wait for the unsolicited report. This functionality would best be offered through a method on `ZWaveNode`, that returns a Promise which resolves when the report comes in.

Does it make sense to offer this method in addition to generating the "value notifications"?  I think I can see use cases that would prefer both options.  The `ZWaveNode` method would be simpler for a client that just wants to start the test and await the result.  The "value notifications" would be really easy to integrate into Home Assistant; the user would just need to kick off a service call and listen on the `zwave_js_event` for the matching "value notification".

AlCalzone: I can see your point, but I prefer having values only for CCs that a user often interacts with. I'd probably remove the SET/POLL_VALUE api completely for this CC. Such management commands are better suited for specialized methods on the node/controller/driver. Examples are the network heal, firmware update.

The API should make sense standalone and not try to work around how HA works internally.

Ikcelaks:
@AlCalzone I moved this reply to the top level, because it was accidentally in a thread about a minor code tweak, and I don't want this discussion to disappear when that's resolved.

> I can see your point, but I prefer having values only for CCs that a user often interacts with. I'd probably remove the SET/POLL_VALUE api completely for this CC. Such management commands are better suited for specialized methods on the node/controller/driver. Examples are the network heal, firmware update.
> 
> The API should make sense standalone and not try to work around how HA works internally.

That makes sense.  Values really should only be for states and events that are part of the operational network.  `Powerlevel CC` tests, like firmware updates, are really out-of-band operations that shouldn't get intermingled with something like the button a user pushed on a device supporting `CentralSceneCC`.

I'll remove the metadata / `persistValues` code and implement methods in the `ZWaveNode` class.

I feel the proper solution is to create a `ZwaveNodePowerlevelNodeTestStatusUpdateCallback` in `ZWaveNodeEventCallbacks` and emit the test status reports as events.  This mirrors how firmware updates give status updates.  This allows listening clients to receive notification when the test has completed (including the actual results of the test).  This would be in addition to a `getPowerlevelNodeTestStatus` method that calls the `Get` command and returns a Promise for the `PowerlevelCCTestNodeReport` response.

The `setPowerlevel` and `getPowerlevel` methods probably don't need an event.

AlCalzone:
> The `setPowerlevel` and `getPowerlevel` methods probably don't need an event.

No, they don't.

As for the proposed API: How often does this CC result in updates? Firmware updates contains thousands of fragments and we emit updates after each one, so the event approach makes sense.
If the Powerlevel updates are limited to one or two it could make sense to have it all in one place, for example with a method that accepts an update callback and returns a Promise that resolves when the process is over - similar to `driver.trySendCommandSupervised`.

AlCalzone:
Hey @Ikcelaks any plans to continue here?

AlCalzone (dismissed):
Approved with my amendments. If a node sends a final unsolicited report of the test, it will be emitted as a `"notification"` event.

AlCalzone (approved):

PR #1841, opened 2021-02-21
feat(config): add Leviton 4 Speed Fan Controller zw4sf

akamel001:
- add zw4sf Leviton 4 Speed Fan Controller config


packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,112 @@
...
+			"maxValue": 99,
+			"defaultValue": 0,
+			"unsigned": true,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true
+		},
+		"6": {
+			"label": "LED Dim Level Indicator Timeout",
+			"description": "The length of time in seconds that the level indicators should stay illuminated after the dimming level is changed.  A value of 0 means the indicator is always off, and a value of 255 means the indicators are always on.",
```
AlCalzone:
```suggestion
			"description": "How long the level indicators should stay illuminated after the dimming level is changed.",
```
The unit "seconds" should be included below. The values 0 and 255 are explained in "options"

packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,112 @@
...
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 255,
+			"defaultValue": 3,
+			"unsigned": true,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": false,
```
AlCalzone:
```suggestion
			"allowManualEntry": true,
```
There are other values allowed besides 0 and 255

packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,112 @@
...
+				},
+				{
+					"label": "Level Indicators Always On",
+					"value": 255
+				}
+			]
+		},
+		"7": {
+			"label": "Locator LED Status",
+			"description": "Change the mode of the LED to Status Mode where the LED is illuminated when the load is On, Locator Mode where the LED is illuminated when the Load is Off, or turns off the LED completely for any other setting values.",
```
AlCalzone:
```suggestion
			"description": "Status Mode: The LED is illuminated when the load is On. Locator Mode: The LED is illuminated when the Load is Off.",
```

AlCalzone (changes requested):

packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,112 @@
...
+			"maxValue": 99,
+			"defaultValue": 99,
+			"unsigned": true,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true
+		},
+		"5": {
+			"label": "Preset Level",
+			"description": "Preset Level for On",
```
blhoward2: Get rid of unnecessary descriptions. These labels can be made as descriptive so you can get rid of them. Preset Level is descriptive enough. Min Level and Max Level is also pretty obvious.

packages/config/config/devices/0x001d/zw4sf.json
blhoward2: I suggest losing the desciption altogether. Change this to LED Indicator for the label. The options can be made more descriptive below.

packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,112 @@
...
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "LED Off",
+					"value": 0
+				},
+				{
+					"label": "Status Mode",
```
blhoward2: Status Mode means "LED on when switch is on" Change to that.

packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,112 @@
...
+					"value": 254
+				},
+				{
+					"label": "Locator Mode",
+					"value": 255
```
blhoward2, line 97: Locator mode means "LED on when switch is off". Change to that.

packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,112 @@
...
+			"minValue": 0,
+			"maxValue": 255,
+			"defaultValue": 255,
+			"unsigned": true,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "LED Off",
```
blhoward2: Change this to Always off

packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,112 @@
+// Leviton ZW4SF
+// 4 Speed Fan Controller
+{
+	"manufacturer": "Leviton",
+	"manufacturerId": "0x001d",
+	"label": "ZW4SF",
+	"description": "Decora Smart 4 Speed Fan Controller",
```
blhoward2: Lose "Decora Smart." If you really want to keep it, fix the comment above to match.

packages/config/config/devices/0x001d/zw4sf.json
akamel001: I accepted the edits to change the language. Perhaps that's a better route so it's consistent with other params that do include a description? Happy to change it if making the options more descriptive is cleaner!

blhoward2: Please do it my way so it matches what we're doing elsewhere.

packages/config/config/devices/0x001d/zw4sf.json
akamel001: Ah, i see what you mean. Fair enough. I'll remove the description. Seems redundant.

packages/config/config/devices/0x001d/zw4sf.json
akamel001: Sounds good, i'll add the `"unit": "seconds"` field

packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,110 @@
...
+			"label": "Lifeline",
+			"maxNodes": 5,
+			"isLifeline": true
+		}
+	},
+	"paramInformation": {
+		"3": {
+			"label": "Min Level",
+			"valueSize": 1,
+			"minValue": 0,
```
blhoward2, line 23: These should be unit "%". Carey through to the other 0-99 level ones. Put the unit directly under valueSize.

packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,110 @@
...
+			"writeOnly": false,
+			"allowManualEntry": true
+		},
+		"6": {
+			"label": "LED Dim Level Indicator Timeout",
+			"description": "How long the level indicators should stay illuminated after the dimming level is changed.",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 255,
+			"unit": "seconds",
```
blhoward2: Move the unit directly under valueSize. We're trying to standardize these.

packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,110 @@
...
+			"defaultValue": 10,
+			"unsigned": true,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true
+		},
+		"4": {
+			"label": "Max Level",
+			"valueSize": 1,
+			"minValue": 0,
```
blhoward2, line 34: Is 0 really a valid value for the max? It may be, just seems odd.

packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,110 @@
...
+			"maxValue": 255,
+			"unit": "seconds",
+			"defaultValue": 3,
+			"unsigned": true,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true,
+			"options": [
+				{
+					"label": "Level Indicators Always Off",
```
blhoward2: Change to Always off and Always on

blhoward2 (changes requested):
This is looking really good! Just a few minor nits left.

packages/config/config/devices/0x001d/zw4sf.json
akamel001: Had to double-check but it looks like it is in their [manual](https://www.leviton.com/en/docs/DG-000-ZW4SF-02A-W.pdf).

blhoward2 (dismissed):

packages/config/config/devices/0x001d/zw4sf.json
```diff
@@ -0,0 +1,109 @@
...
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 5,
+			"isLifeline": true
+		}
+	},
```
AlCalzone: Is this really necessary? The device supports Z-Wave+, so group 1 is always the lifeline.

AlCalzone (changes requested):
Is this really necessary? The device supports Z-Wave+, so group 1 is always the lifeline.

packages/config/config/devices/0x001d/zw4sf.json
akamel001: That's true, it's not necessary. I removed it.

blhoward2 (approved):

AlCalzone (approved):

PR #1871, opened 2021-02-23
fix: only update CCs from NIF during NodeInfo interview stage

### LINKED ISSUE #1865: WD200+ sometimes doesn't respond with updated values after turning on.
**Describe the bug**

ZWaveJS doesn't seem to consistently send a "value updated" event after turning on a Homeseer WD200+.
The WD100+ doesn't seem to suffer from this problem, sending the "value updated" event after 5 seconds (which I understand is due to the following code:

zwave-js/node-zwave-js@6cd0dfa#diff-9397c8948a06d7061964655d80272f34166e53676c4642a76c3a7799b61e4360R322

which only repolls a device if it doesn't support supervision.

The WD200+ is known not to properly support supervision, so zwavejs should treat it as such.
Despite #1531, which should tell zwave-js to do exactly that, the problem still persists.

**Device information**

Homeseer WD200+

**Did you change anything?**
- [ X] **No**

- [ X] **Yes**, in: *(specify application with versions)
  It seems like it sometimes works, but it seems independent of versions.

- [X ] `zwavejs2mqtt` (latest) docker image
- [X ] `HomeAssistant` version core-2021.2.3

**To Reproduce**

You can't reproduce without the device itself but:

Use home assistant `light.turn_on` service on the light associated with the node.
The light turns on (to last dim level), but the state in home assistant flips back to off and never returns to on.

**Additional context**

**Logfile:**

This starts from setting the debug level to `DEBUG` and then after all nodes are interviewed, performing a reinterview on Node 8 which is one of the ones that exhibits the problem.

Please note that after doing all of this and checking the node again, *it's not exhibiting the issue* right now. I will ping back if I can get it back in the state that it's problematic.

[zwavejs_357 4.log](https://github.com/zwave-js/node-zwave-js/files/6029340/zwavejs_357.4.log)

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/issues/1865

### DIFF
--- a/packages/zwave-js/src/lib/node/Node.test.ts
+++ b/packages/zwave-js/src/lib/node/Node.test.ts
@@ -138,7 +138,7 @@ describe("lib/node/Node", () => {
 				controlled: CommandClasses[];
 			}[] = [
 				{
-					supported: [CommandClasses["Anti-theft"]],
+					supported: [CommandClasses["Anti-Theft"]],
 					controlled: [CommandClasses.Basic],
 				},
 			];
@@ -697,7 +697,7 @@ describe("lib/node/Node", () => {
 
 		it("ignores the data in an NIF if it was received already", () => {
 			const node = makeNode();
-			node.updateNodeInfo(emptyNodeInfo as any);
+			node.interviewStage = InterviewStage.Complete;
 			node.updateNodeInfo({
 				controlledCCs: [CommandClasses.Configuration],
 				supportedCCs: [CommandClasses.Battery],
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot pack this

continuousmining:
Call me crazy, I dont think this was ever fixed.  WD-100 and WD-200's still don't work, the way this fix was supposed to fix them.

AlCalzone:
If there is still a problem, please open a new issue and fill out the issue template.

PR #1872, opened 2021-02-23
fix(config): actually load unit from device config files

### LINKED ISSUE #1866: Configuration metadata does not include units from config files

AlCalzone:
fixes: #1866


robertsLando (approved):
LGTM

PR #1873, opened 2021-02-23
refactor(cc): do not expose deltaTime and previousValue as values

AlCalzone:
They don't seem to ever contain anything useful.


blhoward2:
Thank God! I just assumed they were useful somewhere.

PR #1874, opened 2021-02-23
feat(config): add Zooz ZSE29 configuration parameters

NYZack:


NYZack:
@zwave-js-bot fix lint

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,97 @@
...
+			"maxNodes": 4
+		}
+	},
+	"paramInformation": {
+		"1": {
+			"label": "Basic Set Report Value Sent to Associated Devices",
+			"description": "Set the value of basic set report sent to the light associated with the sensor in Group 2 (so the light turns on to the selected brightness level). 99 is full brightness.",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 99,
```
AlCalzone, line 37: Basic Set also supports 255 for "on with last value". Is that possible with this device?

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,97 @@
...
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Disabled",
+					"value": 0
+				},
+				{
+					"label": "Enabled (default)",
```
AlCalzone:
```suggestion
					"label": "Enabled",
```

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,97 @@
...
+			"valueSize": 1,
+			"minValue": 1,
+			"maxValue": 10,
+			"defaultValue": 10,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true
+		},
+		"4": {
+			"label": "Lux Level Trigger",
```
AlCalzone: Sounds like this should be called `Lux Level Threshold`?

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,97 @@
...
+			"description": "Set lux level trigger with 10 being the lowest reported value.  The sensor will report motion only if lux level is below the set value. 0=set manually; 1=ignore lux.",
```
AlCalzone: > . 0=set manually; 1=ignore lux.

I think these should be options

AlCalzone (changes requested):

AlCalzone:
Also requesting review from @blhoward2 - master of consistency

packages/config/config/devices/0x027a/zse29.json
NYZack: I'm not sure.  This capability isn't mentioned in the Open Z-wave configuration file or a Smartthings device handler [https://github.com/krlaframboise/SmartThings/blob/master/devicetypes/krlaframboise/zooz-outdoor-motion-sensor-2-0.src/zooz-outdoor-motion-sensor-2-0.groovy ](url)
Both mention a range of 0-99 only.

AlCalzone: Okay lets stay on 0-99 then.

packages/config/config/devices/0x027a/zse29.json
NYZack: I don't think this is do-able, since allowManualEntry has to be true.  User can select 0 (set manually), 1 (ignore lux), or 10-900 for the trigger (threshold) level.  (Actually unclear what happens if user selects 2-9 - probably deactivates motion sensing.)

packages/config/config/devices/0x027a/zse29.json
NYZack: Changed.

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,96 @@
...
+				}
+			]
+		},
+		"3": {
+			"label": "Motion Sensitivity",
+			"description": "Adjust motion sensitivity where 10 is the most sensitive setting.",
+			"valueSize": 1,
+			"minValue": 1,
+			"maxValue": 10,
+			"defaultValue": 10,
```
blhoward2: The manual doesn't specify a default value. Where did you get 10? Generally, setting the maximum sensitivity is a bad idea for PIR devices. Usually they default to the medium range.

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,96 @@
...
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true
+		},
+		"4": {
+			"label": "Lux Level Threshold",
```
blhoward2: Rename this Brightness Threshold. Change the description to make clear this sends an alert when it falls below the set lux level.

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,96 @@
...
+			"minValue": 0,
+			"maxValue": 900,
+			"defaultValue": 0,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true
+		},
+		"5": {
+			"label": "Motion Re-trigger Time",
+			"description": "Delay before the sensor reports no motion after detecting the last motion activity. 0=set manually.",
```
blhoward2: "set manually" is unclear. Change to "0: Defined by knob on sensor

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,96 @@
...
+			"label": "Lux Reporting Frequency",
+			"unit": "minutes",
+			"valueSize": 2,
+			"minValue": 1,
+			"maxValue": 1440,
+			"defaultValue": 30,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true
+		}
```
blhoward2, line 167: You're missing parameters. Parameter 7 is a read-only report of the motion knob. Parameter 8 is a read-only report of the physical knob setting. https://www.support.getzooz.com/kb/article/321-zse29-outdoor-motion-sensor-ver-2-0-advanced-settings/

You need to make some of these parameters as conditional on the FW version, using the new conditional configuration setting. https://www.support.getzooz.com/kb/article/307-zse29%C2%A0outdoor-motion-sensor-change-log/

Conditional values are described here. Don't worry about templating, I'll handle that. https://zwave-js.github.io/node-zwave-js/#/development/config-files

blhoward2 (changes requested):

packages/config/config/devices/0x027a/zse29.json
NYZack: Got this from the OZW config file, but I don't see this mentioned in the reference link you sent from Zooz (thanks for that).  I'm not sure what this is from personal experience, so I'll remove the default value.

packages/config/config/devices/0x027a/zse29.json
AlCalzone: Yeah, it is. Options may exist always, but if `allowManualEntry` is `true`, they **must** exist.

packages/config/config/devices/0x027a/zse29.json
NYZack: Done.

packages/config/config/devices/0x027a/zse29.json
NYZack: Done.

packages/config/config/devices/0x027a/zse29.json
NYZack: Sorry, I see I can't remove the default value.  I'm keeping it at 10 (unless you recommend an arbitrary value), since the only independent info I have is the OZW config file.

NYZack: BTW @AlCalzone I think this points to an undesired aspect of writing these config files.  For read-only parameters, defaultValue must still be specified.  But I don't think defaultValue makes sense in this context.  I had to pick arbitrary defaultValue settings for the read-only parameters @blhoward2 pointed out (otherwise, I failed the lint check); but these values just return physical settings on the device.

packages/config/config/devices/0x027a/zse29.json
NYZack: Added the read-only parameters.  Will have to work on conditional FW templating later.  (Can I place the individual conditional parameters right in the middle of the parameter specifications, or do I need to completely reproduce the whole parameter set for each firmware version?)

packages/config/config/devices/0x027a/zse29.json
NYZack: OK, I added the options and included some additional ones to be in concordance with the read-only parameters 7 and 8 (i.e., to match what can physically be set by the knobs on the sensor).

packages/config/config/devices/0x027a/zse29.json
AlCalzone: You can just add conditions for individual parameters. If a parameter changes between versions, you need to use an array of objects for that one parameter instead.

NYZack:
@zwave-js-bot fix lint

packages/config/config/devices/0x027a/zse29.json
AlCalzone: Good point. I'll remove this necessity.

packages/config/config/devices/0x027a/zse29.json
NYZack: I limited the parameters to FW >= 2.0.  I started to include conditional parameters for FW 1.3, but it's unclear exactly how they're implemented prior to 2.0 (the parameters are described in the documentation, but I'm not sure whether the numbering was the same).  Parameters prior to 2.0 were quite limited in any event.

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -119,7 +119,15 @@
 		"5": {
 			"$if": "firmwareVersion >= 1.03",
 			"label": "Motion Re-trigger Time",
-			"description": "Delay before the sensor reports no motion after detecting the last motion activity.",
+			"description": [
+				{
+					"$if": "firmwareVersion < 2.00",
```
AlCalzone: That doesn't work, you'll have to repeat the entire parameter if you want to change the description.
Also, firmware versions have no leading zeros, even if manufacturers call them that way.

AlCalzone (changes requested):

packages/config/config/devices/0x027a/zse29.json
NYZack: I'll change the firmware reference.  I think you're looking at an outdated version.  I've just limited all of these parameters to firmware >= 2, as mentioned above.

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,244 @@
...
+			"isLifeline": true
+		},
+		"2": {
+			"label": "PIR Control",
+			"maxNodes": 4
+		}
+	},
+	"paramInformation": {
+		"1": {
+			"$if": "firmwareVersion >= 2",
```
AlCalzone:
```suggestion
			"$if": "firmwareVersion >= 2.0",
```
Same for the others

AlCalzone (changes requested):

packages/config/config/devices/0x027a/zse29.json
NYZack: All changed already.

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,244 @@
...
+			"label": "Manual Motion Re-trigger Time Setting",
+			"description": "Read the value for the Motion Re-trigger Time set by knob on sensor (overridden by setting Motion Re-trigger Time parameter).",
+			"unit": "seconds",
+			"valueSize": 2,
+			"minValue": 5,
+			"maxValue": 720,
+			"defaultValue": 5,
+			"readOnly": true,
+			"writeOnly": false,
+			"allowManualEntry": false,
```
blhoward2: For read-only parameters, please get rid of the options unless it adds a label. So here keep only the test mode one, but change the label to "Test mode".

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,244 @@
...
+			"description": "Read the value for the Brightness Threshold set by knob on sensor (overridden by setting Brightness Threshold parameter).",
+			"unit": "lux",
+			"valueSize": 2,
+			"minValue": 10,
+			"maxValue": 900,
+			"defaultValue": 900,
+			"readOnly": true,
+			"writeOnly": false,
+			"allowManualEntry": false,
+			"options": [
```
blhoward2, line 161: Same for this one

blhoward2 (changes requested):
Some minor changes after discussing it with @AlCalzone

packages/config/config/devices/0x027a/zse29.json
NYZack: I think it's important to include the number (of seconds, for instance) in the label with "Test Mode".  "Test Mode" just specifies that the device is in test mode (obviously), but it also means that the re-trigger time is 5 seconds.

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,244 @@
...
+					"value": 300
+				},
+				{
+					"label": "900 (Test mode)",
+					"value": 900
+				}
+			]
+		}
+	},
 	"metadata": {
```
NYZack, line 169: @AlCalzone With regard to the metadata, this is quoted directly out of the manual and is clearly not specific for the zwavejs2mqtt interface.  I see that the interface allows the user to display the metadata.  Should I edit it to refer to this app?  Or leave it alone?  (For example, instead of saying "Initiate inclusion (pairing) in the app (or web interface)..." it could say something like "Click Add/Remove Device ...")

blhoward2: You can change this one I suppose, but we have around 2,400 files. That's why we just pull them from the manual.

NYZack: Happy to leave it as is; probably not a bad thing for user to see the direct quote from the manual. Wanted to make sure I wasn't leaving anything out.

packages/config/config/devices/0x027a/zse29.json
blhoward2: You missed the parameter above this with the same options. 

If it's important at least make it clearer. "Test mode (Retrigger for 5 seconds)" or something. No one will understand what the 5 means.

NYZack: Not sure if you were looking at the most-recent file; I removed all options except for the "Test mode" option for each read-only parameter.  I *did* keep in the corresponding numerical options for parameters 4 and 5 (although the user can enter them directly), because these correspond to the options that can be set by the knobs on the device.  It's a style point; happy to remove them.

NYZack: Aaaah, strike that.  Knocked out the numerical options; also added a change with how parameter 5 is defined based on the Zooz documentation.

packages/config/config/devices/0x027a/zse29.json
NYZack: So I had a chance to mess with this some, and here is what I discovered (on version 1.3.0 of zwavejs2mqtt):  If I include options and allowManualEntry is true, the options don't show up in the interface:  as far as I could tell, I could *only* enter the value manually, and there was no dropdown.  

Also (with regard to parameters 7 and 8), if I have options set for a read-only parameter (and allowManualEntry is set to true) - if the actual value isn't one of the options, the value doesn't show up.

Filing a bug report on zwavejs2mqtt.  With regard to the first paragraph, it looks to me (when I export the node) like the options data from zwavejs doesn't get picked up in zwavejs2mqtt.

packages/config/config/devices/0x027a/zse29.json
```diff
@@ -16,6 +16,204 @@
...
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 99,
+			"defaultValue": 99,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true
+		},
+		"2": {
+			"$if": "firmwareVersion >= 2",
```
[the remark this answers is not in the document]
AlCalzone: Yes, like I reviewed above :)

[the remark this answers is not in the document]
AlCalzone: @NYZack didn't you say you changed this?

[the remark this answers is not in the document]
NYZack: @AlCalzone and @blhoward2 Not sure I understand. If you're asking why I didn't include conditional parameters for firmware < 2.0, it's because I can't find documentation of the precise details (looked in OZW, Smartthings, OpenHab).  I initially included some conditional parameters for firmware 1.03, but I wasn't confident that the details were correct and thought it better to leave them out.

Basically, there was a hardware version 1, with firmware 1.0 - this had no configuration parameters.  
HW version 1 was updated to FW 1.03 that included some configuration parameters, but I can't find documentation of the details of these (in particular, what the parameter numbers were).
HW version 2 included firmware versions 2.0 and 2.01 (which  have the same configuration parameters as each other) - these are well-documented and included here.

Sorry for the million commits. This is new to me.  (And sorry if I'm not understanding your question.)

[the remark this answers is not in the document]
AlCalzone:
```suggestion
			"$if": "firmwareVersion >= 2.0",
```
I meant this ^

No worries about the commits, this is getting squash-merged anyways.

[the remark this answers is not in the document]
NYZack: That was easy.  Done.

AlCalzone:
@blhoward2 anything to add?

blhoward2 (approved):

blhoward2:
LGTM

AlCalzone (approved):

PR #1880, opened 2021-02-24
feat: add Zooz ZEN17 and ZEN32

blhoward2:
Our first pre-release add. I'm awaiting an answer about one parameter that is being added to a new FW revision.


blhoward2:
@AlCalzone These are ready. Let's import them as is so I can test them and I'll template them when I do the rest of the Zooz devices.

packages/config/config/devices/0x027a/zen17.json
```diff
@@ -0,0 +1,428 @@
...
+				},
+				{
+					"label": "Hours",
+					"value": 2
+				}
+			]
+		},
+		"17": {
+			"label": "Auto Turn-Off Timer Unit (Relay 2)",
+			"description": "Choose between second, minutes, and hours as the unit for Auto Turn-Off time for Relay 2.  Default: minutes",
```
AlCalzone, line 356:
```suggestion
			"description": "Choose between second, minutes, and hours as the unit for Auto Turn-Off time for Relay 2.",
```

AlCalzone (changes requested):

packages/config/config/devices/0x027a/zen17.json
AlCalzone: You can do that while templating

PR #1894, opened 2021-02-24
fix(cc): emit value events when updating color components from hex

### LINKED ISSUE #1885: `hexColor` value change doesn't update single colors currentValues
> @AlCalzone oh - I have `noEvent` set so the value is just updated in the value DB but not emitted

AlCalzone:
fixes: #1885

@robertsLando, can you test?


robertsLando:
One moment

robertsLando:
Confirmed working :ok_hand:

PR #1911, opened 2021-02-27
fix: recreate transports, reconfigure loggers when updating log config

### LINKED ISSUE #1908: Driver.updateLogConfig changes the LogConfig but that doesn't actually change the logging
**Checklist:**

- [x] I am **not** using ZWaveJS2MQTT. **Or:** a developer has told me to come here.

**Describe the bug**

I have attempted to update the log config to set the level to `debug` via a Home Assistant websocket command. When I get the log config, it shows that the change is applied, but I don't see debug statements in the Z-Wave JS server addon logs when the change is applied.

The context of the request to re-add this feature and to report this is that we want to add functionality to the HA UI that will make it easier for us (and you) to get logs from users

**Device information**

Which device(s) is/are affected (make/model)?  N/A
What are the node IDs? N/A

**Did you change anything?**
- [ ] **Yes**: *(please describe)*
- [x] **No**

- [x] **No**, it never worked anywhere
- [ ] **Yes**, in: *(specify application with versions)*
    
- [ ] `zwavejs2mqtt` Manual Docker build
    - `node-zwave-js` branch: 
    - `zwavejs2mqtt` branch: 
- [ ] ioBroker.zwave2 adapter
- [x] `HomeAssistant` version 2021.3.0b2 (using Z-Wave JS server 0.1.10 which has zwave-js-server 1.1.0 embedded
- [ ] Pkg
- [ ] Manual Docker build
    - `node-zwave-js` branch: 
    - `zwavejs2mqtt` branch: 
- [ ] Other: 

**To Reproduce**

2. Click on '...'
3. Scroll down to '...'

**Additional context**

**Logfile:**

```console
$0.hass.callWS({type: "zwave_js/update_log_config", entry_id: "blah", config: {level: "debug"}}).then(response =>
  console.log("response: ", response));
Promise {<pending>}
VM2316:2 response:  null
$0.hass.callWS({type: "zwave_js/get_log_config", entry_id: "blah"}).then(response =>
  console.log("response: ", response));
Promise {<pending>}
VM2322:2 response:  {enabled: true, level: 5, log_to_file: false, filename: "/usr/lib/node_modules/@zwave-js/server/dist/bin/zwave-248.log", force_console: true}
```

AlCalzone:
fixes: #1908

### DIFF
--- a/packages/core/src/log/shared.ts
+++ b/packages/core/src/log/shared.ts
@@ -5,6 +5,10 @@ import * as path from "path";
 import { configs, MESSAGE } from "triple-beam";
 import winston, { Logger } from "winston";
 import type Transport from "winston-transport";
+import type {
+	ConsoleTransportInstance,
+	FileTransportInstance,
+} from "winston/lib/winston/transports";
 import { colorizer } from "./Colorizer";
 
 const { combine, timestamp, label } = winston.format;
@@ -84,7 +88,7 @@ export class ZWaveLoggerBase {
 
 export interface LogConfig {
 	enabled: boolean;
-	level: number;
+	level: string;
 	transports: Transport[];
 	logToFile: boolean;
 	nodeFilter?: number[];
@@ -101,13 +105,13 @@ function stringToNodeList(nodes?: string): number[] | undefined {
 }
 
 export class ZWaveLogContainer extends winston.Container {
-	private fileTransport: Transport | undefined;
-	private consoleTransport: Transport | undefined;
+	private fileTransport: FileTransportInstance | undefined;
+	private consoleTransport: ConsoleTransportInstance | undefined;
 	private loglevelVisibleCache = new Map<string, boolean>();
 
 	private logConfig: LogConfig = {
 		enabled: true,
-		level: getTransportLoglevelNumeric(),
+		level: getTransportLoglevel(),
 		logToFile: !!process.env.LOGTOFILE,
 		nodeFilter: stringToNodeList(process.env.LOG_NODES),
 		transports: undefined as any,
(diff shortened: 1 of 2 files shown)


packages/core/src/log/shared.ts
```diff
@@ -137,19 +141,40 @@ export class ZWaveLogContainer extends winston.Container {
 	}
 
 	public updateConfiguration(config: DeepPartial<LogConfig>): void {
+		const changedLoggingTarget =
+			config.logToFile != undefined &&
+			config.logToFile !== this.logConfig.logToFile;
+
+		const changedLogLevel =
+			config.level != undefined && config.level !== this.logConfig.level;
```
robertsLando, line 152: Is the  `!=` used on purpose instead of `!==` to catch also `null`?

packages/core/src/log/shared.ts
```diff
@@ -137,19 +141,40 @@ export class ZWaveLogContainer extends winston.Container {
...
+
+		const changedFilename =
+			config.filename != undefined &&
```
robertsLando, line 155: Same?

packages/core/src/log/shared.ts
AlCalzone: Yep, I always do that.

packages/core/src/log/shared.ts
AlCalzone: Dito

robertsLando:
The reason to use string loglevels ? More user friendly? BTW I would keep the fallback to numeric to don't introduce breaking changes

AlCalzone:
Winston transports only accept string log levels and I find it extremely awkward to look up what the numeric log level is if I want to set it to "debug". Especially if zwave-js has to reverse the lookup do do anything with the log level.

robertsLando (dismissed):

AlCalzone:
@robertsLando does this break your code?

robertsLando:
> @robertsLando does this break your code?

Sorry @AlCalzone I may have missed this! BTW everything seems ok

AlCalzone:
Its marked as breaking anyways, so you'll have time to test out v7

PR #1913, opened 2021-02-27
fix(config): add additional product ID to Fibaro Roller Shutter 3

Ledfoot775:
Product ID is incorrect at "0x1000", needs to be "0x3000" for correct automated identification


Ledfoot775:
Updated title to make it more specific!

AlCalzone (approved):

PR #1914, opened 2021-02-27
feat(config): add associations for Aeotec ZW130 Wallmote Quad

larstobi:
https://products.z-wavealliance.org/products/2153/assoc?noFilename=True

### DIFF
--- a/packages/config/config/devices/0x0086/zw130.json
+++ b/packages/config/config/devices/0x0086/zw130.json
@@ -24,6 +24,45 @@
 		"max": "255.255"
 	},
 	"supportsZWavePlus": true,
+	"associations": {
+		"1": {
+			"label": "Z-Wave Plus Lifeline. When touching the Button 1 to 4, the Central Scene Notificcation command will be sent to the associated node (1) in Association group 1.",
+			"maxNodes": 1,
+			"isLifeline": true,
+		},
+		"2": {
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
Labels are a bit verbose.
Something like "Button 1: Basic Set" etc. would be better.

blhoward2:
We've been doing these as "Button 2 (Basic Set)" and "Button 2 (Multilevel)" if memory serves.

blhoward2:
Also, I'm planning to template the Aeotec ones some tomorrow so please try to fix this so @AlCalzone can merge it ASAP. Otherwise it's going to create a conflict.

blhoward2:
I've merged this into my upstream work, so I'm going to close it here and it'll merge with my template push.

larstobi:
Sounds good, @blhoward2 ! Looking forward to it. I've received an explanation from Aeotec support saying that the sliding function for the WallMote Quad only works when using direct associations. Groups 3,5,7,9 control the respective assigned device.

larstobi:
@blhoward2 What's the status here ? Maybe we could just merge this, and you could fix the conflict in your code? It shouldn't be too much of a conflict, though, should it?

AlCalzone:
Last time I talked to @blhoward2 he said that he wants to get Aeotec done over the weekend.

PR #1915, opened 2021-02-27
feat(config): improve Leviton dzpd3 parameter metadata and add device metadata

NYZack:
Include units and metadata, and improve descriptions


packages/config/config/devices/0x001d/dzpd3.json
```diff
@@ -30,30 +30,45 @@
 	"paramInformation": {
 		"1": {
 			"label": "Fade On Time",
-			"description": "Time it takes to fade on",
+			"description": "Time it takes to fade on; units are seconds when value is 1-127, and minutes (subtracting 127) when value is 128-253.",
+			"unit": "seconds or -127 minutes",
```
blhoward2: Just omit the unit for these crazy ones.

packages/config/config/devices/0x001d/dzpd3.json
```diff
@@ -30,30 +30,45 @@
...
+			"options": [
+				{
+					"label": "Instant off",
+					"value": 0
+				}
+			]
 		},
 		"3": {
 			"label": "Minimum Light Level",
+			"unit": "%",
```
blhoward2: Please move all units to directly underneath valueSize.

packages/config/config/devices/0x001d/dzpd3.json
```diff
@@ -74,7 +90,6 @@
 		},
 		"7": {
 			"label": "Locator LED Status",
```
blhoward2, line 76: Change this to LED Indicator. And add the options. Status mode is LED on when switch is on. Locator mode is LED on when switch if off. Off is Always Off.

blhoward2 (changes requested):
Please see my comments. We're trying to make these very consistent. I appreciate the attempt to help but these are going to be templated anyway soon so I'd focus only on correcting actual errors.

NYZack (commented):
All changes made

packages/config/config/devices/0x001d/dzpd3.json
```diff
@@ -30,31 +30,44 @@
 	"paramInformation": {
 		"1": {
 			"label": "Fade On Time",
-			"description": "Time it takes to fade on",
+			"description": "Time it takes to fade on; units are seconds when value is 1-127, and minutes (subtracting 127) when value is 128-253.",
```
blhoward2, line 33: Sorry, this is actually info we're trying to remove form descriptions. We don't describe ranges usually. And the unit can in a unit field. I don't think you need a description at all here.

packages/config/config/devices/0x001d/dzpd3.json
```diff
@@ -30,31 +30,44 @@
...
+				{
+					"label": "Instant on",
+					"value": 0
+				}
+			]
 		},
 		"2": {
 			"label": "Fade off time",
-			"description": "Time it takes to fade off",
+			"description": "Time it takes to fade off; units are seconds when value is 1-127, and minutes (subtracting 127) when value is 128-253.",
```
blhoward2, line 51: Same here

packages/config/config/devices/0x001d/dzpd3.json
```diff
@@ -85,15 +98,15 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "LED Off",
+					"label": "Off",
```
blhoward2: Please make this Always off for consistency to what we have elsewhere.

blhoward2 (changes requested):
A few more minor changes. Almost there.

packages/config/config/devices/0x001d/dzpd3.json
NYZack: I think you suggested otherwise above, and, without a description or units, it's impossible to understand:
image

packages/config/config/devices/0x001d/dzpd3.json
NYZack: (See above)

packages/config/config/devices/0x001d/dzpd3.json
NYZack: Done.

packages/config/config/devices/0x001d/dzpd3.json
blhoward2: Ah, disregard. Missed the changing unit. Go back to how you had it. Apologies.

AlCalzone: Funny how they sometimes come so close to using a spec-defined encoding, but fail on the last steps:
grafik

packages/config/config/devices/0x001d/dzpd3.json
```diff
@@ -123,5 +136,10 @@
...
+		"exclusion": "• Enter Programming Mode by holding the control button for 7 seconds; the Locator LED will turn amber.\n1. • Follow directions in the Z-Wave controller to enter exclusion mode.\n2. • Tap the control button one time. The Locator LED will quickly flash green. The Z-Wave controller will exclude the Decora Smart device.\n3. • The Z-Wave controller will confirm successful exclusion from the network.",
```
AlCalzone: These seem to use both 1./2./3. for enumeration as well as bullet points. Please choose one.

AlCalzone (changes requested):

packages/config/config/devices/0x001d/dzpd3.json
NYZack: Done.

AlCalzone (approved):

PR #1917, opened 2021-02-27
feat(config): add Ring Keypad config

WizKid:
I don't have a EU version but I tested the US version.

Wasn't sure if I should leave out the EU version extra config params because I can't test them or do my best to support them anyway.

PR #1985, opened 2021-03-03
fix(cc): length validation in Security S0 Message Encapsulation

WizKid:
- To be honest I don't know if this is the right solution but it seems to work.
- I been trying to add the Ring Alarm Keypad and I noticed that node.firmwareVersion was undefined.
- After some debugging I realize that `await node.commandClasses.Version.get()` didn't work. It just failed every time.
  - Logs https://gist.github.com/WizKid/b5e62faa0a833916e4969830469f986e which shows that we fail to unwrap the message
- After this diff it instead looks like this:
  - https://gist.github.com/WizKid/0ab37503ee8b887e4ee77f48853fb0cf

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/SecurityCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SecurityCC.ts
@@ -496,9 +496,9 @@ export class SecurityCCCommandEncapsulation extends SecurityCC {
 		}
 
 		if (gotDeserializationOptions(options)) {
-			// HALF_NONCE_SIZE bytes iv, 1 byte frame control, 2 bytes CC header, 1 byte nonce id, 8 bytes auth code
+			// HALF_NONCE_SIZE bytes iv, 1 byte frame control, 1 byte CC header, 1 byte nonce id, 8 bytes auth code
 			validatePayload(
-				this.payload.length >= HALF_NONCE_SIZE + 1 + 2 + 1 + 8,
+				this.payload.length >= HALF_NONCE_SIZE + 1 + 1 + 1 + 8,
 			);
 			const iv = this.payload.slice(0, HALF_NONCE_SIZE);
 			const encryptedPayload = this.payload.slice(HALF_NONCE_SIZE, -9);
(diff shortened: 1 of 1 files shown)


AlCalzone:
That doesn't look correct. Can you send me your network key via E-Mail, so I can try to figure out what the underlying problem is?

marcus-j-davies:
@WizKid,

A little off-top.
but i am actually interested in adding the keypad to my network - does this not need the ring hub then?
what version of the keypad are you using?

WizKid:
> That doesn't look correct. Can you send me your network key via E-Mail, so I can try to figure out what the underlying problem is?

I totally agree it doesn't look right. I was really surprised that it made things work. This is just me testing things to the network key is just `Buffer.alloc(16, 1)`

> but i am actually interested in adding the keypad to my network - does this not need the ring hub then?
> what version of the keypad are you using?

Correct. The goal is to use the keypad directly with homeassistant skipping the ring base station. I have the Keypad 1st gen with firmware 2.14. There are more things that needs to happen like #1917 and #1838 . There also probably need to be some work done in home assistant but I haven't gotten that far yet.

marcus-j-davies:
Yeah - support is progressing I see. 👍 
The ring docs overly mention, that it needs the ring base station to work (I have no interest in their offerings)

But its nice to know it is indeed possible, once support is provided.

marcus-j-davies:
Ordered,

I can help do some testing if needed - arriving tomorrow (Thank you Amazon)

EDIT
I'm in the UK - so can test any EU specifics

AlCalzone:
@WizKid I take (almost) everything back. In a sequenced security encapsulated message, the 2nd message does not contain the two CC header bytes. So while the first message must include at least one of those two bytes, after the second one we just need to have a complete CC buffer. So the fix is almost correct.

AlCalzone (approved):

WizKid:
That makes sense. And I'm glad you figured out why it works :) I'll test it after work but I'm guessing my logs were enough for you to test it.

PR #1987, opened 2021-03-03
fix(config): make `allowManualEntry` optional, forbid explicitly specifying `true`

### LINKED ISSUE #1919: Config files: `allowManualEntry` should not exist for `readonly` params

AlCalzone:
This PR makes `allowManualEntry` optional. It now defaults to `true` which makes the most sense for most parameters and `false` when `readOnly` is `true`.
Furthermore, it is now an error to specify anything but `allowManualEntry: false`.

fixes: #1919


AlCalzone:
@blhoward2 this is probably going to conflict with what you're working on. I can hold this back until your Aeotec changes are merged.

blhoward2:
If you could. I'll try to have that wrapped up this weekend if not before.

AlCalzone:
@blhoward2

PR #2003, opened 2021-03-06
add additional product ID to Aeotec Nano Shutter ZW141

Ledfoot775:
Later firmware updates of the ZW141 have updated the product ID on the device - Needs to be reflected here for correct interview results.


AlCalzone:
Looks like we did have that device id in a different file (see error message). Can you verify that it is the same and merge them into one file?

Ledfoot775:
Ok - looks like I stuffed up twice. 

1. Aeon labs and aeotec - same company using two different manufacturer ID's. In the latest firmware update they changed the manufacturer ID. So I'll need to copy the file across to the right directory and clean up the Product ID's.
2. I used the product ID's for the Aeotec Zstick instead of the ZW 141 (facepalm).

I'll revert the file, do the copy etc and then resubmit the pull request.

Ledfoot775:
Ok - looks like I stuffed up twice. 

1. Aeon labs and aeotec - same company using two different manufacturer ID's. In the latest firmware update they changed the manufacturer ID. So I'll need to copy the file across to the right directory and clean up the Product ID's.
2. I used the product ID's for the Aeotec Zstick instead of the ZW 141 (facepalm).

I'll revert the file, do the copy etc and then resubmit the pull request.

Ledfoot775:
I'm also learning how to use github and commits etc - Be gentle...

Ledfoot775:
OK @AlCalzone - Let me know how this one goes...

blhoward2:
Aeotec alerted us there would be a new manufactuerId and I already have a file for it, but a lot more changes than just adding the ID. They told us they did this because it was essentially a new product. Are you sure you installed the correct firmware? I assumed there was a hardware change the way it was described. Maybe not but where did you get the fw from?

In any event, we need to add no less than 7 new parameters and modify other things. I'll handle this one as I already have the file built and it's being templated. I'm working directly with Aeotec to improve their files.

AlCalzone:
@Ledfoot775 thanks for the PR but after discussion with @blhoward2 we should fix this the correct way.

PR #2004, opened 2021-03-06
fix(config): update Qubino Smart Plug 16A, parameter 41 does not exist

ualex73:
It removes parameter (config) 41, and adds 42. Trying to retrieve parameter 41 seems to kill all Z-Wave communication to the Qubino Smart Plug (I need to physically unplug and replug to get it alive again).

### DIFF
--- a/packages/config/config/devices/0x0159/smart_plug_16a.json
+++ b/packages/config/config/devices/0x0159/smart_plug_16a.json
@@ -136,13 +136,13 @@
 			"writeOnly": false,
 			"allowManualEntry": true
 		},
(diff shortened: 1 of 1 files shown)


ualex73:
It is include in #1600, closing.

AlCalzone:
Unless @blhoward2 has started working on this, I'd be fine with changing this. In any case, the label/description should be improved (shorter, no unnecessary mention of the unit, no description that just repeats the label in other words).

ualex73:
ok, let me rebase and send the PR.

Issue #2002, opened 2021-03-05
Inovelli LZW30-SN missing Parameter 13


jayscovill:
**Which device is missing?**

Inovelli LZW30-SN configuration parameter 13

**What are the IDs?**

Manufacturer ID: 0x031e 
Product type: Where do I find this? 
Product ID: 0x0002:0x0001 

- [x ] It is not in the [configuration DB](https://devices.zwave-js.io/)
- [ ] It was not [merged recently](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Apr+is%3Aclosed) or has a [pending PR](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Aopen+is%3Apr)
- [ ] It is not in the [TODO list](https://github.com/zwave-js/node-zwave-js/issues/1600)

As of FW 1.17 Inovelli added a parameter 13 to specify load type but it is not in the configuration database for the device.

https://support.inovelli.com/portal/en/kb/articles/firmware-v1-17-beta-lzw30-sn-switch-red-series-gen-2

ehsman:
I'm guessing that the LZW30-SN might need multiple configs like LZW31-SN. As there is also a parameter 51 (same in the existing LZW31-SN config) added in firmware 1.19

https://support.inovelli.com/portal/en/kb/articles/firmware-v1-19-beta-lzw30-sn-switch-red-series-gen-2

AlCalzone:
FYI: We now have the capability to define config parameters conditionally for specific firmware ranges in one file.
https://zwave-js.github.io/node-zwave-js/#/development/config-files?id=conditional-configuration-settings

maverick1872:
@AlCalzone I assume a PR to update https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x031e/lzw30-sn.json with a conditional configuration on parameters 13 & 51 would be all that's needed?

AlCalzone:
I think so, yes.

AlCalzone:
fixed in https://github.com/zwave-js/node-zwave-js/pull/2025

PR #2025, opened 2021-03-09
feat(config): add config params 13 and 51 to Inovelli LZW30-SN

maverick1872:
These configuration changes add missing parameters for the Inovelli Red Series on/off switch (LZW30-SN) referenced in #2002.

PR will be marked as ready when functional testing is complete.

### DIFF
--- a/packages/config/config/devices/0x031e/lzw30-sn.json
+++ b/packages/config/config/devices/0x031e/lzw30-sn.json
@@ -328,6 +328,48 @@
 			"writeOnly": false,
 			"allowManualEntry": true
 		},
+		"13": {
+			"$if": "firmwareVersion >= 1.17",
+			"label": "",
+			"valueSize": 1,
+			"minValue": 0,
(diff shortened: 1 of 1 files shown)


AlCalzone (approved):

AlCalzone:
@blhoward2 anything to add regarding consistency?

blhoward2:
Nah. And if there is I'll get it when we template that folder.

maverick1872:
Curious if either of you could point me in the direction of the relationship of the `*.values.jsonl` file and those of paired nodes. Since I tested the above using the same configuration dir, after I moved back to my original docker container based on the published image it now lists param013 & param051 respectively. I assume it has something to do with the vue project within `zwavejs2mqtt`?

blhoward2:
I'd guess it's because you haven't reinterviewed. You'd lose it when you did as it reparses the config files. Right now it's loaded from what is already cached.

Issue #1713, opened 2021-02-10
[feat] Alarms mapping compat flag for locks (and more?)


lolouk44:
**Is your feature request related to a problem? Please describe.**
Home Assistant User.  I used to get extra information about the door lock (Yale Conexis L1) such as Locked by RF, Jammed, etc.
I can't see this info present in the logs, but I see the "id": "5-113-0-alarmType" block holds a value that would represent the lock state (found for Yale Conexis L1 so far):
9: Lock is jammed
21: Locked manually
22: Unlocked manually
24: Locked by zwave
25: Unlocked by zwave
27: Auto relock
144: Unlocked by RFID tag

**Describe the solution you'd like**
I haven't found an exhaustive list anywhere yet, ideally have that list exposed to HA, if not at least expose the values so a custom sensor can be created

**Describe alternatives you've considered**
none yet

**Additional context**
Some info here:
https://community.smartthings.com/t/yale-real-living-keyless-deadbolt-displays-incorrect-status/2300/4

AlCalzone:
The Notification CC (former Alarm CC) defines [which values mean what](https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/notifications.json). This information is passed on to applications.

Unfortunately, that was only added in version 2 of the CC and extended upon later. The first version (which your device uses) was not standardized in that regard, so 21 could mean "Locked manually" for your device, but "Flood Alarm" for another one. So a definition of these values would have to be done on a per-device basis and I frankly don't have the motivation nor the capacity to go through thousands of device configs and compare the manuals. Especially since it only affects legacy devices.

lolouk44:
Thanks for this info.
Is there a way to at least expose these values to Home Assistant?
I'm happy creating my own template sensors, but I at least need the values coming into home assistant, ideallly as an attribute.
Thanks

AlCalzone:
They should... originally reported in https://github.com/zwave-js/node-zwave-js/issues/1546, fixed in https://github.com/zwave-js/node-zwave-js/pull/1562

lolouk44:
Just realised I had 2 entitites that were set to unavailable by default and therefore did not show a state.
I have enabled them and I have created my template sensor.
![image](https://user-images.githubusercontent.com/4587746/107633603-36623c80-6c60-11eb-87bf-f7664a478a39.png)
Thanks a lot 👍 
For other Yale Connexis L1 users this is my template sensor:
[...]

PR #2040, opened 2021-03-11
feat: add config for YRD210 with incorrect manufacturerId

blhoward2:
Fixes @Juggler00's issue in #1713. Yale released some locks with an incorrect manufacturerId.

### DIFF
--- a/packages/config/config/devices/0x0109/yrd210.json
+++ b/packages/config/config/devices/0x0109/yrd210.json
@@ -56,48 +56,5 @@
 		"8": {
 			"$import": "../0x0129/templates/yale_template.json#operating_mode"
 		}
-	},
-	"compat": {
-		"alarmMapping": [
-			{
-				"$import": "../0x0129/templates/yale_template.json#alarm_map_keypad_limit"
-			},
-			{
-				"$import": "../0x0129/templates/yale_template.json#alarm_map_manual_unlock"
-			},
-			{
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0109/yrd210.json
```diff
@@ -0,0 +1,60 @@
+// Yale YRD210
+// Push Button Deadbolt
+{
+	"manufacturer": "Yale",
+	"manufacturerId": "0x0109",
```
AlCalzone: Could use a comment here that mentions the incorrect manufacturer ID

packages/config/config/devices/0x0109/yrd210.json
```diff
@@ -0,0 +1,60 @@
...
+			"productType": "0x0002",
+			"productId": "0x0000"
+		},
+		{
+			"productType": "0x0003",
+			"productId": "0x0000"
+		},
+		{
+			"productType": "0x0004",
+			"productId": "0x0000"
```
AlCalzone, line 23: Do they all use the same product ID?

AlCalzone (changes requested):

packages/config/config/devices/0x0109/yrd210.json
blhoward2: As far as I can tell, yes.

AlCalzone (approved):

PR #2044, opened 2021-03-11
fix(config): add compat flag to Ring Keypad

marcus-j-davies:
As titled.

### DIFF
--- a/packages/config/config/devices/0x0346/alarm_keypad_gen1.json
+++ b/packages/config/config/devices/0x0346/alarm_keypad_gen1.json
@@ -19,6 +19,9 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0346/alarm_keypad_gen1.json
```diff
@@ -19,6 +19,9 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
+	"compat": {
+		"disableStrictEntryControlDataValidation": true
+	},
```
AlCalzone: Please move the compat options below the `paramInformation` and before `metadata`

AlCalzone (changes requested):

AlCalzone (approved):

PR #2051, opened 2021-03-12
fix(config): cleanup Fibaro

WizKid:
I went over the Fibaro folder and cleaned up a bunch of stuff. Moved a bunch of stuff over to templates. Also used condition to merge many files into one.

There are still a bunch of more work to do. Lots more params can be templatized. But in this diff I'm renaming every single file so I thought it would probably be good to land something instead of going for perfection. And then more cleanup can be done after this.


packages/config/config/devices/0x010f/fgcd-001.json
```diff
@@ -140,40 +166,37 @@
...
+			"label": "Associations: Send secure commands to 4th group"
+		},
+		"7[0x08]": {
+			"$import": "templates/fibarogroup_templates.json#secure_assoc_bit",
+			"label": "Associations: Send secure commands to 5th group"
+		},
+		"7[0x0f]": {
+			"$import": "templates/fibarogroup_templates.json#secure_assoc_bit",
+			"label": "Associations: Send secure commands to 6th group"
+		},
```
AlCalzone: That's not correct. 0x0f is not a single bit, it selects the lower 4 bits.
This also means the follwing bitmask is incorrect.

packages/config/config/devices/0x010f/templates/fibarogroup_templates.json
```diff
@@ -0,0 +1,294 @@
...
+			{
+				"label": "Enable",
+				"value": 0
+			},
+			{
+				"label": "Disable",
+				"value": 1
+			}
+		]
+	},
```
AlCalzone: There should be already a definition for these in the master_template.json

packages/config/config/devices/0x010f/templates/fibarogroup_templates.json
```diff
@@ -0,0 +1,294 @@
...
+			},
+			{
+				"label": "ALL ON active / ALL OFF disabled",
+				"value": 2
+			},
+			{
+				"label": "ALL ON active / ALL OFF active",
+				"value": 255
+			}
+		]
```
AlCalzone:
```suggestion
		"options": [
			{
				"label": "ALL ON disabled / ALL OFF disabled",
				"value": 0
			},
			{
				"label": "ALL ON disabled / ALL OFF enabled",
				"value": 1
			},
			{
				"label": "ALL ON enabled / ALL OFF disabled",
				"value": 2
			},
			{
				"label": "ALL ON enabled / ALL OFF enabled",
				"value": 255
			}
		]
```
for consistency

packages/config/config/devices/0x010f/templates/fibarogroup_templates.json
```diff
@@ -0,0 +1,294 @@
...
+	},
+	"change_bi_stable_keys": {
+		"label": "Change [On-Off] bi-stable keys",
```
AlCalzone, line 37: Looks weird - is there a nicer way for this?

packages/config/config/devices/0x010f/templates/fibarogroup_templates.json
```diff
@@ -0,0 +1,294 @@
...
+				"value": 0
+			},
+			{
+				"label": "Device status depends on key status",
+				"value": 1
+			}
+		]
+	},
+	"enable_disable_double_click": {
+		"label": "Double-click set lighting at 100%",
```
AlCalzone:
```suggestion
		"label": "Set full brightness on double-click",
```
might be better?

packages/config/config/devices/0x010f/templates/fibarogroup_templates.json
```diff
@@ -0,0 +1,294 @@
...
+		"options": [
+			{
+				"label": "Disable double click",
+				"value": 0
+			},
+			{
+				"label": "Enable double click",
+				"value": 1
+			}
+		]
```
AlCalzone: Then these can be 
```suggestion
		"options": [
			{
				"label": "Disabled",
				"value": 0
			},
			{
				"label": "Enabled",
				"value": 1
			}
		]
```

packages/config/config/devices/0x010f/templates/fibarogroup_templates.json
```diff
@@ -0,0 +1,294 @@
...
+	},
+	"3way_switch": {
+		"label": "The function of 3-way switch",
+		"description": "Key no. 2 controls the Dimmer additionally",
```
AlCalzone:
```suggestion
		"label": "3-way switch function for S2",
```

packages/config/config/devices/0x010f/templates/fibarogroup_templates.json
```diff
@@ -0,0 +1,294 @@
...
+		"options": [
+			{
+				"label": "3-way switch function for S2 disabled",
+				"value": 0
+			},
+			{
+				"label": "3-way switch function for S2 enabled",
+				"value": 1
+			}
+		]
```
AlCalzone:
```suggestion
		"options": [
			{
				"label": "Disabled",
				"value": 0
			},
			{
				"label": "Enabled",
				"value": 1
			}
		]
```

AlCalzone: etc... please try to apply this pattern to as many params as possible (where it makes sense at least)

packages/config/config/devices/0x010f/templates/fibarogroup_templates.json
```diff
@@ -0,0 +1,294 @@
...
+		"options": [
+			{
+				"label": "No reaction",
+				"value": 0
+			},
+			{
+				"label": "Send an alarm frame",
+				"value": 1
+			}
+		]
```
AlCalzone, line 137: Depending on how the labels are, this could also be Enabled/Disabled

packages/config/config/devices/0x010f/templates/fibarogroup_templates.json
```diff
@@ -0,0 +1,294 @@
...
+		"minValue": 0,
+		"maxValue": 65535,
+		"defaultValue": 600,
+		"unsigned": true,
+		"readOnly": false,
+		"writeOnly": false,
+		"allowManualEntry": true
+	},
+	"saving_state_power_failure": {
+		"label": "Saving state before power failure",
```
AlCalzone: Should probably be 
```suggestion
		"label": "Restore state after power failure",
```

packages/config/config/devices/0x010f/templates/fibarogroup_templates.json
```diff
@@ -0,0 +1,294 @@
...
+			},
+			{
+				"label": "Outputs set to previous state upon power restore",
+				"value": 1
+			}
+		]
+	},
+	"alarm_cancelling": {
+		"label": "Alarm cancelling or turning a device off",
+		"description": "Allows for deactivating device turn off and alarm cancellation functions",
```
AlCalzone: This description is redundant

packages/config/config/devices/0x010f/templates/fibarogroup_templates.json
```diff
@@ -0,0 +1,294 @@
...
+				"label": "Alarm (flooding) cancellation inactive",
+				"value": 0
+			},
+			{
+				"label": "Alarm (flooding) cancellation active",
+				"value": 1
+			}
+		]
+	},
+	"key_press_brigthness_level": {
```
AlCalzone:
```suggestion
	"key_press_brightness_level": {
```

packages/config/config/devices/0x010f/templates/fgwpx-1x1.json
```diff
@@ -0,0 +1,489 @@
...
+			"unit": "seconds",
+			"minValue": 0,
+			"maxValue": 32400,
+			"defaultValue": 3600,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true,
+			"options": [
+				{
+					"label": "Disables reporting",
```
AlCalzone, line 97:
```suggestion
					"label": "Disable reporting",
```
Options should be formulated in an active way, not describe what they do.

packages/config/config/devices/0x010f/templates/fgwpx-1x1.json
```diff
@@ -0,0 +1,489 @@
...
+			"allowManualEntry": true
+		},
+		"23[0x01]": {
+			"$import": "fibarogroup_templates.json#enable_disable_bit_default_true",
+			"label": "Power Exceeded controlling 3rd association group devices"
+		},
+		"23[0x02]": {
+			"$import": "fibarogroup_templates.json#enable_disable_bit_default_true",
+			"label": "Power Dropped controlling 3rd association group devices"
+		},
```
AlCalzone: I don't understand these labels 😅

packages/config/config/devices/0x010f/templates/fgwpx-1x1.json
```diff
@@ -0,0 +1,489 @@
...
+				},
+				{
+					"label": "Last State/ON",
+					"value": 255
+				}
+			]
+		},
+		"26": {
+			"label": "Threshold Up",
+			"description": "Command frame sent to the devices in 3rd group",
```
AlCalzone, line 218: Don't understand this either - I think we need to give these labels a good look on your cleanup pass. Some of them are terrible.

packages/config/config/devices/0x010f/templates/fgwpx-1x1.json
```diff
@@ -0,0 +1,489 @@
...
+		"30[0x0f]": {
+			"$import": "fibarogroup_templates.json#enable_disable_bit_default_true",
+			"label": "React to High Temperature Alarm",
+			"description": "Network alarms to which the Wall Plug will respond"
+		},
+		"30[0x10]": {
+			"$import": "fibarogroup_templates.json#enable_disable_bit_default_true",
+			"label": "React to Flood Alarm",
+			"description": "Network alarms to which the Wall Plug will respond"
+		},
```
AlCalzone:
```suggestion
		"30[0x10]": {
			"$import": "fibarogroup_templates.json#enable_disable_bit_default_true",
			"label": "React to High Temperature Alarm",
			"description": "Network alarms to which the Wall Plug will respond"
		},
		"30[0x20]": {
			"$import": "fibarogroup_templates.json#enable_disable_bit_default_true",
			"label": "React to Flood Alarm",
			"description": "Network alarms to which the Wall Plug will respond"
		},
```
I guess this is what you meant.

packages/config/config/devices/0x010f/templates/fgwpx-1x1.json
```diff
@@ -0,0 +1,489 @@
...
+			"minValue": 1,
+			"maxValue": 32400,
+			"defaultValue": 600,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true
+		},
+		"40": {
+			"label": "Power load for violet color",
+			"description": "Maximum active power to flash the LED frame Violet",
```
AlCalzone, line 332: needs a unit

packages/config/config/devices/0x010f/templates/fgs-2xx.json
```diff
@@ -0,0 +1,302 @@
...
+			"label": "Switch 1",
+			"maxNodes": 16
+		},
+		"2": {
+			"label": "Switch 2",
+			"maxNodes": 16
+		},
+		"3": {
+			"label": "Controller Updates",
+			"description": "Reports state of devices.",
```
AlCalzone, line 20:
```suggestion
			"label": "Lifeline",
```
We should stay in Z-Wave slang :) Please check the other association definitions too.

packages/config/config/devices/0x010f/templates/fgs-2xx.json
```diff
@@ -0,0 +1,302 @@
...
+			"options": [
+				{
+					"label": "Manual override disabled",
+					"value": 0
+				},
+				{
+					"label": "Manual override enabled",
+					"value": 1
+				}
+			]
```
AlCalzone, line 48: The label does not seem to match the options

packages/config/config/devices/0x010f/templates/fgwdseu-221.json
```diff
@@ -37,17 +27,33 @@
...
-			"readOnly": false,
-			"writeOnly": false,
-			"allowManualEntry": true
-		},
+		"2": [
+			{
+				"$if": "productType === 0x1b01",
+				"label": "First channel – overload safety switch",
+				"valueSize": 4,
+				"unit": "0.1 watts",
```
AlCalzone:
```suggestion
				"unit": "0.1 W",
```

AlCalzone: same for the others

packages/config/config/devices/0x010f/templates/fgwdseu-221.json
```diff
@@ -173,16 +194,57 @@
...
+						"value": 2
+					},
+					{
+						"label": "1st button toggles 1st channel, 2nd button toggles 2nd channel",
+						"value": 3
+					},
+					{
+						"label": "1st button single click turns 1st channel ON, 1st button double click turns 2nd channel ON, 2nd button single click turns both channels OFF",
+						"value": 4
+					}
```
AlCalzone, line 224: These labels should be shortened.

packages/config/config/devices/0x010f/templates/fgwdseu-221.json
```diff
@@ -173,16 +194,57 @@
...
+				"allowManualEntry": false,
+				"options": [
+					{
+						"label": "1st and 2nd button toggle the load",
+						"value": 1
+					},
+					{
+						"label": "1st button turns the load ON, 2nd button turns the load OFF",
+						"value": 2
+					}
```
AlCalzone, line 244: dito

packages/config/config/devices/0x010f/templates/fgwdseu-221.json
```diff
@@ -204,8 +266,9 @@
 			]
 		},
 		"25": {
+			"$if": "productType === 0x1b01",
 			"label": "Outputs orientation",
-			"description": "Outputs orientation (Only in Double Switch Mode)",
+			"description": "Outputs orientation",
```
AlCalzone: descriptions that match the label should just be removed

packages/config/config/devices/0x010f/templates/fgs-2x4.json
```diff
@@ -24,6 +17,7 @@
 			"maxNodes": 5
 		},
 		"3": {
+			"$if": "productType == 0x0204",
```
AlCalzone, line 20: Does it even make sense to template this if all the params are just defined for one device?

AlCalzone (approved):
I gave the templates a more in-depth look, spot checks of the other files seem fine.

AlCalzone (changes requested):
Whoops wrong button

blhoward2:
@WizKid Earlier today we merged detailed documentation on device files, including a style guide and detailed guidance on templating. Could you please take a look at those, make any changes necessary to conform, and then let us know when you're ready for us to take a look line-by-line?

https://zwave-js.github.io/node-zwave-js/#/

packages/config/config/devices/0x010f/templates/fibarogroup_templates.json
WizKid: To be honest I don't really know what it is. But it was used in two files so I broke it out here so if we change it we update both.

packages/config/config/devices/0x010f/templates/fgs-2xx.json
WizKid: Existing parameter https://github.com/zwave-js/node-zwave-js/blob/910155c88ffb2f8ab6ddce6adc2c0dbb57a135f6/packages/config/config/devices/0x010f/fgs221_2.1-2.3.json#L101-L121

WizKid:
Fixed the main review comments.

Regarding strange labels and descriptions. Yes there are plenty but they are already there. The goal for me with this diff is not to make perfect labels and descriptions. Instead move things around to make it easier to fix those things later.

packages/config/config/devices/templates/master_template.json
```diff
@@ -43,7 +43,7 @@
 		"valueSize": 1,
 		"minValue": 0,
 		"maxValue": 1,
-		"defaultValue": 1,
+		"defaultValue": 0,
```
WizKid, line 46: There was 2 uses of this template and they needed defaultValue = 1.

I had 14 uses of this template that needed defaultValue = 0.

To me it made more sense to change this template but please tell me if I'm wrong.

blhoward2: You cannot change the master template. It's used in dozens of other places. Someone would need to go change everyone of those to override to 0. Just define a local template that changes the default if need be.

blhoward2:
> Fixed the main review comments.
> 
> 
> 
> Regarding strange labels and descriptions. Yes there are plenty but they are already there. The goal for me with this diff is not to make perfect labels and descriptions. Instead move things around to make it easier to fix those things later.

I understand but we're trying to avoid this piecemeal approach. It means that we have to then review the same files two or more times, which over hundreds of PRs adds up. Please try to get them as close to the standard as you can if you're going to template them and start moving things around. You're already much of the way there. 

Worse, if this is templated I'll assume the folder is finished because we're going folder by folder doing these as a complete overhaul. 

If you can't finish it that's fine, but we're likely going to let this PR sit until I can turn to it to finish it.

packages/config/config/devices/templates/master_template.json
WizKid: As my comment said it is used in 2 places. Which I both checked locally and https://github.com/zwave-js/node-zwave-js/search?q=base_enable_disable_inverted . And I updated both those places and added defaultValue=1 in here https://github.com/zwave-js/node-zwave-js/pull/2051/files#diff-cfea19ef962377943d781f23350480971815226082ba0db86dcfbe10d9e1ab01

WizKid:
Sorry I don't follow. When you say close to the standard as possible does that mean labels and descriptions need to be understandable and standarized? Because in that case I probably spent 5% of the time needed to actually be ready for a review. So very far from being there.

blhoward2:
Yeah. We're trying to cleanup all the files in an orderly progression. If you can't finish this set it's fine, but I'm not keen on merging a half finished folder so we may leave the PR until I can finish it. 

I'm on Aeotec now which has taken me weeks for various reasons, then have another group selected. So it would be a couple of weeks most likely. Your work wouldn't be wasted, I'd just build upon it.

Verifying everything you've changed is a lot of work. I'm going to have to go to the manuals and such some. If I'm doing that I'm going to do the labels and stuff all at once.

WizKid:
Good luck

packages/config/config/devices/0x010f/templates/fgs-2xx.json
AlCalzone: That does not mean it is good :)

Issue #2054, opened 2021-03-13
GE 14287 (others?) do not associate the controller with Group 3 automatically despite lifeline flag


blhoward2:
**Describe the bug**

The GE 14287 fan controller has a device config file specifying that Group 3 isLifeline: true. While troubleshooting another issued I noted my controller was nevertheless not associated with Group 3. Adding it manually works fine. Is this a problem on other devices?

**Note we are also ignoring the group names somehow...from the CC?** 

![Screen Shot 2021-03-12 at 11 28 29 PM](https://user-images.githubusercontent.com/33612110/111019046-b1676180-838a-11eb-9d57-94e00879de88.png)

```
	"associations": {
		"1": {
			"label": "Group 1",
			"maxNodes": 5,
			"isLifeline": true
		},
		"2": {
			"label": "Group 2",
			"maxNodes": 5
		},
		"3": {
			"label": "Double Tap",
			"maxNodes": 5,
			"isLifeline": true
		}
	},
```

**Device information**

Which device(s) is/are affected (make/model)?  GE 14287 - others?
What are the node IDs? 15

**How are you using `node-zwave-js`**
App Version
2.1.0+6858a08.6858a08
Zwavejs Version
6.6.0-ccce8b0
Zwavejs-server Version
1.1.0

**Logfile:**

Confirmed that this DOES work correctly on a GE 14294. It DOES work correctly if I manually add the controller to group 3 and the command succeeds (and double tap then works).

Here is it succeeding on the 14294. 
```
23:14:25.529 CNTRLR « [Node 005] received information for association group #3:
                      maximum # of nodes: 5
                      currently assigned nodes: 
23:14:25.530 CNTRLR » [Node 005] Controller missing from lifeline group #3, assinging ourselves...
23:14:25.546 DRIVER » [Node 005] [REQ] [SendData]
                      │ transmit options: 0x25
                      │ callback id:      49
                      └─[AssociationCCSet]
                          group id: 3
                          node ids: 1
23:14:25.556 DRIVER « [RES] [SendData]
                        was sent: true
23:14:25.636 DRIVER « [REQ] [SendData]
                        callback id:     49
                        transmit status: OK
23:14:25.646 DRIVER » [Node 005] [REQ] [SendData]
                      │ transmit options: 0x25
                      │ callback id:      50
                      └─[AssociationCCGet]
                          group id: 3
23:14:25.654 DRIVER « [RES] [SendData]
                        was sent: true
[...]

PR #2055, opened 2021-03-13
fix(config): remove unsupported double tap on GE 26932; add double tap to 12730; fix parameters

### LINKED ISSUE #2042: GE/Jasco 26932 does not support double tap events
**Checklist:**

- [x] I am **not** using ZWaveJS2MQTT. **Or:** a developer has told me to come here.
- [x] I have read the changelog and my problem was **not** mentioned there.

**Describe the bug**

If I look at the zwavejs event stream, a single tap and double tap of my GE switch results in the same "Basic" output.  If I compare that to a switch that works properly, a single tap produces a "Central Scene" event, while a double tap produces a "Basic" event.

**Device information**

Which device(s) is/are affected (make/model)?  GE/Jasco 26932 / 26933 / ZW3008

What are the node IDs? Node ID 8

**Did you change anything?**
- [ ] **Yes**: *(please describe)*
- [x] **No**

- [x] **No**, it never worked anywhere
- [ ] **Yes**, in: *(specify application with versions)*
    
- [x] `zwavejs2mqtt` (latest) docker image
- [ ] `zwavejs2mqtt` Manual Docker build
    - `node-zwave-js` branch: 
    - `zwavejs2mqtt` branch: 
- [ ] ioBroker.zwave2 adapter
- [ ] `HomeAssistant` version XYZ
- [ ] Pkg
- [ ] Manual Docker build
    - `node-zwave-js` branch: 
    - `zwavejs2mqtt` branch: 
- [ ] Other: 

**To Reproduce**

1. Look at event stream in Home Assistant
2. Tap the GE switch once and look at event stream.  Tap the GE twice and look at event stream.
3. Look at event stream and see that single and double taps produce the same "Basic" output.
4. In the attached file switch_event.txt, I tried double tapping and single tapping, but as you can see, it's all either "Basic" 0 or "Basic" 255.

**Additional context**

App Version
2.1.0
Zwavejs Version
6.6.0
Zwavejs-server Version
1.1.0

**Logfile:**
Includes switch event, interview log and node export:
[ge_switch.zip](https://github.com/zwave-js/node-zwave-js/files/6122907/ge_switch.zip)

blhoward2:
Closes #2042. Also removes errnonous parameters reported by @freshcoast, and adds double tap support to the 12730 fan controller.

### DIFF
--- a/packages/config/config/devices/0x0063/ge_12730_zw4002.json
+++ b/packages/config/config/devices/0x0063/ge_12730_zw4002.json
@@ -32,6 +32,27 @@
 		}
 	},
 	"paramInformation": {
+		"3": {
+			"label": "LED Light",
+			"description": "Sets when the LED on the switch is lit.",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 1,
+			"defaultValue": 0,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "LED on when light off",
+					"value": 0
+				},
+				{
+					"label": "LED on when light on",
+					"value": 1
+				}
+			]
+		},
 		"4": {
 			"label": "Invert Switch",
 			"valueSize": 1,
--- a/packages/config/config/devices/0x0063/ge_12730_zw4002.json
+++ b/packages/config/config/devices/0x0063/ge_12730_zw4002.json
@@ -15,32 +15,23 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
-	"paramInformation": {
-		"3": {
-			"label": "LED Light",
-			"description": "Sets when the LED on the switch is lit.",
-			"valueSize": 1,
-			"minValue": 0,
-			"maxValue": 2,
-			"defaultValue": 0,
-			"readOnly": false,
-			"writeOnly": false,
-			"allowManualEntry": false,
-			"options": [
-				{
-					"label": "LED on when light off",
-					"value": 0
-				},
-				{
-					"label": "LED on when light on",
-					"value": 1
-				},
-				{
-					"label": "LED always off",
-					"value": 2
-				}
-			]
+	"associations": {
+		"1": {
+			"label": "Group 1",
+			"maxNodes": 5,
+			"isLifeline": true
 		},
+		"2": {
+			"label": "Local Load",
+			"maxNodes": 5
+		},
+		"3": {
+			"label": "Double Tap",
+			"maxNodes": 5,
+			"isLifeline": true
+		}
+	},
+	"paramInformation": {
 		"4": {
 			"label": "Invert Switch",
 			"valueSize": 1,
@@ -60,81 +51,6 @@
 					"value": 1
 				}
 			]
-		},
-		"7": {
-			"label": "Dim Rate Steps (Z-Wave Command)",
-			"description": "Number of steps or levels",
-			"valueSize": 1,
-			"minValue": 0,
-			"maxValue": 99,
-			"defaultValue": 1,
-			"unsigned": true,
-			"readOnly": false,
-			"writeOnly": false,
-			"allowManualEntry": true
-		},
-		"8": {
-			"label": "Dim Rate Timing (Z-Wave)",
-			"description": "Timing of steps or levels",
-			"valueSize": 2,
-			"unit": "10ms",
-			"minValue": 1,
-			"maxValue": 255,
-			"defaultValue": 3,
-			"unsigned": true,
-			"readOnly": false,
-			"writeOnly": false,
-			"allowManualEntry": true
-		},
-		"9": {
-			"label": "Dim Rate Steps (Manual)",
-			"description": "Number of steps or levels",
-			"valueSize": 1,
-			"minValue": 1,
-			"maxValue": 99,
-			"defaultValue": 1,
-			"unsigned": true,
-			"readOnly": false,
-			"writeOnly": false,
-			"allowManualEntry": true
-		},
-		"10": {
-			"label": "Dim Rate Timing (Manual)",
-			"description": "Timing of steps",
-			"valueSize": 2,
-			"unit": "10ms",
-			"minValue": 1,
-			"maxValue": 255,
-			"defaultValue": 3,
-			"unsigned": true,
-			"readOnly": false,
-			"writeOnly": false,
-			"allowManualEntry": true
-		},
-		"11": {
-			"label": "Dim Rate Steps (All-On/All-Off)",
-			"description": "Number of steps or levels",
-			"valueSize": 1,
-			"minValue": 1,
-			"maxValue": 99,
-			"defaultValue": 1,
-			"unsigned": true,
-			"readOnly": false,
-			"writeOnly": false,
-			"allowManualEntry": true
-		},
-		"12": {
-			"label": "Dim Rate Timing (All-On/All-Off)",
-			"description": "Timing of steps or levels",
-			"valueSize": 2,
-			"unit": "10ms",
-			"minValue": 1,
-			"maxValue": 255,
-			"defaultValue": 3,
-			"unsigned": true,
-			"readOnly": false,
-			"writeOnly": false,
-			"allowManualEntry": true
 		}
 	},
 	"compat": {
--- a/packages/config/config/devices/0x0063/14287.json
+++ b/packages/config/config/devices/0x0063/14287.json
@@ -33,32 +33,6 @@
 		}
 	},
 	"paramInformation": {
-		"3": {
-			"label": "Night Light",
-			"description": "Defines the behavior of the blue LED",
-			"valueSize": 1,
-			"minValue": 0,
-			"maxValue": 2,
-			"defaultValue": 0,
-			"unsigned": true,
-			"readOnly": false,
-			"writeOnly": false,
-			"allowManualEntry": false,
-			"options": [
-				{
-					"label": "LED on when switch is OFF",
-					"value": 0
-				},
-				{
-					"label": "LED on when switch is ON",
-					"value": 1
-				},
-				{
-					"label": "LED always off",
-					"value": 2
-				}
-			]
-		},
 		"4": {
 			"label": "Invert Switch",
 			"description": "Invert the ON/OFF Switch State",
--- a/packages/config/config/devices/0x0063/26932_26933_zw3008.json
+++ b/packages/config/config/devices/0x0063/26932_26933_zw3008.json
@@ -26,13 +26,12 @@
 			"isLifeline": true
 		},
 		"2": {
-			"label": "Local Load",
+			"label": "Group 2",
 			"maxNodes": 5
 		},
 		"3": {
-			"label": "Double Tap",
-			"maxNodes": 5,
-			"isLifeline": true
+			"label": "Group 3",
+			"maxNodes": 5
 		}
 	},
 	"paramInformation": {
@@ -378,8 +377,5 @@
 				}
 			]
 		}
-	},
-	"compat": {
-		"treatBasicSetAsEvent": true
 	}
 }


packages/config/config/devices/0x0063/ge_12730_zw4002.json
```diff
@@ -15,32 +15,23 @@
...
-					"value": 0
-				},
-				{
-					"label": "LED on when light on",
-					"value": 1
-				},
-				{
-					"label": "LED always off",
-					"value": 2
-				}
```
kpine: Parameter 3 is valid, it's only value 2 that does not work. 

```
2021-03-12 20:29:07.588 INFO ZWAVE: Calling api writeValue with args: [ { nodeId: 21, commandClass: 112, property: 3 }, 2, [length]: 2 ]
2021-03-12 20:29:07.589 INFO ZWAVE: Writing 2 to 21-112-0-3
2021-03-12 20:29:07.661 INFO ZWAVE: Success zwave api call writeValue { success: true, message: 'Success zwave api call', result: undefined }
2021-03-12 20:29:08.771 INFO ZWAVE: Node 21: value updated: 112-0-3 2 => 1
```

blhoward2: Can you verify it on the device that it works (in addition to accepting it)? It isn't in the manual for the zw4002 so it would be undocumented. That's possible. I know that you can achieve that with the paddle taps.

packages/config/config/devices/0x0063/ge_12730_zw4002.json
```diff
@@ -60,81 +51,6 @@
 					"value": 1
 				}
 			]
-		},
-		"7": {
```
kpine, line 64: I did confirm 7-12 do not work. Here's setting 7 to a non-zero value, but it reads back 0.
```
2021-03-12 20:37:14.280 INFO ZWAVE: Calling api writeValue with args: [ { nodeId: 21, commandClass: 112, property: 7 }, 3, [length]: 2 ]
2021-03-12 20:37:14.281 INFO ZWAVE: Writing 3 to 21-112-0-7
2021-03-12 20:37:14.345 INFO ZWAVE: Success zwave api call writeValue { success: true, message: 'Success zwave api call', result: undefined }
2021-03-12 20:37:15.455 INFO ZWAVE: Node 21: value updated: 112-0-7 3 => 0
```

blhoward2: Thanks

packages/config/config/devices/0x0063/ge_12730_zw4002.json
kpine: Yes, 0 and 1 work fine. I use them in an automation so I can sleep at night. 😄 I wish I could permanently turn it off...
```
2021-03-12 20:39:38.190 DEBUG SOCKET: Event ZWAVE_API emitted to fxJmwvmEqXR2QrSgAAAJ
2021-03-12 20:39:38.192 INFO ZWAVE: Calling api writeValue with args: [ { nodeId: 21, commandClass: 112, property: 3 }, 0, [length]: 2 ]
2021-03-12 20:39:38.192 INFO ZWAVE: Writing 0 to 21-112-0-3
2021-03-12 20:39:38.259 INFO ZWAVE: Success zwave api call writeValue { success: true, message: 'Success zwave api call', result: undefined }
2021-03-12 20:39:39.372 INFO ZWAVE: Node 21: value updated: 112-0-3 0 => 0
2021-03-12 20:39:46.200 DEBUG SOCKET: Event ZWAVE_API emitted to fxJmwvmEqXR2QrSgAAAJ
2021-03-12 20:39:46.203 INFO ZWAVE: Calling api writeValue with args: [ { nodeId: 21, commandClass: 112, property: 3 }, 1, [length]: 2 ]
2021-03-12 20:39:46.204 INFO ZWAVE: Writing 1 to 21-112-0-3
2021-03-12 20:39:46.278 INFO ZWAVE: Success zwave api call writeValue { success: true, message: 'Success zwave api call', result: undefined }
2021-03-12 20:39:47.397 INFO ZWAVE: Node 21: value updated: 112-0-3 1 => 1
```

blhoward2: Pushed back in. Online it looks like people paint it with black nail polish.

packages/config/config/devices/0x0063/ge_12730_zw4002.json
```diff
@@ -60,81 +72,6 @@
...
-			"minValue": 1,
-			"maxValue": 255,
-			"defaultValue": 3,
-			"unsigned": true,
-			"readOnly": false,
-			"writeOnly": false,
-			"allowManualEntry": true
 		}
 	},
 	"compat": {
```
kpine, line 77: I guess it doesn't hurt, but the Supervision CC class removal is not necessary for this device. It's Z-Wave Classic, so no Supervision CC, and it's therefore not a Fan Switch Device Type. The removal is only required for the GE Z-Wave Plus devices that report as Fan Switch DT and don't support Supervision CC.

The treatBasicSetAsEvent is apparently unnecessary as well since it looks like double taps are not supported, due to no association groups?

packages/config/config/devices/0x0063/ge_12730_zw4002.json
```diff
@@ -15,13 +15,29 @@
...
+		"2": {
+			"label": "Local Load",
+			"maxNodes": 5
+		},
+		"3": {
+			"label": "Double Tap",
+			"maxNodes": 5,
+			"isLifeline": true
+		}
+	},
```
kpine, line 33: This doesn't seem to be doing anything. I downloaded this file, mapped it into my docker container, and did a re-interview, and no groups show up.

docker-compose:
```
    volumes:
      - ./zjs/data:/data
      - ./zjs/settings.json:/data/settings.json:ro
      - "./zjs/custom/ge_12730_zw4002.json:/usr/src/app/node_modules/@zwave-js/config/config/devices/0x0063/ge_12730_zw4002.json"
```

Association CC is not listed as a supported CC:
```
11:06:04.207 CNTRLR « [Node 021] node info received
                      supported CCs:
                      · Multilevel Switch
                      · All Switch
                      · Configuration
                      · Version
                      · Manufacturer Specific
                      controlled CCs:
```

No groups in the UI:
image

The configuration settings were updated correctly so I know I mapped the file correctly:
image

packages/config/config/devices/0x0063/ge_12730_zw4002.json
blhoward2: It is necessary. The zw4002 hardware was breaking with the 14287 and this device due to supervision. This fixes a prior issue.

kpine: The 12730 is not a 14287. `ge_12730_zw4002.json `.

packages/config/config/devices/0x0063/ge_12730_zw4002.json
blhoward2: This may be due to https://github.com/zwave-js/node-zwave-js/issues/2054 as it is the same hardware.

kpine: The quotes don't matter. Please explain how the Configuration parameters would have changed if I didn't update the file.

The 14287 is not the same thing as the 12730.

blhoward2: I edited my post as I missed that part of yours. They are both build on the zw4002 hardware according to the specs.

kpine: Did you not notice you have edited both a config file for 12730 and 14287 in this PR? They have different product ids. The 12730 is Z-Wave Classic and the 14287 is Z-Wave Plus. The zw4002 identifier is absolutely irrelevant.

https://products.z-wavealliance.org/products/1202?selectedFrequencyId=2
https://products.z-wavealliance.org/products/2103?selectedFrequencyId=2

blhoward2: They are different models. They share the same Jasco model base and the SAME FCC ID. the 14287 is a rebranded 12730 with an upgraded firmware adding zwave plus.

kpine: That's great, but one is Z-Wave Classic and one is Z-Wave Plus. That makes them not the same with respect to Z-Wave, which is what matters.

blhoward2: I'm not sure why everything needs to be so argumentative. Do you want to take over maintenance of the device files?

blhoward2: Look, I'm done arguing about this. I'm not adding the compat flag, its already there. It isn't the subject of this PR. If you think it should be removed, push a PR doing so and then we can talk about it on your PR.

blhoward2: Look, I'm done arguing about this. I'm not adding the compat flag, it already exists. If you think it should be removed push your own PR and we'll talk about it there. It isn't the subject of this PR.

AlCalzone (approved):

AlCalzone:
So as far as I understand this: There are two different versions of a device we cannot distinguish based on the information we currently use for that. One is ZW+, supports Associations and has broken Supervision support. One is legacy Z-Wave, does not support associations and (probably) has no Supervision support.

In that case, tough luck for the legacy device. The supervision flag is a no-op then, so it doesn't hurt either.

kpine:
> 
> 
> So as far as I understand this: There are two different versions of a device we cannot distinguish based on the information we currently use for that. 

For the record, these devices have different product IDs and different config files. They are distinguished just fine.

AlCalzone:
Okay, can you summarize the problem in a way that someone who doesn't know anything about those devices (i.e. me) can grasp it?

kpine:
I don't think there's any problem at this point, as the additional items in the config file are inconsequential as has been discussed. The conversation just devolved into an argument about whether the devices are the same or not, and I'm stubborn enough not to concede on that point.

If you guys want to keep the config files as it is just to be safe, go for it, there was no objection from me, just observations. If the argument is that there's a 12730 somewhere out there in the world that, despite all documentation, is Z-Wave Plus and identifies as a Fan Switch, well you got me, that's impossible to disprove.

In my defense I did not ask for any changes to this PR (aside from the removed config parameters which was quickly fixed).

[GE 12730](https://products.z-wavealliance.org/products/1202?selectedFrequencyId=2):
- It is a fan controller marketed as a ZW4002.
- It is Z-Wave Classic, using a 300-series chip.
- It has its own product ID of `0x3034`.
- It uses config file `packages/config/config/devices/0x0063/ge_12730_zw4002.json`.
- Because it is Z-Wave Classic, the specific device type is "Multilevel Power Switch", not "Fan Switch". Thus the Supervision CC is not a mandatory CC and it is unaffected by the Supervision CC bug.
- It does not support Supervision CC according to the logs I posted.
- It does not support Association CC according to the logs I posted.

[GE 14287](https://products.z-wavealliance.org/products/2506?selectedFrequencyId=2): 
- It is also a fan controller marketed as a ZW4002.
- It is Z-Wave Plus, using a 500-series chip.
- It has its own product ID of `0x3131`.
- It uses config file `packages/config/config/devices/0x0063/14287.json`
- It presumably supports Association CC, I don't have anything to say in regards to that because I don't have the switch installed and can't check it.
- It is a Z-Wave Plus "Fan Switch" DT, but doesn't support Supervision CC (as others have pointed out). Thus, it is affected by the mandatory CC problem which requires removing that CC.

There are many of these fan controllers marketed as a ZW4002. I won't argue that they aren't the same hardware or don't share FCC IDs, that's fine. I will continue to state they are not identical in functionality:

- Different product IDs
- Various z-wave chips (300 vs 500)
- Different versions of Z-Wave: Classic, Plus without S2, Plus with S2
- Different config parameters, some documented and some undocumented

Adding the associations for the compat flag to the non-zwave-plus 12730 may not have any consequences. Removing the Supervision CC from the fans that do support S2 and Supervision would be! For example, (according to the alliance DB) the 55259 version of the switch supports Supervision, yet the config file removes it. It also supports Central Scene (according to the alliance DB an manual), but it's automatically enrolled into the Group 3 Double Taps association group. Are those the right things to do? I don't know, just saying not everything is identical... 🤷

blhoward2:
We purposefully include double tap for central scene controllers. Both work and server different functions. Many people that control devices by direct association must react to the double tap to refresh devices that don't call on their change via a scene event. While you could do this via central scene, most people have multiple generations of switches and so doing it the same way is easier. 

More importantly, however, if we don't setup for double tap and the controller somehow gets added to group 3...and double tap up or down takes the state out of sync with HA and causes havoc.

kpine:
Again, I was not making a judgement about whether it is right or wrong, and not asking for any changes, just pointing out that the "same" devices are in fact different.

> More importantly, however, if we don't setup for double tap by unmapping the basic sets and the controller somehow gets added to group 3...any double tap up or down takes the state out of sync with HA

I am probably not understanding this use case, but it sounds like a problem exclusive to zwavejs2mqtt? For HA's zwave_js integration, these Basic Sets are just events. The double taps would have no effect on the switch states, which use Binary Switch or Multilevel CCs, not Basic CC.

I personally don't think it's right to force every user to have the controller in group 3 if central scene is available, just my opinion.  It generates unnecessary network traffic. There's no way to opt out is there, I'd have to manage my own config file?

> if we can assume that every device reporting itself as a 12730 doesn't have the firmware or features intended for the 14287. I'm just not sure I'm willing to take that bet.

I've seen the same product with different firmware having different functionality, but if you're telling me devices lie about their identity, well I have no response to that. I guess chalk it up to another example of a poor certification process.

blhoward2:
If not unmapped, the basic set is mapped to the primary CC and will be sent through as a value change by the node. This affects every front end that uses node as the backend. It's only sent as an event if we do the compat flag...that's what it does. 

Whether to force the lifeline is of course a point for debate. It doesn't actually cause any additional traffic on the zwave network, and a minuscule difference in traffic for the event bus. I suppose that we could setup the compat flag but not the lifeline flag for central scene devices, meaning those users would have an additional step to enable double tap. That creates a difference in apparent functionality that is best avoided to reduce service calls, though. I'd entertain that if you wanted to track down each and every device that is central scene, and for which the model was never sold without central scene, and then PR those changes. I have enough on my plate.

kpine:
> If not unmapped, the basic set is mapped to the primary CC (Binary or MuliLevel) per the spec and will be sent through as a value change by the node. This affects every front end that uses node as the backend. It's only sent as an event if we do the compat flag...that's what it does. It unmaps it and sends it as an event.

Fair enough, thanks for the explanation, I forget zwave-js maps the Basic Sets. I doubt GE intended for the double taps to be used in this way, which is why it's a strange implementation.

> We all get it. You know the zwave spec very well and I'm sure will be a great resource. But we're trying to foster a more positive community than what ultimately led to OZW's demise. 

Wow, all that was unnecessary. I'm sorry I struck such a nerve. 😫 I never asked for any changes to be made nor demanded anything from you.

blhoward2:
I'm not trying to attack you, I'm really not. But that's my point exactly. If you aren't raising a concern or asking for changes to be made that why is it relevant in a PR discussion thread?

kpine:
Your comments are quite unfair, and you are clearly attacking me. Normally I'd just let this pass, but you continue to misrepresent my comments and are suggesting things I've never said; therefore I feel I need to defend myself.

First, after I did "raise a concern" about the association groups, you became [hostile](https://github.com/zwave-js/node-zwave-js/pull/2055#discussion_r593798527) merely because I challenged your assertion that the devices were identical. You falsely accused me of arguing about the compat flag, and blamed me for wasting your time. That comment thread was about the association groups! I also did not force you to engage and **I stopped** after you became upset. Let me repeat, I stopped engaging in any arguments at this point.

Second, in the actual [conversation](https://github.com/zwave-js/node-zwave-js/pull/2055#discussion_r593788820) about the compat flag I didn't even argue about anything! All I did was make an observation that it probably wasn't necessary! No demands, nothing. I didn't even challenge your response, I only mentioned the devices files were different.

My comments may have been blunt, so my apologies for that, but it wasn't deserving of the hostility.

> If you aren't raising a concern or asking for changes to be made then why is it relevant in a PR discussion thread? This isn't the proper forum for debate.

Now you accuse me of wanting to debate, but I was literally [asked](https://github.com/zwave-js/node-zwave-js/pull/2055#issuecomment-798801309) a question by Al. I replied to the question I was asked. I did not start any debate. I did not push for any changes. I did explain my position that not all of the fan devices are the same. I did give an opinion that maybe not all config files should be the same. Again, I was asked a question! The fact is, **you were the one who [inserted yourself](https://github.com/zwave-js/node-zwave-js/pull/2055#issuecomment-798813306) into that second conversation, engaging _me_** and starting a "debate". I had no intention of commenting further until you replied to me, where I assumed you were open for a conversation at the point. That I regret.

Out of nowhere you [attacked](https://github.com/zwave-js/node-zwave-js/pull/2055#issuecomment-798819605) me a second time, accusing me of grandstanding, being toxic and creating a negative environment, wanting to "start a debate", and blaming me for wasting your time. For what? My [comments](https://github.com/zwave-js/node-zwave-js/pull/2055#issuecomment-798816009) were not disrespectful at all, you just seem to have gotten angry that I had a difference of opinion? I'm honestly confused about that one.

You're right, this PR is no place for a debate, but I'll remind you again that you started this new conversation, so accusing me of the one debating is in bad faith.

> This non-issue argument over which you aren't objecting to what we're doing has delayed the entire Aeotec set by a day and frankly put me out of the mood of working on it tomorrow.

I agree it's a non-issue argument and I'm truly sorry for wasting your time, because I feel the same! However, once again I didn't force you to engage in a conversation that **you** decided to initiate with me. To blame me for your delays and mood is not fair at all.

Feel free to lock this conversation, it's not doing anyone any good.

PR #2057, opened 2021-03-13
fix(config): correct description in FGBS222

### LINKED ISSUE #2049: FGBS222: discrepancy between label, description and manual
See https://github.com/zwave-js/node-zwave-js/issues/2048#issuecomment-797466039 for details.

blhoward2:
Closes #2049. The descriptions in here are generally terrible, but I'll fix those when I template.


AlCalzone (changes requested):
As discussed, this should go in to #2051

PR #2061, opened 2021-03-13
fix(config): Update lzw31-sn parameter 16 to match documentation.

mattsch:
This updates the LED effect parameter per docs and the [OZW](https://github.com/InovelliUSA/OpenZWave/blob/master/config/inovelli/lzw31-sn.xml) file.

@AlCalzone A related question,  my read of #1411 is that single params like this are less helpful than splitting each out like [lzw30-sn](packages/config/config/devices/0x031e/lzw30-sn.json). If that's the case, would it be better for me to update this to match that instead of updating this single parameter?

### DIFF
--- a/packages/config/config/devices/0x031e/lzw31-sn_0.0-1.45.json
+++ b/packages/config/config/devices/0x031e/lzw31-sn_0.0-1.45.json
@@ -261,7 +261,7 @@
 			"label": "LED Strip Effect",
 			"valueSize": 4,
 			"minValue": 0,
-			"maxValue": 83823359,
+			"maxValue": 100600575,
 			"defaultValue": 0,
 			"readOnly": false,
 			"writeOnly": false,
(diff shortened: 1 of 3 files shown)


AlCalzone:
@mattsch yes in the end we want to have partial parameters for these complicated params so users don't need an external calculator to determine the value. If you're willing to do that here it would be perfect. Otherwise, a correct value is better than what we have now.

AlCalzone (approved):

mattsch:
Sounds good, I'll work that up and do a separate PR for it. One inconsistency I noticed was in the naming of the params in lzw30-sn vs lzw36. The former uses `LED Strip Effect (Color)` while the latter uses `LED Strip Effect Color`. Personally, I think I prefer the second but curious if you have a preference so I can clean that up as well while I'm in here.

mattsch:
Closing in favor of #2069

PR #2067, opened 2021-03-14
fix(docs): add `ready` property for node

### LINKED ISSUE #2063: Document `node.ready` property
Need to know the ready status of all nodes after the nodes have already emitted their ready event.

There should be a `.ready` or `.isReady` property on each node. It is my understanding that `.interviewStage == "Complete"` does not exactly equate to being "ready". But perhaps I am wrong.

Storing a local array that is appended when the ready event is emitted. And then referencing that array later. However, this is only good for one scope. If the driver/controller is shared among more than one scope, than an array is needed for each - or some way of linking the array among the scopes.

marcus-j-davies:
fixes https://github.com/zwave-js/node-zwave-js/issues/2063

### DIFF
--- a/docs/api/node.md
+++ b/docs/api/node.md
@@ -419,6 +419,14 @@ enum ZWavePlusNodeType {
 }
 ```
 
+### `ready`
+
+```ts
+readonly ready: boolean
+```
+
+Whether this node is a ready, and safe to interact with.
+
 ### `roleType`
 
 ```ts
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
Thanks but you might wanna spell check that. Probably best if the description from the code is used.

AlCalzone (approved):

PR #2068, opened 2021-03-14
fix(config): update LZW30 parameters to match documentation/latest firmware

smitterson:
LZW30 does not have parameter 8 (that is for lzw30-sn only). Added parameter 13 added in fw version 1.17. Added parameter 51 added in fw version 1.19. Also added White as an LED option also added in fw 1.19.

For reference: 
https://support.inovelli.com/portal/en/kb/articles/firmware-v1-17-beta-lzw30-switch-black-series-gen-2
https://support.inovelli.com/portal/en/kb/articles/firmware-v1-19-beta-lzw30-switch-black-series-gen-2

### DIFF
--- a/packages/config/config/devices/0x031e/lzw30.json
+++ b/packages/config/config/devices/0x031e/lzw30.json
@@ -123,6 +123,10 @@
 				{
 					"label": "Pink",
 					"value": 234
+				},
+				{
+					"label": "White (firmware 1.19+)",
+					"value": 255
 				}
 			]
 		},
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x031e/lzw30.json
```diff
@@ -123,6 +123,10 @@
 				{
 					"label": "Pink",
 					"value": 234
+				},
+				{
```
marcus-j-davies, line 127: Just a couple of pointers that will likely be requested.
where a param/option only applies to a certain firmware, use the "$if" property, before the other bits of info.

```
{
  "$if": "firmwareVersion >= 1.19",
  "label": "White",
  "value": 255
}
```

see.
https://zwave-js.github.io/node-zwave-js/#/development/config-files?id=conditional-configuration-settings

packages/config/config/devices/0x031e/lzw30.json
```diff
@@ -238,16 +242,48 @@
 				}
 			]
 		},
-		"8": {
-			"label": "LED Strip Effect",
-			"description": "Calculated sum of parameters as per overview section",
-			"valueSize": 4,
+		"13": {
+			"label": "Special Load Type",
```
marcus-j-davies, line 248: And here.
It can then be removed from the descriptions
```
{
  "$if": "firmwareVersion >= 1.17",
  "label": "Special Load Type",
   ......
}
```

packages/config/config/devices/0x031e/lzw30.json
```diff
@@ -238,16 +242,48 @@
...
+					"value": 0
+				},
+				{
+					"label": "Manually set for special load type",
+					"value": 1
+				}
+			]
+		},
+		"51": {
+			"label": "Instant On",
```
marcus-j-davies, line 270: And here.

```
{
  "$if": "firmwareVersion >= 1.19",
  "label": "Instant On",
   ......
}
```

packages/config/config/devices/0x031e/lzw30.json
smitterson: Thanks for the pointer! I went ahead and made the change. I was looking at other config files instead of the documentation for guidance. If this gets accepted I'll see what I can do to clean up config files for other inovelli devices that I have.

marcus-j-davies: @blhoward2, will likely have something on his radar. So I would check in with him before going full throttle.

packages/config/config/devices/0x031e/lzw30.json
```diff
@@ -238,16 +243,50 @@
...
+				{
+					"label": "Manually set for special load type",
+					"value": 1
+				}
+			]
+		},
+		"51": {
+			"$if": "firmwareVersion >= 1.19",
+			"label": "Instant On",
+			"description": "Enables instant on (ie: disables the 700ms button delay).",
```
AlCalzone: The description does not seem to add value in addition to the label and the options. Please remove.

AlCalzone (changes requested):

packages/config/config/devices/0x031e/lzw30.json
smitterson: Removed

AlCalzone (approved):

AlCalzone:
Thanks!

PR #2082, opened 2021-03-15
fix(config): add missing Sunricher device configs

taakberg:
This PR adds the following Sunricher devices:
* `ZV9001K4-DIM` Z-Wave Dim Remote Control
* `ZV9001K4-DIM-G2` 2 group single color wall mounted remote
* `ZV9001K12-DIM-Z5` 5 Channel Dimmer Remote Control

It also updates the existing config file for the `ZV9001K12-DIM-Z4` 4 Channel Dimmer Remote Control, adding some missing information and making it more in line with other Sunricher device configs.

### DIFF
--- a/packages/config/config/devices/0x0330/zv9001k12-dim-z4.json
+++ b/packages/config/config/devices/0x0330/zv9001k12-dim-z4.json
@@ -1,13 +1,23 @@
-// ShenZhen Sunricher Technology, Ltd. SR-ZV9001K12-DIM-Z4
-// Z-Wave Dim Remote Control
+// Sunricher ZV9001K12-DIM-Z4
+// 4 Channel Dimmer Remote Control
 {
-	"manufacturer": "ShenZhen Sunricher Technology, Ltd.",
+	"manufacturer": "Sunricher",
 	"manufacturerId": "0x0330",
-	"label": "SR-ZV9001K12-DIM-Z4",
-	"description": "Z-Wave Dim Remote Control",
+	"label": "ZV9001K12-DIM-Z4",
+	"description": "4 Channel Dimmer Remote Control",
 	"devices": [
 		{
 			"productType": "0x0300",
+			"productId": "0xb302",
+			"zwaveAllianceId": 3907
+		},
+		{
+			"productType": "0x0301",
+			"productId": "0xb302",
+			"zwaveAllianceId": 3907
+		},
+		{
+			"productType": "0x0302",
 			"productId": "0xb302",
 			"zwaveAllianceId": 3907
 		}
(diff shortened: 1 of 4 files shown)


packages/config/config/devices/0x0330/zv9001k4-dim.json
```diff
@@ -0,0 +1,46 @@
...
+			"label": "Factory Reset",
+			"description": "Factory reset",
+			"valueSize": 2,
+			"minValue": 0,
+			"maxValue": 21930,
+			"defaultValue": 0,
+			"unsigned": true,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": true
```
AlCalzone: In other devices that have a factory reset parameter, we set `allowManualEntry` to `false` and provide two options - one for the default state and one for the value to factory reset. Please do the same here (and in the other file above).

packages/config/config/devices/0x0330/zv9001k12-dim-z4.json
```diff
@@ -1,21 +1,54 @@
...
+		},
+		"4": {
+			"label": "Launch 3",
+			"maxNodes": 5
+		},
+		"5": {
+			"label": "Launch 4",
+			"maxNodes": 5
+		}
+	},
```
AlCalzone: These devices support Association Group Info CC and Z-Wave+, so associations can be auto-discovered. Unless the auto-discovered labels are worse than this, the entire block here is unnecessary.

packages/config/config/devices/0x0330/zv9001k12-dim-z5.json
```diff
@@ -0,0 +1,59 @@
...
+		},
+		"5": {
+			"label": "Launch 4",
+			"maxNodes": 5
+		},
+		"6": {
+			"label": "Launch 5",
+			"maxNodes": 5
+		}
+	},
```
AlCalzone: These devices support Association Group Info CC and Z-Wave+, so associations can be auto-discovered. Unless the auto-discovered labels are worse than this, the entire block here is unnecessary.

packages/config/config/devices/0x0330/zv9001k4-dim-g2.json
```diff
@@ -0,0 +1,37 @@
...
+		},
+		"2": {
+			"label": "Launch 1",
+			"maxNodes": 5
+		},
+		"3": {
+			"label": "Launch 2",
+			"maxNodes": 5
+		}
+	},
```
AlCalzone: These devices support Association Group Info CC and Z-Wave+, so associations can be auto-discovered. Unless the auto-discovered labels are worse than this, the entire block here is unnecessary.

packages/config/config/devices/0x0330/zv9001k4-dim.json
```diff
@@ -0,0 +1,46 @@
...
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 1,
+			"isLifeline": true
+		},
+		"2": {
+			"label": "Launch 1",
+			"maxNodes": 5
+		}
+	},
```
AlCalzone: These devices support Association Group Info CC and Z-Wave+, so associations can be auto-discovered. Unless the auto-discovered labels are worse than this, the entire block here is unnecessary.

AlCalzone (changes requested):

packages/config/config/devices/0x0330/zv9001k4-dim.json
taakberg: I've looked trough the manufacturer documentation, and it actually doesn't state any factory reset parameter. I've based this on the [Z-Wave Alliance docs](https://products.z-wavealliance.org/products/3783?selectedFrequencyId=1) of the SR-ZV9001K4-DIM-G2, but I think the information there is incorrect. So I've just opted to remove it instead.

AlCalzone (approved):

PR #2089, opened 2021-03-15
feat: provide API to lookup device definitions with preserved conditions

AlCalzone:
@marcus-j-davies This is what we discussed. You'll want to use `lookupDevicePreserveConditions` to look up the device config.
The returned object can have a `condition` property on association groups, config parameters (which are all arrays with 1 or more definitions) and config parameter options.

Let me know if you can work with that.

### DIFF
--- a/docs/api/config-manager.md
+++ b/docs/api/config-manager.md
@@ -263,6 +263,14 @@ Looks up the definition of a given device in the configuration DB. It is not nec
 
 For details on the available properties, refer to the [config file documentation](development/config-files.md).
 
+### `lookupDevicePreserveConditions`
+
+```ts
+lookupDevicePreserveConditions(manufacturerId: number, productType: number, productId: number, firmwareVersion?: string): Promise<ConditionalDeviceConfig | undefined>;
+```
+
+Like `lookupDevice`, but does not evaluate `$if` conditions. The resulting object will contain all settings that are defined regardless if they apply or not.
+
 ### `loadNotifications`
 
 ```ts
--- a/docs/api/config-manager.md
+++ b/docs/api/config-manager.md
@@ -280,7 +280,10 @@ interface ConditionalDeviceConfig {
 	readonly manufacturerId: number;
 	readonly label: string;
 	readonly description: string;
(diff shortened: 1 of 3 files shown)


marcus-j-davies:
I'll give this a deeper look tomorrow. but initially.

>Like lookupDevice, but does not evaluate $if conditions.  

>The returned object can have a condition property on association groups, config parameters (which are all arrays with 1 or more definitions) and config parameter options.

is good for what we are trying to achieve. 👍

docs/api/config-manager.md
```diff
@@ -263,6 +263,14 @@ Looks up the definition of a given device in the configuration DB. It is not nec
 
 For details on the available properties, refer to the [config file documentation](development/config-files.md).
 
+### `lookupDevicePreserveConditions`
+
+```ts
+lookupDevicePreserveConditions(manufacturerId: number, productType: number, productId: number, firmwareVersion?: string): Promise<ConditionalDeviceConfig | undefined>;
+```
+
+Like `lookupDevice`, but does not evaluate `$if` conditions. The resulting object will contain all settings that are defined regardless if they apply or not.
```
marcus-j-davies: Maybe add here, the slight difference, with how the params are delivered.
Each param object being wrapped in an array block (1 object in the array for each condition assumably?)
where as `lookupDevice `returns an unwrapped object for each param

My best effot...

------------------------------
Like `lookupDevice`, but does not evaluate `$if` conditions. The resulting object will contain all settings that are defined regardless if they apply or not. Each Association Group, Config Parameter and Parameter Option, will include a `condition` statement, providing one has been provided.

Each Config parameter will come wrapped in an array (1 for each condition)

marcus-j-davies (changes requested):
Have tested - seems to work.

What I shall be doing here, is collapsing each param array, But storing the different conditions, then I can use those list of conditions per param, to do something beautiful on the UI.

My only suggestion is to highlight the difference in how params are structured

docs/api/config-manager.md
AlCalzone: I've opted to include the types instead.

PR #2094, opened 2021-03-17
feat(config): add Fakro ZWS230 chain actuator

gyorsok:
Based on OZW import (0x0085-0x0002-0x0003) and
my real device (0x0085-0x0003-0x0002).

### DIFF
--- a/packages/config/config/devices/0x0085/zws230.json
+++ b/packages/config/config/devices/0x0085/zws230.json
@@ -1,0 +1,29 @@
+// Fakro ZWS230
+// Chain actuator - window opener
+{
+	"manufacturer": "Fakro",
+	"manufacturerId": "0x0085",
+	"label": "ZWS230",
+	"description": "Chain actuator 230VAC",
+	"devices": [
+		{
+			"productType": "0x0002",
+			"productId": "0x0003"
+		},
+		{
+			"productType": "0x0003",
+			"productId": "0x0002"
+		}
+	],
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 5,
+			"isLifeline": true
+		}
+	}
+}
--- a/packages/config/config/devices/0x0085/zws230.json
+++ b/packages/config/config/devices/0x0085/zws230.json
@@ -9,10 +9,6 @@
 		{
 			"productType": "0x0002",
 			"productId": "0x0003"
-		},
-		{
-			"productType": "0x0003",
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0085/zws230.json
```diff
@@ -0,0 +1,29 @@
...
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 5,
+			"isLifeline": true
+		}
+	}
```
AlCalzone, line 24: Do you know if the device supports Z-Wave Plus? In that case this definition shouldn't be necessary.

packages/config/config/devices/0x0085/zws230.json
```diff
@@ -0,0 +1,29 @@
...
+	"description": "Chain actuator 230VAC",
+	"devices": [
+		{
+			"productType": "0x0002",
+			"productId": "0x0003"
+		},
+		{
+			"productType": "0x0003",
+			"productId": "0x0002"
+		}
```
AlCalzone: According to 
https://github.com/OpenZWave/open-zwave/blob/e7b1705403fced2c2d2bc8aa92b559f6351f5542/config/manufacturer_specific.xml#L618-L620
there is type 3, id 2 and type 2, id 5. Where's type 2, id 3 taken from?

AlCalzone (changes requested):

packages/config/config/devices/0x0085/zws230.json
gyorsok: Type 2 ID 3 (0x0085-0x0002-0x0003) is the actual device that I own.

I attached a nodes-dump snippet of that one node.
[fakro-zws230-nodes.json.gz](https://github.com/zwave-js/node-zwave-js/files/6157184/fakro-zws230-nodes.json.gz)

packages/config/config/devices/0x0085/zws230.json
gyorsok: I think these are not Zwave Plus devices, there is no mention of "Plus" in their manuals.

Also, although it has a switch_multilevel, you cannot really open it partially that way, any value above 0 opens the window all the way. The device has to be refreshed to see the new opening level (which immediatelly jumps to max btw),

AlCalzone: From your dump:
```
  "zwavePlusVersion": 1,
```
Guess it doesn't hurt to leave this in though.

AlCalzone (approved):

AlCalzone:
Thanks!

Issue #2099, opened 2021-03-16
[bug] Qubino Flush Dimmer - using I2 status updates are not triggered after restart


aveseli:
### Version

**Checklist:**

- [x] I am **not** using HomeAssistant. **Or:** a developer has told me to come here.
- [x] I have checked the troubleshooting section and my problem is **not** described there.
- [x] I have read the changelog problem was **not** mentioned there.

Build/Run method

- [x] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build )

zwavejs2mqtt version: 2.3.0
zwave-js version: 6.6.2

#### Context / Setup

I am using a Qubino Flush Dimmer. It has 3 different inputs for physical switches. I1 is used to control the dimmer itself. I have hooked up I2 to a second switch. In the configuration I have set up `[10-112-0-100] Enable / Disable Endpoints I2` to `[9] Sensor binary`. 

![Screenshot 2021-03-16 at 22 49 56](https://user-images.githubusercontent.com/1632225/111384626-05838600-86aa-11eb-94a3-947d637017a5.png)

This gives me another `Basic` report `[10-32-2-currentValue] Current value` which toggles from the values 0 to 255 and vice versa whenever I press the mono-stable switch on I2. 

![Screenshot 2021-03-16 at 22 50 56](https://user-images.githubusercontent.com/1632225/111384721-2350eb00-86aa-11eb-8f4c-e7ff5040b207.png)

#### The Problem
Whenever I restart zwavejs2mqtt it doesnt report anymore. I can see in the debug log that the zwave event is coming in, but I dont see the `Basic` section in the control panel nor it isn't reported to mqtt. When I re-interview the node everything works fine after that.

### To Reproduce

1. restart zwavejs2mqtt
2. Click the switch connected on I2, no mqtt message nor update on control panel.
3. re-interview the node and the problem is gone.

### Expected behavior

After restarting zwavejs2mqtt it should still report the toggle on I2 without re-interviewing the node.

AlCalzone:
Found it... the endpoint has a different device class than the node itself, causing the mapping to fail.
https://github.com/zwave-js/node-zwave-js/issues/2100

AlCalzone:
Potential fix at #2101

AlCalzone:
@aveseli Piecing together the response from Qubino and your log:
You have configured the Basic CC association group 5 (under root device), which is why the Basic CC commands are being sent:
[...]

PR #2101, opened 2021-03-17
feat: handle device class of endpoints

### LINKED ISSUE #2100: Basic CC mapping does not work if node has a different device class than the endpoint
```
20:54:42.383 CNTRLR « [Node 010] received response for protocol info:
                      basic device class:    Routing Slave
                      generic device class:  Multilevel Switch
                      specific device class: Multilevel Power Switch
...
20:54:43.656 CNTRLR « [Node 010] received response for endpoint capabilities (#2):
                      generic device class:  Binary Sensor
                      specific device class: Unused
```

AlCalzone:
fixes: #2100
and was probably due long ago 😅

### DIFF
--- a/docs/getting-started/migrating-to-v7.md
+++ b/docs/getting-started/migrating-to-v7.md
@@ -32,3 +32,22 @@ This event serves a similar purpose as the `"value notification"` event (which w
 Since the original implementation, the need for a more versatile CC-specific notification event has arisen. Therefore we decided to rework this event and decouple it from the `Notification CC`. As a result, the event callback now indicates which CC raised the event and its arguments are moved into a single object parameter.
 
 See the [`"notification"` event](../api/node#quotnotificationquot) docs for a detailed description what changed.
+
+## `Endpoint` constructor changed
+
+The `Endpoint` needed to be modified in order to include the device class information. To be in line with the existing `ZWaveNode` constructor, the additional argument was added **before** the last one:
+
+```diff
+	public constructor(
+		/** The id of the node this endpoint belongs to */
+		public readonly nodeId: number,
+		/** The driver instance this endpoint belongs to */
+		protected readonly driver: Driver,
+		/** The index of this endpoint. 0 for the root device, 1+ otherwise */
+		public readonly index: number,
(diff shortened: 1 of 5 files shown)


AlCalzone:
@aveseli can you test if this branch fixes your problem with the Flush Dimmer?
You'll probably need to build a custom container https://zwave-js.github.io/zwavejs2mqtt/#/troubleshooting/docker_custom_builds

And re-interview the node for the changes to have an effect.

aveseli:
It reports the state after a restart here:

Screenshot 2021-03-18 at 22 25 10

I guess this is expected.

One strange thing I noticed: it toggles the value to `true` and after 2-5 seconds it switches back to `false` without me pressing the switch.

I have build a new docker image like this:

`DOCKER_BUILDKIT=1 docker build --build-arg SRC=git-clone-src --build-arg Z2M_BRANCH=master --build-arg ZWJ_BRANCH=endpoint-device-class --no-cache -f zwavejs2mqtt/docker/Dockerfile.contrib -t zwavejs2mqtt .`

I am not sure if using  `Z2M_BRANCH=master ` is correct, since the control panel changes a bit and probably this is still under development.

AlCalzone:
Could you make a log so I can check what exactly happens?

aveseli:
[zwavejs2mqtt-store-2.zip](https://github.com/zwave-js/node-zwave-js/files/6173503/zwavejs2mqtt-store-2.zip)

What I did:
- starting the new test version of zwavejs2mqtt
- hitting the switch 2-3 times
- re-interviewing the node 10
- hitting the switch 2-3 times

Please note: I configured the zwave logging to only include Node 10.

AlCalzone:
Ok so it looks like the mapping works now, but the device actually sends a followup Binary Sensor Report with state `false`. You could try adding the controller node to association group 7 to see if this changes things.

I'm actually not sure how this is supposed to work. I'll ask my contact at Qubino and let you know.

Let's continue this discussion in your original issue #2099 - we might have to do something else to make the device work correctly.

This PR is working as intended and fixes #2100.

PR #2102, opened 2021-03-18
docs: clarify description label in config file documentation

blhoward2:
Clarifies difference between device description and parameter description. Minor grammar corrections.

### DIFF
--- a/docs/config-files/style-guide.md
+++ b/docs/config-files/style-guide.md
@@ -60,7 +60,7 @@ Sometimes a manufacturer makes a device for another company. The field `manufact
 
 ## Device Descriptions
 
-These should generally conform to the name under which the device is sold. If the description merely mirrors the label, it should be omitted instead. Remove the manufacturer name from the description, if present. For example:
+These should generally conform to the name under which the device is sold. If the description merely mirrors the label, it should be omitted. Remove the manufacturer name from the description, if present. For example:
 
 ```diff
 	// Yale YDM3109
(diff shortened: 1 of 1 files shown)


docs/config-files/style-guide.md
```diff
@@ -261,7 +261,7 @@ Whenever a parameter only allows two options, formulate the label and/or descrip
 
 A parameter must define the range of min/max values, however, that range should only be as large as is necessary. Frequently, imported config files have a range of 0-255 when the only possible range is 0-1. Please check the manual.
 
-An exception are parameters which accept a single special value outside the normal range, e.g. `10-255` and `0 (disable)`. There isn't a simple way to represent these gaps, so the value range should be `0-255` in this case.
+An exception is parameters that accept a single special value outside the normal range, e.g. `10-255` and `0 (disable)`. There isn't a simple way to represent these gaps, so the value range should be `0-255` in that case.
```
AlCalzone, line 264: The "is" seems weird, but then I'm not a native speaker...

AlCalzone (approved):

docs/config-files/style-guide.md
robertsLando: I think that's good as it's referred to the exception not the parameters but yeah seems werid to read

AlCalzone: Parameters that (...) are an exception.
😅 idk

blhoward2: It's correct. Parameters is plural but there is only one exception. You could also say "Parameters that ... are an exception." That's correct too because then parameters is the object and is plural.

AlCalzone: Like I said, I'm not a native speaker :)

PR #2119, opened 2021-03-20
fix(config): minor update for some Inovelli switches and dimmer

tprelog:
I updated the `LED Indicator Color` parameter to be more consistent for the Inovelli switches and dimmers
- allow manual entry and add parameter description for the `lzw30`
- add "White" option for the `lzw30-sn`
- add color options and add parameter description for the `lzw31`

I also updated the device description for the `lzw30` and `lzw30-sn` switches to Title Case, and fixed minor typo


tprelog:
@zwave-js-bot fix lint

AlCalzone (approved):

PR #2120, opened 2021-03-20
feat: emit an event when node status changes to unknown

raman325:
As I had mentioned in Discord (https://discord.com/channels/330944238910963714/800356888827002880/822887257983811647) it appears that a node can go from dead to unknown if a message is received when a node is in a dead state. It is useful to capture this state change as an event so that we can handle this scenario downstream.

I wasn't sure whether we wanted to log something or perform an action when this event is fired like we do for all other status change events, but I can add that in if that's desired

### DIFF
--- a/docs/api/node.md
+++ b/docs/api/node.md
@@ -621,6 +621,14 @@ A non-sleeping node has stopped responding or just started responding again. The
 (node: ZWaveNode) => void
 ```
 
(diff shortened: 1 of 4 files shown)


raman325:
per AlCalzone, this state transition is not valid anymore so there is nothing to capture an event for

PR #2135, opened 2021-03-22
fix(config): Aeotec ZW116

SantaFox:
Updated config file for Aeotec ZW116 based on printed manual + OpenZWave config + data from Alliance database


SantaFox:
Retest please

PR #2137, opened 2021-03-22
fix(config): Aeotec zw116

SantaFox:
Updated based on paper device manual & OpenZWave config file for this device

### DIFF
--- a/packages/config/config/devices/0x0086/zw116.json
+++ b/packages/config/config/devices/0x0086/zw116.json
@@ -12,25 +12,35 @@
 		},
 		{
 			"productType": "0x0003",
-			"productId": "0x0074"
+			"productId": "0x0074",
+			"zwaveAllianceId": 2289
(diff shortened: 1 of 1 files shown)


AlCalzone:
Thanks, but @blhoward2 is currently working on all Aeotec device files. I'll leave this PR open as a reference for him, but I suspect the work is already done in one of his branches.

SantaFox:
Thank you. This specific device seems not to be updated in his commits so far, however it may be included in other batches. I also saw his templates solution and it looks really nice; of course my amendments weren't used them.

I will appreciate Brian's decision if he needs this PR or not.

blhoward2:
I'm going back through every file in both Aeotec folders, and going back to the manual for each to clean them up.  We're trying to make things consistent across manufacturers so the labels and such I'm using are the same or similar to what we've used elsewhere.

PR #2145, opened 2021-03-24
New device file for MCO Home - MH-S511 AU version

kanebullen:
Australian Version of MH-S511 1 way Touch Panel

Based on mh-s511-il.json, but with additional parameter information from mh-s513.json


AlCalzone:
Do you know how this is different from mh-s511-il that we also have a file for? Could it be that those are the same devices and we should combine the files?

AlCalzone:
This does look quite a bit like https://products.z-wavealliance.org/products/3674 but that lists a different product type/id

kanebullen:
I would suggest that this would cover both - it's just the device productIDs that are different.     

It was mentioned to me to create new files - but I'm thinking it'd be better to add the device definitions to the existing one instead  (and I can also add the parameter information which is missing)

And yes, the device you linked to is the same switch (a third configuration)  Seems MCO Home have different identifiers for each region (different zwave frequencies)

Would you recommend I cancel this PR and submit an update for mh-s511-il ?

AlCalzone:
> Would you recommend I cancel this PR and submit an update for mh-s511-il ?

I'd suggest you add the existing product ID to your new file, delete the old one and call the new file `mh-s511.json`. That way we can also get rid of the `IL version` and `AU version` suffixes.

While you're at it, please also check out our config style guide:
- https://zwave-js.github.io/node-zwave-js/#/config-files/contributing-files
- https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide

kanebullen:
Ok perfect - I'll do that, and also look at ensuring the formatting fits the style guide

Might also add the AU versions against the files for the rest of the family (2 and 3 way switch models).

PR #2147, opened 2021-03-24
Delete mh-s511-il.json

kanebullen:
Replaced with non "il" specific version


AlCalzone:
Superseded by https://github.com/zwave-js/node-zwave-js/pull/2151

PR #2151, opened 2021-03-24
fix(config): update configuration for MCOHome MH-S510 series

AlCalzone:
@kanebullen I've put your PRs together in this one. However I'm not sure we're done yet. The more I search for these files, the more I'm confused...

The Z-Wave Alliance website lists this manual, which has the same parameters you provided, but also has a param 7:
https://products.z-wavealliance.org/ProductManual/File?folder=&filename=product_documents/3674/MH%20S510%20user%20manual.pdf

The MCOHome website lists many more config parameters (and different ones): http://mcohome.com/ProductDetail/3894276.html

These look to be the same, but I'm not sure if it maybe a new series?


packages/config/config/devices/0x015f/mh-s511.json
```diff
@@ -0,0 +1,124 @@
...
+		{
+			"productType": "0x5111",
+			"productId": "0x1406"
+		},
+		{
+			"productType": "0x5111",
+			"productId": "0x5103"
+		},
+		{
+			"productType": "0x5131",
```
AlCalzone, line 18: This might be a S513 instead? There is a pattern to these productTypes (5111 for S511, 5121 for S512, 5131 for S513)

kanebullen:
@AlCalzone Yes, I've also found them to be very confusing - I've seen different PDF manuals for the same model, but with slightly different details, which make it very hard!           Also, they seem to have a manual for the series (for 1, 2 or 3 switch), so some of the details in the manual don't apply to all.

And regarding the 0x5131 - I think you might be right - the product type should be 0x5111 (but the productID seems to vary with different regions)

packages/config/config/devices/0x015f/mh-s511.json
kanebullen: Agreed

AlCalzone:
I've asked the MCOHome support for some clarification, let's see how they respond.

AlCalzone:
OK, Update from the support. Ignore the files on Z-Wave Alliance. The PDF on their website is correct. These are the correct deviceIDs:

```
S511:
productType: 0x5111,
productId: 0x5103
 
S512:
productType: 0x5121,
productId: 0x5103
 
S513:
productType: 0x5131,
productId: 0x5103
```

@kanebullen would you be willing to update one of the files to match? I'll apply it to the others. You can just open a PR onto this PR.

kanebullen:
@AlCalzone I'm willing to help - I just think I'll need to learn a bit about GitHub first!     So it looks like mh-s512.json and mh-s513.json are correct - it's just the new mh-s511.json that needs to be fixed up

AlCalzone:
I was also talking about the configuration parameters (see the PDF on MCOHome's website). There are many other different parameters than what we have now.

kanebullen:
Aha - no problems, I'll get that sorted in next couple of days.   Cheers

kanebullen:
@zwave-js-bot fix lint

blhoward2:
I think that #2178 is the latest incarnation of this draft. If true, this should be closed. If not, there are conflicts.

PR #2153, opened 2021-03-24
feat: collect telemetry information for identified devices without a config file

AlCalzone:
With this PR, we detect at the end of an interview whether the device could be identified but we don't have a config file for it. In that case, we collect some basic information that helps us create said config file:

- Whether the device supports Configuration CC V3+ - if yes, the detected config parameters are recorded
- Whether the device supports AGI CC - if yes, the association group names are recorded
- Whether the device supports Z-Wave+ and which version of the standard it implements

/cc @blhoward2


blhoward2 (dismissed):
Can this be disabled by the user or UI? LGTM

AlCalzone:
Currently not. But it does not collect any information that can be used to identify a user. Even the anonymous fingerprint we use for crash reporting gets removed.

PR #2169, opened 2021-03-27
Update mh-s511.json

kanebullen:
Not sure what the product IDs in the "-IL" versions are, but perhaps they are for a different product (they mention a relay, rather than a touch panel switch).    So I think this is now correct.

(not sure how I can submits this pull request against the previous?)

### DIFF
--- a/packages/config/config/devices/0x015f/mh-s511.json
+++ b/packages/config/config/devices/0x015f/mh-s511.json
@@ -7,14 +7,6 @@
 	"description": "Touch Panel Switch 1-Way",
 	"devices": [
 		{
-			"productType": "0x5111",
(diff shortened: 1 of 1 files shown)


AlCalzone:
> (not sure how I can submits this pull request against the previous?)

you did :)

grafik

kanebullen:
@AlCalzone Hopefully I've done correctly - have done all three files, with all the parameters relevant for each.        (think I might have committed to a separate pull request )

Really need to find a decent amount of time to get my head around github!

blhoward2:
You're creating a conflict by modifying a file you're modifying in #2178. Please push the changes to the file in that PR.

PR #2173, opened 2021-03-27
fix: add compat flag to zen25

DaveCo1701:
zen25 has a similar issue as described in https://github.com/zwave-js/node-zwave-js/pull/2126. Applied compatibility flag and tested. Feedback from zwave device now works properly.

### DIFF
--- a/packages/config/config/devices/0x027a/zen25.json
+++ b/packages/config/config/devices/0x027a/zen25.json
@@ -291,5 +291,8 @@
 				}
 			]
 		}
+	},
+	"compat": {
+		"preserveRootApplicationCCValueIDs": true
 	}
 }


blhoward2:
This is no longer necessary after #2160 (merged yesterday) and in fact we're removing the compat flag from the zen16/zen17. We no longer map the reports this was designed to undo. Do you have the ability to build a docker off the master branch and test it again without the compat flag?

https://zwave-js.github.io/zwavejs2mqtt/#/troubleshooting/docker_custom_builds

DaveCo1701:
I don’t have that ability, but I can wait until the docker container with the updated zwave-js version comes out and report back.

PR #2203, opened 2021-03-31
fix(config): update Radio Thermostat CT101 config

jjlawren:
Updated the CT101 thermostat configuration file based on the existing `ct100_plus.json` config.

Used documentation available at https://static.wixstatic.com/ugd/94b5a7_23ba26a52a1e4f5498afa36d1552ed07.pdf (linked from https://www.radiothermostat.com/support).


packages/config/config/devices/0x0098/ct101.json
```diff
@@ -52,29 +52,40 @@
...
 					"value": 3
 				},
 				{
-					"label": "2.0°",
+					"label": "2.0° F",
 					"value": 4
 				}
 			]
 		},
+		"2": {
```
blhoward2: Please check the manual and list out these options. If there are only four they should be presented as options through the UIs.

packages/config/config/devices/0x0098/ct101.json
```diff
@@ -91,9 +102,72 @@
 				}
 			]
 		},
+		"4": {
+			"label": "Power Status",
+			"description": "C-Wire / Battery Status",
+			"valueSize": 1,
```
blhoward2, line 160: This isn't valid. The min and max cannot both be zero. if this is a limited set of options, please list out the options.

packages/config/config/devices/0x0098/ct101.json
```diff
@@ -103,42 +177,67 @@
...
-					"label": "3.5°",
+					"label": "3.5° F",
 					"value": 7
 				},
 				{
-					"label": "4.0°",
+					"label": "4.0° F",
 					"value": 8
 				}
 			]
```
blhoward2, line 270: This label and description don't make sense. What does this do?

packages/config/config/devices/0x0098/ct101.json
```diff
@@ -148,14 +243,65 @@
...
 				{
-					"label": "Economy recovery Mode",
+					"label": "Economy recovery mode",
 					"value": 2
 				}
 			]
+		},
+		"10": {
+			"label": "Temperature Reporting Filter",
+			"description": "Upper/Lower bounds for thermostat temperature reporting",
```
blhoward2: This description doesn't appear to make sense. How can this be both the upper and lower bounds? Please referecnce the manual.

blhoward2 (changes requested):
I corrected some things myself, but please take a look at the comments and request a new reivew when complete.

packages/config/config/devices/0x0098/ct101.json
jjlawren: This param is a read-only byte array. I'm not quite sure how to represent this using options, as some of them are dynamic based on the number or stages:
```
2. HVAC Settings (32-bit)
Byte 1 = HVAC Setup: Normal (0x01) or Heat Pump (0x02)
Byte 2 = Aux Setup(Gas (0x01) or Electric (0x02)) & Number of Auxiliary Stages (Heat Pump)/Number of Heat Stages (Normal)
Byte 3 = Number of Heat Pump Stages
Byte 4 = Number of Cool Stages
```

Edit: Discovered partials, updated the config to reflect.

packages/config/config/devices/0x0098/ct101.json
jjlawren: If this is for param 8, I'm honestly not sure. The documentation does not describe the parameter in detail. This was copied from the `ct100_plus.json` file.

packages/config/config/devices/0x0098/ct101.json
jjlawren: Updated.

packages/config/config/devices/0x0098/ct101.json
jjlawren: Looks like this may need to be broken down into partial parameters. Reading up on that now.

AlCalzone:
@blhoward2 you would have loved this manual and the partials... not 🤣

blhoward2:
I pushed some style changes. I'm not sure if there is more to do with this one. There are still some odd labels but I know you were both running into difficulties?

jjlawren:
Thanks, those are good changes. One more adjustment left to be completed.

packages/config/config/devices/0x0098/ct101.json
jjlawren: This param is complicated with different layouts between get & set. Since there doesn't appear to be a known real-world use for it, I've removed it from the config.

blhoward2 (dismissed):
I haven't verified the bitmasks as I know @AlCalzone was working on this with you.

packages/config/config/devices/0x0098/ct101.json
AlCalzone: It sounds like this is the hysteresis around the setpoint in which the device does not switch on or off. Might be worth calling it "Temperature Hysteresis" instead

AlCalzone (dismissed):
LGTM

PR #2207, opened 2021-03-31
feat: implement reporting of usage statistics

AlCalzone:


robertsLando (dismissed):
As always, LGTM 🚀

packages/zwave-js/src/lib/telemetry/metrics.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export function compileMetrics(
+	driver: Driver,
+	appInfo: AppInfo,
+): Record<string, any> {
+	return {
+		// Hash the homeId, so it cannot easily be tracked
+		id: crypto
+			.createHash("md5")
```
chrisns: why md5, not sha (SHA-256)?

packages/zwave-js/src/lib/telemetry/metrics.ts
```diff
@@ -0,0 +1,90 @@
+import { formatId } from "@zwave-js/shared";
+import { isObject } from "alcalzone-shared/typeguards";
+import axios from "axios";
+import * as crypto from "crypto";
+import type { Driver } from "../driver/Driver";
+
+const baseUrl = "https://metrics.zwave-js.io:9998"; // No leading slash!
+// The shared "secret" between metrics server and zwave-js. Can be used to disable reporting for misbehaving clients.
+const authKey = "Og9O}$[wKk=<$Z*qUduK";
```
chrisns: how? why not use the application version since changing this would require a release?

chrisns (changes requested):

packages/zwave-js/src/lib/telemetry/metrics.ts
```diff
@@ -0,0 +1,90 @@
+import { formatId } from "@zwave-js/shared";
+import { isObject } from "alcalzone-shared/typeguards";
+import axios from "axios";
+import * as crypto from "crypto";
+import type { Driver } from "../driver/Driver";
+
+const baseUrl = "https://metrics.zwave-js.io:9998"; // No leading slash!
```
chrisns: why on a non standard port? you may find the client gets blocked or you just draw unnecessary attention from any network DLP/IDP type thing

packages/zwave-js/src/lib/telemetry/metrics.ts
robertsLando: We could use an api to provide the auth key IMO (it's not very secret). Yeah changing this in this way could break old instances instantly

chrisns: an api to provide an authkey to an api?
seems like unnecessary calls and just something else to get clobbered.
hash of the homeid should be enough to track and block bad clients that aren't out to get you

chrisns: failing that a WAF might not be a bad idea

packages/zwave-js/src/lib/telemetry/metrics.ts
AlCalzone: Doesn't really matter, we can do sha. The thought was: MD5 is shorter 🤷🏻‍♂️

packages/zwave-js/src/lib/telemetry/metrics.ts
AlCalzone: Like we discussed earlier - this should rather be a "client id" or "schema id" or "token". Just a simple publicly known value that can easily be changed in a config file and gets used to ignore every request by a client that's not on the latest version.

This does not mean that specific implementations shouldn't do some on-top validation serverside.

packages/zwave-js/src/lib/telemetry/metrics.ts
AlCalzone: gonna change that to `https://stats.zwave-js.io` without port

AlCalzone:
Everything addressed :)

docs/getting-started/telemetry.md
```diff
@@ -23,7 +23,7 @@ We also use [Sentry](https://sentry.io) to capture basic information about devic
 
 ## Usage statistics
 
-In order to gain insight how `zwave-js` is used, which manufacturers and devices are most prevalent and where to best focus our efforts in order to improve `zwave-js` the most, we collect statistics. That is **if the application developer has opted in** to this functionality. Since this data may seem more sensitive, we are extra careful - therefore we've built our own metrics stack which is self-hosted next to the Sentry instances. The reports include the following data:
+In order to gain insight how `zwave-js` is used, which manufacturers and devices are most prevalent and where to best focus our efforts in order to improve `zwave-js` the most, we collect statistics. That is **if the application developer has opted in** to this functionality. Since this data may seem more sensitive, we are extra careful - therefore we've built our own statistics stack which is self-hosted next to the Sentry instances. The reports include the following data:
```
robertsLando: May add a link to the repo here

robertsLando:
🚀

PR #2232, opened 2021-04-02
feat(config): add NIE Tech / Eva Logik ZW97

akennerly:


akennerly:
@zwave-js-bot fix lint

packages/config/config/devices/0x0312/z97.json
```diff
@@ -0,0 +1,95 @@
+// NIE Technology Co., Ltd. ZW97 / Eva Logik
+// Outdoor Z-Wave Plug-In Module (2-Channel)
+// TODO: ZW97 support Over temperature protection, When the temperature is greater than 105 degrees Celsius, the Heat Alarm signal is sent out Notification Type : Heat Alarm 0x04 . Event :Overheat detected 0x02
```
blhoward2: It's inappropriate to leave a TODO at the top of the file like this. Please remove it. If an alarm goes off it will be passed through automatically anyway.

packages/config/config/devices/0x0312/z97.json
```diff
@@ -0,0 +1,95 @@
...
+	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 1,
+			"isLifeline": true
+		},
+		"2": {
+			"label": "Outlet 1 (Left) State Notification",
+			"description": "When the state of Outlet 1 (Left) is changed, On (0xFF) or Off (0x00). The ZW97 will automatically send out a related command, On (0xFF) or Off (0x00) to its associated group.",
```
blhoward2: We don't generally use descriptions for lifelines and I don't beleive any of the UIs do anything with them. These also don't conform to the style guide.

packages/config/config/devices/0x0312/z97.json
```diff
@@ -0,0 +1,95 @@
...
+		"max": "255.255"
+	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 1,
+			"isLifeline": true
+		},
+		"2": {
+			"label": "Outlet 1 (Left) State Notification",
```
blhoward2: Change to just "Outlet 1 (Left)". Change the other one too.

packages/config/config/devices/0x0312/z97.json
```diff
@@ -0,0 +1,95 @@
...
+		},
+		"3": {
+			"label": "Outlet 2 (Left) State Notification",
+			"description": "When the state of Outlet 2 (Right) is changed, On (0xFF) or Off (0x00). The ZW97 will automatically send out a related command, On (0xFF) or Off (0x00) to its associated group.",
+			"maxNodes": 5
+		}
+	},
+	"paramInformation": {
+		"1": {
+			"label": "LED Invert/Off",
```
blhoward2: Change to "LED Indicator" for consistency with other files.

packages/config/config/devices/0x0312/z97.json
```diff
@@ -0,0 +1,95 @@
...
+				{
+					"label": "Outlet On / LED On",
+					"value": 0
+				},
+				{
+					"label": "Outlet On / LED Off",
+					"value": 1
+				},
+				{
+					"label": "LED Disabled",
```
blhoward2: Change to "LED disabled". Option labels are sentence case per the style guide.

packages/config/config/devices/0x0312/z97.json
```diff
@@ -0,0 +1,95 @@
...
+					"value": 2
+				}
+			]
+		},
+		"2": {
+			"label": "Countdown",
```
blhoward2: Make the label descriptive so it is clear what it applies to, then lose the description. Do the same for the others.

blhoward2 (changes requested):
Please look at the style guide and incorporate. Request a new review when you're ready. https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide

packages/config/config/devices/0x0312/z97.json
akennerly: OK

packages/config/config/devices/0x0312/z97.json
akennerly: Done

packages/config/config/devices/0x0312/z97.json
akennerly: I don't believe Groups 2 and 3 are considered lifelines.  "basic set command" is what the zwave alliance calls it. The style guide says to include a description if it adds value.  https://products.z-wavealliance.org/products/3701/assoc?noFilename=True  The description is rather verbose, I'll admit it can be simplified.
I wanted to convey that outlet 1 and 2 state notifications are sent on the respective groups.  I believe the intention was to allow other zw97s that are joined to group 2 or 3 to synchronize their outlet state.

AlCalzone: "Lifeline" is not the correct term, but we sometimes use it for additional association groups that should be associated to the controller

akennerly: I don't believe groups 2 and 3 would be associated with the controller. Most likely just other zw97s.

akennerly: @blhoward2 @AlCalzone Should I change the descriptions on groups 2 and 3? 

I just want those association groups to be have a meaningful description. Groups 2 and 3 wouldn't (or at least can't see a valid reason why it would) be associated with the controller.  Therefore not a "lifeline"(?)

AlCalzone: If there's a better label, go for it. Btw, I'm sure @blhoward2 meant "association group", not "lifeline".

blhoward2: I meant Lifeline as the name of Group 1 per the style guide.

blhoward2: Yes. I meant groups.

akennerly: @blhoward2  I removed the group descriptions. I think that's the last recommended change.

blhoward2 (approved):
Approved with my corrections pushed.

akennerly:
@blhoward2 Good to merge now?

AlCalzone (approved):

PR #2233, opened 2021-04-02
feat: add new node events to monitor interview progress

raman325:
As discussed in Discord, I added three new events:
- `interview started`
- `interview stage completed`
- ~~`not ready`~~

This will allow us to update the ready status in `zwave-js-server-python` as needed, and also will allow us to display information to the user about the last completed stage of interview during a re-interview or during startup

### DIFF
--- a/docs/api/node.md
+++ b/docs/api/node.md
@@ -166,6 +166,12 @@ Resets all information about this node and forces a fresh interview.
 
 > [!WARNING] Take care NOT to call this method when the node is already being interviewed. Otherwise the node information may become inconsistent.
 
+In `zwave-js-server`, calling this command will trigger a `not ready` event for the node. The node is passed as the single argument to the callback:
+
+```ts
+(node: ZWaveNode) => void
+```
+
 ### `interviewCC`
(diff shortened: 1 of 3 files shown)


AlCalzone (changes requested):
I don't like the `"not ready"` event. This only happens when a node is manually re-interviewed. If you need that event I'd prefer you add it in `zwave-js-server` and raise it whenever you call `refreshInfo` on a node. This could be mentioned in the docs for `refreshInfo` though.

The rest is fine.

raman325:
Do you have any guidance on the expected behavior of a node when it is not ready? Should we assume that the node needs to be left alone until the interview finishes, e.g. disable all interaction with the node?

AlCalzone:
> Should we assume that the node needs to be left alone until the interview finishes, e.g. disable all interaction with the node?

Yes. Like during the initial interview, you should not interact with a node that is not ready. So after calling `refreshInfo`, you need to wait for the `ready` event.

AlCalzone (approved):

PR #2235, opened 2021-04-03
fix(config): Fix Inovelli LZW31-SN Parameter 51

ahertz:
Parameter 51 ("Instant On") on the Inovelli LZW31-SN is currently described backwards: "Enable" disables instant on, and vice-versa.  Changed the description of the parameter to "Physical On/Off Delay", copied from the LZW31, which behaves correctly.  This way the two switches are also described consistently.


blhoward2:
This parameter has already been corrected in pending PR #2247.

PR #2236, opened 2021-04-03
fix(config): update Aeotec ZW056 config

jamesdawson3:
To allow playing ringtones on demand.  Allowable range for parameter 6 should be 0-100, not a set of options:

http://manuals-backend.z-wave.info/make.php?lang=en&sku=ZW056-C&cert=ZC10-15110019

### DIFF
--- a/packages/config/config/devices/0x0086/zw056.json
+++ b/packages/config/config/devices/0x0086/zw056.json
@@ -70,12 +70,10 @@
 		"6": {
 			"$import": "../templates/master_template.json#base_0-100_nounit",
 			"label": "Play Ringtone",
-			"options": [
-				{
-					"label": "Stop playing",
-					"value": 0
-				}
-			]
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 100,
+			"defaultValue": 0,
 		},
 		"7": {
 			"label": "Control Items",
--- a/packages/config/config/devices/0x0086/zw056.json
+++ b/packages/config/config/devices/0x0086/zw056.json
@@ -73,7 +73,7 @@
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 100,
-			"defaultValue": 0,
+			"defaultValue": 0
 		},
 		"7": {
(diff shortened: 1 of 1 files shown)


jamesdawson3:
@zwave-js-bot fix lint

AlCalzone (changes requested):
The referenced $import is defined as follows:
```
	"base_0-100_nounit": {
		"valueSize": 1,
		"minValue": 0,
		"maxValue": 100,
		"defaultValue": 1,
		"unsigned": true,
		"readOnly": false,
		"writeOnly": false,
		"allowManualEntry": true
	},
```
so all you'd have to change is `defaultValue: 0` if that is correct. Because of `allowManualEntry: true`, the option is just a hint/predefined value.

jamesdawson3:
Thanks for pointing me at the $import, that's helpful to understand.  Also largely makes my PR unnecessary.  Perhaps my issue is a Home Assistant implementation problem.  

The following is from my zwave_js_dump - which correctly matches the config file (with the $import):
`              {
                "endpoint": 0,
                "commandClass": 112,
                "commandClassName": "Configuration",
                "property": 6,
                "propertyName": "Play Ringtone",
                "ccVersion": 1,
                "metadata": {
                  "type": "number",
                  "readable": true,
                  "writeable": true,
                  "label": "Play Ringtone",
                  "default": 1,
                  "min": 0,
                  "max": 100,
                  "states": {
                    "0": "Stop playing"
                  },
                  "valueSize": 1,
                  "format": 1,
                  "allowManualEntry": true,
                  "isFromConfig": true
                }
              },`

However, when I try to set the parameter in HA, it fails if I set the value to anything other than 0:

`Failed to call service zwave_js/set_config_parameter. Must provide a value that represents a valid state key or label from {"0": "Stop playing"}`

Seems like HA is inappropriately requiring one of the defined states, if states are defined.

AlCalzone:
> Seems like HA is inappropriately requiring one of the defined states, if states are defined.

Looks like it. I think that's changed in the `.4` release, but I'm not 100% sure.

PR #2247, opened 2021-04-06
fix(config): clarify parameters 50 and 51 on inovelli lzw31

### LINKED ISSUE #2245: Instant On Setting is Backwards
https://github.com/zwave-js/node-zwave-js/blob/f295a1d041318e2b91c7664c9f91c3fd16eaf69f/packages/config/config/devices/0x031e/lzw31-sn.json#L703-L724

I haven't looked up the reference for this at Inovelli, but it seems to be backwards. As it works now, "on" allows for scenes and multi-tap functionality, "off" disables that and the switch functions instantly like a plain LZW31.

justindthomas:
fixes: https://github.com/zwave-js/node-zwave-js/issues/2245

These changes could make these settings more comprehensible.

### DIFF
--- a/packages/config/config/devices/0x031e/lzw31-sn.json
+++ b/packages/config/config/devices/0x031e/lzw31-sn.json
@@ -690,7 +690,8 @@
 		},
 		"50": {
 			"$if": "firmwareVersion >= 1.49",
-			"label": "Configure Physical On/Off Delay",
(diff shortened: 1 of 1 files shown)


blhoward2 (approved):
Approved with changes I pushed.

AlCalzone (approved):

PR #2254, opened 2021-04-06
fix: add statisticsEnabled property, salt homeID hash

AlCalzone:


packages/zwave-js/src/lib/telemetry/statistics.ts
```diff
@@ -13,15 +13,18 @@ export interface AppInfo {
...
+	const salt = await driver.getUUID();
 	return {
-		// Hash the homeId, so it cannot easily be tracked
+		// Salt and hash the homeId, so it cannot easily be tracked
+		// It is no state secret, but since we're collecting it, make sure it is anonymous
 		id: crypto
 			.createHash("sha256")
 			.update(driver.controller.homeId!.toString(16))
+			.update(salt)
 			.digest("hex"),
```
frenck, line 28: This is inefficient. The Home ID has 0 value anymore in this code.

Storing the SHA256 UUID and just send that, would work. there is no need to calculate a hash here at all in that case.

PR #2265, opened 2021-04-07
fix(config): update configuration for MCOHome Devices (various)

kanebullen:
Updated configuration files for various MCOHome devices


kanebullen:
Slight error - will try again.

PR #2283, opened 2021-04-08
fix(config): update FGRGBW-442 config

DKeeper:
- updates list of devices, updates some labels, adds unit type for time fields


DKeeper:
@zwave-js-bot rename commit

AlCalzone (changes requested):
The labels are already very long and your changes make them even longer.

I do think it is valuable to know which CCs are being sent, but we should find a middle ground between information and label length. Consider that these often go into dropdown fields that are just a few centimeters wide.

DKeeper:
@AlCalzone I reverted some labels, thanks for the clarification.

AlCalzone (approved):

PR #2291, opened 2021-04-09
fix(config): enable Basic Set mapping for FGBS001

XMaarten:
Adding enableBasicSetMapping to fix FGBS001. After the latest update it was not working anymore. I tried and applied the suggested fix by frescoast in the URL below, and this fixes the problem.

https://community.home-assistant.io/t/fibaro-universal-binary-sensor-fgbs001-not-working-when-upgrading-z-wave-js-to-0-1-16/296802

### DIFF
--- a/packages/config/config/devices/0x010f/fgbs001.json
+++ b/packages/config/config/devices/0x010f/fgbs001.json
@@ -353,6 +353,7 @@
 		}
 	},
 	"compat": {
+		"enableBasicSetMapping": true,
 		"commandClasses": {
 			"remove": {
 				// Alarm Sensor CC is reported as supported, but we either get no response


AlCalzone:
Maybe I'm missing something, but can you not just configure the device to send a better fitting command?
Like
grafik
or one of those
grafik

XMaarten:
I am not familiar with those settings, but I am not using the device to trigger any scenes or automations. The purpose for this device for me is to show the status of a particular switch (open or closed) within HomeAssistant. The previous version showed this status correctly, but after an update the Binary Sensor values (23-48-1-Any and 23-48-2-Any) didn't change anymore when changing the inputs on the FGBS001. The fix used in #2256 caused those values to change again, so I assumed that was the right fix for this problem as well.

AlCalzone:
I find it weird that a device that is called UNIVERSAL **BINARY** SENSOR is not using the `Binary Sensor CC` to communicate its status. But it looks like there isn't any other option for this.

AlCalzone (dismissed):

AlCalzone:
@zwave-js-bot rename commit

AlCalzone (approved):

tofuSCHNITZEL:
sorry to ask this here... but could we do the same config change for https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x0109/zd2102-5.json?
it seems to have a similar issue

PR #2292, opened 2021-04-09
feat: add driver method to return whether statistics are enabled

raman325:
This would be useful for both `zwavejs2mqtt` and `zwave-js-server`

### DIFF
--- a/docs/api/driver.md
+++ b/docs/api/driver.md
@@ -56,6 +56,14 @@ disableStatistics(): void
 
 Disable sending usage statistics.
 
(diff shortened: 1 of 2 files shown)


AlCalzone:
What's wrong with
https://github.com/zwave-js/node-zwave-js/blob/7425f4aead09a3f07b564ed23a00167d45560be6/packages/zwave-js/src/lib/driver/Driver.ts#L1243-L1245
?
:)

raman325:
doh!

PR #2294, opened 2021-04-09
docs: add `statisticsEnabled` property

raman325:
If I missed this in the docs as well I am packing it up for the day

### DIFF
--- a/docs/api/driver.md
+++ b/docs/api/driver.md
@@ -56,6 +56,14 @@ disableStatistics(): void
 
 Disable sending usage statistics.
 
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
You didn't miss it, but the property should be put where the `enable` and `disable` methods are :)

raman325:
> You didn't miss it, but the property should be put where the `enable` and `disable` methods are :)

that was the intention... 🤦🏾

AlCalzone (approved):

PR #2296, opened 2021-04-09
fix(config): add options to device status after power failure for hank switch

cgarwood:


firstof9:
Try `fix(config): add options to device status after power failure for hank switch` as the title.

AlCalzone (approved):

PR #2313, opened 2021-04-11
fix(config): update Remotec ZXT-600  config

geirsf:
Updated the Remotec ZXT-600 AC Master config file base on its [manual](https://github.com/zwave-js/node-zwave-js/files/6291775/6dacfa_70e9d19869b44299b27f501fd1673f85.pdf).

### DIFF
--- a/packages/config/config/devices/0x5254/zxt-600_1.3.json
+++ b/packages/config/config/devices/0x5254/zxt-600_1.3.json
@@ -30,20 +30,175 @@
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 32,
-			"defaultValue": 32,
+			"defaultValue": 0,
 			"readOnly": false,
 			"writeOnly": false,
-			"allowManualEntry": true
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "OFF",
+					"value": 0
+				},
+				{
+					"label": "ON (RESUME)",
+					"value": 1
+				},
+				{
+					"label": "17°C COOL ",
+					"value": 2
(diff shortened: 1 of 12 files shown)


packages/config/config/devices/0x5254/zxt-600_1.3.json
```diff
@@ -104,11 +311,62 @@
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 255,
-			"defaultValue": 5,
+			"defaultValue": 0,
 			"unsigned": true,
```
AlCalzone: Actually, if you remove `"unsigned": true`, this can be represented entirely without options.
Change the range to be -5 to 5 and restore `allowManualEntry` to `true`.
While you're at it, add `"unit": "°C"`

AlCalzone (changes requested):
Please check out our [Style Guide](https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide) - some of the labels need to be optimized a bit.

packages/config/config/devices/0x5254/zxt-600_1.3.json
geirsf: Just to be sure, will this work when -5 uses 251(0xFB)  as its value?

AlCalzone: Yes, -5 will be transmitted as 251

geirsf: Ah cool. Updated according to your suggestion.

PR #2333, opened 2021-04-13
fix: respect rate limiter response for usage statistics backoff

AlCalzone:


packages/zwave-js/src/lib/telemetry/statistics.ts
```diff
@@ -55,6 +61,16 @@ export async function sendStatistics(
...
+			if (
+				isObject(e.response.headers) &&
+				"retry-after" in e.response.headers
+			) {
+				const retryAfter = parseInt(e.response.headers["retry-after"]);
+				if (Number.isInteger(retryAfter)) return retryAfter;
+			}
+		}
 		// didn't work, try again later
 		return false;
```
robertsLando, line 75: What about adding a log here in case of error? Maybe just ignore network errors

AlCalzone: That's logged at the call site anyways.

PR #2349, opened 2021-04-15
fix(config): cleanup and template Aeotec configurations (part 2)

### LINKED ISSUE #2129: Aeotec ZW111/ZW116 Config - Basic Set Compat Flag needed
The Aeotec ZW111 nano single dimmer allows for two physical switches to be connected. They can be configured to either control the direct load, send an event or do both. This is configured in config parameter 22-112-0-81 and 22-112-0-82 and then completed by adding the destination node (in my case Node 1) to either Group - Control:Key 1 or Group - Control:Key 2 depending on what you would like each switch to do.

If configured to send an event only, they send a Basic Set (only option) which by default with zwavejs gets mapped into a report. In order to use this event to perform a secondary function it needs to be an event so that Home Assistant can be triggered. I was able to get this working by making a local change to the zw111_2.3.json device config adding the compat flag.

```
"compat": {
		"treatBasicSetAsEvent": true
	}
```

Can this be added to this configuration? This will likely also be needed for ZW116 nano switch which supports the same functionality.

Thanks

blhoward2:
Aeotec templating group 2. Closes #2129


packages/config/config/devices/0x0086/templates/aeotec_template.json
```diff
@@ -125,11 +125,11 @@
 		"allowManualEntry": false,
 		"options": [
 			{
-				"label": "Send Basic Set CC",
+				"label": "Basic CC",
 				"value": 1
 			},
 			{
-				"label": "Send Sensor Binary Report CC",
+				"label": "Sensor Binary Report CC",
```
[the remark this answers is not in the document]
AlCalzone: Let's try to be consistent with naming CC commands. I changed some of the names a bit from the specs to make them sound more natural.
https://github.com/zwave-js/node-zwave-js/blob/master/packages/core/src/capabilities/CommandClasses.ts#L7

So the schema should be
```
<CC name> CC <command name>
```

packages/config/config/devices/0x0086/templates/aeotec_template.json
```diff
@@ -150,11 +150,64 @@
...
+			{
+				"label": "Disable",
+				"value": 0
+			},
+			{
+				"label": "Basic CC",
+				"value": 1
+			},
+			{
+				"label": "Binary Report CC",
```
AlCalzone: Binary Sensor or Binary Switch?

AlCalzone: Are these all reports or is it using `Basic CC Set` and `Binary Sensor(?) CC Report` for example?

packages/config/config/devices/0x0086/templates/aeotec_template.json
```diff
@@ -150,11 +150,64 @@
...
+			{
+				"label": "Disable",
+				"value": 0
+			},
+			{
+				"label": "Basic CC",
+				"value": 1
+			},
+			{
+				"label": "MultiLevel CC",
```
AlCalzone: Multilevel Sensor or Multilevel Switch

packages/config/config/devices/0x0086/templates/aeotec_template.json
```diff
@@ -168,6 +221,39 @@
 		"writeOnly": false,
 		"allowManualEntry": true
 	},
+	"celsius_farenheit": {
```
AlCalzone:
```suggestion
	"celsius_fahrenheit": {
```

packages/config/config/devices/0x0086/templates/aeotec_template.json
```diff
@@ -168,6 +221,39 @@
...
+				"label": "Celsius",
+				"value": 1
+			},
+			{
+				"label": "Fahrenheit",
+				"value": 2
+			}
+		]
+	},
+	"celsius_farenheit_0": {
```
AlCalzone:
```suggestion
	"celsius_fahrenheit_0": {
```

packages/config/config/devices/0x0086/templates/aeotec_template.json
```diff
@@ -1127,5 +1323,529 @@
...
+		"minValue": 8,
+		"maxValue": 127,
+		"defaultValue": 15,
+		"unsigned": true,
+		"readOnly": false,
+		"writeOnly": false,
+		"allowManualEntry": true
+	},
+	"current_power_mode": {
+		"$if": "firmwareVersion >= 1.7 && firmwareVersion <= 1.9",
```
AlCalzone: This seems to be specific to one device. Let's at least mention which one, either as a comment or in the template name.

packages/config/config/devices/0x0086/templates/aeotec_template.json
```diff
@@ -1127,5 +1323,529 @@
...
+				"label": "Asleep",
+				"value": 0
+			},
+			{
+				"label": "Awake",
+				"value": 1
+			}
+		]
+	},
+	"switch_disconnet_s1": {
```
AlCalzone:
```suggestion
	"switch_disconnect_s1": {
```
but that doesn't seem to match the label? Same for the next one.

packages/config/config/devices/0x0086/templates/aeotec_template.json
```diff
@@ -1127,5 +1323,529 @@
...
+			},
+			{
+				"label": "Only recognize once when first power on",
+				"value": 1
+			},
+			{
+				"label": "Recognize the load once power on",
+				"value": 2
+			}
+		]
```
AlCalzone:
```suggestion
		"options": [
			{
				"label": "Never",
				"value": 0
			},
			{
				"label": "Only on first power on",
				"value": 1
			},
			{
				"label": "Every power on",
				"value": 2
			}
		]
```
would be shorter or is that not specific enough?

packages/config/config/devices/0x0086/templates/aeotec_template.json
```diff
@@ -1127,5 +1323,529 @@
...
+	},
+	// Appointment times
+	"base_appointment_day": {
+		"description": "Add values to select day(s) of week: Disable = 0; Monday = 1; Tuesday = 2; Wednesday = 4; Thursday = 8; Friday = 16; Saturday = 32; Sunday = 64",
```
AlCalzone: Maybe "Sum values" instead of "Add values"?

packages/config/config/devices/0x0086/zw120.json
```diff
@@ -56,22 +56,26 @@
...
 					"value": 1
 				},
 				{
-					"label": "Send Basic Set CC",
+					"label": "Sensor Binary Report CC",
 					"value": 16
+				},
+				{
+					"label": "Basic CC and Sensor Binary Report CC",
+					"value": 17
```
AlCalzone: See comments in template

AlCalzone (changes requested):

packages/config/config/devices/0x0086/templates/aeotec_template.json
```diff
@@ -125,11 +125,11 @@
 		"allowManualEntry": false,
 		"options": [
 			{
-				"label": "Send Basic Set CC",
+				"label": "Basic CC",
```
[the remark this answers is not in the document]
AlCalzone: Or is that clear from context?

AlCalzone (approved):

PR #2350, opened 2021-04-15
feat(config): add Zooz ZSE11

devbdavis:
I'm not sure if I did this correctly, but I took a stab at creating an initial config for the newly launched Zooz multi sensor. It isn't in any of the databases just yet, and is in somewhat of a "beta" release. I tested this with a physical ZSE11 device, and the parameters seem to take effect as expected.

If I can improve this PR, I am very welcome to feedback, or if this is something that should be handled in a different way, totally understand. I can work on a template if that is truly preferred, but it seems each Zooz sensor has pretty different configs. This unit would compare with the ZSE40, which has different inclusion/exclusion/wake instructions, different parameters, etc. Same manufacturer, same type of device, but fairly different in the configs.

**Which device is missing?**

Zooz ZSE11 "Q Sensor"

**What are the IDs?**

Manufacturer ID: 0x027a
Product type: 0x0201
Product ID: 0x0006

**Did you check if we really don't have it?**
- [ x ] It is not in the [configuration DB](https://devices.zwave-js.io/)
- [ x ] It was not [merged recently](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Apr+is%3Aclosed) or has a [pending PR](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Aopen+is%3Apr)
- [ x ] It is not in the [TODO list](https://github.com/zwave-js/node-zwave-js/issues/1600)

**Do you have a link to the PDF manual?**

Manual: https://cdn.shopify.com/s/files/1/0218/7704/files/zooz-q-sensor-zse11-ver1-manual-online.pdf
Store Page: https://www.thesmartesthouse.com/products/zooz-z-wave-plus-q-sensor-zse11-motion-temp-humidity-light

### DIFF
--- a/packages/config/config/devices/0x027a/zse11.json
+++ b/packages/config/config/devices/0x027a/zse11.json
@@ -1,0 +1,111 @@
+// Zooz ZSE11
+// Q Sensor
+{
+	"manufacturer": "Zooz",
+	"manufacturerId": "0x027a",
+	"label": "ZSE11",
+	"description": "Q Sensor",
+	"devices": [
+		{
+			"productType": "0x0201",
+			"productId": "0x0006"
+		}
+	],
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"metadata": {
+		"inclusion": "Put your Z-Wave hub into inclusion mode and click the Z-Wave button 3 times as quickly as possible. The LED indicator will start blinking to confirm inclusion mode. The sensor will automatically pair as a repeater if connected to USB power, no special button sequence required.",
+		"exclusion": "Bring the sensor within direct range of your Z-Wave gateway (hub). Put the Z-Wave hub into exclusion mode. Press and release the Z-Wave button 3 times quickly.",
+		"wakeup": "Press and hold the Z-Wave Button for 3 seconds to wake the sensor up manually"
+	},
+	"paramInformation": {
+		"12": {
+			"label": "Motion Sensor Sensitivity",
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x027a/zse11.json
```diff
@@ -0,0 +1,111 @@
...
+				{
+					"label": "Off",
+					"value": 0
+				},
+				{
+					"label": "Flash LED on Motion",
+					"value": 1
+				}
+			]
+		},
```
AlCalzone: Just a small nit:
```suggestion
		"19": {
			"label": "Flash LED on Motion",
			"valueSize": 1,
			"minValue": 0,
			"maxValue": 1,
			"defaultValue": 1,
			"readOnly": false,
			"writeOnly": false,
			"allowManualEntry": false,
			"options": [
				{
					"label": "Disable",
					"value": 0
				},
				{
					"label": "Enable",
					"value": 1
				}
			]
		},
```
allows us to make these on/off parameters booleans (toggles) in future versions.

AlCalzone (changes requested):

packages/config/config/devices/0x027a/zse11.json
devbdavis: Good eye, and thank you for using the suggestion feature. I haven't seen these tools before, but you make PRs very easy to manage with the bots + suggestions!

AlCalzone (approved):

Issue #2311, opened 2021-04-10
Missing device config: 0x0063:0x4952:0x3036:5.22


AlCalzone:
**What are the IDs?**

Manufacturer ID: 0x0063  
Product type: 0x4952  
Product ID: 0x3036  
Firmware version: 5.22

**Additional info:**

supportsAGI: true
associationGroups: Lifeline, On/Off control, On/Off control
supportsConfigCCV3: false
supportsZWavePlus: true
zWavePlusVersion: 1

marcus-j-davies:
mis reported

https://devices.zwave-js.io/?jumpTo=0x0063:0x4952:0x3036:5.22

PR #2353, opened 2021-04-15
feat(conf): add GE/Jasco 14291/ZW4005

geirsf:
As requested here #2311,  based on  https://products.z-wavealliance.org/products/2501/ and https://manual.zwave.eu/backend/make.php?lang=en&sku=14291/ZW4005&cert=ZC10-17075672


geirsf:
Seems like its already here: packages/config/config/devices/0x0063/ge_14291_zw4005.json

PR #2378, opened 2021-04-16
fix(config): incorrect selective reporting labels for aeon home energy meters

brimmasch:
Amps are a measurement of current.  Watts are a measurement of power.  These parameters are all measuring watts or percentage of watts change so the label must be power especially since there are other reports that deliver current in amps messages.

### DIFF
--- a/packages/config/config/devices/0x0086/dsb09.json
+++ b/packages/config/config/devices/0x0086/dsb09.json
@@ -59,16 +59,16 @@
 			"label": "Selective Reporting"
 		},
 		"4": {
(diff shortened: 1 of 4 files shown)


brimmasch:
I don't know what a "Semantic PR title / commit message" is.  If somebody wouldn't mind doing an edit so I can get in and out and contribute a fix without having to learn all of the culture.

blhoward2:
I'll give you that this is true and should be fixed. A lesson in not trying to remember high school electronics class material at 2 AM.

blhoward2 (approved):

AlCalzone (approved):

PR #2382, opened 2021-04-16
feat(conf): Add assoc. groups to Namron 2 kanaler bryter

jmgiaever:
Added the missing association groups for this device.

2. https://www.elektroimportoren.no/docs/lib/4512720-Brukerveiledning-5.pdf
See page 2 for groups.

### DIFF
--- a/packages/config/config/devices/0x0438/2_kanaler_bryter.json
+++ b/packages/config/config/devices/0x0438/2_kanaler_bryter.json
@@ -15,6 +15,21 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 5,
+			"isLifeline": true
+		},
+		"2": {
+			"label": "Launch 1",
+			"maxNodes": 5
+		},
+		"3": {
+			"label": "Launch 2",
+			"maxNodes": 5
+		}
+	},
 	"paramInformation": {
 		"1": {
 			"label": "Factory Reset",


AlCalzone (commented):
Maybe this is not necessary. The device supports Z-Wave Plus and Association Group Info CC, so this info can be discovered automatically. We do still specify the groups to improve bad labels, so it depends on how the associations are shown without this change. Can you share that?

jmgiaever:
Hi,

It's seems like you are correct. However, the reason I added it was because the assoc.groups list showed up empty yesterday when I added the switch. But now they're there! 

I haven't re-interviewed the nodes, so why they didn't show up yesterday is a mystery to me.

jmgiaever:
Can add: Turned out it was zwavejs2mqtt that needed a fullpage refresh to get the list of associations populated. Just re-tried with another remote now.

AlCalzone:
Can you still share how the associations are discovered? Wonder if the labels are okay or if we need your change after all.

jmgiaever:
Can you elaborate? Attach the log from the inclusion process?

AlCalzone:
Just a Screenshot of the available association group should be fine

jmgiaever:
Screenshot from 2021-04-17 08-58-34
After inclusion :arrow_up: 

Screenshot from 2021-04-17 11-38-43
After a page refresh **1 kanal** :arrow_up: 

Screenshot from 2021-04-17 11-36-52
After a page refresh **2 kanaler** :arrow_up: 

But looks like @robertsLando already pushed a PR for a fix. I'm testing it out asap and keep you posted. Thanks!

EDIT: Sorry, took screenshot of the wrong node. However this occurred for all of them. I have 3, 1 «1 kanal» and 2 «2 kanaler» remotes.

jmgiaever:
Just a quick notice though; Looks like the «product name» is in the «product code», and the product name is empty? 

The codes should be
1. kanal: 4512712
2. kanaler: 4512720

PR #2383, opened 2021-04-16
feat(conf): Add missing assoc. groups for 1 kanal bryter

jmgiaever:
Added the missing association groups.

1. https://www.elektroimportoren.no/docs/lib/4512712-Brukerveiledning-5.pdf


AlCalzone (commented):
See https://github.com/zwave-js/node-zwave-js/pull/2382#pullrequestreview-638085118

PR #2391, opened 2021-04-17
fix(config): correct Linear wd500z

cdnninja:
Corrected to spec for the Linear wd500z.  Could you please confirm zwave js will identify this as different than the nortek?  Only difference is a z-wave alliance ID.  Device IDs are identical.  Only updating as it already existing in zwave JS. https://github.com/zwave-js/node-zwave-js/blob/541ddb486173882b99f908cd437982367e15379a/packages/config/config/devices/0x014f/wd500z-1.json

### DIFF
--- a/packages/config/config/devices/0x0014/wd500z.json
+++ b/packages/config/config/devices/0x0014/wd500z.json
@@ -1,7 +1,7 @@
-// Cyberhouse WD500Z
+// Linear WD500Z
 // Dimmer Switch
 {
-	"manufacturer": "Cyberhouse",
+	"manufacturer": "Linear",
 	"manufacturerId": "0x0014",
 	"label": "WD500Z-1",
 	"description": "Dimmer Switch",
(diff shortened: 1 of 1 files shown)


AlCalzone:
> Could you please confirm zwave js will identify this as different than the nortek?

Yes, the manufacturerId is different - so it is a different device as far as zwave-js is concerned.

AlCalzone (changes requested):
I know you didn't touch that, but the device descriptions reads
`"description": "Linear -1",`
Is that correct? Seems weird.

cdnninja:
Hard to get details on it.  From what I can tell they are actually called WD500Z-1 so I assume that is the reference.  I will update the details. https://products.z-wavealliance.org/products/1032?selectedFrequencyId=2

blhoward2:
This is a dimmer switch, right? I've changed the description as we try not to repeat the model number. Also, this is marketed under the Linear name, right? If so that should be the brand indicated.

blhoward2 (approved):

AlCalzone (approved):

PR #2392, opened 2021-04-18
feat(config): add compat flag to GE zw3008 to re-enable basic command events to replicate central scene functionality

DaveCo1701:
Although the GE zw3008 does not support double taps, it is still useful to trigger an event on a single button push if the user want to manually trigger actions within Home Assistant.

### DIFF
--- a/packages/config/config/devices/0x0063/26932_26933_zw3008.json
+++ b/packages/config/config/devices/0x0063/26932_26933_zw3008.json
@@ -377,5 +377,8 @@
 				}
 			]
 		}
+	},
+	"compat": {
+		"treatBasicSetAsEvent": true
 	}
 }
--- a/packages/config/config/devices/0x0063/26932_26933_zw3008.json
+++ b/packages/config/config/devices/0x0063/26932_26933_zw3008.json
@@ -378,8 +378,8 @@
 			]
 		}
 	},
-	// Add compat flag to re-enable basic command events to replicate central scene functionality. 
-	// Allows user to detect switch button presses as events when zwave controller is added to Group 2 or 3. 
+	// Add compat flag to re-enable basic command events to replicate central scene functionality.
+	// Allows user to detect switch button presses as events when zwave controller is added to Group 2 or 3.
 	"compat": {
 		"treatBasicSetAsEvent": true
 	}
--- a/packages/config/config/devices/0x0063/26932_26933_zw3008.json
+++ b/packages/config/config/devices/0x0063/26932_26933_zw3008.json
@@ -378,6 +378,8 @@
 			]
 		}
 	},
+	// Add compat flag to re-enable basic command events to replicate central scene functionality. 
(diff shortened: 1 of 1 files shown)


AlCalzone:
This **seems** incorrect. How are you getting the updated switch status then?

DaveCo1701:
The goal is to have an event fire off when a button is pushed (even if it is only a single button press - association group 2 or 3 on this switch). Normal status updates still take place via the lifeline association group back to the hub. What I am trying to reproduce is the "central scene" events that newer switches generate. For example, on a newer switch, when I turn it it on locally, the switch physically turns on the light, but I also get a single tap central scene notification. This can be used in various automatons (in my use case, it disables automatic light control).

I've tested this and it works for my use case. I am admittedly unsure if there are any knock on consequences I'm not seeing or considering.

AlCalzone:
I see. We could do this, but I still want to understand the use case a bit better. Adding the controller to all groups seems redundant and you're generating extraneous messages in your network.

Is this not possible by simply watching `currentValue` for changes?

blhoward2:
I'm not sure about this. Why wouldn't you just fire off the state engine instead of the event service? New state is on or new state is off will fire off any state change.

DaveCo1701:
Here's my use case:

I have a light in the bedroom that turns on when it is dark and motion is detected. If it is before 9 PM, it turns on to 100% and stays on. If it is after 9 PM it will turn on to 10% when motion is detected and then turn off. I call this auto mode.

If someone goes to bed early, I want to disable this auto mode. The most user friendly way I found to do this was to disable the automation for the remainder of the day if the off button is pushed within 15 seconds of the light being automatically turned on. 

Since there is an automation that turns off the light, I need a way to uniquely detect that the off button was physically pushed. I don't want to use feedback as that would change whether the light was turned off locally or by an automation. I don't want to disable the automation if the light was turned off by an automation.

I realize this is likely a fringe use case, which is why I think that the controller shouldn't be automatically associated with GROUP 2 (or GROUP 3, as both do the same thing for this switch). Just enabling this compatibility flag allows those that need to detect unique button pushes a route to enable it.

If you see another alternative to achieve the same thing I'm open to ideas. I used to run OpenZWave and Home Assistant and successfully used this setup, which is why I'm trying to recreate the functionality. If this change doesn't align with the design philosophy of zwave-js, I understand and will figure out something else.

Side note: zwave_js has been running very well on my system. Just want to say a quick thank you to you and the other devs!

blhoward2:
You can just trigger off state and leave new/old blank and it'll fire off both. Whether it was local or not I suppose wouldn't show that way.

blhoward2:
@AlCalzone It's your call. The compat flag doesn't do anything here unless the controller is added to groups two or three, which the user would have to do. So it is still an opt in feature. They aren't losing functionality.

blhoward2 (dismissed):

AlCalzone (dismissed):

AlCalzone:
@zwave-js-bot fix lint

AlCalzone (approved):

PR #2399, opened 2021-04-19
fix(config): correct LZW31-SN dimming and ramp labels

eddiezane:
After some hair pulling and digging it turns out that the original documentation for parameter 2 and 4 are mislabeled for the Inovelli LZW31-SN (Red Dimmer).

I was really confused why setting these parameters wasn't changing what I expected until I realized that they were swapped. 1 & 3 are for z-wave control and 2 & 4 are for wall switch control.

After some digging I came across these forums posts. Not sure why the [official docs](https://support.inovelli.com/portal/api/kbArticles/407409000001873397/attachments/1abnfa5822ce11f1e4dfe9a7306effa38a8cf/content?portalId=1ce1db55e4825f16146555b24f0f1753cbf5f7d83ca0a3d8b31013b2a2d69e1f&inline=true) haven't been updated yet.

https://community.inovelli.com/t/lzw31-manual-dimming-not-working/1315/9
https://community.inovelli.com/t/device-page-inovelli-dimmer-red-series-lzw31-sn/2272
https://community.inovelli.com/t/dimming-speed-and-ramp-rate-parameter1-and-parameter3-are-swapped-fixed-in-2020-08-12-driver/3529/14

### DIFF
--- a/packages/config/config/devices/0x031e/lzw31-sn.json
+++ b/packages/config/config/devices/0x031e/lzw31-sn.json
@@ -18,7 +18,7 @@
 	"supportsZWavePlus": true,
 	"paramInformation": {
 		"1": {
-			"label": "Dimming Speed",
+			"label": "Dimming Speed (Z-Wave)",
 			"unit": "seconds",
 			"valueSize": 1,
 			"minValue": 0,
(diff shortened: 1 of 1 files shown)


eddiezane:
cc @InovelliUSA

AlCalzone (changes requested):
We're using `Manual` instead of `From Switch` in other locations. Please do the same here for consistency.
Please also add ` (Z-Wave)` to the parameters 1 and 3, so it is clear which param changes what.

eddiezane:
Changes made

AlCalzone (approved):

AlCalzone:
Thanks!

PR #2402, opened 2021-04-20
fix(cc): persist Basic CC values when requested using Basic CC API

trantamjp:
Possibly a bug when doing interview as below:

My device is `WADWAZ-1`. I am pretty sure it supports BasicCC (use Z-Wave PC Controller and has confirmed it). But whatever I interview it, I always get the message `No response to Basic Get command, assuming the node does not support Basic CC...`

Now, I checked the code [Node.js](https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/node/Node.ts) and I found that during the interview, it refreshes the values (without storing the value) and right after that, it checks if there is a value from the `ValueDB`.

Here is the detail: the method at the line 174 [await this.refreshValues()](https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/BasicCC.ts#L174) does not store the value into `ValueDB` so at the line 186 [this.getValueDB().getValue](https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/BasicCC.ts#L186), it will return `undefined` and think it doesn't support BasicCC.

Below is the fix for `BasicCC.ts`. I haven't check other `CC` if it actually uses the values **_right after_** doing `refreshValues`. Please verify other `CC`.

Because of this issue, many devices have to inject the `enableBasicSetMapping`. Without `enableBasicSetMapping`, because the driver already removes the BasicSet for that device already, the driver just caches the value and does nothing (which means HA doesn't get any state update)

My device `WADWAZ-1` didn't work until 13 days ago until this commit [add enableBasicSetMapping compat flag to wadwaz-1 and wapirz-1](https://github.com/zwave-js/node-zwave-js/commit/ff376539643b33e572d93617827f709168d922f3). But, if the driver had handled interview correctly, we didn't need to set `enableBasicSetMapping` for wadwaz-1 and wapirz-1

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/BasicCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/BasicCC.ts
@@ -139,6 +139,11 @@ export class BasicCCAPI extends CCAPI {
 			this.commandOptions,
 		);
 		if (response) {
+			const valueDB = this.endpoint.getNodeUnsafe()?.valueDB;
+			valueDB?.setValue(
+				getCurrentValueValueId(this.endpoint.index),
+				response.currentValue,
+			);
 			return pick(response, ["currentValue", "targetValue", "duration"]);
 		}
 	}
@@ -224,10 +229,6 @@ current value:      ${basicResponse.currentValue}`;
 target value:       ${basicResponse.targetValue}
 remaining duration: ${basicResponse.duration?.toString() ?? "undefined"}`;
 			}
-			this.getValueDB().setValue(
-				getCurrentValueValueId(this.endpointIndex),
-				basicResponse.currentValue,
-			);
 			this.driver.controllerLog.logNode(node.id, {
 				endpoint: this.endpointIndex,
 				message: logMessage,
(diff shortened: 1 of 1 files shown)


blhoward2:
Have you tried building this and running it, without the compat flag, to see if it affects the need for the mapping? Have you tried under the most current version of node-zwave-js?  We reverted the change after the PR you cite that undid the mapping in #2288.

AlCalzone:
Oh, I must have messed that up when I refactored the interview to be separate form refreshing values. The fix is **almost** correct, but IMO it should go into `BasicCCAPI.get` instead of `refreshValues`. Inside the `if (response)` branch.

This way, every `get` for the Basic CC values will result in the value being stored.

AlCalzone (approved):

AlCalzone:
Thanks!

trantamjp:
Now, next question is ... should `enableBasicSetMapping` be reverted on some devices config because now we have a correct interview if a device supports BasicCC.

AlCalzone:
`enableBasicSetMapping` enables us to treat `Basic Set` as a `Binary Sensor Report` for example, so it does make sense to keep that. Unless there's one device where it doesn't actually make sense to map.

PR #2449, opened 2021-04-24
feat: update embedded configuration on demand

AlCalzone:
For a while now, we were already releasing nightly updates for `@zwave-js/config` if the only changes since the last release affect config files. Now we have the ability to install those on demand. To do so, two new methods were added to the driver:

```ts
checkForConfigUpdates(): Promise<string | undefined>
```
can be used to check if there is a compatible update and use that information in UIs.

```ts
installConfigUpdate(): Promise<boolean>
```
will attempt to install the available update. If it succeeds, the new configuration files are automatically loaded.
Both `npm` and `yarn v1` are supported and can be used to install the update.

Note that although the updated config gets loaded after the update, bugfixes and changes to device configuration might still require either a driver restart or re-interview of the changed devices to take effect.


AlCalzone:
@zwave-js-bot pack this

AlCalzone:
@zwave-js-bot pack this

AlCalzone:
Ironed out a few kinks, but I cannot actually test this without a new release and a followup nightly config release.

So we'll release it as experimental.

docs/api/driver.md
```diff
@@ -236,6 +236,24 @@ getLogConfig(): LogConfig
 
 Returns the current logging configuration.
 
+### `checkForConfigUpdates`
+
+```ts
+checkForConfigUpdates(): Promise<string | undefined>
```
robertsLando, line 242: @AlCalzone Forgot to document `silent` var here

AlCalzone: That's okay. It is only used internally so we don't get duplicate logging output if you first check and then update.

PR #2457, opened 2021-04-25
fix(config): update zw97 device config for 2nd gen EvaLogik hardware

akennerly:
NIE Technology Co., Ltd. / Eva Logik ZW97 - Update for new hardware revision

### DIFF
--- a/packages/config/config/devices/0x0312/z97.json
+++ b/packages/config/config/devices/0x0312/z97.json
@@ -24,11 +24,11 @@
 		},
 		"2": {
 			"label": "Outlet 1 (Left)",
-			"maxNodes": 5
+			"maxNodes": 2
 		},
 		"3": {
 			"label": "Outlet 2 (Right)",
-			"maxNodes": 5
+			"maxNodes": 2
 		}
 	},
 	"paramInformation": {
--- a/packages/config/config/devices/0x0312/z97.json
+++ b/packages/config/config/devices/0x0312/z97.json
@@ -57,11 +57,11 @@
 			]
 		},
 		"2": {
-			"label": "Outlet 1 (Left) Auto Shutoff",
-			"valueSize": 2,
-			"unit": "seconds",
+			"label": "Outlet 1 (Left) Auto Off Timer",
+			"valueSize": 4,
+			"unit": "minutes",
 			"minValue": 0,
-			"maxValue": 36000,
+			"maxValue": 65535,
 			"defaultValue": 0,
 			"unsigned": true,
 			"readOnly": false,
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
If this is a new hardware revision, there surely should be something to distinguish the old and new one?
Like a different device identifier or a new firmware version?

akennerly:
@AlCalzone I contacted the manufacturer and couldn't get a straight answer on differences between firmware versions and/or hardware revisions. You took a look in the statistics and found firmwares 1.0 and 1.1.  "Under the device ID you posted we only have reports for firmwares 1.0 and 1.1.  No reports for missing device config that could tell us more".  

@blhoward2 "We can try to look in the stats. If we absolutely can’t figure it out I’d go with the more inclusive version as it’s less likely the older devices are still in use"

For now it seems like we should change the config to the version of a product that I have and wait to see if there are any issues submitted.  I'll take it upon myself to submit a PR for the differences if someone has a ZW97 with the same code 0x0312:0xc000:0xc007 and the other set of parameters from the older documentation the manufacturer has posted in the past.

AlCalzone:
Ah, I didn't make the connection here. Remind me again, where do your changes come from? A manual?

Which firmware version does your device have? The manual I find here matches our old config file https://products.z-wavealliance.org/products/3701?selectedFrequencyId=2 and that one wasn't certified too long ago.

Could you share a log of a re-interview of your device, so I can take a look if there's anything else we could use to distinguish these?

akennerly:
@AlCalzone 

Latest parameters come from here: https://github.com/akennerly/zw97_manuals/blob/main/ZW97-manual-2020.7.15.pdf
I tested all of these parameters on my device except for parameter 6. I have firmware 1.1 on both of my devices.  Purchased about 1 month apart in November and December 2020.

I based the first version of the ZW97 device config off of the z-wave alliance link and it's not correct at all. Parameters 2 and 3 are 4 bytes instead of 2 on my devices. Parameters 4 thru 6 don't exist on the Z-wave alliance version but they do on mine.

There is potentially a newer version of the product too with the same parameters but slightly different plug shape. I have the manual for that revision too. I'm placing all of the versions of this product in one place: https://github.com/akennerly/zw97_manuals

Re-interview: https://gist.github.com/akennerly/3fcfbf2160716f46db3f626883dc75f5

AlCalzone:
Hm, I think the main problem is that we cannot detect whether a device is the old or the new revision. In that case I'm okay with having a config for only the new one.
@blhoward2 what do you think?

AlCalzone (approved):

blhoward2:
Could someone explain why we think fw 1.1 isn't always the new hardware model? That wouldn't show as an unknown device because the current file matches for all fw versions. Shouldn't we condition the new parameters to fw > 1.1 and then see if anyone complains?

AlCalzone:
Because there is an old manual which also references Firmware 1.1 but has the old parameters.

Issue #1596, opened 2021-02-04
Add support for Fantem Door/Window Sensor and MultiSensor


akoolenbourke:
Fantem OEM Aeotec devices so should have the same capabilities. These are the values for the sensor that ZWave JS shows in the Manufacturer Specific section (Using the ZWaveJS2MQTT addon)

Door/Window Sensor (Fantem)
Man: 362 == 0x16a 
Prod Type: 514 == 0x202
Prod ID: 112 == 0x4a

Motion Sensor
Man: 362 == 0x16a
Prod Type: 514 == 0x202
prod Id: 100 == 0x64

AlCalzone:
I've added a new issue template to get a hold of these kinds of requests to make it easier for us. Could you please add the missing information to your request:

---

(first two points are done)

**Did you check if we really don't have it?**

- [ ] It is not in the [configuration DB](https://devices.zwave-js.io/)
- [ ] It was not [merged recently](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Apr+is%3Aclosed) or has a [pending PR](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Aopen+is%3Apr)
- [ ] It is not in the [TODO list](https://github.com/zwave-js/node-zwave-js/issues/1600)

**Do you have a link to the PDF manual?**

If so, it helps in creating the config file.

akoolenbourke:
Thanks.

Not in DB
Not merged or pending PR
It's not in the TODO List. The zwave alliance site lists it for China and US/Canada/Mexic but not for AUS/NZ which is my territory.

The door/window sensor and the multisensor are there for the Aeotec (OEM) items for my territory but they use different man, type, and prod ids it seems.

Door/Window
https://products.z-wavealliance.org/products/1616?selectedFrequencyId=4

MultiSensor
https://products.z-wavealliance.org/products/1404?selectedFrequencyId=4


[OomiMultiSensor.pdf](https://github.com/zwave-js/node-zwave-js/files/5928998/OomiMultiSensor.pdf)
[Oomi Door Window Sensor manual.pdf](https://github.com/zwave-js/node-zwave-js/files/5928999/Oomi.Door.Window.Sensor.manual.pdf)

cdnninja:
I have added the door window sensor as a pull request.  I will work on the Mutli sensor another day.

cdnninja:
[...]

PR #2458, opened 2021-04-25
feat(config): add support for Omni OOMIDW1

cdnninja:
Added support for Omni OOMIDW1 / Fantem FT112-A. Fixes a part of #1596.

### DIFF
--- a/packages/config/config/devices/0x016A/OOMIDW1.json
+++ b/packages/config/config/devices/0x016A/OOMIDW1.json
@@ -1,9 +1,9 @@
 // Omni OOMIDW1 and FT112-A
 // Door/Window Sensor
 {
-	"manufacturer": "Omni",
+	"manufacturer": "Fantem",
 	"manufacturerId": "0x016A",
-	"label": "OOMIDW1",
+	"label": "FT112-A",
 	"description": "Door/Window Sensor",
 	"devices": [
 		{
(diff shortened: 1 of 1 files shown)


AlCalzone:
I'm confused a bit by these imports, but I **think** they do use the Aeotec hardware, right?

packages/config/config/devices/0x016A/OOMIDW1.json
```diff
@@ -0,0 +1,55 @@
...
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 5,
+			"isLifeline": true
+		}
+	},
+	"paramInformation": {
+		"1": {
+			"$import": "../0x0086/templates/aeotec_template.json#invert_state_report"
```
AlCalzone, line 28: Just to avoid confusion, a short comment explaining why you import from Aeotec templates would be good. E.g.:
```suggestion
		// Fantem is an OEM for Aeotec devices
		"1": {
			"$import": "../0x0086/templates/aeotec_template.json#invert_state_report"
```

AlCalzone (changes requested):

packages/config/config/devices/0x016A/OOMIDW1.json
AlCalzone: Looks good otherwise

cdnninja: Great thanks.  I will do similar for the second device in the issues log and submit.  I wanted to get this one approved before moving on.

cdnninja:
Converting to draft.  I found two files that look to be duplicates.   This guy: https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x016a/ft112-k.json and https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x016a/ft100.json

cdnninja:
Replaced with #2489.

PR #2460, opened 2021-04-26
feat: add method to get all groups of a node and its endpoints

AlCalzone:
/cc @robertsLando

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -113,6 +113,7 @@ The following methods can be used to manage associations between nodes and/or en
 
 ```ts
 getAssociationGroups(source: AssociationAddress): ReadonlyMap<number, AssociationGroup>;
+getAllAssociationGroups(nodeId: number): ReadonlyMap<number, ReadonlyMap<number, AssociationGroup>>;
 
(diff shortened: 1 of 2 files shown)


robertsLando (approved):

robertsLando:
wondering if `getAllAssociationGroups` could be implemented in a better way, I mean I think that now there could be some repeated groups on endpoints right?

If node has multiple endpoints I could have group X be listed in both endpoits

wouldn't it be better to have a `endpoints`  property in metadata along with `maxNodes`, `isLifeline` and `multiChannel` that will be a `number[]` array of endpoints that support that group?

AlCalzone:
> If node has multiple endpoints I could have group X be listed in both endpoits

Yes, and they could be different association groups. Here's a Qubino device for example:
grafik

The problem is we don't know which group on the endpoints correspond to which group on the root without reading and understanding the manual. Then there are some devices which have the groups only on the endpoints, but they are not related.

Also, there **could** be a difference in how a device handles an association that's defined on the root versus when its defined on an endpoint instead.

robertsLando:
Ok no problem got it :) Thanks for the explaination

PR #2463, opened 2021-04-26
fix(config): correct manufacturer name for mp20z

highaltidude:
Manufacturer was wrong in config file


blhoward2:
@zwave-js-bot rename commit

blhoward2 (approved):

PR #2468, opened 2021-04-27
feat(config): add new variant of param 52 for LZW31-SN v1.54+

kreene1987:
Inovelli has released beta firmware 1.54 which adds a 3rd option (value 2) for Smart Bulb Mode which keeps power on even if the switch is turned off. Wording matches exactly from SmartThings Device Handler revised documentation:

https://github.com/InovelliUSA/SmartThingsInovelli/blob/master/devicetypes/inovelliusa/inovelli-dimmer-red-series-lzw31-sn.src/inovelli-dimmer-red-series-lzw31-sn.groovy

Related community link:
https://community.inovelli.com/t/firmware-v1-54-beta-lzw31-sn-dimmer-red-series-gen-2/8132.

Note I did not split this out from 1.47+ (which is original options set of 0 or 1) as I don't know syntax on how to split this off for only firmware 1.54+. Option 2 is invalid for current production firmware.

### DIFF
--- a/packages/config/config/devices/0x031e/lzw31-sn.json
+++ b/packages/config/config/devices/0x031e/lzw31-sn.json
@@ -723,24 +723,28 @@
 			]
 		},
 		"52": {
-			"$if": "firmwareVersion >= 1.47",
+			"$if": "firmwareVersion >= 1.54",
 			"label": "Smart Bulb Mode",
-			"description": "Optimize power output to be more compatible with smart bulbs. This prevents the dimmer from being able to dim & makes it act like an ON / OFF switch.",
+			"description": "Change the working mode of the dimmer. Choose either on / off only which makes the device work like an on / off switch. Or choose smart bulb mode which optimized the device for interaction with smart bulbs. (firmware 1.54+)",
 			"valueSize": 1,
 			"minValue": 0,
-			"maxValue": 1,
+			"maxValue": 2,
 			"defaultValue": 0,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Disable",
+					"label": "Disabled",
 					"value": 0
 				},
 				{
-					"label": "Enable",
+					"label": "On / Off Only",
 					"value": 1
+				}
+				{
+					"label": "Smart Bulb",
+					"value": 2
 				}
 			]
 		}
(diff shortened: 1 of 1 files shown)


AlCalzone:
> I don't know syntax on how to split this off for only firmware 1.54+.

You have to use an array for the parameter
```jsonc
"52": [
	{
		"$if": "firmwareVersion >= 1.54",
		// definition for 1.54+
	},
	{
		"$if": "firmwareVersion >= 1.47",
		// definition for 1.47-1.53
	},
	// potentially others
]
```

kreene1987:
> > I don't know syntax on how to split this off for only firmware 1.54+.
> 
> You have to use an array for the parameter
> 
> ```js
> "52": [
> 	{
> 		"$if": "firmwareVersion >= 1.54",
> 		// definition for 1.54+
> 	},
> 	{
> 		"$if": "firmwareVersion >= 1.47",
> 		// definition for 1.47-1.53
> 	},
> 	// potentially others
> ]
> ```

@cgarwood Can you confirm the ordering will apply the first one it encounters that is true? Or do I need to order them 1.47+ THEN 1.54+?

AlCalzone:
> Can you confirm the ordering will apply the first one it encounters that is true?

Yes, but no application ever sees the conditional definitions. `node-zwave-js` resolves to the first matching one before exposing the final device config.

AlCalzone:
Oh and btw, fuck GitHubs new security feature. I have to manually approve every single PR run here.

kreene1987:
> Oh and btw, fuck GitHubs new security feature. I have to manually approve every single PR run here.

This is literally my first-ever pull request. So it might be related to my noobness.

Thanks for the fix!

AlCalzone:
> So it might be related to my noobness.

No, it is actually a new "feature" to avoid cryptomining in other people's actions. But no way to opt-out or whitelist uncritical files right now. 🙄

blhoward2:
While the driver may return the first match, I think the conditions in this manner will be confusing to users. There is also a risk we ever accidentially resort these while processing the files especially since we'd be likely to sort them low to high. @AlCalzone what are your thoughts as to whether we should make the applicable range clear by requiring the range be defined? 

`"$if": "firmwareVersion >= 1.47 && firmwareVersion <= 1.53",`

AlCalzone:
I intentionally made it so the conditions are evaluated from first to last. Basically `if, elseif, ..., else`.
If you're afraid that ranges are not explicit enough this way, we can keep an eye on it, but I don't think making that a lint rule is worth the effort.

AlCalzone (approved):

PR #2473, opened 2021-04-28
fix: make ObjectKeyMap iterable

AlCalzone:
it was missing the [Symbol.iterator] signature


robertsLando:
I was feeling so stupid wondering why this wasn't working 😆

PR #2484, opened 2021-04-28
fix(config): manual review for Ring Alarm keypad v2 config

xhemp:
Ring Keypad Gen2 configuration amended

### DIFF
--- a/packages/config/config/devices/0x0346/alarm_keypad_gen2.json
+++ b/packages/config/config/devices/0x0346/alarm_keypad_gen2.json
@@ -24,7 +24,7 @@
 	"paramInformation": {
 		"1": {
 			"label": "Heartbeat Interval",
-			"description": "This parameter is the number of minutes between heartbeats",
+			"description": "Number of minutes between heartbeats",
 			"valueSize": 1,
 			"unit": "minutes",
 			"minValue": 1,
--- a/packages/config/config/devices/0x0346/alarm_keypad_gen2.json
+++ b/packages/config/config/devices/0x0346/alarm_keypad_gen2.json
@@ -309,14 +309,24 @@
 		},
 		"22": {
 			"label": "System Security Mode Display",
-			"description": "Options are: 601 = Always On, 1 - 600 = periodic interval, 0 = Always Off, except when there's activity",
+			"description": "Options are: 601 = Always On (default), 1 - 600 = periodic interval, 0 = Always Off, except when there's activity",
 			"valueSize": 2,
 			"minValue": 0,
 			"maxValue": 601,
 			"defaultValue": 0,
 			"readOnly": false,
 			"writeOnly": false,
-			"allowManualEntry": true
+			"allowManualEntry": true,
+                        "options": [
+                                {
+                                        "label": "Always off",
+                                        "value": 0
+                                },
+                                {
+                                        "label": "Always on",
+                                        "value": 601
+                                }
+                        ]
 		},
 		"23[0x01000000]": {
 			"$import": "templates/ring_template.json#supported_language",
--- a/packages/config/config/devices/0x0346/alarm_keypad_gen2.json
+++ b/packages/config/config/devices/0x0346/alarm_keypad_gen2.json
@@ -384,6 +384,7 @@
 			"defaultValue": 0,
 			"readOnly": false,
 			"writeOnly": false,
+                        "allowManualEntry": false,
                         "options": [
                                 {
                                         "label": "Idle",
--- a/packages/config/config/devices/0x0346/alarm_keypad_gen2.json
+++ b/packages/config/config/devices/0x0346/alarm_keypad_gen2.json
@@ -28,6 +28,7 @@
 			"minValue": 1,
 			"maxValue": 70,
 			"defaultValue": 70,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
@@ -38,6 +39,7 @@
 			"minValue": 0,
 			"maxValue": 5,
 			"defaultValue": 1,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
@@ -49,6 +51,7 @@
 			"minValue": 1,
 			"maxValue": 60,
 			"defaultValue": 5,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
@@ -59,6 +62,7 @@
 			"minValue": 0,
 			"maxValue": 10,
 			"defaultValue": 7,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
@@ -69,6 +73,7 @@
 			"minValue": 0,
 			"maxValue": 10,
 			"defaultValue": 6,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
@@ -79,6 +84,7 @@
 			"minValue": 0,
 			"maxValue": 10,
 			"defaultValue": 10,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
@@ -90,6 +96,7 @@
 			"minValue": 2,
 			"maxValue": 5,
 			"defaultValue": 3,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
@@ -101,6 +108,7 @@
 			"minValue": 2,
 			"maxValue": 5,
 			"defaultValue": 3,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
@@ -112,6 +120,7 @@
 			"minValue": 0,
 			"maxValue": 30,
 			"defaultValue": 5,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
@@ -123,6 +132,7 @@
 			"minValue": 0,
 			"maxValue": 30,
 			"defaultValue": 5,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
@@ -134,6 +144,7 @@
 			"minValue": 1,
 			"maxValue": 30,
 			"defaultValue": 5,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
@@ -145,17 +156,19 @@
 			"minValue": 0,
 			"maxValue": 100,
 			"defaultValue": 100,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
 		},
 		"13": {
 			"label": "Key Backlight: Brightness",
-                        "description": "Used for key backlight LEDs only",
+			"description": "Used for key backlight LEDs only",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 100,
 			"defaultValue": 100,
+			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -285,42 +328,79 @@
...
 			"defaultValue": 0,
-			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
 		}
 	},
+        "compat": {
+                "disableStrictEntryControlDataValidation": true
+        },
```
AlCalzone: Are you sure that the V2 keypad still suffers from the same problem?

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -285,42 +328,79 @@
...
+			"$import": "templates/ring_template.json#supported_language",
+			"label": "Arabic"
 		},
 		"24": {
 			"label": "Calibrate Speaker",
 			"description": "This parameter is the speaker calibration value",
 			"valueSize": 1,
 			"minValue": 0,
-			"maxValue": 0,
+			"maxValue": 1,
```
AlCalzone, line 405: I assume 0 is idle and 1 is calibrate? I think we should add options for these.

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -273,9 +317,8 @@
 			"valueSize": 2,
 			"unit": "seconds",
```
AlCalzone: The description states milliseconds

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -223,34 +220,82 @@
...
 			"defaultValue": 30,
-			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
 		},
 		"18": {
+                        "$if": "productId === 0x0401",
 			"label": "Languages Set",
 			"description": "This parameter sets the language on the keypad",
```
AlCalzone:
```suggestion
			"label": "Keypad Language",
```

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -199,20 +188,28 @@
...
+                                        "label": "Enable",
+                                        "value": 1
+                                }
+                        ]
 		},
 		"16": {
 			"label": "Ramp Time Config",
 			"description": "This parameter is the ramp time to turn the leds on/off in milliseconds",
 			"valueSize": 1,
 			"unit": "seconds",
```
AlCalzone: The description states milliseconds

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -1,16 +1,19 @@
...
 	"manufacturerId": "0x0346",
-	"label": "KEYPAD V2",
-	"description": "Ring Keypad v2",
+	"label": "- / 4AK-1SZ-OEU0",
+	"description": "Alarm Keypad (2nd generation)",
 	"devices": [
 		{
 			"productType": "0x0101",
-			"productId": "0x0301",
-			"zwaveAllianceId": 3870
```
AlCalzone, line 12: We did import this from the ZWA website, so please keep the ID.

AlCalzone (changes requested):
Several of these parameter descriptions simply restate what the label already says. Those should be removed. And please make sure that 1. the configured units match what the description used to say and 2. the labels don't contain "(in seconds)" or similar.

AlCalzone:
@zwave-js-bot fix lint

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
xhemp: Added on both productIds

AlCalzone: ZWA only has the product ID 0x0301, so we should only add the reference to that one

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -19,302 +24,385 @@
 	"paramInformation": {
 		"1": {
 			"label": "Heartbeat Interval",
-			"description": "This parameter is the interval between heartbeats in minutes",
+			"description": "Number of minutes between heartbeats",
```
AlCalzone: I wasn't clear enough it seems. This description does not add any value, so it should be removed. The label already states that this is an interval between heartbeats. And the unit below states that the parameter is in minutes.

The same is true for several of the other descriptions.

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -19,302 +24,385 @@
...
 			"minValue": 0,
 			"maxValue": 10,
 			"defaultValue": 10,
-			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
 		},
 		"7": {
 			"label": "Long Press Emergency Duration (seconds)",
```
AlCalzone:
```suggestion
			"label": "Long Press Emergency Duration",
```

AlCalzone: What's the difference between this and param 8? If these reference specific buttons, I prefer labels like
`Emergency Button: Long Press Duration`
and
`Number Pad: Long Press Duration`

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -19,302 +24,385 @@
...
 			"defaultValue": 100,
-			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
 		},
 		"13": {
 			"label": "Key Backlight Brightness",
-			"description": "This parameter adjusts the brightness of the keypad backlight. used for key backlight leds only",
+			"description": "Adjusts the brightness of the keypad backlight. used for key backlight leds only",
```
AlCalzone:
```suggestion
			"description": "Used for key backlight LEDs only",
```

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -19,302 +24,385 @@
...
 			"minValue": 0,
 			"maxValue": 100,
 			"defaultValue": 100,
-			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
 		},
 		"14": {
 			"label": "Ambient Light Sensor Level",
```
AlCalzone:
```suggestion
			"label": "Key Backlight: Ambient Light Threshold",
```

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -19,302 +24,385 @@
...
 			"minValue": 0,
 			"maxValue": 100,
 			"defaultValue": 100,
-			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
 		},
 		"13": {
 			"label": "Key Backlight Brightness",
```
AlCalzone:
```suggestion
			"label": "Key Backlight: Brightness",
```

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -19,302 +24,385 @@
...
 			"defaultValue": 100,
-			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
 			"allowManualEntry": true
 		},
 		"14": {
 			"label": "Ambient Light Sensor Level",
-			"description": "This parameter is the light threshold where keypad will stop backlighting if higher than",
+			"description": "Light threshold where keypad will stop backlighting if higher than",
```
AlCalzone:
```suggestion
			"description": "Backlighting is turned off when the ambient light level is higher than this value",
```

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -19,302 +24,385 @@
...
+					"value": 0
+				},
+				{
+					"label": "Enable",
+					"value": 1
+				}
+			]
 		},
 		"16": {
 			"label": "Ramp Time Config",
```
AlCalzone:
```suggestion
			"label": "LED Ramp Time",
```

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -19,302 +24,385 @@
...
-			"description": "This parameter is the ramp time to turn the leds on/off in milliseconds",
+			"description": "Ramp time to turn the leds on/off",
```
AlCalzone: remove

AlCalzone (changes requested):
The labels and descriptions will need a second pass. I've added some suggestions that would be better - please carry throughout.

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
xhemp: No worries, I will change that. Some of them though are probably needed but let me know if you think they are not.

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
xhemp: Yes, that's it,  will change that.

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -326,8 +308,8 @@
 			"allowManualEntry": true
 		},
 		"22": {
-			"label": "System Security Mode Display - 601 = Always On,1 - 600 = periodic interval, 0 = Always Off, except when there's activity",
-			"description": "System security mode display options",
+			"label": "System Security Mode Display",
+			"description": "Options are: 601 = Always On, 1 - 600 = periodic interval, 0 = Always Off, except when there's activity",
```
AlCalzone: 0 and 601 should be added as predefined `"options"` instead. Also, this sounds like there should be a unit.

AlCalzone (changes requested):
Almost done - unless @blhoward2 has something to add

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
xhemp: Cool, I added the options but what about 1 - 600 ( Just leave it there as manual entry?)?  I also can't see a unit mentioned on the PDF.

AlCalzone: Yup, just leave it as manual entry. 0 and 601 will appear as suggestions in UIs which support this.
Since there doesn't seem to be a unit, the description or label should somehow mention that this configures an interval

packages/config/config/devices/0x0346/alarm_keypad_gen2.json
```diff
@@ -19,302 +23,367 @@
...
-			"description": "This parameter is the speaker calibration value",
 			"valueSize": 1,
 			"minValue": 0,
-			"maxValue": 0,
+			"maxValue": 1,
 			"defaultValue": 0,
-			"unsigned": true,
 			"readOnly": false,
 			"writeOnly": false,
-			"allowManualEntry": true
```
AlCalzone, line 317: allowManualEntry needs to exist (`false` in this case)

AlCalzone:
You're still missing allowManualEntry for param 24

AlCalzone (changes requested):
You're stumbling across functional and stylistic lint checks every time. I suggest working on this locally by following https://zwave-js.github.io/node-zwave-js/#/development/intro.
If you set up the environment correctly, VSCode will auto-format the config files for you, at the very least when you run the "format document" command.
If you execute `yarn run lint` (for style) and `yarn run lint:config` (for functional errors), you'll see what needs to be changed locally before pushing.

AlCalzone:
Just so we're clear: `unsigned` doesn't need to exist for every parameter - only the ones that go outside the signed range. It doesn't hurt to have it though, so you don't need to remove the unnecessary ones.

blhoward2:
Closed in favor of #2477 as I already verified that one against the manual.

PR #2491, opened 2021-04-29
fix: consider custom transports to determine loglevel visibility

### LINKED ISSUE #2490: Custom File Log Transports: No Entries when there is no interactive console.
As per discussion.

since 7.2.4 (where I first noticed).
Z-Wave JS is not writing into custom file transports.

BUT, the really odd thing is, it doesn't when there is no interactive console running the driver.
if the console is interactive, it all works fine (and pre 7.2.4 behaviour is restored)

Examples (all using a custom, shared file transport)
| Environment (macOS & RPI)                                                | Result                           |
| ----------------------------------------- | ---------------------- |
| Systemd Service                                        | Not Working                 |
| Interactive console                                          | Working                         |
| Interactive console (Redirected output)        | Not Working                    |

Oddly, when its working, Z-Wave JS is still outputting into console (despite no such transport).
this seems to be a new behaviour 

image

AlCalzone:
With this PR, we also consider custom transports to determine whether a log message will be visible.

fixes: #2490

### DIFF
--- a/packages/core/src/log/shared.ts
+++ b/packages/core/src/log/shared.ts
@@ -200,15 +200,16 @@ export class ZWaveLogContainer extends winston.Container {
 
 	/** Tests whether a log using the given loglevel will be logged */
 	public isLoglevelVisible(loglevel: string): boolean {
-		// If we are not connected to a TTY, not unit testing and not logging to a file, we won't see anything
-		if (isUnitTest) return true;
+		// If we are not connected to a TTY, not logging to a file and don't have any custom transports, we won't see anything
 		if (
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot pack this

marcus-j-davies:
@AlCalzone,

Logging seems to work again: tested using `app.js > /dev/null`
When not redirecting console, it still prints in console (and the transport) - but not sure if that is much of a problem

AlCalzone:
> When not redirecting console, it still prints in console (and the transport)

Yup, that is intended

PR #2492, opened 2021-04-29
feat(conf): add zso7300 to logic group

Roagert:


Roagert:
@zwave-js-bot fix lint

packages/config/config/devices/0x0234/templates/logic_group_template.json
```diff
@@ -526,5 +526,77 @@
...
+		"minValue": 0,
+		"maxValue": 500,
+		"defaultValue": 50,
+		"unsigned": true,
+		"readOnly": false,
+		"writeOnly": false,
+		"allowManualEntry": true
+	},
+	"watt_threshold_power": {
+		"label": "Watt Threshold: Power Off",
```
blhoward2: Does this label make sense? What's the difference between the two? What does Power Off mean?

blhoward2 (changes requested):

packages/config/config/devices/0x0234/templates/logic_group_template.json
Roagert: You can set an upper limit of wattage consumption. 
If this number is reached lets say 2900 out of 3000 as its max outpu. it can then trigger and turn off

packages/config/config/devices/0x0234/templates/logic_group_template.json
```diff
@@ -564,9 +557,9 @@
 		"allowManualEntry": true
 	},
 	"watt_threshold_power": {
-		"label": "Watt Threshold: Power Off",
-		"unit": "watt",
+		"label": "Power Change Threshold: Power Off",
```
AlCalzone: This refers to the total power, not the power change, does it?

Roagert: This refers to the total power output of the socket
I guess it should be called upper or upper limit or so

AlCalzone: Maybe just `Upper Power Limit`?

AlCalzone (approved):

PR #2532, opened 2021-05-03
feat(conf): add new power reporting parameter to ZEN25

jtscott:
Hello, this PR adds index 18 to the zen25 device. Parameter 18 was added in [firmware 2.0 by zooz](https://www.support.getzooz.com/kb/article/304-zen25%C2%A0s2-double-plug-change-log/) to control the reporting, as it tends to spam the zwave network.

### DIFF
--- a/packages/config/config/devices/0x027a/zen25.json
+++ b/packages/config/config/devices/0x027a/zen25.json
@@ -290,42 +290,41 @@
 					"value": 3
 				}
 			]
+		},
+		"18": {
+			"$if": "firmwareVersion >= 2.0",
+			"label": "Enable/Disable Power Reports",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 4,
+			"defaultValue": 0,
+			"readOnly": false,
+			"writeOnly": false,
+			"allowManualEntry": false,
+			"options": [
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x027a/zen25.json
```diff
@@ -292,6 +292,40 @@
 			]
 		}
 	},
+	"18": {
+		"label": "Enable/Disable Power Reports",
+		"description": "Enable or disable power reporting (Requires Firmware 2.0)",
```
AlCalzone: This should be modeled as a conditional parameter instead of the description: https://zwave-js.github.io/node-zwave-js/#/config-files/conditional-settings

AlCalzone (changes requested):

packages/config/config/devices/0x027a/zen25.json
jtscott: Awesome, I didn't know that. I pushed 5af7044 that I think resolves this. Thank you!

AlCalzone:
Okay with my small fix. Lets see what the bot has to say about this.

AlCalzone (approved):

PR #2546, opened 2021-05-05
fix(config): update Devolo Siren param 31

marcus-j-davies:
This value is actually used on the devices as multiples of 30 seconds.
1 = 30 seconds
2 = 60 seconds
3 = 90 seconds etc etc

Not sure if unit is best here, but let me know and I'll remove one or the other.
Have also added an Option (0) - which causes the siren to never stop on its own.

### DIFF
--- a/packages/config/config/devices/0x0175/ph-pse02.json
+++ b/packages/config/config/devices/0x0175/ph-pse02.json
@@ -80,7 +80,13 @@
 			"defaultValue": 6,
 			"readOnly": false,
 			"writeOnly": false,
-			"allowManualEntry": true
+			"allowManualEntry": true,
+			"options": [
+				{
+					"label": "Never stop.",
+					"value": 0
+				}
+			]
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0175/ph-pse02.json
```diff
@@ -72,14 +72,21 @@
 		},
 		"31": {
 			"label": "Alarm Duration",
-			"description": "Play alarm sound duration.",
+			"description": "Play alarm sound duration (30s * this value)",
+			"unit": "*30 seconds",
```
AlCalzone:
```suggestion
			"unit": "30 seconds",
```

packages/config/config/devices/0x0175/ph-pse02.json
```diff
@@ -72,14 +72,21 @@
...
 			"minValue": 0,
 			"maxValue": 127,
 			"defaultValue": 6,
 			"readOnly": false,
 			"writeOnly": false,
-			"allowManualEntry": true
+			"allowManualEntry": true,
+			"options": [
+				{
+					"label": "Never stop.",
```
AlCalzone:
```suggestion
					"label": "Never stop",
```

AlCalzone (changes requested):

AlCalzone (approved):
Okay with my changes

PR #2555, opened 2021-05-05
fix(config): change Aeotec Minimote config to writeOnly

ualex73:


ualex73:
@zwave-js-bot fix lint

blhoward2:
It's quite possible. I'll check tonight. Can you remove the readonly override though? It should already be false in the template. In fact, if the template is only used here the change should be made to the template not in the file.

ualex73:
@blhoward2 i have removed the readOnly lines.

blhoward2 (dismissed):
This template isn't used for any other files, so changing the write only in the template is the better way to do this. The manual is silent on whether this is write only but I do see that OZW has theirs as such. Approved with my changes.

AlCalzone (changes requested):
@blhoward2 maybe we should give the `base_mode_5` a better name that makes it clear that it's intended for remotes?

AlCalzone (approved):

PR #2571, opened 2021-05-08
fix(config): add Clamp 3 meters to DSB28, fix bitmask

booroondook-eng:
Added Clamp 3 meters and corrected 1 mistake

### DIFF
--- a/packages/config/config/devices/0x0086/dsb28.json
+++ b/packages/config/config/devices/0x0086/dsb28.json
@@ -98,11 +98,17 @@
 		"101[0x0200]": {
 			"$import": "templates/aeotec_template.json#auto_report_group1_watt_clamp2"
 		},
+		"101[0x0400]": {
+			"$import": "templates/aeotec_template.json#auto_report_group1_watt_clamp3"
+		},
 		"101[0x0800]": {
 			"$import": "templates/aeotec_template.json#auto_report_group1_kwh_clamp1"
 		},
 		"101[0x1000]": {
 			"$import": "templates/aeotec_template.json#auto_report_group1_kwh_clamp2"
+		},
+		"101[0x2000]": {
+			"$import": "templates/aeotec_template.json#auto_report_group1_kwh_clamp3"
 		},
 		"101[0x010000]": {
 			"$import": "templates/aeotec_template.json#auto_report_group1_v_clamp1"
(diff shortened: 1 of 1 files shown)


blhoward2:
Not all devices have three clamps and defining these for ones that don't can cause issues. Can you confirm your deviceType and deviceId? Firmware version? We're going to have to see if we can condition these somehow.

booroondook-eng:
> Not all devices have three clamps and defining these for ones that don't can cause issues. Can you confirm your deviceType and deviceId? Firmware version? We're going to have to see if we can condition these somehow.

2021-05-09 14_44_13-IMG_20210509_143852_9 jpg - IrfanView (Zoom_ 1224 x 918)
As you can see n the attached photo, my device has name DSB28-ZWEU and it has 3 clamps for 3 phases.
Also you can read its user manual: https://aeotec.freshdesk.com/support/solutions/articles/6000166671-home-energy-meter-g2-user-manual- (please read "Diagram of EU/AU Version(3 Phases,4 Wires)" article), and you can download  Excel table (HEM Calculator) with its parameters: https://aeotec.freshdesk.com/helpdesk/attachments/6103459970

blhoward2:
That's not the info I need. From zjs2mqtt or HA, what's the deviceType, deviceID and firmware version? I need the device's fingerprint.

booroondook-eng:
> That's not the info I need. From zjs2mqtt or HA, what's the deviceType, deviceID and firmware version? I need the device's fingerprint.

I don't fully understand where I can get the information you need. Maybe these screenshots will answer your question?
2021-05-09 15_36_25-Window
2021-05-09 15_39_10-Z-Wave JS - Home Assistant

AlCalzone:
@blhoward2 there were already other settings for clamp 3 in that file. Looks like an oversight to me. Device ID is correct.

blhoward2:
Aeotec confirmed there is no way to distinguish these or the ZW095. Some have two clamps and some three. Do we hide the parameter when it eventually times out on the read?

AlCalzone:
No we don't. Could just be temporary RF problems. Don't think that's much of a problem though.

AlCalzone:
sorry, I'll need to hijack your PR for a test

AlCalzone:
hm, that's not what I expected to happen 😅

AlCalzone (approved):

PR #2572, opened 2021-05-09
feat: add node deviceDatabaseUrl property

raman325:
As I mentioned in Discord, I think it would be useful to have the device DB URL in node-zwave-js. That way each downstream project doesn't have to generate it on its own.

### DIFF
--- a/packages/zwave-js/src/lib/node/Node.ts
+++ b/packages/zwave-js/src/lib/node/Node.ts
@@ -615,6 +615,14 @@ export class ZWaveNode extends Endpoint {
 		return this._deviceConfig?.label;
 	}
 
+	public get dbUrl(): string {
+		const manufacturerId = num2hex(this.manufacturerId);
+		const productType = num2hex(this.productType);
+		const productId = num2hex(this.productId);
+		const firmwareVersion = this.firmwareVersion || "0.0";
+		return `https://devices.zwave-js.io/?jumpTo=${manufacturerId}:${productType}:${productId}:${firmwareVersion}`;
+	}
+
 	private _neighbors: readonly number[] = [];
 	/**
 	 * The IDs of all direct neighbors of this node
--- a/packages/zwave-js/src/lib/node/Node.ts
+++ b/packages/zwave-js/src/lib/node/Node.ts
@@ -617,15 +617,16 @@ export class ZWaveNode extends Endpoint {
 
 	public get deviceDatabaseUrl(): string | undefined {
 		if (
-			(this.manufacturerId && this.productType && this.productId) ==
-			undefined
-		)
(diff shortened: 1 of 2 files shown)


packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -615,6 +615,14 @@ export class ZWaveNode extends Endpoint {
 		return this._deviceConfig?.label;
 	}
 
+	public get deviceDatabaseUrl(): string {
```
AlCalzone: If manufacturerId, productType or productId is `undefined`, this should also return `undefined` IMO.

AlCalzone (changes requested):

marcus-j-davies:
Can I add a suggestion here?

There is a Public API endpoint, that checks if the device is found on the DB.
for a  better user experience (instead of landing on the Browser without any reason nothing is listed)

Call this API, before returning the URL, so the front end can alert the user, that it is not found.
This API call, can be done within  Z-Wave JS or the front end before grabbing the URL?

It is just a suggestion, thoughts?

raman325:
> Can I add a suggestion here?
> 
> There is a Public API endpoint, that checks if the device is found on the DB.
> for a better user experience (instead of landing on the Browser without any reason nothing is listed)
> 
> Call this API, before returning the URL, so the front end can alert the user, that it is not found.
> This API call, can be done within Z-Wave JS or the front end before grabbing the URL?
> 
> It is just a suggestion, thoughts?

can you point me to the API? Couldn't find it, but I could easily add that check in and return undefined if it doesn't pass

marcus-j-davies:
`/public_api/getdeviceinfo/{ManufacturerID}/{ProductType}/{ProductID}/{FirmwareVersion}`

deviceFound - is what you are interested in.
it will also return as to what DB version the check was made.

```javascript
{
  "configVersion": "7.2.3",
  "deviceFound": true,
  "device": {
    "manufacturerId": "0x031e",
    "productType": "0x0001",
    "productId": "0x0001",
    "firmwareRange": {
      "min": "0.0",
      "max": "255.255"
    },
    "label": "LZW31-SN",
    "description": "Red Series Dimmer",
    "brand": "Inovelli"
  }
}
```

marcus-j-davies:
The guts are there so to speak.
Leaving it to the front end, should mean nothing more on this PR - freeing your time 😉

AlCalzone (approved):

AlCalzone:
> The guts are there so to speak.

Yeah, but if I've learned something, developers will do mistakes if they can. I guarantee that someone out there will find an unintended way to DDoS the DB by trying to check if the link exists 🤣

PR #2574, opened 2021-05-09
Adjusted parameters in fgt001.json according to manual

andreasfar:
Adjusted the allowed parameters according to the manual for the device.

[https://manuals.fibaro.com/content/manuals/en/FGT-001/FGT-001-EN-T-v1.4.pdf#page=23](url)


AlCalzone (changes requested):
If we're touching them now, they should be correctly modeled as [partial parameters](https://zwave-js.github.io/node-zwave-js/#/config-files/file-format?id=partial-parameters):
grafik

PR #2601, opened 2021-05-12
fix(serial): detect when socket disconnects and destroy driver

### LINKED ISSUE #1167: [bug] TCP connection drop leads to undefined behavior
### Version

**Checklist:**

Build/Run method

- [x] Docker
- [ ] PKG
- [ ] Snap package
- [ ] Manually built (git clone - yarn install - yarn run build )

zwavejs2mqtt version: 4.0.1
zwave-js version: 7.3.0

### Describe the bug

I'm using ser2net to connect my exposed Z-Wave USB dongle to zwavejs2mqtt. This works very well as long as the connection is stable. I had to restart the device with the dongle and this is where zwavejs2mqtt started to misbehave:

As long as no command is sent to the dongle, nothing is logged (but the controller is actually dead!). If I start inclusion, I end up with the Add/Remove device modal window that cannot be closed (i.e. no button is visible) and the following in the log:

```
2021-05-10 16:07:53.719 INFO ZWAVE: Calling api startInclusion with args: [ false, [length]: 1 ]
2021-05-10 16:07:54.956 INFO ZWAVE: Failed to send the message after 3 attempts startInclusion {
success: false,
message: 'Failed to send the message after 3 attempts'
}
```
Any further action after reloading the web page, including changing the ZWave settings, leads to the following in the log:
```
2021-05-10 16:10:00.614 ERROR ZWAVE: undefined
```

The only way I was able to fix this behavior was to restart the zwavejs2mqtt container.

I suppose that there are actually two problems: 
1. The modal window that cannot be closed
2. A communication problem on the zwavejs side.

### To Reproduce

1. Connect the Z-Wave dongle via TCP, i.e. `tcp://dongle.lan:5000`
2. Drop the connection or just shutdown ser2net on the host with the dongle
3. Start inclusion inside zwavejs2mqtt

### Expected behavior

1. The dropped connection should be automatically detected and restored by the zwavejs driver
2. The add device modal window should contain a close or cancel button

### Logs
Slightly redacted logs (removed homeid and ip address)

[zwavejs2mqtt.log.zip](https://github.com/zwave-js/zwavejs2mqtt/files/6460060/zwavejs2mqtt.log.zip)

AlCalzone:
With this change, serial-over-tcp ports will be monitored with the keepAlive option. Once the connection goes down, this causes the socket to be closed after a few seconds.

Because the driver and controller state may go out of sync in this case, a closed socket will destroy the driver instance, which must be handled by the application - e.g. by restarting or shutting down.

fixes: https://github.com/zwave-js/zwavejs2mqtt/issues/1167

### DIFF
--- a/packages/serial/src/ZWaveSocket.ts
+++ b/packages/serial/src/ZWaveSocket.ts
@@ -1,4 +1,4 @@
-import type { ZWaveLogContainer } from "@zwave-js/core";
+import { ZWaveError, ZWaveErrorCodes, ZWaveLogContainer } from "@zwave-js/core";
 import * as net from "net";
 import { ZWaveSerialPortBase } from "./ZWaveSerialPortBase";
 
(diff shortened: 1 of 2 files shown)


robertsLando (approved):
LGTM

PR #2606, opened 2021-05-12
feat: add ability to specify string as CC set duration instead of only Duration

matejdro:
This is a continuation of the https://github.com/zwave-js/zwavejs2mqtt/pull/1159. Idea is to be able to pass duration to the `sendCommand/set` mqtt command, as a workaround for the https://github.com/zwave-js/node-zwave-js/issues/1321.

For example, this mqtt command is supposed to set the level of the light to the 99 in 10 seconds:

```
{
    "args": [
        {
            "nodeId": 11,
            "commandClass": 38,
            "property": ""
        },
        "set",
        [
            99,
	    "10s"
        ]
    ]
}
```

However, I'm a bit stumped here. From what I can see, `sendCommand` just calls the `set` method directly, so I modified the arguments of the set method and added duration string parsing. But upon trying to send above command, I get this error message:

```
2021-05-12 16:56:39.946 INFO ZWAVE: Message dropped because of an unexpected error: this.duration.serializeSet is not a function sendCommand {
  success: false,
  message: 'Message dropped because of an unexpected error: this.duration.serializeSet is not a function'
}
```

Any idea what did I miss? Thanks.

### DIFF
--- a/packages/core/src/values/Duration.test.ts
+++ b/packages/core/src/values/Duration.test.ts
@@ -154,34 +154,53 @@ describe("lib/util/Duration", () => {
 		});
 
 		it(`should return a duration for seconds`, () => {
-			expect(Duration.parseString("10s")).toEqual(new Duration(10, "seconds"));
-			expect(Duration.parseString("25s")).toEqual(new Duration(25, "seconds"));
-			expect(Duration.parseString("33S")).toEqual(new Duration(33, "seconds"));
+			expect(Duration.parseString("10s")).toEqual(
+				new Duration(10, "seconds"),
+			);
+			expect(Duration.parseString("25s")).toEqual(
+				new Duration(25, "seconds"),
+			);
+			expect(Duration.parseString("33S")).toEqual(
+				new Duration(33, "seconds"),
+			);
 		});
 
 		it(`should return a duration for minutes`, () => {
-			expect(Duration.parseString("17m")).toEqual(new Duration(17, "minutes"));
-			expect(Duration.parseString("22m")).toEqual(new Duration(22, "minutes"));
-			expect(Duration.parseString("99M")).toEqual(new Duration(99, "minutes"));
+			expect(Duration.parseString("17m")).toEqual(
+				new Duration(17, "minutes"),
+			);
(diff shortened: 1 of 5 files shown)


blhoward2:
@AlCalzone will have to comment on the error but I approved you for all of the bot jobs for now.

matejdro:
@zwave-js-bot fix lint

robertsLando (dismissed):
From a code side seems good but let's see @AlCalzone thoughts

AlCalzone:
> But upon trying to send above command, I get this error message

Is that still happening with the latest commit? I can try this out later - thanks so far, looking good!

matejdro:
Yes. But it is possible that bungled up something when I built the release and it did not use right core version. I can try again a bit later.

AlCalzone:
Looks like it works:
```ts
		const node = driver.controller.nodes.get(15)!;
		await node.commandClasses["Binary Switch"].set(false, "1m");
```
gave me
```
00:27:08.097 DRIVER » [Node 015] [REQ] [SendDataBridge]
                      │ source node id:   1
                      │ transmit options: 0x25
                      │ route:            0, 0, 0, 0
                      │ callback id:      1
                      └─[BinarySwitchCCSet]
                          target value: false
                          duration:     [Duration: 1 minute]
```

I'll clean up a little bit and merge this - thanks for the starter!

AlCalzone (dismissed):

matejdro:
Awesome, thanks!

packages/core/src/values/Duration.test.ts
```diff
@@ -177,9 +177,18 @@ describe("lib/util/Duration", () => {
...
+			);
+			expect(Duration.parseString("1h8s")).toEqual(
+				new Duration(60, "minutes"),
+			);
+		});
+
+		it(`should return the nearest possible duration for combined seconds and minutes`, () => {
 			expect(Duration.parseString("19m18s")).toEqual(
-				new Duration(19 * 60 + 18, "seconds"),
+				new Duration(19, "minutes"),
```
matejdro, line 191: Is that okay? Why do we remove accuracy?

AlCalzone (approved):

PR #2614, opened 2021-05-14
chore(config): remove redundant leading comments from config files

AlCalzone:
I've wanted to do this for a while. Now that we can use `comment-json` to automatically edit these files, things are a lot easier.


blhoward2:
I kind of like the leading comments for human readability when you’re flipping between files, but obviously if you want them gone...

AlCalzone:
I found them to be inconsistent sometimes, especially when new external contributors are involved.

blhoward2:
Eh. Lint could just as easily compare them to the fields and rewrite them, but I don't feel that strongly.

blhoward2 (approved):
LGTM

PR #2616, opened 2021-05-14
fix(config): fix several warnings in config files

AlCalzone:
Trying to make a dent into the several hundred warnings the `lint:config` script is spitting out. Some are false positives or cannot be solved, but I'd like to get the number down a bit.

Help is welcome - just make a PR on the `config-warnings` branch or push directly into it.


AlCalzone:
Down to 470 warnings 😅

AlCalzone:
Aside from the hundreds of "unit in description" warnings, this now fixes all but these:
#2623
#2625
#2626

PR #2617, opened 2021-05-15
fix(config): corrected partial paramaters in Philio Technology Corp PST02-C

### LINKED ISSUE #2575: Philio Technology Corp PST02-C: Config file needs partial parameters
Reported in https://github.com/AlCalzone/ioBroker.zwave2/issues/627
Link to device DB: https://devices.zwave-js.io/?jumpTo=0x013c:0x0002:0x000e:0.0
Link to manual: http://manuals-backend.z-wave.info/make.php?lang=en&type=&sku=PHI_PST02-1C

cdnninja:
closes #2575

### DIFF
--- a/packages/config/config/devices/0x013c/pst02c.json
+++ b/packages/config/config/devices/0x013c/pst02c.json
@@ -48,7 +48,7 @@
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 1,
-			"defaultValue": 0
+			"defaultValue": 0,
 			"options": [
 				{
 					"label": "Disable",
@@ -65,7 +65,7 @@
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 1,
-			"defaultValue": 0
+			"defaultValue": 0,
 			"options": [
 				{
 					"label": "Disable",
@@ -82,7 +82,7 @@
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 1,
-			"defaultValue": 0
+			"defaultValue": 0,
 			"options": [
 				{
 					"label": "Fahrenheit",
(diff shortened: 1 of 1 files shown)


cdnninja:
@zwave-js-bot fix lint

packages/config/config/devices/0x013c/pst02c.json
```diff
@@ -29,43 +29,249 @@
...
+			"options": [
+				{
+					"label": "Disable",
+					"value": 1
+				},
+				{
+					"label": "Disable",
+					"value": 0
+				}
+			]
```
AlCalzone: One of these is not correct :D

packages/config/config/devices/0x013c/pst02c.json
```diff
@@ -29,43 +29,249 @@
...
 			"label": "Basic Set Level",
-			"description": "Setting the BASIC command value to turn on the light.",
+			"description": "Setting the BASIC command value to turn on the light",
 			"valueSize": 1,
-			"minValue": -1,
-			"maxValue": 100,
-			"defaultValue": -1
+			"minValue": 0,
+			"maxValue": 255,
+			"defaultValue": 255
```
AlCalzone:
```suggestion
			"defaultValue": 255,
			"unsigned": true
```

AlCalzone (changes requested):
Some of the descriptions are suboptimal - either broken english or unnecessary because they just repeat the label in more words. Plus two small errors

cdnninja:
I have some more labels to edit however interested on thoughts on this.  The manual isn't clear if these are actually double negatives or not.  An example: `Disable send out BASIC OFF after door closed. (1:Disable, 0:Enable)` So do you enable it to disable send basic off?   I assume this is defaulted to 1 due to the order as well.  I have a few assumptions I have to make when reading this manual.

AlCalzone:
I think this configures whether a Basic Set is sent or not, with the default being "No".
I don't think these are double negatives, just bad labels.

AlCalzone (changes requested):

AlCalzone (approved):
Ok with my changes

PR #2618, opened 2021-05-15
Update zw15s.json

mt98987987:


AlCalzone:
I'm guessing this was a mistake and the other PR #2619 is the correct one?

PR #2665, opened 2021-05-21
refactor(config): template "state after power failure" setting

AlCalzone:
I've added more variants (2 and 3 way) for the "state after power failure" and $import-ed them where possible.


blhoward2:
I'm fine with doing this awhile, but we will need to note to re-verify those as we do those folders. I've found several instances where we were missing options or an option was added at a specific fw verison.

blhoward2 (dismissed):

AlCalzone:
Mental note added 😅

PR #2669, opened 2021-05-22
feat(config): add support for Weiser SmartCode 5

mw-white:
Added a config for the Weiser SmartCode 5 Lever Lock:  https://products.z-wavealliance.org/products/2378

Very similar to the existing Weiser and Kwikset configs.

### DIFF
--- a/packages/config/config/devices/0x0090/ged1455.json
+++ b/packages/config/config/devices/0x0090/ged1455.json
@@ -1,0 +1,143 @@
+{
+	"manufacturer": "Weiser",
+	"manufacturerId": "0x0090",
+	"label": "GED1455",
+	"description": "SmartCode 5 Touchpad Electronic Lever",
+	"devices": [
+		{
+			"productType": "0x0006",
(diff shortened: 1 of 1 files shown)


AlCalzone:
Do you know by any chance if that's an OEM for Kwikset? We avoid cross-importing from other manufacturers unless this is the case. Granted, here it seems suspiciously similar, so it might aswell be.

mw-white:
I suspect that it is - they all seem to fall under the same manufacturer ID.  I believe the Weiser lock is just a rebranded Kwikset 912 (one of the codes for the quickset is productType 0x0003, Id 0x0339 and the Weiser is 0x0001, 0x0339.

Would it make more sense to add this as another device in the 912 entry?

AlCalzone:
> Would it make more sense to add this as another device in the 912 entry?

No, it's fine here - otherwise its going to confuse users.

AlCalzone:
@zwave-js-bot rename commit

AlCalzone (approved):

PR #2672, opened 2021-05-23
feat: emit a driver event when the log config is updated

raman325:
With `zwave-js-server`, there can be multiple clients that could each be storing the state of the log config internally (we do this in HA). If one client updates the log config, there's currently no way for the other clients to know without polling the log config regularly.


AlCalzone:
I don't see why this needs to be in the driver, as it does not care about multiple clients. From the driver POV, the application calls `updateLogConfig`, which is the only way to update the log config. 
IMO, if `zwavejs-server` needs to sync this info between clients, it should do so when calling `updateLogConfig`

raman325:
Fair point

PR #2689, opened 2021-05-25
feat(config): add additional identifier for Eaton 5-Scene Keypad

### LINKED ISSUE #2640: Eaton RFWDC
**Which device is missing?**

Eaton RFWDC (same as RFWC5)

**What are the IDs?**

Manufacturer ID: 26 
Product type: 22349
Product ID: 52233 

**Did you check if we really don't have it?**

- [x] It is not in the [configuration DB](https://devices.zwave-js.io/)
- [x] It was not [merged recently](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Apr+is%3Aclosed) or has a [pending PR](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Aopen+is%3Apr)
- [x] It is not in the [TODO list](https://github.com/zwave-js/node-zwave-js/issues/1600)

**Do you have a link to the PDF manual?**

When I was using Smartthings I was using the device handler for the RFWC5 and everything functioned correctly.

If so, it helps in creating the config file.

marcus-j-davies:
- Update label
 - Add Identifier
 
 Fixes https://github.com/zwave-js/node-zwave-js/issues/2640


AlCalzone (approved):

blhoward2:
While they’re very similar devices, I’d keep them separate. They have different manuals and identify differently. As of now we don’t know of any differences but one very may well require a compat flag the other doesn’t, or something. I’m also not clear that the old one is zwave plus, though it may be.

I think different model numbers should be kept apart unless there is a very good reason to combine, which usually means identical except for variations between frequencies or re-releases of new hardware under the same model number.

AlCalzone (changes requested):
Fine with me. @marcus-j-davies can you split the file please?

marcus-j-davies:
Yup,

About to engulf myself in food, then will get to it.

marcus-j-davies:
@zwave-js-bot fix lint

AlCalzone:
> but files changed is 1

That's correct though, right? You don't touch the other file anymore.

AlCalzone (approved):

marcus-j-davies:
I removed the addition, and put it into its own.

dgoodwin430:
Thanks guys for your work on this! Should this appear in the device database now?

marcus-j-davies:
It will appear on the DB Browser, when I update the config DB version, that is uses.
Your environment, will get support when you update Z-Wave JS.

PR #2696, opened 2021-05-25
fix(config): change GED2150 extension to json

Chris-V:
1. Renamed GED2150 extension from `sjon` to `json`
2. Fixed a type in the description


AlCalzone:
Ah damn, I should look more closely...
Thanks!

marcus-j-davies:
> I should look more closely...

So should I, that was my doing 🙈

AlCalzone (approved):

PR #2726, opened 2021-05-28
feat(config): define partial parameters for Fibaro FGT001, add additional parameters for MH9 CO2

kvaster:
This is my first PR and it is considered as draft for now.

## Changes:
* Partial parameters for fgt001 according to manuals and tested them on my own devices.
* Additional parameters for mcohome mh9-co2 sensors.

### Associations and notifications

It seems that I don't completely understand how associations works in zwavejs or I don't undestand how they are shown in zwavejs2mqtt. fgt001 have only one association group (lifeline) and I've added associations merge for all endpoints, but I'm not sure if I'm doing this right, cause nothing changed in zwavejs2mqtt panel.

As I understand (am I right?) associations are related to 'notification' section in zwavejs2mqtt. And I can see following for fgt001 devices:

- [5-113-1-Power Management-unknown] Power Management (property) - editable field with value 254
- [5-113-1-System-unknown] System (property) - editable field with value 254
- [5-113-1-Power Management-Battery load status] Battery load status - r/o field with value undefined
- [5-113-1-Power Management-Battery level status] Battery level status - r/o field with value undefined
- [5-113-1-System-Hardware status] Hardware status - r/o field with value undefined
- [5-113-2-Power Management-Battery maintenance status] Battery maintenance status - r/o field with value undefined
- [5-113-2-Power Management-unknown] Power Management (property) - editable field with value 254 AND (!) this field is not available after first interview - it will appear only after clicking 'refresh' button in zwavejs2mqtt.

I think that it is expected to have editable association groups with single or multiple subscribers inside....

Also fgt001 manual have following specs for notification events:

For endpoint 1:
- power management, events:
  - charge battery soon (0x0e)
  - charge battery now (0x0f)
  - battery is charging (0x0c)
  - battery is full charged (0x0d)
- system, events:
  - system hardware failure, event parameters:
    - external sensor remove (0x02)
    - motor error (0x03)
    - calibration error (0x04)

For endpoint 2:
- power management, events:
  - replace battery soon (0x0a)
  - replace battery now (0x0b)

And I can't correlate all values in notification view with this information...

### Mapping root to endpoint

We've already discussed this in 'question' issue. This devices duplicates root (single channel) in endpoint 1. I just checked open-zwave code and they have option in device description to map root to endpoint. For now I've added command class remove for root channel for battery class, cause it was duplicated in panel.

### MH9-CO2

This is simple. I have 5 such devices and two of them recognized as '-WA' and others as '-WD' devices. Generally all of them are '-WA' devices, cause difference between '-WA' and '-WD' is only in power supply (12V vs 220V) according to mcohom site. Real difference is in device hw revision. One ('-WA' in zwavejs) is zwave plus with additional features/parameters and second one is older without 'plus'.

### Basic command class

Don't know how to deal with this right now. For fgt001 current value is always undefined, for mh9 current value have 'some' integer value... maybe it is related to 'air quality', but I'm not able to correlate unfortunatelly.... Also I can see writeable target value for both....

### Endpoints

It seems that open-zwave have configurable endpoint names - maybe it's good idea for zwavejs also ? :)

### DIFF
--- a/packages/config/config/devices/0x010f/fgt001.json
+++ b/packages/config/config/devices/0x010f/fgt001.json
@@ -19,7 +19,7 @@
 	},
 	"paramInformation": {
 		"1": {
-			"label": "Override Schedule duration",
+			"label": "Override Schedule Duration",
 			"description": "Duration of Override Schedule after turning knob",
 			"unit": "minutes",
 			"valueSize": 4,
@@ -29,18 +29,18 @@
 		},
 		"2[0x01]": {
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Open window detector (normal)",
+			"label": "Open Window Detector (Normal)",
 			"defaultValue": 1,
 			"valueSize": 4
 		},
 		"2[0x02]": {
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Open window detector (rapid)",
+			"label": "Open Window Detector (Rapid)",
 			"valueSize": 4
 		},
 		"2[0x04]": {
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Increase receiver sensitivity",
+			"label": "Increase Receiver Sensitivity",
 			"description": "Increase receiver sensitivity, but shortens battery life",
 			"valueSize": 4
 		},
@@ -52,59 +52,59 @@
 		},
 		"2[0x10]": {
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Protect setting Full ON/OFF",
+			"label": "Protect Setting Full ON/OFF",
 			"description": "Protect from setting Full ON and Full OFF mode by turning the knob manually",
 			"valueSize": 4
 		},
 		"2[0x20]": {
 			"$if": "firmwareVersion >= 4.7",
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Vertical mount",
+			"label": "Vertical Mount",
 			"description": "Device mounted in vertical position",
 			"valueSize": 4
 		},
 		"2[0x40]": {
 			"$if": "firmwareVersion >= 4.7",
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Moderate regulator behaviour",
+			"label": "Moderate Regulator Behaviour",
 			"description": "Set moderate regulator behaviour instead of rapid",
 			"valueSize": 4
 		},
 		"2[0x80]": {
 			"$if": "firmwareVersion >= 4.7",
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Inverted knob operation",
+			"label": "Inverted Knob Operation",
 			"valueSize": 4
 		},
 		"2[0x100]": {
 			"$if": "firmwareVersion >= 4.7",
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Heating medium demand reports",
+			"label": "Heating Medium Demand Reports",
 			"valueSize": 4
 		},
 		"2[0x200]": {
 			"$if": "firmwareVersion >= 4.7",
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Detecting heating system failures",
+			"label": "Detecting Heating System Failures",
 			"valueSize": 4
 		},
 		"3[0x01]": {
 			"$import": "../templates/master_template.json#base_true_false",
-			"label": "Temperature sensor",
+			"label": "Temperature Sensor",
 			"description": "Optional temperature sensor connected and operational",
 			"valueSize": 4,
 			"readOnly": true
 		},
 		"3[0x02]": {
 			"$import": "../templates/master_template.json#base_true_false",
-			"label": "Open window detected",
+			"label": "Open Window Detected",
 			"valueSize": 4,
 			"readOnly": true
 		},
 		"3[0x04]": {
 			"$if": "firmwareVersion >= 4.7",
 			"$import": "../templates/master_template.json#base_true_false",
-			"label": "Provide heat",
+			"label": "Provide Heat",
 			"description": "Provide heat in order to maintain set temperature",
 			"valueSize": 4,
 			"readOnly": true
@@ -112,7 +112,7 @@
 		"3[0x08]": {
 			"$if": "firmwareVersion >= 4.7",
 			"$import": "../templates/master_template.json#base_true_false",
-			"label": "Malfunctioning heating system",
+			"label": "Malfunctioning Heating System",
 			"description": "Malfunctioning heating system - cannot reach set temperature",
 			"valueSize": 4,
 			"readOnly": true
--- a/packages/config/config/devices/0x010f/fgt001.json
+++ b/packages/config/config/devices/0x010f/fgt001.json
@@ -18,6 +18,15 @@
 		"max": "255.255"
 	},
 	"supportsZWavePlus": true,
+	"compat": {
+		"commandClasses": {
+			"remove": {
+				"0x80": {
+					"endpoints": [0]
+				}
+			}
+		}
+	},
 	"paramInformation": {
 		"1": {
 			"label": "Override Schedule duration",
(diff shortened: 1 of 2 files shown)


packages/config/config/devices/0x010f/fgt001.json
```diff
@@ -18,6 +18,15 @@
 		"max": "255.255"
 	},
 	"supportsZWavePlus": true,
```
AlCalzone:
```suggestion
```
We no longer use that

packages/config/config/devices/0x010f/fgt001.json
```diff
@@ -18,6 +18,15 @@
...
+	"compat": {
+		"commandClasses": {
+			"remove": {
+				"0x80": {
+					"endpoints": [0]
+				}
+			}
+		}
+	},
```
AlCalzone: The `"compat"` section should be moved to the end of the file (before "metadata" if that exists")

packages/config/config/devices/0x010f/fgt001.json
```diff
@@ -18,6 +18,15 @@
 		"max": "255.255"
 	},
 	"supportsZWavePlus": true,
+	"compat": {
+		"commandClasses": {
+			"remove": {
+				"0x80": {
+					"endpoints": [0]
```
AlCalzone: Please leave a comment why this is being removed.

packages/config/config/devices/0x010f/fgt001.json
```diff
@@ -28,22 +37,121 @@
 			"maxValue": 10000,
 			"defaultValue": 240
 		},
-		"2": {
-			"label": "Additional functions",
-			"description": "This parameter allows to enable different additional functions of the device.",
+		"2[0x01]": {
+			"label": "Enable open window detector (normal)",
+			"$import": "../templates/master_template.json#base_enable_disable",
```
AlCalzone, line 31: `$import` should come before labels and such, otherwise the label might get overwritten.

packages/config/config/devices/0x010f/fgt001.json
```diff
@@ -28,22 +37,121 @@
...
+			}
+		},
+		"2": {
+			"associations": {
+				"1": {
+					"$import": "#endpoints/0/associations/1",
+					"isLifeline": false
+				}
+			}
+		}
```
AlCalzone: Since the device supports Z-Wave Plus, defining the associations shouldn't be necessary.

packages/config/config/devices/0x015f/mh9-co2-wa.json
```diff
@@ -19,8 +19,45 @@
 			"label": "CO2 Notification Threshold",
 			"valueSize": 2,
 			"minValue": 1,
-			"maxValue": 2000,
+			"maxValue": 5000,
 			"defaultValue": 1000
+		},
+		"2": {
+			"label": "CO2 reporting threshold",
+			"description": "0 report disabled, base on 5ppm unit, 10 by default, 10*5ppm=50ppm",
```
AlCalzone: Please don't mention units or default values in the description. We have the `"unit"` and `"defaultValue"` properties for that. Same goes for the other parameters.

AlCalzone: The special meaning of `0 = report disabled` should be added as a single entry under `"options"` instead. Those are used as suggestions in UIs.

packages/config/config/devices/0x015f/mh9-co2-wa.json
```diff
@@ -19,8 +19,45 @@
...
+		},
+		"255": {
+			"label": "Factory setting",
+			"description": "Restore the factory setting - write only, set to 85 to reset",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 255,
+			"defaultValue": 0,
+			"unsigned": true,
+			"writeOnly": true
```
AlCalzone: We normally enter those parameters with `allowManualEntry: false` and two options - one for idle and one for resetting

AlCalzone (changes requested):
I left you some comments. Please check your labels, they don't conform to our [style guide](https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide?id=labels)

AlCalzone:
> As I understand (am I right?) associations are related to 'notification' section in zwavejs2mqtt

No. Associations are used for direct communication between devices as well as sending reports to the controller. Notifications are just part of that communication.

> I think that it is expected to have editable association groups with single or multiple subscribers inside....

See [manual](https://manuals.fibaro.com/content/manuals/en/FGT-001/FGT-001-EN-T-v1.4.pdf):
> The  device  supports  only  “Lifeline”  association  group  that  reports  the  device  status  and  allows  for  assigning  single  device  only  (main  controller by default)

> It seems that open-zwave have configurable endpoint names - maybe it's good idea for zwavejs also ? :)

Could be worth a feature request

> ### Basic command class
> 
> Don't know how to deal with this right now. For fgt001 current value is always undefined

That might be mapped to a better CC instead of being stored. I'd need to look at a log to know

> Also fgt001 manual have following specs for notification events:

You don't need to define those, they are queried from the device.

kvaster:
I've updated my PR. And I just realized that 'groups' in zwavejs2mqtt is for associations and it's all OK with them.

I still have some 'questions' (hopefully I'm not bothering you too much).
* For MH9-CO2 we have factory reset configuration. For now I've added two options - idle (0) and reset (85), but value is w/o and in control panel you see no any setting and you'll be able to see three options inteadof two. Third one will be `[undefined] Custom`. But I think it's UI question and probably will make PR also :)
* For FGT001 thermostat mode shows two options: `[0] Off` and `[1] Heat` and in manual:
 * 0 - Off
 * 99 - Heat
 * 255 - manufacturer specific (valve fully opened)
* For FGT001 thermostat mode have write only property - manufacturerData. Don't undesrtand what this is...
* For FGT001 don't understand 'notification' properties. Specially editable properties with value 254...

AlCalzone:
>     * For MH9-CO2 we have factory reset configuration. For now I've added two options - idle (0) and reset (85), but value is w/o and in control panel you see no any setting and you'll be able to see three options inteadof two. Third one will be `[undefined] Custom`. But I think it's UI question and probably will make PR also :)

Hm, that doesn't really make sense. Maybe it is best if we leave out the idle value?

> For FGT001 thermostat mode shows two options: [0] Off and [1] Heat and in manual:
> 0 - Off
> 99 - Heat
> 255 - manufacturer specific (valve fully opened)

I think you're confusing the `Thermostat Mode CC` (Off/Heat) with `Basic CC` which has the values you mentioned. You cannot configure this anyways, so no need to discuss it here.

> For FGT001 thermostat mode have write only property - manufacturerData

That's a special property that the `Thermostat Mode CC` (or was it Setpoint?) offers. I don't think I've ever seen it used, but there's no way to discover whether it is. So it needs to stay.

> For FGT001 don't understand 'notification' properties. Specially editable properties with value 254...

I don't follow - what do you mean?

packages/config/config/devices/0x015f/mh9-co2-wa.json
```diff
@@ -13,14 +13,79 @@
...
+			]
+		},
+		"3": {
+			"label": "Temperature Reporting Threshold",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 255,
+			"defaultValue": 1,
+			"unsigned": true,
+			"unit": "0.5C",
```
AlCalzone:
```suggestion
			"unit": "0.5 °C",
```

packages/config/config/devices/0x015f/mh9-co2-wa.json
```diff
@@ -13,14 +13,79 @@
...
+			"maxValue": 255,
+			"defaultValue": 0,
+			"unsigned": true,
+			"writeOnly": true,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Idle",
+					"value": 0
+				},
```
AlCalzone:
```suggestion
```
as discussed - if this value isn't ever used, there's not much sense in having it.

packages/config/config/devices/0x015f/mh9-co2-wa.json
```diff
@@ -13,14 +13,79 @@
...
 	},
-	"supportsZWavePlus": true,
 	"paramInformation": {
 		"1": {
 			"label": "CO2 Notification Threshold",
 			"valueSize": 2,
 			"minValue": 1,
-			"maxValue": 2000,
+			"maxValue": 5000,
 			"defaultValue": 1000
```
AlCalzone: Is there a unit for this?

AlCalzone (changes requested):
Please make sure the labels conform to the [style guide](https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide?id=labels) - including the part 

> Labels should be **Title Case**.

kvaster:
> > ```
> > For now I've added two options - idle (0) and reset (85)
> > ```
> 
> Hm, that doesn't really make sense. Maybe it is best if we leave out the idle value?

Fixed. For UI with 'undefined' option is a question to zwavejs2mqtt.
 
> > For FGT001 thermostat mode shows two options: [0] Off and [1] Heat and in manual:
> > 0 - Off
> > 99 - Heat
> > 255 - manufacturer specific (valve fully opened)
> 
> I think you're confusing the `Thermostat Mode CC` (Off/Heat) with `Basic CC` which has the values you mentioned. You cannot configure this anyways, so no need to discuss it here.

Checked documentation and you're completely right. Thanks.
 
> > For FGT001 thermostat mode have write only property - manufacturerData
> 
> That's a special property that the `Thermostat Mode CC` (or was it Setpoint?) offers. I don't think I've ever seen it used, but there's no way to discover whether it is. So it needs to stay.

For thermostat mode I can see following and I think `manufacturerData` is property you're talking about.

image
 
> > For FGT001 don't understand 'notification' properties. Specially editable properties with value 254...
> 
> I don't follow - what do you mean?

This is screenshot I can see in control panel:

image

And if I click 'refresh' button then one more editable property with default value 254 will appear: [5-113-2-Power Management-unknown] Power Management (property).

Attached zwavejs log:

[zwavejs_2021-06-02.log](https://github.com/zwave-js/node-zwave-js/files/6582525/zwavejs_2021-06-02.log)

AlCalzone:
@kvaster The notification part doesn't look like it is working as intended. For some reason the node is detected as being in "pull" mode, but the manual indicates it is pushing notifications. I'll have to look into that (and #2670 which is also suboptimal).

packages/config/config/devices/0x010f/fgt001.json
```diff
@@ -17,33 +17,117 @@
...
 			"maxValue": 10000,
 			"defaultValue": 240
 		},
-		"2": {
-			"label": "Additional functions",
-			"description": "This parameter allows to enable different additional functions of the device.",
+		"2[0x01]": {
+			"$import": "../templates/master_template.json#base_enable_disable",
+			"label": "Open Window Detector (Normal)",
+			"defaultValue": 1,
```
AlCalzone:
```suggestion
			"defaultValue": 1, // Enable
```

AlCalzone (changes requested):

AlCalzone (approved):
Ok with my changes

kvaster:
Great! Thanks! Waiting for PR to be merged and thanks for excelent project!

PR #2731, opened 2021-05-29
feat(config): add treatBasicSetAsEvent flag to Aeotec ZW132

AlCalzone:
As mentioned in https://github.com/zwave-js/node-zwave-js/issues/2129#issuecomment-850811338

### DIFF
--- a/packages/config/config/devices/0x0086/zw132.json
+++ b/packages/config/config/devices/0x0086/zw132.json
@@ -202,9 +202,6 @@
 			"$if": "firmwareVersion >= 2.0",
 			"$import": "templates/aeotec_template.json#switch_destination_s2"
 		},
-		"252": {
-			"$import": "templates/aeotec_template.json#lock_configuration"
-		},
 		// WallSwipe Settings
 		"144": {
 			"$if": "firmwareVersion >= 2.0",
@@ -339,7 +336,9 @@
 		}
 	},
 	"compat": {
-		"preserveRootApplicationCCValueIDs": true
+		"preserveRootApplicationCCValueIDs": true,
+		// The device can report S1/S2 switch operation via Basic Set to groups 3 and 4
+		"treatBasicSetAsEvent": true
 	},
 	"metadata": {
 		"inclusion": "Turn the primary controller of Z-Wave network into inclusion mode, short press the product’s Action button that you can find on the product's housing",


AlCalzone:
Pending verification

luke-barnett:
> 2021-05-29T20:24:35.107Z CNTRLR [Node 002] [~] [Basic] currentValue: 0 => 255 [Endpoint 2]
> 2021-05-29T20:25:00.568Z SERIAL « 0x010d0004000207600d0200200100bd (15 bytes)
> 2021-05-29T20:25:00.571Z SERIAL » [ACK] (0x06)
> 2021-05-29T20:25:00.574Z DRIVER « [Node 002] [REQ] [ApplicationCommand]
> └─[MultiChannelCCCommandEncapsulation]
> │ source: 2
> │ destination: 0
> └─[BasicCCSet]
> target value: 0
> 2021-05-29T20:25:00.577Z CNTRLR [Node 002] treating BasicCC::Set as a report [Endpoint 2]
> 2021-05-29T20:25:00.578Z CNTRLR [Node 002] [~] [Basic] currentValue: 255 => 0 [Endpoint 2]
> 2021-05-29T20:25:04.388Z SERIAL « 0x010d0004000207600d02002001ff42 (15 bytes)
> 2021-05-29T20:25:04.390Z SERIAL » [ACK] (0x06)
> 2021-05-29T20:25:04.393Z DRIVER « [Node 002] [REQ] [ApplicationCommand]
> └─[MultiChannelCCCommandEncapsulation]
> │ source: 2
> │ destination: 0
> └─[BasicCCSet]
> target value: 255
> 2021-05-29T20:25:04.397Z CNTRLR [Node 002] treating BasicCC::Set as a report [Endpoint 2]
> 2021-05-29T20:25:04.398Z CNTRLR [Node 002] [~] [Basic] currentValue: 0 => 255 [Endpoint 2]

luke-barnett:
@AlCalzone what's the easiest way for me to validate the flag being useful?

I'm running zwavejs2mqtt using the official docker image. Do I just add the edited file to app/store/config as mentioned here https://zwave-js.github.io/zwavejs2mqtt/#/development/intro?

Will this then give me back the sensor entity in the home assistant integration? This is what I was using before I updated and the sensor became unavailable. Or will this come through as an event to home assistant? If so what type?

AlCalzone:
> [Node 002] treating BasicCC::Set as a report [Endpoint 2]

Okay, the flag makes sense here. You don't want it being treated as a report.

> Do I just add the edited file to app/store/config as mentioned here

Yes, make sure that the log prints "User-provided device config loaded". The docs have a typo there.

> Will this then give me back the sensor entity in the home assistant integration?

I think so. You might need to re-interview the device.

luke-barnett:
@AlCalzone is there a particular folder structure?

I've just stuck it as a file in the folder and then also tried copying the structure of devices/0x0086 but I can't get it loading the specific config.

image

luke-barnett:
Something isn't quite right here.

Either the configuration setting for zwavejs isn't being passed from zwavejs2mqtt properly or I'm looking at the wrong folder.

Based on this line: https://github.com/zwave-js/node-zwave-js/blob/e2e10c61b6fda5436c11e8e1d0473c8fe73c6efe/packages/config/src/ConfigManager.ts#L465

After deleting the folder, I'm expecting to see that show up in the logs but I'm not at all.

But it is the folder that I'm expecting based on this line and it does look to be being passed into the config.
https://github.com/zwave-js/zwavejs2mqtt/blob/1a35379bb733abfa1132a9dd894bc70d11b3afdd/lib/ZwaveClient.js#L750

I've got logging on silly, so everything should be showing up but this is all I get:

> 2021-05-30T05:39:00.218Z DRIVER   destroying driver instance...
> 2021-05-30T05:39:07.037Z DRIVER   ███████╗ ██╗    ██╗  █████╗  ██╗   ██╗ ███████╗             ██╗ ███████╗
>                                   ╚══███╔╝ ██║    ██║ ██╔══██╗ ██║   ██║ ██╔════╝             ██║ ██╔════╝
>                                     ███╔╝  ██║ █╗ ██║ ███████║ ██║   ██║ █████╗   █████╗      ██║ ███████╗
>                                    ███╔╝   ██║███╗██║ ██╔══██║ ╚██╗ ██╔╝ ██╔══╝   ╚════╝ ██   ██║ ╚════██║
>                                   ███████╗ ╚███╔███╔╝ ██║  ██║  ╚████╔╝  ███████╗        ╚█████╔╝ ███████║
>                                   ╚══════╝  ╚══╝╚══╝  ╚═╝  ╚═╝   ╚═══╝   ╚══════╝         ╚════╝  ╚══════╝
> 2021-05-30T05:39:07.039Z DRIVER   version 7.7.0
> 2021-05-30T05:39:07.040Z CONFIG   version 7.7.0
> 2021-05-30T05:39:07.041Z DRIVER   
> 2021-05-30T05:39:07.041Z DRIVER   starting driver...
> 2021-05-30T05:39:07.073Z DRIVER   opening serial port /dev/aeoteczwave
> 2021-05-30T05:39:07.110Z DRIVER   serial port opened
> 2021-05-30T05:39:07.113Z SERIAL » [NAK]                                                                   (0x15)
> 2021-05-30T05:39:07.121Z DRIVER   Checking for configuration updates...
> 2021-05-30T05:39:07.685Z DRIVER   No configuration update available...
> 2021-05-30T05:39:08.711Z DRIVER   loading configuration...
> 2021-05-30T05:39:09.970Z DRIVER   beginning interview...

Am I running into the same issue mentioned here? https://github.com/zwave-js/zwavejs2mqtt/issues/1223

I've tried both creating the folder via a command prompt within the container and from the zwavejs2mqtt UI.

AlCalzone:
I'll try to debug it later. For now you can volume map the file to its original location.

AlCalzone:
zwavejs2mqtt passes the priority config dir incorrectly.
https://github.com/zwave-js/zwavejs2mqtt/pull/1294 fixes that.

luke-barnett:
@AlCalzone volume mapped it in.

The basic values now never change (undefined until a refresh and then always 99) and the resulting sensors I get in home assistant are thus relatively useless so I've left them disabled.

image

But what I am getting is a value event coming through consistently on the zwave_js_value_notification namespace. Which I've then connected to my automation as the trigger point. I like this method a lot more than the senor approach, and it now also matches what I was getting from the previous zwave integration in home assistant.

So all in all, it's working perfectly. I do wonder however what value there is in the basic values being created and if the compat flag you've added should not create these values. They're useless (at least for this device) and so it would make sense for there to just be the generated event and not the value.

AlCalzone:
> I do wonder however what value there is in the basic values being created and if the compat flag you've added should not create these values. They're useless (at least for this device) and so it would make sense for there to just be the generated event and not the value.

I agree but this is currently not possible. When `Basic CC` isn't hidden/disabled, all its values are visible. Not hiding/disabling Basic CC is required for the Basic Event value to be usable.

Issue #985, opened 2020-09-22
Add statistics


AlCalzone:
Things that should be considered (total and per node):

- [x] Transmitted messages per node (#2736)
- [x] Transmitted messages (controller) (#2736)
- [x] Received messages per node (#2736)
- [x] Received messages (controller) (#2736)
- [x] Dropped messages from nodes
- [x] Dropped messages from the controller (#2736)
- [x] Controller errors (CAN, NAK, timeout) (#2736)
- [ ] Node errors (NoACK, timeout, ...) (done in #2736, but currently no distinction between not sent and NoACK)
- [x] Response time (#4022)
- [x] Background RSSI, current and average (done in #5545)
- [ ] Failures during interview
- [x] When was a node last seen (alive/awake/unsolicited msg) (#6008)
- [ ] Queue: size, what is in the queue? (#1270)

LordMike:
> Bonus to having events per message, rather than just counters, is also being able to see how long it actually took to deliver that "turn on" message. Would help immensely in tracking down where delays are occurring, when lights appear slow.

https://github.com/zwave-js/node-zwave-js/issues/1270#issuecomment-757505434

nicoh88:
+1 for `last seen`

jaimevisser:
Especially on bigger z-wave networks this would be a big help in tuning sensor data output.

Masterz69:
Missing ```last seen``` property on Home Assistant for ZWave devices.

Just today had an situation: FGBS-222 was powered off for ~25 minutes.
Device is configured to report every 5 minutes, couple of messages not received by controller.
ZWAVEJS2MQTT device ```node status``` still ```alive``` :( 
- as no configuration on device level: when to declare device  to not ```alive``` when no messages received from.

IMHO I can't relay on ```node status``` to determine actual device status.
```last seen``` will help to react on devices sending no info long time.

AlCalzone:
`lastSeen` will be added in https://github.com/zwave-js/node-zwave-js/pull/6008

PR #2736, opened 2021-05-30
feat: implement transmission statistics for the controller and nodes

AlCalzone:
Implements part of #985 

Both the controller and all nodes now have an event `"statistics updated"`, which is emitted whenever something in this regard changes (but throttled). It is likely that not all edge cases are covered here, but this should be a good start to get some insight into what the network does.

Example usage:
```ts
		driver.controller.on("statistics updated", (s) => {
			console.debug(s);
		});
		for (const node of driver.controller.nodes.values()) {
			node.on("statistics updated", (s) => {
				console.debug({
					node: node.id,
					...s,
				});
			});
		}
```
would cause these logs to go into the debugging console:
grafik


AlCalzone:
@robertsLando Any questions/thoughts?

robertsLando:
Nope, it's clear :)

PR #2737, opened 2021-05-30
fix(config): add missing param 1 config option to Jasco 35931

airdrummingfool:
Parameter 1 controls the power-on behavior of the light (i.e. after a power outage). Right now the available options both involve the light coming on after power is restored, but there is another option that restores the light's previous state (which means if it was off before losing power, it will stay off after power is restored).

I found the 3rd option listed here: http://manuals-backend.z-wave.info/make.php?lang=en&sku=39723&cert=ZC10-17075703, but was unable to find any other detailed documentation for this device. However, I own multiple of these devices and have already tested this change on my local zwave-js instance.

This is an important option IMO - I've woken up in the middle of the night because my bedroom lights came on after a power fluctuation 😆

### DIFF
--- a/packages/config/config/devices/0x0063/35931.json
+++ b/packages/config/config/devices/0x0063/35931.json
@@ -28,11 +28,11 @@
 					"value": 0
 				},
 				{
-					"label": "Last brightness setting",
+					"label": "Last non-zero brightness",
 					"value": 1
 				},
 				{
-					"label": "Last lamp status",
+					"label": "Previous state",
 					"value": 2
 				}
 			]
--- a/packages/config/config/devices/0x0063/35931.json
+++ b/packages/config/config/devices/0x0063/35931.json
@@ -19,7 +19,7 @@
 			"label": "Dim level when light is turned ON",
 			"valueSize": 1,
 			"minValue": 0,
-			"maxValue": 1,
+			"maxValue": 2,
 			"defaultValue": 0,
 			"allowManualEntry": false,
 			"options": [
@@ -30,6 +30,10 @@
 				{
 					"label": "Last brightness setting",
 					"value": 1
+				},
+				{
+					"label": "Last lamp status",
+					"value": 2
 				}
 			]
 		},
--- a/packages/config/config/devices/0x0063/35931_39723.json
+++ b/packages/config/config/devices/0x0063/35931_39723.json
@@ -1,7 +1,7 @@
 {
 	"manufacturer": "GE/Jasco",
 	"manufacturerId": "0x0063",
-	"label": "35931",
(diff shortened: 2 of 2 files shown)


packages/config/config/devices/0x0063/35931.json
```diff
@@ -30,6 +30,10 @@
 				{
 					"label": "Last brightness setting",
 					"value": 1
+				},
+				{
+					"label": "Last lamp status",
+					"value": 2
 				}
```
AlCalzone: So 1 is the last non-zero brightness and 2 is the previous state?

airdrummingfool: Correct. With `2`, a lamp that is off will remain off after a power cycle. If the lamp is on before power is cycled, the behavior is the same as `1`.

AlCalzone:
@blhoward2 the Z-Wave alliance lists this device as Enbrighten 39723, not GE/Jasco 35931. Any clue why we call it differently?

airdrummingfool:
As a data point, my bulbs have the Enbrighten logo (no GE/Jasco markings) and are labeled `35931` (i.e. a mix of the two).

gregsheremeta:
As far as I can tell, Jasco makes every zwave thing that's branded GE or Enbrighten. They also self brand some stuff. They also make a lot of Honeywell branded stuff.

Forgive me if this is already well-known -- I'm new :)

blhoward2:
Yes, Jasco makes all of those and some for other companies. @AlCalzone they reuse device fingerprints in some other areas I believe but I'll look into this more.

AlCalzone:
@blhoward2 let me know what you find out

blhoward2:
So, it appears that both the 35931 and the 39723 share the same fingerprint.

https://products.z-wavealliance.org/products/1942?selectedFrequencyId=-1

https://products.z-wavealliance.org/products/2528?selectedFrequencyId=-1

packages/config/config/devices/0x0063/35931.json
```diff
@@ -30,6 +30,10 @@
 				{
 					"label": "Last brightness setting",
```
AlCalzone:
```suggestion
					"label": "Last non-zero brightness",
```

packages/config/config/devices/0x0063/35931.json
```diff
@@ -30,6 +30,10 @@
...
 					"value": 1
+				},
+				{
+					"label": "Last lamp status",
```
AlCalzone:
```suggestion
					"label": "Previous state",
```

AlCalzone (changes requested):

AlCalzone (approved):

PR #2798, opened 2021-06-07
fix: make `startLevelChange` options optional in MultilevelSwitchCC

robertsLando:


AlCalzone:
What is this supposed to achieve? A level change needs a direction (up/down) and either a start level or the flag to ignore the start level.

robertsLando:
@AlCalzone I always called this without args and it was working, I noticed options was not optional when switching to ts. I was wondering if them could be optional and why it was working even without options specified

AlCalzone:
> why it was working even without options specified

Probably some implicit JS weirdness, because the direction flag is determined by `LevelChangeDirection[direction] << 6`, so in your case `undefined << 6`. You could also never perform a "down" level change that way.

Anyways, the options are required and `zwave-js` even fails to build when they aren't. Let's book this under "TypeScript knows better".

PR #2823, opened 2021-06-09
fix(config): enable auto-idling notifications for MH9-CO2

kvaster:
mh9-co2 will not reset notification to idle, it will only resend alert each 30 seconds if threshold is crossed.

Seems that this is cause notification version is 3:

```
      "ccSpecific": {
        "notificationType": 3
      },
```

### DIFF
--- a/packages/config/config/devices/0x015f/mh9-co2-wa.json
+++ b/packages/config/config/devices/0x015f/mh9-co2-wa.json
@@ -84,5 +84,8 @@
 				}
 			]
 		}
+	},
+	"compat": {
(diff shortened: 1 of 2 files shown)


AlCalzone (approved):
Fix looks good, but 
```
"notificationType": 3
```
is not the CC version :)

kvaster:
> Fix looks good, but
> 
> ```
> "notificationType": 3
> ```
> 
> is not the CC version :)

Thanks :)
Is CC version logged ?

AlCalzone:
during the interview and its saved in `<homeid>.json`

PR #2842, opened 2021-06-12
fix(config): add additional productID for the Yale YKFCON

disforw:


disforw:
@zwave-js-bot fix lint

packages/config/config/devices/0x0129/ykfcon.json
```diff
@@ -8,12 +8,17 @@
...
+		{
+			"productType": "0x0600",
+			"productId": "0x0000"
 		}
 	],
 	"firmwareVersion": {
 		"min": "0.0",
 		"max": "255.255"
 	},
+	"supportsZWavePlus": true,
```
AlCalzone:
```suggestion
```
We don't use this anymore

packages/config/config/devices/0x0129/ykfcon.json
```diff
@@ -29,7 +34,7 @@
 			"$import": "templates/yale_template.json#auto_relock"
 		},
 		"3": {
-			"$import": "templates/yale_template.json#auto_relock_time_180"
+			"$import": "templates/yale_template.json#manual_relock_time"
```
AlCalzone, line 36: Did this change?
/cc @blhoward2

AlCalzone (changes requested):

packages/config/config/devices/0x0129/ykfcon.json
disforw: As per documentation, param 3 is 7-60 with default value of 7. This matches the “manual_relock_time” template.
http://manuals-backend.z-wave.info/make.php?lang=en&sku=YKFCON&cert=ZC10-18015918

disforw (commented):
No changes made.

packages/config/config/devices/0x0129/ykfcon.json
blhoward2: This looks right but that parameter is inconsistent in the template itself. It should be renamed to Auto Relock Time to match the others and the description should be removed. The template reference itself should also change to match the others (and fixed in any other files).

I can look at this later unless someone beats me to it.

AlCalzone: This manual also mentions the auto relock time of max. 180:
https://products.z-wavealliance.org/ProductManual/File%3Ffolder%3D%26filename%3DMarketCertificationFiles/3211/YaleRealLivingAssureSL_Z-WavePlus_SystemIntegratorsGuide.pdf

Could this be the difference between Yale's Z-Wave Module 1 and 2?

Edit: The manual I found isn't for the keyfree lock though, so maybe they do have two versions of the parameter 🤷🏻‍♂️

blhoward2 (approved):

AlCalzone (approved):
We'll clean up the template in another PR

PR #2845, opened 2021-06-13
fix(config): add v1 alarm mapping for Yale YRD120

mw-white:
The Yale YRD120 lock uses the same v1 notifications as the YRD220/240 and YRL210.  This patch adds the compat section to the YRD120 config.


AlCalzone:
@zwave-js-bot rename commit

PR #2852, opened 2021-06-14
fix(driver): prevent simultaneous config updates

### LINKED ISSUE #1304: [bug] Config update to 7.7.2-20210602 failed
### Version

**Checklist:**

- [ ] I am **not** using HomeAssistant. **Or:** a developer has told me to come here.
- [X] I have checked the troubleshooting section and my problem is **not** described there.
- [X] I have read the changelog problem was **not** mentioned there.

Build/Run method

- [X] Docker
- [ ] PKG
- [ ] Snap package
- [ ] Manually built (git clone - yarn install - yarn run build )

zwavejs2mqtt version: 4.5.1
zwave-js version: 7.7.1

### Describe the bug

Updating the device config files fails.

There are also some usability issues with the update process. When you click the Install button, the dialog closes but the notification is still apparent, and you can install many times in parallel.

```
2021-06-03T23:09:26.634Z DRIVER   Installing version 7.7.2-20210602 of configuration DB...
2021-06-03T23:09:38.087Z DRIVER   Installing version 7.7.2-20210602 of configuration DB...
```

The update process is fairly opaque, maybe this is a node-zwave-js problem, but there is no indication of what's going on. Running on my RPi4, it consumes all the CPU and disk I/O. Is it expected that config file updates are so resource intensive? Is this because I'm running on the aarch64 arch? It took 6 mins to see the failure.

### To Reproduce

1. Click Install in the Device Config updates
2. Watch zwave-js logs for failure

### Expected behavior

The update succeeds.

I should not be allowed to issue multiple Installs at the same time, it should be disabled while one is in progress.

### Additional context

Here are the zwave-js logs:
```
2021-06-03T22:56:34.634Z DRIVER   Checking for configuration updates...
2021-06-03T22:56:34.792Z DRIVER   Configuration update available: 7.7.2-20210602
2021-06-03T22:56:38.305Z DRIVER   Installing version 7.7.2-20210602 of configuration DB...
2021-06-03T23:02:31.526Z DRIVER   Config update failed: Package manager exited with code 1
                                  warning Resolution field "@zwave-js/config@7.7.2-20210602" is incompatible wit
                                  h requested version "@zwave-js/config@7.7.1"
                                  error https://registry.yarnpkg.com/prettier-standard/-/prettier-standard-16.4.
                                  1.tgz: ESOCKETTIMEDOUT
                                  error https://registry.yarnpkg.com/rxjs/-/rxjs-6.6.7.tgz: ESOCKETTIMEDOUT
                                  warning " > eslint-loader@4.0.2" has unmet peer dependency "webpack@^4.0.0 ||
                                  ^5.0.0".
                                  warning "prettier-standard > prettierx > @typescript-eslint/typescript-estree
                                  > tsutils@3.21.0" has unmet peer dependency "typescript@>=2.8.0 || >= 3.2.0-de
                                  v || >= 3.3.0-dev || >= 3.4.0-dev || >= 3.5.0-dev || >= 3.6.0-dev || >= 3.6.0-
                                  beta || >= 3.7.0-dev || >= 3.7.0-beta".
                                  error An unexpected error occurred: "ENOENT: no such file or directory, open '
                                  /usr/local/share/.cache/yarn/v6/npm-prettier-standard-16.4.1-964e7e3f3029427ab
                                  8e01bdffb72dea9a87f62c8-integrity/node_modules/prettier-standard/.yarn-metadat
                                  a.json'".
```

Here are the zwavejs2mqtt logs, it seems to think it was successful?
```
z2m        | 2021-06-03 15:56:38.173 INFO ZWAVE: Calling api installConfigUpdate with args: [ [length]: 0 ]
z2m        | 2021-06-03 16:02:31.528 INFO ZWAVE: Success zwave api call installConfigUpdate { success: true, message: 'Success zwave api call', result: false }
```

robertsLando:
Fixes https://github.com/zwave-js/zwavejs2mqtt/issues/1304

cc @kpine

### DIFF
--- a/packages/zwave-js/package.json
+++ b/packages/zwave-js/package.json
@@ -85,8 +85,5 @@
     "clean": "tsc -b tsconfig.build.json --clean",
     "prewatch": "gulp prebuild",
     "watch": "tsc -b tsconfig.build.json --watch --pretty"
-  },
-  "devDependencies": {
-    "@types/proper-lockfile": "^4.1.1"
   }
 }
--- a/packages/zwave-js/package.json
+++ b/packages/zwave-js/package.json
@@ -72,6 +72,7 @@
     "ansi-colors": "^4.1.1",
     "axios": "^0.21.1",
     "fs-extra": "^9.0.1",
+    "proper-lockfile": "^4.1.2",
     "reflect-metadata": "^0.1.13",
     "serialport": "^9.1.0",
     "source-map-support": "^0.5.19",
@@ -84,5 +85,8 @@
     "clean": "tsc -b tsconfig.build.json --clean",
     "prewatch": "gulp prebuild",
     "watch": "tsc -b tsconfig.build.json --watch --pretty"
+  },
+  "devDependencies": {
+    "@types/proper-lockfile": "^4.1.1"
   }
 }
--- a/packages/zwave-js/src/lib/driver/UpdateConfig.ts
+++ b/packages/zwave-js/src/lib/driver/UpdateConfig.ts
@@ -68,20 +68,28 @@ export async function installConfigUpdate(newVersion: string): Promise<void> {
 			requireLockfile: false,
 		});
 	} catch {
+		_installConfigLock = false;
 		throw new ZWaveError(
 			`Config update failed: No package manager detected or package.json not found!`,
 			ZWaveErrorCodes.Config_Update_PackageManagerNotFound,
 		);
 	}
 
-	// And install it
-	const result = await pak.overrideDependencies({
-		"@zwave-js/config": newVersion,
-	});
-	if (result.success) return;
-	throw new ZWaveError(
-		`Config update failed: Package manager exited with code ${result.exitCode}
+	try {
+		// And install it
+		const result = await pak.overrideDependencies({
+			"@zwave-js/config": newVersion,
+		});
+
+		if (result.success) return;
+
+		throw new ZWaveError(
+			`Config update failed: Package manager exited with code ${result.exitCode}
 ${result.stderr}`,
-		ZWaveErrorCodes.Config_Update_InstallFailed,
-	);
+			ZWaveErrorCodes.Config_Update_InstallFailed,
+		);
+	} catch (error) {
+		_installConfigLock = false;
+		throw error;
+	}
 }
--- a/packages/zwave-js/src/lib/driver/UpdateConfig.ts
+++ b/packages/zwave-js/src/lib/driver/UpdateConfig.ts
@@ -2,9 +2,8 @@ import { detectPackageManager, PackageManager } from "@alcalzone/pak";
 import { ZWaveError, ZWaveErrorCodes } from "@zwave-js/core";
 import { isObject } from "alcalzone-shared/typeguards";
 import axios from "axios";
+import * as lockfile from "proper-lockfile";
 import * as semver from "semver";
-
-let _installConfigLock = false;
 
 /**
  * Checks whether there is a compatible update for the currently installed config package.
@@ -52,14 +51,6 @@ export async function checkForConfigUpdates(
  * Installs the update for @zwave-js/config with the given version.
  */
 export async function installConfigUpdate(newVersion: string): Promise<void> {
-	if (_installConfigLock) {
-		throw new ZWaveError(
-			`Config update failed: another installation is in progress`,
-			ZWaveErrorCodes.Config_Update_InstallFailed,
-		);
-	}
-
-	_installConfigLock = true;
 	// Check which package manager to use for the update
 	let pak: PackageManager;
 	try {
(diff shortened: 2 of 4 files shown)


packages/zwave-js/src/lib/driver/UpdateConfig.ts
```diff
@@ -58,20 +68,28 @@ export async function installConfigUpdate(newVersion: string): Promise<void> {
...
-	if (result.success) return;
-	throw new ZWaveError(
-		`Config update failed: Package manager exited with code ${result.exitCode}
+	try {
+		// And install it
+		const result = await pak.overrideDependencies({
+			"@zwave-js/config": newVersion,
+		});
+
+		if (result.success) return;
```
AlCalzone: Need to release the lock here too

packages/zwave-js/src/lib/driver/UpdateConfig.ts
```diff
@@ -50,6 +52,14 @@ export async function checkForConfigUpdates(
  * Installs the update for @zwave-js/config with the given version.
  */
 export async function installConfigUpdate(newVersion: string): Promise<void> {
+	if (_installConfigLock) {
+		throw new ZWaveError(
+			`Config update failed: another installation is in progress`,
+			ZWaveErrorCodes.Config_Update_InstallFailed,
+		);
```
AlCalzone: Not sure if I want to throw an error here. In other locations I've indicated a busy status by returning `false` and `true` if something was done.

AlCalzone (changes requested):
Two minor things. Additionally, I think the lock should live in the driver instance, which capsules these methods.

robertsLando:
> Additionally, I think the lock should live in the driver instance, which capsules these methods.

I don't think so as the update is done on the same lock file, this means that even if there are multiple drivers active there should be only one update at a time running

packages/zwave-js/src/lib/driver/UpdateConfig.ts
robertsLando: Ok so should I return false here? I think that with an error there are more informations about the reason

packages/zwave-js/src/lib/driver/UpdateConfig.ts
robertsLando: Yep right :(

AlCalzone:
Alternatively we could use a real lockfile like the DB does now does to persist across potential crashes.

robertsLando:
> Alternatively we could use a real lockfile like the DB does now does to persist across potential crashes.

I dunno if/why this should be required. I mean if an installation is stopped by a crash you can simply restart it, or not?

AlCalzone:
Not entirely sure, but it seems like one cause of the currupted DB files was multiple separate driver instances accessing the same files. Theoretically it would be possible to have two separate processes, both trying to do an update at the same time, spawning multiple concurrent `yarn` processes.

That could be avoided like this:
https://github.com/AlCalzone/jsonl-db/pull/180/files?file-filters%5B%5D=.ts#diff-196dc6cfdb581213173439841d9e85ada909cf40face80a3ac1b6739bbd85bf6

robertsLando:
How do I get the path to the `.lock` file using `pak` ? Should I lock the `.lock` file or `package.json`?

packages/zwave-js/src/lib/driver/UpdateConfig.ts
```diff
@@ -64,11 +65,38 @@ export async function installConfigUpdate(newVersion: string): Promise<void> {
 		);
 	}
 
+	const pkgFilepath = `${pak.cwd}/package.json`;
```
AlCalzone: Add this to the call to `await detectPackageManager({...})`:
```
	setCwdToPackageRoot: true,
```
so `pak.cwd` actually contains the package root.

packages/zwave-js/src/lib/driver/UpdateConfig.ts
```diff
@@ -64,11 +65,38 @@ export async function installConfigUpdate(newVersion: string): Promise<void> {
...
+
+	await lockfile.lock(pkgFilepath, {
+		// We cannot be sure that the file exists before acquiring the lock
+		realpath: false,
```
AlCalzone:
```suggestion
```
Here, we can be sure that package.json exists

packages/zwave-js/src/lib/driver/UpdateConfig.ts
```diff
@@ -64,11 +65,38 @@ export async function installConfigUpdate(newVersion: string): Promise<void> {
...
+
+		stale:
+			// Avoid timeouts during testing
+			process.env.NODE_ENV === "test"
+				? 100000
+				: /* istanbul ignore next - this is impossible to test */ undefined,
```
AlCalzone:
```suggestion
```
And I don't think we need this either.

packages/zwave-js/src/lib/driver/UpdateConfig.ts
```diff
@@ -64,11 +65,38 @@ export async function installConfigUpdate(newVersion: string): Promise<void> {
 		);
 	}
 
+	const pkgFilepath = `${pak.cwd}/package.json`;
+
+	await lockfile.lock(pkgFilepath, {
```
AlCalzone: This can throw if we cannot acquire the lock. In that case we should return a better (ZWave) error. The code can be `Config_Update_InstallFailed`.

packages/zwave-js/src/lib/driver/UpdateConfig.ts
```diff
@@ -64,11 +65,38 @@ export async function installConfigUpdate(newVersion: string): Promise<void> {
...
+
 	// And install it
 	const result = await pak.overrideDependencies({
 		"@zwave-js/config": newVersion,
 	});
+
+	// Free the lock
+	try {
+		if (await lockfile.check(pkgFilepath, { realpath: false }))
+			await lockfile.unlock(pkgFilepath, { realpath: false });
```
AlCalzone: See comment above regarding realpath
```suggestion
		if (await lockfile.check(pkgFilepath))
			await lockfile.unlock(pkgFilepath);
```

packages/zwave-js/package.json
```diff
@@ -84,5 +85,8 @@
     "clean": "tsc -b tsconfig.build.json --clean",
     "prewatch": "gulp prebuild",
     "watch": "tsc -b tsconfig.build.json --watch --pretty"
+  },
+  "devDependencies": {
+    "@types/proper-lockfile": "^4.1.1"
   }
```
AlCalzone: In a monorepo, the `devDependencies` go into the root package.
```
yarn add --dev -W <package-name>
```

AlCalzone (changes requested):

AlCalzone (approved):
OK with my changes

robertsLando:
Perfect :)

PR #2854, opened 2021-06-14
fix(config): update Fibaro FGD212 configuration

glyph-se:
Adjustments to value ranges, units, casing and some usages of the master_template.json

I went through all of the parameters and updated them. It contains a lot of changes to units and value ranges, the inclusion of the master_template for enable/disable were appropriate and updates to the label and description in order to make it more clear.

This is my first PR in this project so sorry if I missed anything or did anything wrong. 

There are probably still improvements that can be made since this is my first try at changing anything like this. But I still felt like this is an improvement to the current file and didn't want to spend to much time If I am on going in the wrong direction.

Is there a way to validate / test the configuration file?
EDIT: Saw that there are a bunch of unit tests for this.

### DIFF
--- a/packages/config/config/devices/0x010f/fgd212_3.5.json
+++ b/packages/config/config/devices/0x010f/fgd212_3.5.json
@@ -345,19 +345,19 @@
 		},
 		"27[0x01]": {
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Send secure Z-wave commands to 2nd association group"
+			"label": "Send Secure Z-wave Commands to 2nd Association Group"
 		},
 		"27[0x02]": {
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Send secure Z-wave commands to 3rd association group"
+			"label": "Send Secure Z-wave Commands to 3rd Association Group"
 		},
 		"27[0x04]": {
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Send secure Z-wave commands to 4th association group"
+			"label": "Send Secure Z-wave Commands to 4th Association Group"
 		},
 		"27[0x08]": {
 			"$import": "../templates/master_template.json#base_enable_disable",
-			"label": "Send secure Z-wave commands to 5th association group"
+			"label": "Send Secure Z-wave Commands to 5th Association Group"
 		},
 		"28": {
 			"$import": "../templates/master_template.json#base_enable_disable",
--- a/packages/config/config/devices/0x010f/fgd212_3.5.json
+++ b/packages/config/config/devices/0x010f/fgd212_3.5.json
@@ -212,7 +212,6 @@
 		},
 		"19": {
 			"label": "Forced Switch On Brightness Level",
-			"description": "Percentage value.",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 99,
--- a/packages/config/config/devices/0x010f/fgd212_3.5.json
+++ b/packages/config/config/devices/0x010f/fgd212_3.5.json
@@ -547,7 +547,7 @@
 			]
 		},
 		"39": {
-			"label": "Power Limit (Overload)",
+			"label": "Power Limit",
 			"description": "Defined value turns off load",
 			"valueSize": 2,
 			"unit": "W",
@@ -672,7 +672,7 @@
 			"defaultValue": 600
 		},
 		"45": {
-			"label": "Overload Alarm Report",
+			"label": "Power Limit Alarm Report",
 			"description": "Load power consumption too high",
 			"valueSize": 1,
 			"minValue": 0,
--- a/packages/config/config/devices/0x010f/fgd212_3.5.json
+++ b/packages/config/config/devices/0x010f/fgd212_3.5.json
@@ -48,7 +48,7 @@
 		},
 		"3": {
 			"label": "Incandescence Level of Dimmable CFL*",
-			"description": "Value relative to the range defined by minimum and maximum brightness level configuration",
+			"description": "Relative to the range defined by minimum and maximum brightness level configuration",
 			"valueSize": 1,
 			"minValue": 1,
 			"maxValue": 99,
@@ -64,7 +64,7 @@
 		},
 		"5": {
 			"label": "Automatic Control: Dimming Step Size",
-			"description": "Value relative to the range defined by minimum and maximum brightness level configuration",
+			"description": "Relative to the range defined by minimum and maximum brightness level configuration",
 			"valueSize": 1,
 			"minValue": 1,
 			"maxValue": 99,
@@ -80,7 +80,7 @@
 		},
 		"7": {
 			"label": "Manual Control: Dimming Step Size",
-			"description": "Value relative to the range defined by minimum and maximum brightness level configuration",
+			"description": "Relative to the range defined by minimum and maximum brightness level configuration",
 			"valueSize": 1,
 			"minValue": 1,
 			"maxValue": 99,
@@ -212,7 +212,7 @@
 		},
 		"19": {
 			"label": "Forced Switch On Brightness",
-			"description": "Value relative to the range defined by minimum and maximum brightness level configuration",
+			"description": "Relative to the range defined by minimum and maximum brightness level configuration",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 99,
--- a/packages/config/config/devices/0x010f/fgd212_3.5.json
+++ b/packages/config/config/devices/0x010f/fgd212_3.5.json
@@ -123,19 +123,19 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "ALL ON not Active, ALL OFF not Active",
+					"label": "ALL ON not active, ALL OFF not active",
 					"value": 0
 				},
 				{
-					"label": "ALL ON not Active, ALL OFF Active",
+					"label": "ALL ON not active, ALL OFF active",
 					"value": 1
 				},
 				{
-					"label": "ALL ON Active, ALL OFF not Active",
+					"label": "ALL ON active, ALL OFF not active",
 					"value": 2
 				},
 				{
-					"label": "ALL ON Active, ALL OFF Active",
+					"label": "ALL ON active, ALL OFF active",
 					"value": 255
 				}
 			]
@@ -154,11 +154,11 @@
 					"value": 0
 				},
 				{
-					"label": "Force Auto-calibration Without Fibaro Bypass",
+					"label": "Force auto-calibration without Fibaro Bypass",
 					"value": 1
 				},
 				{
-					"label": "Force Auto-calibration With Fibaro Bypass",
+					"label": "Force auto-calibration with Fibaro Bypass",
 					"value": 2
 				}
 			]
@@ -173,11 +173,11 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "No Calibration or Manual Settings",
+					"label": "No calibration or manual settings",
 					"value": 0
 				},
 				{
-					"label": "Dimmer Operates on Auto-calibration",
+					"label": "Dimmer operates on auto-calibration",
 					"value": 1
 				}
 			]
@@ -218,7 +218,7 @@
 			"defaultValue": 0,
 			"options": [
 				{
-					"label": "Disable, Previous Value is Used",
+					"label": "Disabled, previous value is used",
 					"value": 0
 				}
 			]
@@ -233,21 +233,21 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Mono-stable Input (Button)",
+					"label": "Mono-stable input (button)",
 					"value": 0
 				},
 				{
-					"label": "Bi-stable Input (Switch)",
+					"label": "Bi-stable input (switch)",
 					"value": 1
 				},
 				{
-					"label": "Roller Blind Switch (UP / DOWN)",
+					"label": "Roller blind switch (UP / DOWN)",
 					"value": 2
 				}
 			]
 		},
 		"21": {
-			"label": "Value Sent to Associated Devices on Single Click",
+			"label": "Value sent to associated devices on Single Click",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 1,
@@ -256,7 +256,7 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Set Associated Devices to Last State",
+					"label": "Set associated devices to last state",
 					"value": 0
 				},
 				{
@@ -275,11 +275,11 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Output Changes on Input Switch Change",
+					"label": "Output changes on input switch change",
 					"value": 0
 				},
 				{
-					"label": "Output Synced with Input Switch Position",
+					"label": "Output synced with input switch position",
 					"value": 1
 				}
 			]
@@ -515,11 +515,11 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Standard Mode",
+					"label": "Standard mode",
 					"value": 0
 				},
 				{
-					"label": "S1 Operates as S2, S2 Operates as S1",
+					"label": "S1 operates as S2, S2 operates as S1",
 					"value": 1
 				}
 			]
@@ -534,15 +534,15 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Forced Leading Edge",
+					"label": "Forced leading edge",
 					"value": 0
 				},
 				{
-					"label": "Forced Trailing Edge",
+					"label": "Forced trailing edge",
 					"value": 1
 				},
 				{
-					"label": "Select from Auto-calibration",
+					"label": "Select from auto-calibration",
 					"value": 2
 				}
 			]
@@ -557,11 +557,11 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Leading Edge",
+					"label": "Leading edge",
 					"value": 0
 				},
 				{
-					"label": "Trailing Edge",
+					"label": "Trailing edge",
 					"value": 1
 				}
 			]
@@ -577,15 +577,15 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Disabled (Dimming is Possible)",
+					"label": "Disabled (dimming is possible)",
 					"value": 0
 				},
 				{
-					"label": "Enabled (Dimming is not Possible)",
+					"label": "Enabled (dimming is not possible)",
 					"value": 1
 				},
 				{
-					"label": "Select From Auto-calibration",
+					"label": "Select from auto-calibration",
 					"value": 2
 				}
 			]
@@ -600,11 +600,11 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Dimming Possible",
+					"label": "Dimming possible",
 					"value": 0
 				},
 				{
-					"label": "Dimming not Possible",
+					"label": "Dimming not possible",
 					"value": 1
 				}
 			]
@@ -620,15 +620,15 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "No Soft-start",
+					"label": "No soft-start",
 					"value": 0
 				},
 				{
-					"label": "Short Soft-start (0,1s)",
+					"label": "Short soft-start (0,1s)",
 					"value": 1
 				},
 				{
-					"label": "Long Soft-start (0,5s)",
+					"label": "Long soft-start (0,5s)",
 					"value": 2
 				}
 			]
@@ -647,19 +647,19 @@
 					"value": 0
 				},
 				{
-					"label": "After First Power On",
+					"label": "After first power On",
 					"value": 1
 				},
 				{
-					"label": "After Each Power On",
+					"label": "After each power On",
 					"value": 2
 				},
 				{
-					"label": "After First Power Up or Each LOAD ERROR, SURGE or OVERCURRENT",
+					"label": "After first power up or each LOAD ERROR, SURGE or OVERCURRENT",
 					"value": 3
 				},
 				{
-					"label": "After Each Power Up or Each LOAD ERROR, SURGE or OVERCURRENT",
+					"label": "After each power up or each LOAD ERROR, SURGE or OVERCURRENT",
 					"value": 4
 				}
 			]
@@ -674,11 +674,11 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Device Permanently Disabled",
+					"label": "Device permanently disabled",
 					"value": 0
 				},
 				{
-					"label": "Three Attempts to Turn On the Load",
+					"label": "Three attempts to turn on the load",
 					"value": 1
 				}
 			]
@@ -696,7 +696,7 @@
 					"value": 0
 				},
 				{
-					"label": "Always Enabled",
+					"label": "Always enabled",
 					"value": 255
 				}
 			]
@@ -720,19 +720,19 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "No Reaction",
+					"label": "No reaction",
 					"value": 0
 				},
 				{
-					"label": "Turn On the Load",
+					"label": "Turn on the load",
 					"value": 1
 				},
 				{
-					"label": "Turn Off the Load",
+					"label": "Turn off the load",
 					"value": 2
 				},
 				{
-					"label": "Load Blinking",
+					"label": "Load blinking",
 					"value": 3
 				}
 			]
@@ -747,19 +747,19 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "No Reaction",
+					"label": "No reaction",
 					"value": 0
 				},
 				{
-					"label": "Turn On the Load",
+					"label": "Turn on the load",
 					"value": 1
 				},
 				{
-					"label": "Turn Off the Load",
+					"label": "Turn off the load",
 					"value": 2
 				},
 				{
-					"label": "Load Blinking",
+					"label": "Load blinking",
 					"value": 3
 				}
 			]
@@ -774,19 +774,19 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "No Reaction",
+					"label": "No reaction",
 					"value": 0
 				},
 				{
-					"label": "Turn On the Load",
+					"label": "Turn on the load",
 					"value": 1
 				},
 				{
-					"label": "Turn Off the Load",
+					"label": "Turn off the load",
 					"value": 2
 				},
 				{
-					"label": "Load Blinking",
+					"label": "Load blinking",
 					"value": 3
 				}
 			]
@@ -801,19 +801,19 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "No Reaction",
+					"label": "No reaction",
 					"value": 0
 				},
 				{
-					"label": "Turn On the Load",
+					"label": "Turn on the load",
 					"value": 1
 				},
 				{
-					"label": "Turn Off the Load",
+					"label": "Turn off the load",
 					"value": 2
 				},
 				{
-					"label": "Load Blinking",
+					"label": "Load blinking",
 					"value": 3
 				}
 			]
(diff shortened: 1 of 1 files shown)


AlCalzone:
> Is there a way to validate / test the configuration file?

You can execute `yarn run lint:config` locally or just trust the Github checks.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -57,37 +57,39 @@
 		},
 		"4": {
 			"label": "Incandescence time of dimmable CFL*",
-			"description": "time required for switching. 0s - 25,5s",
+			"description": "Each step is 100 ms, e.g. 5 = 500 ms",
```
AlCalzone: This should be a unit instead.
```suggestion
			"unit": "100 ms",
```

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -57,37 +57,39 @@
...
 			"valueSize": 1,
 			"unit": "%",
 			"minValue": 1,
 			"maxValue": 99,
 			"defaultValue": 1
 		},
 		"6": {
-			"label": "Dimming Step Timing at Automatic Control",
+			"label": "Automatic control: Dimming step timing",
+			"description": "Each step equals 10 ms, e.g. 5 = 50 ms",
```
AlCalzone: This should also be a unit

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -57,37 +57,39 @@
...
 			"valueSize": 1,
+			"unit": "%",
 			"minValue": 1,
 			"maxValue": 99,
 			"defaultValue": 1
 		},
 		"8": {
-			"label": "Dimming Step Timing at Manual Control",
+			"label": "Manual control: Dimming step timing",
+			"description": "Each step equals 10 ms, e.g. 5 = 50ms",
```
AlCalzone: This should also be a unit

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -98,15 +100,15 @@
 		},
 		"10": {
 			"label": "Timer functionality (auto - off)",
-			"description": "Defines the setting of the timer",
+			"description": "Will automatically turn off after the specified time from switching on (0 = disable)",
```
AlCalzone: The special value `0` should instead be an `"option"`. With `allowManualEntry` set to `true` (the default), this is interpreted as a suggestion in addition to the free entry.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -171,41 +171,41 @@
...
+					"label": "Dimmer operates on auto-calibration",
 					"value": 1
 				}
 			]
 		},
 		"15": {
-			"label": "Burnt out bulb detection",
-			"description": "Percentage value of power variation",
+			"label": "Burnt out bulb detection power variation (LOAD ERROR)",
+			"description": "Percentage value of standard power consumption as measured during calibration, 0 = detection disabled",
```
AlCalzone: "0 = detection disabled" should be an `"option"`

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -171,41 +171,41 @@
...
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 99,
 			"defaultValue": 30
 		},
 		"16": {
-			"label": "Time delay of a burnt out bulb",
-			"description": "Time of delay for power variation detection",
+			"label": "Burnt out bulb detection time delay (LOAD ERROR)",
+			"description": "0 = detection disabled",
```
AlCalzone: this should be an `"option"`

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -171,41 +171,41 @@
...
 			"valueSize": 2,
+			"unit": "seconds",
 			"minValue": 0,
 			"maxValue": 255,
 			"defaultValue": 5
 		},
 		"19": {
 			"label": "Forced switch on brightness level",
-			"description": "Dimmer 2 will always set this brightness level.",
+			"description": "Percentage value, 0 = disabled, previous value is used.",
```
AlCalzone: `0` should be an `"option"`

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -238,151 +238,118 @@
...
-					"label": "do not send on double click",
+					"label": "Do not send on double click",
 					"value": 8
 				},
 				{
-					"label": "send 0xFF value on double click",
+					"label": "Send 0xFF value on double click",
 					"value": 16
 				}
 			]
```
AlCalzone: This looks like it should be possible to combine these options. To model that, we use [partial parameters](https://zwave-js.github.io/node-zwave-js/#/config-files/partial-parameters).

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -238,151 +238,118 @@
...
```
AlCalzone: dito

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -238,151 +238,118 @@
...
-				}
-			]
+			"label": "3-way switch function",
+			"description": "Key S2 also controls the dimmer",
+			"$import": "../templates/master_template.json#base_enable_disable"
 		},
 		"27": {
 			"label": "Associations in Z-Wave network security mode",
-			"description": "Defines commands are sent as secure or non-secure",
+			"description": "Defines commands are sent as secure or non-secure. May be combined, e.g. 1 + 2 = 3 means that 2nd and 3rd are sent as secure",
```
AlCalzone: This parameter definitely should be split into partial parameters.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -541,62 +491,61 @@
...
 			"minValue": 0,
-			"maxValue": 255,
+			"maxValue": 4,
 			"defaultValue": 1,
 			"unsigned": true,
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "No auto-calibration of the load after power on",
+					"label": "No auto-calibration of after power on",
```
AlCalzone: I think we have a good opportunity to shorten the labels here. If the parameter label is just called "Auto-calibration", this could be called
```suggestion
					"label": "Never",
```

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -541,62 +491,61 @@
...
 					"value": 0
 				},
 				{
-					"label": "Auto-calibration performed after first power on",
+					"label": "Auto-calibration after first power on",
```
AlCalzone: And this 
```suggestion
					"label": "After first power on",
```

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -541,62 +491,61 @@
...
 					"value": 1
 				},
 				{
-					"label": "Auto-calibration performed after each power on",
+					"label": "Auto-calibration after each power on",
```
AlCalzone: and this
```suggestion
					"label": "After each power on",
```
etc.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -606,32 +555,44 @@
...
 				{
-					"label": "three attempts to turn on the load",
+					"label": "Three attempts to turn on the load",
 					"value": 1
 				}
 			]
 		},
 		"38": {
 			"label": "Brightness level correction for flickering loads",
+			"description": "The duration in seconds. 0 = disabled, 255 = always enabled",
```
AlCalzone: This description is redundant, please remove.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -846,75 +805,80 @@
 			]
 		},
 		"50": {
-			"label": "Active power reports",
-			"description": "Pwr level change resulting in power report sent",
+			"label": "Active power report",
+			"description": "The power level change as a percentage from the previous report that will result in a new power report being sent",
```
AlCalzone: This is quite verbose and could probably be shortened, for example:
```suggestion
			"label": "Power Report: Relative Change Threshold",
```

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -846,75 +805,80 @@
...
+					"label": "Disabled",
+					"value": 0
+				}
+			]
 		},
 		"53": {
-			"label": "Energy reports",
-			"description": "Energy level change resulting in energy report sent",
+			"label": "Energy report",
+			"description": "The energy level change resulting in energy report sent. Each step is 0.01 kWh, e.g. 10 = 0.1 kWh",
```
AlCalzone: Likewise:
```suggestion
			"label": "Energy Report: Change Threshold",
```
and the unit should be put in the `"unit"` property.

AlCalzone (changes requested):
Lots of good changes here - a few things need to be changed though. And please make sure that the labels conform with the [style guide](https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide?id=labels), especially this part:
> Labels should be Title Case.

glyph-se:
> Lots of good changes here - a few things need to be changed though.

Thanks.

> And please make sure that the labels conform with the [style guide](https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide?id=labels), especially this part:
> 
> > Labels should be Title Case.

I read the style guide carefully and tried to follow it to my best. I missed the information about the casing on the labels, I only saw the part saying that descriptions should be sentence case and then I guesses the labels from a few other files. But no problem, I will update it.

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: Thanks, reading carefully I now see that this is also mentioned in the style guide. I probably missed this caused there was no example of it there.

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: Ok, I was unsure if this was to be used even if the value 0 had a totally different meaning from the others.

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: `0` moved to `option`. Would it be correct to use the `"unit" : "%"` here? I guess not? So we leave that in the description?

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: Nice suggestion of formulating it different.

I went with "Perform Auto-calibration" instead of just "Auto-calibration". Since I thought that was a bit short and couldn't be understood without reading the different options. Ok?

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: Would `Active Power Report: Relative Change Since Previous Report` be ok, or still to long? I wanted to keep something saying that the previous report was the reference since that might not be obvious.

Could move that to the description though and use your suggestion as the label.

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: Here I thought about if I should use `"unit" : "0.01 kWh",` or `"unit" : "10 Wh",` any guidelines for that?

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: It is possible (and I did so). But the labels became very long, any suggestions?

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: done

packages/config/config/devices/0x010f/fgd212_3.5.json
AlCalzone: I'd prefer that to be in the description. If the label looks like it could be a sentence, it is too long IMO.

packages/config/config/devices/0x010f/fgd212_3.5.json
AlCalzone: No real guidelines, but we normally use the `kWh` variant.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -97,16 +99,22 @@
...
+				}
+			]
 		},
 		"11": {
-			"label": "ALL ON/ALL OFF function",
-			"description": "The parameter specifies the available functions.",
+			"label": "ALL ON/ALL OFF Function",
 			"valueSize": 2,
 			"minValue": 0,
 			"maxValue": 65535,
```
AlCalzone: I must have missed that on the last review. This should be 
```suggestion
			"maxValue": 255,
```
since the option values don't go above 255. Or is the value size 2 a mistake?

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -167,315 +173,383 @@
...
+					"label": "Dimmer Operates on Auto-calibration",
 					"value": 1
 				}
 			]
 		},
 		"15": {
-			"label": "Burnt out bulb detection",
-			"description": "Percentage value of power variation",
+			"label": "Burnt Out Bulb Detection Power Variation (LOAD ERROR)",
+			"description": "Percentage value of standard power consumption as measured during calibration",
```
AlCalzone, line 186: There should be a `"unit": "%"` too.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -167,315 +173,383 @@
...
+					"label": "Disable",
+					"value": 0
+				}
+			]
 		},
 		"19": {
-			"label": "Forced switch on brightness level",
-			"description": "Dimmer 2 will always set this brightness level.",
+			"label": "Forced Switch On Brightness Level",
+			"description": "Percentage value.",
```
AlCalzone:
```suggestion
```
Refer to the note here: https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide?id=units
This is not a percentage ;)

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -167,315 +173,383 @@
...
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Set associated devices to last state",
+					"label": "Set Associated Devices to Last State",
 					"value": 0
 				},
 				{
-					"label": "Dimmer 2 state synchronizes brightness level",
+					"label": "This Dimmers Current State Synchronizes Brightness Level",
```
AlCalzone: The previous label confused me, what about this?
```suggestion
					"label": "This dimmer's current brightness",
```

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -167,315 +173,383 @@
...
+			"label": "Send Command Frames in 2nd and 3rd Association Groups when Switching ON (Single Click)",
 			"valueSize": 1,
 			"minValue": 0,
-			"maxValue": 255,
-			"defaultValue": 1,
+			"maxValue": 1,
+			"defaultValue": 0,
 			"unsigned": true,
 			"allowManualEntry": false,
 			"options": [
```
AlCalzone: I think you should be able to base these partials off
`"$import": "../templates/master_template.json#base_enable_disable"`
so you'd just have to override the label and default value where necessary.

AlCalzone (changes requested):
A few more details. And I guess we need to highlight that a bit better:

> Option labels should be Sentence case.

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: The valueSize 2 is correct. Changed maxValue to 255.

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: I read that, thus I put it as a description and not a unit :)

My reasoning here was that my interpretation of the Fibaro was that you can set a percentage between 1 and 99 and that will be used as the default brightness value, with 0% and 100% being unavailable. But I guess it could also be that the scale 1-99 is transformed to 0-100 or 1-100 percentages.

We also have parameter 1 and 2 that talks about brightness levels, and the value here is a different brightness. If parameter 1 and 2 specified the brightness level to be 10-50, we can still specify 75 on parameter 19 here and that will be a "real" brightness level of 40.

You don't think it is confusing if we have the same name when they mean different things?

glyph-se:
> A few more details. And I guess we need to highlight that a bit better:
> 
> > Option labels should be Sentence case.

Or I need to read better... I'm also thinking if this could be automated somehow, casing should be easy to automate if the rules are this simple.

Anyway, it is fixed now I went through all your comments and checked everything myself again. Also added a metadata section at the end.

Please take another look

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: Or do I need to add something like?

`"description": "Scale refers to the range set by minimum and maximum brightness level configuration"`

blhoward2:
> I'm also thinking if this could be automated somehow, casing should be easy to automate if the rules are this simple.

We do automate this on import with the scripts, but it isn't ALWAYS truly sentence case. LED is always capitlized. Command Class Names follow their convention, etc.

glyph-se:
> > I'm also thinking if this could be automated somehow, casing should be easy to automate if the rules are this simple.
> 
> We do automate this on import with the scripts, but it isn't ALWAYS truly sentence case. LED is always capitlized. Command Class Names follow their convention, etc.

I see, guess there could be an exclude list, but could be quite much work to maintain it... I guess this shows that I'm not so used to this domain :)

Other projects I work with have similar rules which are also enforced by manual code review, after a while you get used to them and follow them automatically.

AlCalzone (dismissed):

AlCalzone:
@blhoward2 got anything to add?

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -32,62 +32,64 @@
 	"supportsZWavePlus": true,
```
blhoward2: Remove supprsZWavePlus. We don't use that anymore.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -32,62 +32,64 @@
...
 			"valueSize": 1,
 			"minValue": 1,
 			"maxValue": 99,
 			"defaultValue": 1
 		},
 		"4": {
-			"label": "Incandescence time of dimmable CFL*",
-			"description": "time required for switching. 0s - 25,5s",
+			"label": "Incandescence Time of Dimmable CFL*",
+			"unit": "100 ms",
```
blhoward2: Nitpicking, but please move the unit to after the valueSize throughout.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -157,8 +164,7 @@
 			]
 		},
 		"14": {
-			"label": "Auto-calibration status (read-only parameter)",
-			"description": "Operating mode of the Dimmer",
+			"label": "Auto-calibration Status",
```
blhoward2: The C should also be capitlized.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -171,41 +177,59 @@
...
+				{
+					"label": "Disable",
+					"value": 0
+				}
+			]
 		},
 		"16": {
-			"label": "Time delay of a burnt out bulb",
-			"description": "Time of delay for power variation detection",
+			"label": "Burnt Out Bulb Detection Time Delay (LOAD ERROR)",
```
blhoward2: Remove the (LOAD ERROR). Its not necessary to understand the parameter.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -238,244 +262,139 @@
...
 				{
-					"label": "Dimmer 2 state synchronizes brightness level",
+					"label": "This dimmer's current brightness",
 					"value": 1
 				}
 			]
 		},
 		"22": {
-			"label": "Change [On-Off] bi-stable keys",
+			"label": "Change Bi-stable Input Switch Behavior",
```
blhoward2: S should be capitilized. Standard rules for Title Case with hyphenated words.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -238,244 +262,139 @@
...
 				{
-					"label": "Control mode selected automatically",
+					"label": "Select from auto-calibration",
 					"value": 2
 				}
 			]
 		},
 		"31": {
-			"label": "Load recognized during auto-calibration",
+			"label": "Dimmer Mode Recognized During Auto-calibration",
```
blhoward2: Also capitilize C. Please do any other hyphenated words I missed.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -606,32 +524,43 @@
...
+				},
+				{
+					"label": "Always enable",
+					"value": 255
+				}
+			]
 		},
 		"39": {
-			"label": "Power limit - OVERLOAD",
+			"label": "Power Limit - OVERLOAD",
```
blhoward2: Nothing should be in all caps. For this specific one usually we just call the parameter Overcurrent Protection, or just Power Limit.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -787,11 +713,11 @@
 			]
 		},
 		"47": {
-			"label": "OVERCURRENT alarm report",
-			"description": "(short circuit, burnt out bulb causing overcurrent)",
+			"label": "OVERCURRENT Alarm Report",
```
blhoward2: Again, lose the all caps.

blhoward2 (changes requested):
Some stylistic changes.

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: Sure. Is the order specified anywhere in the style guide? Is it required properties first and then optional?

EDIT: Probably not, since then description wouldn't be after label...

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: Fixed in all places

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: Removed "overload" from here and parameter 45.

Overcurrent I think is something different since the dimmer have 2 separate alarms for overload and that (parameter 45 and 47)

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: Done in all places, I thought that OVERCURRENT etc might be some Z-Wave term that was supposed to be written like this.

packages/config/config/devices/0x010f/fgd212_3.5.json
AlCalzone: @blhoward2 I don't think we have this requirement written in the style guide yet. It is only there for the top level properties:
https://zwave-js.github.io/node-zwave-js/#/config-files/file-format?id=configuration-file-format

glyph-se:
@blhoward2 can you please take another look? There are still two open conversations but they are more like questions for the future.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -238,244 +261,139 @@
...
+			"$import": "../templates/master_template.json#base_enable_disable",
+			"label": "Send secure Z-wave commands to 2nd association group"
+		},
+		"27[0x02]": {
+			"$import": "../templates/master_template.json#base_enable_disable",
+			"label": "Send secure Z-wave commands to 3rd association group"
+		},
+		"27[0x04]": {
+			"$import": "../templates/master_template.json#base_enable_disable",
+			"label": "Send secure Z-wave commands to 4th association group"
```
blhoward2: Capitilization is off. The labels should be Title Case. Please carry throughout.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -29,66 +29,67 @@
...
 			"valueSize": 1,
-			"minValue": 1,
+			"minValue": 2,
 			"maxValue": 99,
 			"defaultValue": 99
 		},
 		"3": {
-			"label": "Incandescence level of dimmable CFL*",
+			"label": "Incandescence Level of Dimmable CFL*",
 			"description": "Virtual value (percentage between 1% and 99%)",
```
blhoward2: Take out the percentage description. @AlCalzone maintains it can't be a percentage if the range is 1-99 and it is his show. Just call it levels. I think you can lose this whole description, though.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -29,66 +29,67 @@
...
 			"minValue": 0,
 			"maxValue": 255,
 			"defaultValue": 1
 		},
 		"7": {
-			"label": "Dimming Step at Manual Control",
-			"description": "The parameter defines the percentage of a dimming step at manual control. 1-99",
+			"label": "Manual Control: Dimming Step Size",
 			"valueSize": 1,
+			"unit": "%",
```
blhoward2: See above. It can't be a percentage if its 1-99.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -171,41 +176,59 @@
...
+			"label": "Forced Switch On Brightness",
+			"description": "Scale is relative to the range set by minimum and maximum brightness level configuration",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 99,
-			"defaultValue": 0
+			"defaultValue": 0,
+			"options": [
+				{
+					"label": "Disable, previous value is used",
```
blhoward2: Just say "Previous level"

blhoward2 (dismissed):
Minor nits, then it LGTM

packages/config/config/devices/0x010f/fgd212_3.5.json
AlCalzone: Would make sense to mention that this is a value relative to the range defined by parameters 1/2.
1 = minimum brightness (param 1), 99 = maximum brightness (param 2)

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: Correct, these used to be option labels.

packages/config/config/devices/0x010f/fgd212_3.5.json
glyph-se: While I fully agree with your point that percentages should go from 0 to 100 and not from 1 to 99 or something else. I don't understand why it's not allowed to call the unit for a parameter as being a percentage point when setting it to 50 means that it is 50% of the maximum value. Just because values 0 and 100 is not allowed for the parameter I don't believe it is wrong to call it a percentage.

Anyway, I changed them all to this:

`Value relative to the range defined by minimum and maximum brightness level configuration`

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -212,7 +212,7 @@
 		},
 		"19": {
 			"label": "Forced Switch On Brightness",
-			"description": "Scale is relative to the range set by minimum and maximum brightness level configuration",
+			"description": "Scale is the range defined by minimum and maximum brightness level configuration",
```
AlCalzone: I like a mix of the two variants - what do you think?
```suggestion
			"description": "Relative to the range defined by minimum and maximum brightness level configuration",
```
(same for the others if we want to do that).

glyph-se: You commented on an old revision. The current wording is

`Value relative to the range defined by minimum and maximum brightness level configuration`

Your version only removes the "Value" part. I guess that is redundant though. I can remove that.

AlCalzone (approved):
LGTM

PR #2855, opened 2021-06-14
docs: fix links to config-files documents

glyph-se:
While working on https://github.com/zwave-js/node-zwave-js/pull/2854 I checked the CONTRIBUTING.md and saw some invalid links, just helping out and fixing them :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,7 +15,7 @@ Open an issue with the **feature request template** and fill it out. For complic
 
 ## Do you want to contribute device configuration files?
 
-Thanks, just go ahead and open a PR! Please make sure to follow about our [standards and contribution guidelines for configuration files](https://zwave-js.github.io/node-zwave-js/#/config-files/contributing-files).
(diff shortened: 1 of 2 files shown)


AlCalzone:
Thanks - do you think it would make more sense to link to the documentation instead of its source files?
=> https://zwave-js.github.io/node-zwave-js/#/config-files/overview etc.

glyph-se:
> Thanks - do you think it would make more sense to link to the documentation instead of its source files?
> => https://zwave-js.github.io/node-zwave-js/#/config-files/overview etc.

Yes, I think this makes. The online documentation is easier to with the left menu and there exists links directly to GitHub for editing / updating it.

I changed the PR slightly, now pointing just to the general instructions for device configuration files, less duplication of content.

I also moved the style guide description further to the top on that page as I believe that to be more important and general than the others. To me partial parameters and conditional parameters seems more like edge cases that are not always relevant, while the style guide always is important.

What do you think?

AlCalzone (approved):

AlCalzone:
Sounds good!

PR #2874, opened 2021-06-17
fix: detect of serial disconnection, destroy driver on serial/socket failure

### LINKED ISSUE #2873: [bug] Detect serial disconnection
### Version

zwavejs2mqtt: 5.0.4.ebfd0b3
zwave-js: 7.7.5
home id: 4266870589
home hex: 0xfe53473d

**Checklist:**

Build/Run method

- [ ] Docker
- [ ] PKG
- [ ] Snap package
- [x] Manually built (git clone - yarn install - yarn run build )

zwavejs2mqtt version: 5.0.4.ebfd0b3
zwave-js version: 7.7.5

### Describe the bug

the health is still OK event when the adapter is disconnected (not in the /dev/ttyACM... )

### To Reproduce

1. Check the health without service

```
curl localhost:8091/health -H "Accept: text/plain" -v
*   Trying 127.0.0.1...
* TCP_NODELAY set
* connect to 127.0.0.1 port 8091 failed: Connection refused
* Failed to connect to localhost port 8091: Connection refused
* Closing connection 0
curl: (7) Failed to connect to localhost port 8091: Connection refused
```
OK

2. Start service without adapter connected
```
yarn start
``` 

OK 

3. Check the health
```
curl localhost:8091/health -H "Accept: text/plain"
Error
```

OK

4. Connect the adapter wait a minute and check health

```
curl localhost:8091/health -H "Accept: text/plain"
Ok
```

OK

5. Remove the adapter wait x minutes and check health

```
curl localhost:8091/health -H "Accept: text/plain"
Ok
```

NOK

### Expected behavior

After removing adapter the health should be Error 

### Additional context

I'm trying to add to my app some info about problem with zWave adapter connection. 
In zigbee2mqtt it can be easy done via mqtt -> ``state_topic: "zigbee2mqtt/bridge/state"``

image

Is there any way (via mqtt or http) to have the real status about zwave?

 :pray:  Thanks 

BTW
this is the greatest web app I ever saw :+1:

robertsLando:
Fixes #2873

### DIFF
--- a/packages/serial/src/ZWaveSerialPort.ts
+++ b/packages/serial/src/ZWaveSerialPort.ts
@@ -1,6 +1,10 @@
-import type { ZWaveLogContainer } from "@zwave-js/core";
+import { ZWaveError, ZWaveErrorCodes, ZWaveLogContainer } from "@zwave-js/core";
 import type SerialPort from "serialport";
 import { ZWaveSerialPortBase } from "./ZWaveSerialPortBase";
+
+interface DisconnectError extends Error {
+	disconnected: boolean;
+}
 
 /** The default version of the Z-Wave serial binding that works using node-serialport */
 export class ZWaveSerialPort extends ZWaveSerialPortBase {
@@ -27,6 +31,18 @@ export class ZWaveSerialPort extends ZWaveSerialPortBase {
 					}),
 				open: (serial: SerialPort) =>
 					new Promise((resolve) => {
+						// detect serial disconnection errors
+						serial.on("close", (err: DisconnectError) => {
+							if (err?.disconnected === true) {
+								this.emit(
+									"error",
+									new ZWaveError(
+										`The socket closed unexpectedly!`,
+										ZWaveErrorCodes.Driver_Failed,
+									),
+								);
+							}
+						});
 						serial.once("open", resolve).open();
 					}),
 				close: (serial: SerialPort) =>
--- a/packages/serial/src/ZWaveSerialPort.ts
+++ b/packages/serial/src/ZWaveSerialPort.ts
@@ -32,12 +32,12 @@ export class ZWaveSerialPort extends ZWaveSerialPortBase {
 				open: (serial: SerialPort) =>
 					new Promise((resolve) => {
 						// detect serial disconnection errors
-						serial.on("close", (err: DisconnectError) => {
+						serial.on("close", (err?: DisconnectError) => {
 							if (err?.disconnected === true) {
 								this.emit(
 									"error",
 									new ZWaveError(
-										`The socket closed unexpectedly!`,
+										`The serial port closed unexpectedly!`,
 										ZWaveErrorCodes.Driver_Failed,
 									),
 								);
--- a/packages/serial/src/ZWaveSocket.ts
+++ b/packages/serial/src/ZWaveSocket.ts
@@ -6,6 +6,10 @@ export type ZWaveSocketOptions =
 	| Omit<net.TcpSocketConnectOpts, "onread">
 	| Omit<net.IpcSocketConnectOpts, "onread">;
 
+export interface DisconnectError extends Error {
+	disconnected: boolean;
+}
+
 /** A version of the Z-Wave serial binding that works using a socket (TCP or IPC) */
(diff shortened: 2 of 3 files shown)


packages/serial/src/ZWaveSocket.ts
```diff
@@ -17,8 +21,8 @@ export class ZWaveSocket extends ZWaveSerialPortBase {
 				create: () => new net.Socket(),
 				open: (serial: net.Socket) =>
 					new Promise((resolve) => {
-						serial.on("close", () => {
-							if (this.isOpen) {
+						serial.on("close", (err: DisconnectError) => {
+							if (err?.disconnected === true) {
```
AlCalzone, line 25: Are you sure about this? The type declarations claim that the parameter for the close event is a boolean. And the serialport code throws an error, so I would expect this to come through the `"error"` event.

packages/serial/src/ZWaveSocket.ts
```diff
@@ -6,6 +6,10 @@ export type ZWaveSocketOptions =
 	| Omit<net.TcpSocketConnectOpts, "onread">
 	| Omit<net.IpcSocketConnectOpts, "onread">;
 
+export interface DisconnectError extends Error {
```
AlCalzone:
```suggestion
interface DisconnectError extends Error {
```
No need to export.

AlCalzone (changes requested):

packages/serial/src/ZWaveSocket.ts
AlCalzone: Ok, seems like the type declarations are incorrect:
https://serialport.io/docs/api-stream#close

packages/serial/src/ZWaveSocket.ts
```diff
@@ -6,6 +6,10 @@ export type ZWaveSocketOptions =
...
+	disconnected: boolean;
+}
+
 /** A version of the Z-Wave serial binding that works using a socket (TCP or IPC) */
 export class ZWaveSocket extends ZWaveSerialPortBase {
```
AlCalzone: That's the wrong file! This is the
> serial binding that works using a socket (TCP or IPC)

You wanted to edit `ZWaveSerialPort.ts`, which uses the serialport binding.

robertsLando: I think it could be applied in both but ok I will move it to the other file

packages/serial/src/ZWaveSocket.ts
```diff
@@ -18,7 +18,7 @@ export class ZWaveSocket extends ZWaveSerialPortBase {
 				open: (serial: net.Socket) =>
 					new Promise((resolve) => {
 						serial.on("close", () => {
-							if (this.isOpen) {
+							if (!this.isOpen) {
```
AlCalzone: Uhm, why? If the socket **is** open and suddenly out of nowhere the `"close"` event is emitted - that's unexpected IMO.

packages/serial/src/ZWaveSerialPort.ts
```diff
@@ -27,6 +31,18 @@ export class ZWaveSerialPort extends ZWaveSerialPortBase {
 					}),
 				open: (serial: SerialPort) =>
 					new Promise((resolve) => {
+						// detect serial disconnection errors
+						serial.on("close", (err?: DisconnectError) => {
+							if (err?.disconnected === true) {
+								this.emit(
```
AlCalzone, line 37: Did you test this? Does it work?

AlCalzone (changes requested):

packages/serial/src/ZWaveSocket.ts
robertsLando: Revert error, sorry

packages/serial/src/ZWaveSerialPort.ts
robertsLando: Not tested yet

AlCalzone:
Works like a charm. I've changed it so the driver destroys itself when the socket or serial port fails during usage. Previously it only destroyed itself when it failed during opening.

AlCalzone (approved):

PR #2888, opened 2021-06-20
feat(config): add alarm mapping for First Alert ZCOMBO

Laikulo:
Marking this as draft for two reasons:
I'd like to add whatever the rest of the values for `AlarmType`, if I can find documentation on what they are. (types 3-11 inclusive)
It's late and I don't trust myself.

Random Notes:
The device shows alarm 13 at 255 on reboot It does not appear to set alarmtype to 13 at any time after reboot
The devices only has one test state, so I've assigned it to the smoke alarm, since I don't think I can map it to both.

### DIFF
--- a/packages/config/config/devices/0x0138/zcombo.json
+++ b/packages/config/config/devices/0x0138/zcombo.json
@@ -45,5 +45,79 @@
 				}
 			]
 		}
+	},
+	"compat":{
+		"alarmMapping":[
+			{
+				"from":{
+					"alarmType": 1,
+					"alarmLevel": 255
+				},
+				"to":{
+					"notificationType": 1,
+					"notificationEvent": 2
+				}
+			},
+			{
+				"from":{
+					"alarmType": 1,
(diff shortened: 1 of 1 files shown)


Laikulo:
@zwave-js-bot help

Laikulo:
@zwave-js-bot fix lint

AlCalzone (approved):

kpine:
It looks like these mappings don't cause values to be created until the V1 alarm is actually reported by the node. It's a little bit of a challenge to generate real smoke and CO alarms. Is it possible to create the values from the mappings themselves ahead of time?

AlCalzone:
Should be possible

Laikulo:
These particular devices appear to send 13 255 on startup, and running the standard test sends 12 255, followed by 12 0 once the test is complete.
I assume if the test fails, it probably sends something else.

Worst case, I'll order a can of test smoke, and test them that way.

However, it's probably worth covering devices that don't squawk on boot.

PR #2891, opened 2021-06-21
feat: add options to configure preferred sensor scales

### LINKED ISSUE #2882: Add driver option to select preferred temperature scale
https://github.com/home-assistant/core/issues/45967

AlCalzone:
This PR adds a new option to configure the preferred sensor scales to be used in Multilevel Sensor queries if no sensor scale was provided manually. By default, this is passed as a new driver option `preferences.scales`. From an application perspective it might be preferable to dynamically detect which scales are available, so this is also configurable using the new `driver.setPreferredScales` method.

The option accepts an object where the key is either:
- the name of a named scale group, e.g. "temperature", which applies to every sensor type that uses this scale group.
- or the numeric sensor type to specify the scale for a single sensor type

The value must match what is defined in the sensor type config file and contain either:
- the label (e.g. "Celsius", "Fahrenheit")
- the unit (e.g. "°C", "°F")
- or the numeric key of the scale (e.g. 0 or 1).

Single-type preferences have a higher priority than named ones. For example, the following preference
```js
{
    temperature: "°F",
    0x01: "°C", // Can also use decimal numbers as a key
}
```
will result in using the Fahrenheit scale for all temperature sensors, except the air temperature (`0x01`).

---

The `get` method for the `Multilevel Sensor CC API` now has an additional overload that just takes the sensor type and automatically determines the scale to use.

---

To determine the configured sensor types, the driver's `ConfigurationManager` instance (`driver.configManager`) can be used. It provides the properties `namedScales` for the named scale definitions and `sensorTypes` for the sensor types and their scale definitions.

---

fixes: #2882
fixes: https://github.com/home-assistant/core/issues/45967

### DIFF
--- a/docs/api/CCs/MultilevelSensor.md
+++ b/docs/api/CCs/MultilevelSensor.md
@@ -13,6 +13,10 @@ async get(): Promise<
 
 async get(
 	sensorType: number,
+): Promise<MultilevelSensorValue | undefined>;
+
+async get(
+	sensorType: number,
 	scale: number,
 ): Promise<number | undefined>;
 ```
--- a/docs/api/config-manager.md
+++ b/docs/api/config-manager.md
@@ -371,3 +371,52 @@ getNotificationName(notificationType: number): string;
 Returns the defined label for a given notification type
 
 > [!NOTE] `loadNotifications` must be used first.
+
+## ConfigManager properties
+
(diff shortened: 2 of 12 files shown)


AlCalzone:
@zwave-js-bot fix lint

PR #2898, opened 2021-06-22
fix(config): add support for clearing inovelli notifications

Laikulo:
0 is actually a valid value for inovelli notification timeouts, it was silently allowed on all but the lzw36.
Setting 0 clears a notification if one is present.

### DIFF
--- a/packages/config/config/devices/0x031e/lzw30-sn.json
+++ b/packages/config/config/devices/0x031e/lzw30-sn.json
@@ -441,7 +441,7 @@
 		},
 		"8[0xff0000]": {
 			"label": "LED Effect Duration",
-			"description": "1 to 60 = seconds, 61 to 120 = minutes (minus 60), 121 - 254 = hours (minus 120), 255 = indefinitely",
+			"description": "0 = disabled, 1 to 60 = seconds, 61 to 120 = minutes (minus 60), 121 - 254 = hours (minus 120), 255 = indefinitely",
 			"valueSize": 4,
 			"minValue": 0,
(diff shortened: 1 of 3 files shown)


Laikulo:
Workaround: Set a notification with all fields zero except time, with time as one.
This has the side effect of turning off the light bar for one second when clearing a notification.

AlCalzone:
Seems reasonable, but I'm confused by the wording "notification". What does that have to do with LED effect configuration?

Laikulo:
Inovelli uses the term "Notification" to refer to the Effect/Color/Timeout/Brightness config paramaters. They are meant to send notifications to the end user, by temporarily overriding the color/brightness of one of the LED strips on the switch for a given amount of time. They can also be dismissed manually with a button combo on the device.

For example, you could have your controller send a config parameter update when a mailbox sensor is tripped, or whenever a given door is opened (and another update to clear when it is closed).

Right now, I am using this as a "Pilot" light for a light that can't be seen from its switch. My controller sends Pulse/Yellow/Indefinite Time/Full brightness whenever the lamp is turned on, and Solid/Red/1 Second/0 brightness whenever the lamp is turned off.

Laikulo:
An example of this behavior: https://www.youtube.com/watch?v=fPiDrBpCSlY
(may still be processing)

AlCalzone:
I see - would the parameter still be understandable if we change that to `0 = disabled`? At least it would confuse **me** less, but I'm not familiar with those devices.

AlCalzone (approved):
Thanks!

PR #2906, opened 2021-06-24
feat: add valueChangeOptions as an optional metadata parameter

raman325:
As discussed here: https://github.com/zwave-js/node-zwave-js/issues/1321#issuecomment-760086568

This would make it easier for us to build logic to determine when we can use the `transitionDuration` option

### DIFF
--- a/docs/api/valueid.md
+++ b/docs/api/valueid.md
@@ -51,6 +51,7 @@ interface ValueMetadataAny {
 	description?: string;
 	label?: string;
 	ccSpecific?: Record<string, any>;
+	setValueOptions?: string[];
 }
 ```
 
@@ -60,6 +61,7 @@ interface ValueMetadataAny {
 -   `description`: A description of the value
 -   `label`: A human-readable label for the property
 -   `ccSpecific`: CC specific information to help identify this value [(see below)](#CC-specific-fields)
+-   `setValueOptions`: Parameters that can be passed as `options` to the [`setValue`](api/node.md#setvalue) command for this value.
 
 ### Value types
 
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -42,6 +42,9 @@
     "semver": "^7.3.2",
     "winston": "^3.3.3"
   },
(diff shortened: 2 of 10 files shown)


raman325:
@zwave-js-bot fix lint

raman325:
@zwave-js-bot fix lint

raman325:
@zwave-js-bot fix lint

AlCalzone:
@raman325 We're going to do this, but this needs some (linguistic) revision.

I want to strongly-type the array options, for that reason the interface needs to be moved into `@zwave-js/core`. We cannot call it `SetValueOptions` there, because that interface already exists for the `ValueDB.setValue` call options. I'm hesitant to call it its current name `SetValueAPIOptions` there, because that API is a concept of the `zwave-js` module, not the `core` module.

I'm open to suggestions, maybe `RelatedValues` or something with `Meta` in the name (although that can be confusing with metadata)? Ideally, the name should make sense for the Metadata property (currently called `setValueOptions`) aswell.

AlCalzone:
I think `valueChangeOptions` could work. Thoughts?

raman325:
Works for me!

AlCalzone (approved):

docs/api/valueid.md
```diff
@@ -60,6 +61,7 @@ interface ValueMetadataAny {
 -   `description`: A description of the value
 -   `label`: A human-readable label for the property
 -   `ccSpecific`: CC specific information to help identify this value [(see below)](#CC-specific-fields)
+-   `setValueOptions`: Parameters that can be passed as `options` to the [`setValue`](api/node.md#setvalue) command for this value.
```
MartinHjelmare, line 64: Is this name correct, `setValueOptions`?

raman325: good catch, way to be on vacation. Will submit a new PR

AlCalzone: Whoops I forgot to change that when I renamed things

PR #2913, opened 2021-06-26
build: migrate to Yarn v3

AlCalzone:
TODOs:

- [x] ~~figure out if we should replace lerna with https://yarnpkg.com/features/release-workflow -> see https://github.com/yarnpkg/berry/issues/1229#issuecomment-784416753~~ (we shouldn't for now)
- [x] ~~blocked by https://github.com/dsherret/ts-morph/issues/1168~~ (found a workaround)


.github/workflows-disabled/approve-workflow-runs.yml
```diff
@@ -16,42 +16,42 @@ jobs:
...
+        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
+
+      - name: Use Yarn cache
+        uses: actions/cache@v2
+        id: yarn-cache
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ matrix.node-version }}-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
```
merceyz:
```suggestion
          key: yarn-${{ hashFiles('**/yarn.lock') }}
          restore-keys: |
            yarn-
```
The cache is identical between OS and node versions so you can share it between all runners

(Applies to all the workflow files but I wont make that many suggestions)

AlCalzone: Does that apply to native (compiled) dependencies as well?

AlCalzone: And dependencies that are not available for Windows for example?

merceyz: You're only caching the cache, it doesn't contain any of the postinstall output, under PnP that's located in `.yarn/unplugged`

.github/workflows/test-and-release.yml
```diff
@@ -318,28 +318,29 @@ jobs:
...
-    - name: Publish package to npm
+    - name: Publish packages to npm
       run: |
-        npm config set //registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}
-        npm whoami
-        npx lerna publish from-package --yes
+        yarn config set npmPublishRegistry 'https://registry.npmjs.org'
+        yarn config set 'npmRegistries["//registry.npmjs.org"].npmAuthToken' "${{ secrets.NPM_TOKEN }}"
+        yarn npm whoami
+        yarn workspaces foreach exec npm publish --tolerate-republish
```
merceyz:
```suggestion
        yarn workspaces foreach npm publish --tolerate-republish
```

Otherwise it will use `npm` to do the publish

AlCalzone: Yeah thanks. Was late last night, I forgot to remove that.

package.json
```diff
@@ -30,98 +31,85 @@
...
-    "postversion": "git push && git push --tags",
+    "release": "yarn run build && yarn release-script",
     "config": "yarn ts packages/config/maintenance/importConfig.ts",
     "docs": "docsify serve docs",
     "docs:generate": "yarn ts packages/maintenance/src/generateTypedDocs.ts",
     "test:perf": "yarn ts test/valuedb-perf.ts",
     "test:firmware": "yarn ts test/firmware-extraction.ts",
-    "ts": "node -r esbuild-register"
+    "//": "remove cross-env when https://github.com/evanw/esbuild/issues/1406 is fixed",
+    "ts": "cross-env ESBUILD_WORKER_THREADS=0 yarn node -r esbuild-register"
```
merceyz: Yarn ships with a portable shell so you don't actually need `cross-env` to do this anymore
https://github.com/yarnpkg/berry/tree/1bea22b3fab51b4c7757616708c511580c007947/packages/yarnpkg-shell

AlCalzone: Good to know, thanks!

Since you're here, there's one more blocker I need to solve before I can actually make the switch.

With `lerna` (fixed mode) it was trivial to bump all changed workspaces to a common version. I can't seem find an easy way to do this with the `version` plugin, except by controlling it with an external script:
https://github.com/yarnpkg/berry/discussions/3067

PR #2925, opened 2021-06-30
feat: include Node.js version, OS and arch in telemetry

AlCalzone:


robertsLando:
@AlCalzone Would suggest to add also arch. Remember to also update validation rules on statistics server :)

AlCalzone:
done and done

Issue #2286, opened 2021-04-09
📣 Meta-Issue: Missing reports from devices 📣


AlCalzone:
Since version 7.1.0, more and more issues are popping up regarding missing reports from devices. It is on our radar and we might finally have a solution for once and all. **Please don't open any new issues regarding this**, as they only cause noise.

---

### A little background

So called unsolicited updates (that weren't requested by `zwave-js`) are sent by devices via associations - usually the "Lifeline". The Z-Wave specifications do a terrible job at explaining how these should be configured and used, so as a result, most manufacturers got it wrong. Missing updates are usually caused by incorrect associations or devices that send incorrect or ambiguous reports or simply devices that are unnecessarily using endpoints.

There are (what feels like) 20 different ways to configure these associations to the controller, all of them very dependent on the actual device that should be configured. But only a handful of variations can be automatically decided on, so it is more likely that we get it wrong than right.

When a device sent us a report via the root device, we used to guess which endpoint that report was supposed to come from. No matter how we changed the guessing strategy, some use cases were broken. Our plan going forward is to detect when endpoints are unnecessary and should be ignored - as opposed to ignoring the root device when there are any endpoints like we do now. 

This should massively reduce our chances to get the associations wrong, remove the need to make these guesses and make the remaining edge cases easier to handle.

---

##### If all endpoints have different device classes
* ignore the endpoints and work only with the root device
* do not set up multi channel associations, set up a node lifeline association on the root device

This needs a compat flag to opt out and restore the current behavior:
```json
"preserveEndpoints": "*", // to preserve all endpoints
"preserveEndpoints": [2, 3], // to preserve endpoints 2 and 3, but ignore endpoint 1. There are some thermostats which expose an external sensor with battery through a seemingly unnecessary endpoint, but it needs to be preserved
```

##### If some endpoints share a device class
* ignore the root device's application CCs (current behavior), opt out possible with `preserveRootApplicationCCValueIDs`
* Try to set up normal lifeline associations on each endpoint separately, fall back to multi channel associations on each endpoint, then fall back to the current behavior of a multi channel association to endpoint 0 on the root device

The lifeline association strategy should be controllable with the `multiChannel` flag in the association config. New default: `"auto"`, `true` forces a multi channel association (if possible), `false` forces a node association (if possible)

This needs a compat flag to define which endpoint the root reports should be mapped to (for devices like the FGS-222, which work half)
```json
"mapRootReportsToEndpoint": 1
```

---

[...]

PR #2932, opened 2021-07-01
fix(cc): do not query all user codes during interview

### LINKED ISSUE #2725: Yale lock code query disables device, and kills the battery due to looping NodeInfo
**Checklist:**

-   [x] My problem is **not** within Home Assistant or the ZWave JS integration. **Or:** a Home Assistant developer has told me to come here.
-   [x] My problem is **not** within ZWaveJS2MQTT. **Or:** a ZWaveJS2MQTT developer has told me to come here.
-   [x] I have checked the troubleshooting section and my problem is **not** described there.
-   [x] I have read the changelog and my problem was **not** mentioned there.

**Describe the bug**
I use ZWaveJS2MQTT but the zwave logic for this I believe comes from  ZWave JS directly hence filing here.

Yale lock node user code update / interview is strobing the device batteries to death.

I have 5 Yale Assure Touchscreen Deadbolts on my network and due to a bug in ZWaveJS2MQTT (possibly https://github.com/zwave-js/zwavejs2mqtt/issues/998) after restarting, sometimes they have to be re-interviewed to work again (but just adding them the first time was the same hassle) as they all go to "Unknown manufacturer 0xXXXX".

It seems that the NodeInfo loop gets all the relevant details from the lock, then proceeds to query all user codes, I believe there are 254 (?) or so, but it gets to 15-20 when the node doesn't respond quickly enough, and it gives up on the interview, then restarts it several times, even eventually declaring the node dead, or getting stuck in an interview loop until it depletes all the batteries. While this is going on the lock is unresponsive to commands.

This makes the locks effectively useless as I can't control them and can't even use them as "dumb" locks (keypad or auto relock) as the interview eats the 4 AA batteries in a day or two.

Could the code query be deferred so the device can finish NodeInfo without it, and then in a separate, more lenient loop fetch them without maybe restarting from the start every single time?

**Device information**

Which device(s) is/are affected (manufacturer/model)? 
They report in as: Yale | Assure Touchscreen Deadbolt | YRD226,  YRD246 and YRL256

What is/are the node IDs?
Node 24
Node 37
Node 43
Node 45
Node 50

**Did you change anything?**

-   [ ] **Yes**: _(please describe)_
-   [x] **No**

-   [ ] **Don't know**, this is a new device
-   [x] **No**, it never worked anywhere
-   [ ] **Yes**, in: _(specify application with versions)_

-   [ ] `zwavejs2mqtt` (latest) docker image
-   [ ] `zwavejs2mqtt` (dev) docker image
-   [ ] `zwavejs2mqtt` Manual Docker build
    -   `node-zwave-js` branch: 
    -   `zwavejs2mqtt` branch: 
-   [ ] ioBroker.zwave2 adapter
-   [x] `HomeAssistant` version was zwavejs2mqtt 0.17.0 now 0.18.0 (zwavejs2mqtt: 4.5.0 zwave-js: 7.7.0)
-   [ ] Pkg
-   [ ] Manual Docker build
    -   `node-zwave-js` branch: 
    -   `zwavejs2mqtt` branch: 
-   [ ] `node-red-contrib-zwave-js`
    -   Module version (double click node): 
-   [ ] Manually built (as described in the [docs](https://zwave-js.github.io/node-zwave-js/#/development/installing-from-github))
-   [ ] Other: 

**To Reproduce**

Try to add a Yale lock, or re-interview one.
[zwavejs2mqtt.log](https://github.com/zwave-js/node-zwave-js/files/6563288/zwavejs2mqtt.log)

### LINKED ISSUE #2727: Do not query all user codes during interview

sybohy:
This PR is a proposed solution for #2727. More specifically, we expose the ability to skip the interviewing of user code by setting a `skipUserCodeInterview` flag in the ZWaveOptions being passed to the driver.

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
@@ -756,7 +756,9 @@ export class UserCodeCC extends CommandClass {
 		}
 
 		// Synchronize user codes and settings
-		await this.refreshValues();
+		if (!this.driver.options.skipUserCodeInterview) {
+			await this.refreshValues();
+		}
 
 		// Remember that the interview is complete
 		this.interviewComplete = true;
--- a/packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
@@ -770,7 +770,7 @@ export class UserCodeCC extends CommandClass {
 		}
 
 		// Synchronize user codes and settings
-		if (this.driver.options.interview?.queryAllUserCodes) {
+		if (this.driver.options.interview.queryAllUserCodes) {
 			await this.refreshValues();
 		}
 
--- a/packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
@@ -236,6 +236,48 @@ function validateCode(code: string, supportedChars: string): boolean {
 	return [...code].every((char) => supportedChars.includes(char));
 }
 
+function setUserCodeMetadata(this: UserCodeCC, userId: number) {
+	const valueDB = this.getValueDB();
+	const statusValueId = getUserIdStatusValueID(this.endpointIndex, userId);
+	const codeValueId = getUserCodeValueID(this.endpointIndex, userId);
+	const supportedUserIDStatuses =
+		valueDB.getValue<UserIDStatus[]>(
+			getSupportedUserIDStatusesValueID(this.endpointIndex),
+		) ??
+		(this.version === 1
+			? [
+					UserIDStatus.Available,
+					UserIDStatus.Enabled,
+					UserIDStatus.Disabled,
+			  ]
+			: [
+					UserIDStatus.Available,
+					UserIDStatus.Enabled,
+					UserIDStatus.Disabled,
+					UserIDStatus.Messaging,
+					UserIDStatus.PassageMode,
+			  ]);
+	if (!valueDB.hasMetadata(statusValueId)) {
+		valueDB.setMetadata(statusValueId, {
+			...ValueMetadata.Number,
+			label: `User ID status (${userId})`,
+			states: enumValuesToMetadataStates(
+				UserIDStatus,
+				supportedUserIDStatuses,
+			),
+		});
+	}
+	const codeMetadata: ValueMetadata = {
+		...ValueMetadata.String,
+		minLength: 4,
+		maxLength: 10,
+		label: `User Code (${userId})`,
+	};
+	if (valueDB.getMetadata(codeValueId)?.type !== codeMetadata.type) {
+		valueDB.setMetadata(codeValueId, codeMetadata);
+	}
+}
+
 function persistUserCode(
 	this: UserCodeCC,
 	userId: number,
@@ -271,29 +313,8 @@ function persistUserCode(
 		valueDB.setMetadata(statusValueId, undefined);
 		valueDB.setMetadata(codeValueId, undefined);
 	} else {
-		// Always create metadata if it does not exist
-		if (!valueDB.hasMetadata(statusValueId)) {
-			valueDB.setMetadata(statusValueId, {
-				...ValueMetadata.Number,
-				label: `User ID status (${userId})`,
-				states: enumValuesToMetadataStates(
-					UserIDStatus,
-					supportedUserIDStatuses,
-				),
-			});
-		}
-		const codeMetadata: ValueMetadata = {
-			...(typeof userCode === "string"
-				? ValueMetadata.String
-				: ValueMetadata.Buffer),
-			minLength: 4,
-			maxLength: 10,
-			label: `User Code (${userId})`,
-		};
-		if (valueDB.getMetadata(codeValueId)?.type !== codeMetadata.type) {
-			valueDB.setMetadata(codeValueId, codeMetadata);
-		}
-
+		// Always create metadata in case it does not exist
+		setUserCodeMetadata.call(this, userId);
 		valueDB.setValue(statusValueId, userIdStatus);
 		valueDB.setValue(codeValueId, userCode);
 	}
@@ -755,8 +776,14 @@ export class UserCodeCC extends CommandClass {
 			return;
 		}
 
+		let nextUserId = 1;
+		while (nextUserId > 0 && nextUserId <= supportedUsers) {
+			setUserCodeMetadata.call(this, nextUserId);
+			nextUserId += 1;
+		}
+
 		// Synchronize user codes and settings
-		if (!this.driver.options.skipUserCodeInterview) {
+		if (this.driver.options.queryAllUserCodes) {
 			await this.refreshValues();
 		}
 
--- a/packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
@@ -272,7 +272,7 @@ function setUserCodeMetadata(
 		});
 	}
 	const codeMetadata: ValueMetadata = {
-		...(typeof userCode === "object"
+		...(Buffer.isBuffer(userCode)
 			? ValueMetadata.Buffer
 			: ValueMetadata.String),
 		minLength: 4,
--- a/packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
@@ -236,7 +236,11 @@ function validateCode(code: string, supportedChars: string): boolean {
 	return [...code].every((char) => supportedChars.includes(char));
 }
 
-function setUserCodeMetadata(this: UserCodeCC, userId: number) {
+function setUserCodeMetadata(
+	this: UserCodeCC,
+	userId: number,
+	userCode?: string | Buffer,
+) {
 	const valueDB = this.getValueDB();
 	const statusValueId = getUserIdStatusValueID(this.endpointIndex, userId);
 	const codeValueId = getUserCodeValueID(this.endpointIndex, userId);
@@ -268,7 +272,9 @@ function setUserCodeMetadata(this: UserCodeCC, userId: number) {
 		});
 	}
 	const codeMetadata: ValueMetadata = {
-		...ValueMetadata.String,
+		...(typeof userCode === "object"
+			? ValueMetadata.Buffer
+			: ValueMetadata.String),
 		minLength: 4,
 		maxLength: 10,
 		label: `User Code (${userId})`,
@@ -287,23 +293,6 @@ function persistUserCode(
 	const statusValueId = getUserIdStatusValueID(this.endpointIndex, userId);
 	const codeValueId = getUserCodeValueID(this.endpointIndex, userId);
 	const valueDB = this.getValueDB();
-	const supportedUserIDStatuses =
-		valueDB.getValue<UserIDStatus[]>(
-			getSupportedUserIDStatusesValueID(this.endpointIndex),
-		) ??
-		(this.version === 1
-			? [
-					UserIDStatus.Available,
-					UserIDStatus.Enabled,
-					UserIDStatus.Disabled,
-			  ]
-			: [
-					UserIDStatus.Available,
-					UserIDStatus.Enabled,
-					UserIDStatus.Disabled,
-					UserIDStatus.Messaging,
-					UserIDStatus.PassageMode,
-			  ]);
 
 	// Check if this code is supported
 	if (userIdStatus === UserIDStatus.StatusNotAvailable) {
@@ -314,7 +303,7 @@ function persistUserCode(
 		valueDB.setMetadata(codeValueId, undefined);
 	} else {
 		// Always create metadata in case it does not exist
-		setUserCodeMetadata.call(this, userId);
+		setUserCodeMetadata.call(this, userId, userCode);
 		valueDB.setValue(statusValueId, userIdStatus);
 		valueDB.setValue(codeValueId, userCode);
 	}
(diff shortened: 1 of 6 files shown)


sybohy:
@AlCalzone I have a couple of questions here.

First, do we want the default behavior to be NOT querying all the user codes? If so, I need to fix this because right now the default behavior is still to scan them.

Second, if we merge this as a 8.0 release and are ok with breaking changes, then I would suggest restructuring the zwave options from this:

```typescript
...
/**
* @internal
* Set this to true to skip the controller interview. Useful for testing purposes
*/
skipInterview?: boolean;

/** Allows you to skip user code interview */
skipUserCodeInterview?: boolean;
...
```


to maybe this?

```typescript
interviewBehavior: {
  skipInterview?: boolean;
  skipUserCodeInterview?: boolean;
}
```

packages/zwave-js/src/lib/driver/ZWaveOptions.ts
```diff
@@ -59,6 +59,9 @@ export interface ZWaveOptions {
 	 */
 	skipInterview?: boolean;
 
+	/** Allows you to skip user code interview */
+	skipUserCodeInterview?: boolean;
```
sybohy: if we want to flip the default behavior to skip by default, we can maybe rename this `doNotSkipUserCodeInterview` or `interviewUserCodes`.

AlCalzone: I'd make **not scanning** user codes the default for 8.0.0 and call the flag to restore the previous behavior `queryAllUserCodes`.
Moving `skipInterview` there is okay, we just need to update tests that use it.

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
```diff
@@ -756,7 +756,9 @@ export class UserCodeCC extends CommandClass {
 		}
 
 		// Synchronize user codes and settings
-		await this.refreshValues();
+		if (!this.driver.options.skipUserCodeInterview) {
```
AlCalzone: When we end up not querying the codes, we still need to create default metadata for each possible user code. Applications like zwavejs2mqtt and HomeAssistant operate mostly on values and need the metadata to discover the user codes.

Currently, this happens here https://github.com/zwave-js/node-zwave-js/blob/f5a353baa13334ea0c20f3ef3368710b9415e0e0/packages/zwave-js/src/lib/commandclass/UserCodeCC.ts#L274-L284, which is called when decoding `[Extended]UserCodeReports`.

We don't know for sure if the user codes are strings or Buffers, but I think it is safe to default to `string` before knowing what the user codes are like.

AlCalzone (changes requested):

AlCalzone:
Well, here's a test build for you to play with while I'm asleep:
@zwave-js-bot pack this

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
sybohy: hm interesting. i see what the problem is. Let me take a crack at writing a function that sets the metadata for each code even if we haven't queried the actual value from the device. I'm guessing this function will need to be used in a couple of spots (`persistUserCode` and `interview`).

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
```diff
@@ -236,6 +236,48 @@ function validateCode(code: string, supportedChars: string): boolean {
 	return [...code].every((char) => supportedChars.includes(char));
 }
 
+function setUserCodeMetadata(this: UserCodeCC, userId: number) {
```
sybohy: @AlCalzone I've moved the setting of the user code metadata to a separate function because it appears we need to call this functionality in a couple of different places. I've also gone ahead and **assumed** that the default type for a code is a string. Let me know if this is ok.

Actually quick question: do we want to be able to override the metadata type for the user code to Buffer in the event that it is in fact a buffer?

AlCalzone: > Actually quick question: do we want to be able to override the metadata type for the user code to Buffer in the event that it is in fact a buffer?

Yeah, we need to be able to do that, when a device responds with a user code that needs to be represented as a buffer. I'd add an optional `userCode` parameter here and use that to distinguish. The current code will always set the metadata to string.

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
```diff
@@ -236,6 +236,48 @@ function validateCode(code: string, supportedChars: string): boolean {
...
+			...ValueMetadata.Number,
+			label: `User ID status (${userId})`,
+			states: enumValuesToMetadataStates(
+				UserIDStatus,
+				supportedUserIDStatuses,
+			),
+		});
+	}
+	const codeMetadata: ValueMetadata = {
+		...ValueMetadata.String,
```
AlCalzone: This needs to distinguish the type based on the actual user code (and default to string if none was passed)

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
```diff
@@ -236,6 +236,48 @@ function validateCode(code: string, supportedChars: string): boolean {
...
+		minLength: 4,
+		maxLength: 10,
+		label: `User Code (${userId})`,
+	};
+	if (valueDB.getMetadata(codeValueId)?.type !== codeMetadata.type) {
+		valueDB.setMetadata(codeValueId, codeMetadata);
+	}
```
AlCalzone, line 284: Then this takes care of updating the metadata if necessary.

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
```diff
@@ -271,29 +313,8 @@ function persistUserCode(
...
-				: ValueMetadata.Buffer),
-			minLength: 4,
-			maxLength: 10,
-			label: `User Code (${userId})`,
-		};
-		if (valueDB.getMetadata(codeValueId)?.type !== codeMetadata.type) {
-			valueDB.setMetadata(codeValueId, codeMetadata);
-		}
-
+		// Always create metadata in case it does not exist
```
AlCalzone, line 305: Can't comment on the correct line, but `const supportedUserIDStatuses` isn't used in this method any longer and can be removed.

packages/zwave-js/src/lib/driver/ZWaveOptions.ts
```diff
@@ -59,8 +59,8 @@ export interface ZWaveOptions {
 	 */
 	skipInterview?: boolean;
 
-	/** Allows you to skip user code interview */
-	skipUserCodeInterview?: boolean;
+	/** Allows you to query all user code during the interview of the UserCode CC */
+	queryAllUserCodes?: boolean;
```
AlCalzone:
```suggestion
	/**
	 * Whether all user code should be queried during the interview of the UserCode CC.
	 * Note that enabling this can cause a lot of traffic during the interview.
	 */
	queryAllUserCodes?: boolean; // default: false
```

AlCalzone: Also, do we still want to merge the interview settings under a common property?

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
```diff
@@ -755,8 +776,14 @@ export class UserCodeCC extends CommandClass {
 			return;
 		}
 
+		let nextUserId = 1;
+		while (nextUserId > 0 && nextUserId <= supportedUsers) {
+			setUserCodeMetadata.call(this, nextUserId);
+			nextUserId += 1;
+		}
```
AlCalzone:
```suggestion
		for (let userId = 1; userId <= supportedUsers, userId++) {
			setUserCodeMetadata.call(this, userId);
		}
```

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
sybohy: i can't believe i wrote `nextUserId += 1` 🤦🤦🤦

sybohy:
@zwave-js-bot fix lint

AlCalzone:
Damn Github requiring me to approve your workflow runs. About time we get this merged so you're whitelisted.

packages/zwave-js/src/lib/driver/ZWaveOptions.ts
```diff
@@ -53,12 +53,19 @@ export interface ZWaveOptions {
...
+		 * Set this to true to skip the controller interview. Useful for testing purposes
+		 */
+		skipInterview?: boolean;
 
+		/**
+		 * Whether all user code should be queried during the interview of the UserCode CC.
+		 * Note that enabling this can cause a lot of traffic during the interview.
+		 */
+		queryAllUserCodes?: boolean;
+	};
```
sybohy, line 68: this needs to be added to `8.0.0` migration notes as a breaking change. Not sure if there's a process for this?

sybohy:
@AlCalzone actually i'd love another build with these latest changes

packages/zwave-js/src/lib/driver/ZWaveOptions.ts
AlCalzone: Last time I added a migration guide in the docs and extended it in each PR.

AlCalzone:
Doesn't seem to build - I think you forgot to update the default options.

sybohy:
@AlCalzone are you ok with expanding the scope of this PR just a bit to get the build to go through? See this commit: https://github.com/zwave-js/node-zwave-js/pull/2932/commits/32b5f53220f7c9aa9ce3e93c39c129e7ed2342d2

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -753,7 +756,7 @@ export class Driver extends TypedEventEmitter<DriverEventCallbacks> {
 			}
 		};
 
-		if (!this.options.skipInterview) {
+		if (!this.options.interview?.skipInterview) {
```
AlCalzone, line 759: The `?.` is unnecessary, but at least this should build now.

AlCalzone:
Yup, definitely

AlCalzone:
@zwave-js-bot pack this

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
```diff
@@ -262,6 +261,38 @@ function persistUserCode(
...
+			states: enumValuesToMetadataStates(
+				UserIDStatus,
+				supportedUserIDStatuses,
+			),
+		});
+	}
+	const codeMetadata: ValueMetadata = {
+		...(typeof userCode === "object"
+			? ValueMetadata.Buffer
+			: ValueMetadata.String),
```
AlCalzone:
```suggestion
		...(Buffer.isBuffer(userCode)
			? ValueMetadata.Buffer
			: ValueMetadata.String),
```
since I tend to stay away from `typeof userCode === "object"`

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
```diff
@@ -755,8 +765,14 @@ export class UserCodeCC extends CommandClass {
 			return;
 		}
 
+		for (let userId = 1; userId <= supportedUsers; userId++) {
+			setUserCodeMetadata.call(this, userId);
+		}
+
 		// Synchronize user codes and settings
-		await this.refreshValues();
+		if (this.driver.options.interview?.queryAllUserCodes) {
```
AlCalzone:
```suggestion
		if (this.driver.options.interview.queryAllUserCodes) {
```
After applying the default options, this is always defined.

AlCalzone: Same goes for the other locations.

AlCalzone (changes requested):
Just a few nits then I think this is ready if your tests work out.

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
sybohy: oh duh, i just realized you're doing a deep merge in the Driver constructor using the default options. Got it. Let me roll back these

sybohy: fixed, via git commit --amend + force push

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
sybohy: i like this. 👍

packages/zwave-js/src/lib/driver/ZWaveOptions.ts
sybohy: If that's ok with you, I will add a `migrating-to-v8.md` file now that follows roughly the structure of prior file. I will then add this change in zwave options to it.

sybohy: e.g. https://github.com/zwave-js/node-zwave-js/pull/2932/commits/fd4ad42b7a5fcb6ea4446920650c7668b6b6a92c

sybohy:
@zwave-js-bot pack this

sybohy:
@AlCalzone I've been testing this for the past 2 hours and it works like a charm on my end. Assuming there are no other issues, we can probably merge. Curious what the `v8.0.0` release process is like.

AlCalzone:
@zwave-js-bot pack this

packages/zwave-js/src/lib/commandclass/UserCodeCC.ts
AlCalzone: > git commit --amend + force push

You generally don't need to do that here - just makes following the changes in a PR harder. I squash merge every PR for a clean history.

AlCalzone:
> Curious what the v8.0.0 release process is like.

I'm preparing the 8.x version in the `next` branch. I'm going to do a few more breaking changes, including the handling of endpoints as outlined in https://github.com/zwave-js/node-zwave-js/issues/2286 and changes to the build process.

From that branch I'm going to release a few alpha/beta prereleases for applications to test, then at some point make it official.

AlCalzone (approved):
I can't push changes to your branch - you probably disabled the "allow maintainer to make changes" checkbox. I'll just change the nits I have after merging.

sybohy:
my apologies. I'll take a look at the settings in a minute

PR #2965, opened 2021-07-02
fix(config): correct valueSize for param 64 on Aeotec ZW100 / Fantem FT100

crxporter:
Template value size was different from the expected size on the device. Thank's @marcus-j-davies for helping me track this one down.


AlCalzone:
@blhoward2 this was the only location where we used that template and did not specify a value size. Maybe we should remove the value size from that template altogether?

AlCalzone (approved):

PR #2970, opened 2021-07-02
feat: implement invokeCCAPI to generically/dynamically invoke CC API 

### LINKED ISSUE #2799: Expose a single "entry point" to access all CC APIs
like z2m does: https://github.com/zwave-js/zwavejs2mqtt/blob/58c067ebe073e826dd20b079f86d51954d44719b/lib/ZwaveClient.js#L1495

AlCalzone:
With the power of auto generated types and some TypeScript
tenor
we now have a dynamic/generic way to call CC APIs

fixes: #2799

/cc @raman325 
/cc @robertsLando you have that already but it might be worth migrating to this method

---

The method has the following signature where `...args` and the return type depend on the actual method that is used.
```ts
invokeCCAPI(
	cc: CommandClasses,
	method: string, // any of the CC's methods
	...args: unknown[], // that method's arguments
): Promise<unknown> // that method's return type
```

VSCode/TypeScript will help you with that:
grafik

---

#### Example - `UserCodeCC.set`
The parameters can be found at https://zwave-js.github.io/node-zwave-js/#/api/CCs/UserCode?id=set, so this call translates to
```ts
invokeCCAPI(CommandClasses["User Code"], "set",  userId, userIdStatus, userCode)
```
which returns a `Promise<void>`.


raman325:
Might be helpful to include an example of a call using the `invokeCCAPI` method. I know it would help me 🙂 

For the User Code CC `set` function (looking for something with multiple parameters to make sure I get it), would it be `invokeCCAPI(CommandClasses["User Code"], "set", <user_id>, <user_id_status>, <user_code>)`?

AlCalzone:
Updated the description

raman325:
Thanks, that's pretty nifty that the typing works so well on that

marcus-j-davies:
2IC

So! this will allow me to remove a (not so proud of) lookup table to do such thing in the Node red Plugin.
is this due for the next release?

The NR plugin can use Both APIs - so this will shorten my code massively!

AlCalzone:
> is this due for the next release?

This is going into 7.11, yeah

robertsLando:
Awesome! I could replace my sendCommand implementation with this :)

robertsLando:
Only question is, actually in my implementation I also get the enpoint instance, here I don't see any reference to the endpoint, why?

marcus-j-davies:
If I am not mistaken, `getEndpoint(n).invokeCCAPI` will do that, whereas `Controller.nodes.get(n).invokeCCAPI` will result in the root endpoint of node `n` being used?

Thats my undestanding on it.

AlCalzone:
Yeah, that's basically the gist. You can just use getEndpoint which will return the node itself when 0 is passed.

PR #2971, opened 2021-07-02
feat(config): update MP20Z config

emichael:
This adds all of the configuration options described by the manufacturer. It also updates the manual to a more complete one from the manufacturer's website.

If you don't want to have a manual link that's not the z-wave alliance, I can back that line out.

Also, this was done entirely in-browser and without testing. I don't have a decent dev environment setup. Should be straightforward enough, though. I just looked at the manual I linked below and started from the MP21Z paramInformation. The one thing I was uncertain about was why the MP21Z has a maxValue of 3 for param 1. Is the maxValue supposed to be exclusive? If so, all of those should be bumped up by one.

### DIFF
--- a/packages/config/config/devices/0x0312/mp20z.json
+++ b/packages/config/config/devices/0x0312/mp20z.json
@@ -99,27 +99,7 @@
 			]
 		},
 		"6": {
-			"label": "State After Power Failure",
-			"valueSize": 1,
-			"minValue": 0,
-			"maxValue": 2,
-			"defaultValue": 2,
-			"unsigned": true,
-			"allowManualEntry": false,
-			"options": [
-				{
-					"label": "Output Off",
-					"value": 0
-				},
-				{
-					"label": "Output On",
-					"value": 1
-				},
-				{
-					"label": "Restore Outlets to Previous State",
-					"value": 2
-				}
-			]
+			"$import": "templates/master_template.json#state_after_power_failure_off_on_prev"
 		}
 	},
 	"metadata": {
--- a/packages/config/config/devices/0x0312/mp20z.json
+++ b/packages/config/config/devices/0x0312/mp20z.json
@@ -99,7 +99,7 @@
 			]
 		},
 		"6": {
-			"$import": "templates/master_template.json#state_after_power_failure_off_on_prev"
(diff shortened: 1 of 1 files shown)


emichael:
Can't see any more details, so I'm not sure what's causing the "PR title / commit messages" failure. Seems pretty semantic to me!

glyph-se:
> 
> 
> Can't see any more details, so I'm not sure what's causing the "PR title / commit messages" failure. Seems pretty semantic to me!

Try changing the commit message to the same as the title on the PR

AlCalzone:
Yeah that's it. Unfortunately, GitHub has the stupid habit of ignoring the PR title in its merge commit messages if there is only one commit.

packages/config/config/devices/0x0312/mp20z.json
```diff
@@ -14,10 +14,118 @@
...
+					"label": "Output On",
+					"value": 1
+				},
+				{
+					"label": "Restore Outlets to Previous State",
+					"value": 2
+				}
+			]
+		}
+	},
```
AlCalzone, line 104: We have [several templates](https://github.com/zwave-js/node-zwave-js/blob/086ac3e67a36f74e4157c46c3511529d22700520/packages/config/config/devices/templates/master_template.json#L256) for these, please choose the matching one and [use it](https://zwave-js.github.io/node-zwave-js/#/config-files/using-templates?id=using-templates).

AlCalzone (changes requested):

packages/config/config/devices/0x0312/mp20z.json
AlCalzone: ...that comment was meant to go on the "State after power loss" parameter.

emichael: Done!

emichael:
Aggg, sorry about the import path. Really should have set something up to test it locally before pushing. (Side note: there probably should be a way to refer to the templates directory independent of the current directory, no? I get that the folder structure of the project is pretty mature at this point, but it still seems unnecessarily brittle?)

AlCalzone (approved):

PR #2993, opened 2021-07-07
fix(config): allow manual entry for tone duration for Aeotec Doorbell and Siren

viljasenville:
Allow manual entries for Aeotec tone duration settings. Allowed values according to the device manuals:
- 0 = Original tone duration
- 1-254 = Tone duration in seconds
- 255 = Last used configuration value.


AlCalzone (approved):

AlCalzone:
Thanks!

Issue #2910, opened 2021-02-10
ZMNHDA Qubino Flush Dimmer multilevel switch current value not propagating from zwavejs to zwavejs2mqtt


NINE78:
### Version

Build/Run method

- [X] Docker -> running a Home Assistant supervised add-on

zwavejs2mqtt version: 1.1.0
zwavejs version: 6.1.3

### Describe the bug

For my Qubino (GOAP) ZMNHDA dimmers, I see multilevel switch ("current value") updates sent to Endpoints 0 and 1, but only Endpoint 1 is received in zwavejs2mqtt, this when manually toggling the light switch.

From zwavejs log:
```
15:34:46.396 SERIAL « 0x0109000400400326030094                                            (11 bytes)
15:34:46.398 CNTRLR   [Node 064] [~] [Multilevel Switch] currentValue: 63 => 0          [Endpoint 0]
15:34:46.399 SERIAL » [ACK]                                                                   (0x06)
15:34:46.402 DRIVER « [Node 064] [REQ] [ApplicationCommand]
                      └─[MultilevelSwitchCCReport]
                          current value: 0
15:34:46.407 CNTRLR   [Node 064] [~] [Multilevel Switch] currentValue: 63 => 0          [Endpoint 1]
15:34:47.336 SERIAL « 0x0110000400400a320221340000023b0000bd                              (18 bytes)
```

However in zwavejs2mqtt log, only 38-1 value is updated, which does not get picked up by the "Current Value" entity in HA (there is only one for the light):
`2021-02-10 15:34:46.411 INFO ZWAVE: Node 64: value updated: 38-1-currentValue 63 => 0`

### To Reproduce

1. toggle lights
2. observe only endpoint 1 coming through to zwavejs2mqtt in the logs

### Expected behavior

Also expect 64-38-0-currentValue to be updated

### Additional context

Logs attached
[zwavejs2mqtt.log](https://github.com/zwave-js/zwavejs2mqtt/files/5959164/zwavejs2mqtt.log)
[zwavejs_578.log](https://github.com/zwave-js/zwavejs2mqtt/files/5959168/zwavejs_578.log)

AlCalzone:
This **should** be fixed with https://github.com/zwave-js/node-zwave-js/pull/2997 in the upcoming 8.x release. If not, feel free to reopen with new logs.

bratanon:
Also seeing this issue with all ZMNHDD. 

[...]

Issue #2714, opened 2021-05-18
[feat] Proper Multi Channel support


andras-tim:
**Is your feature request related to a problem? Please describe.**

The current group assignment has several issues, but I think these are in strong connection, therefore I write all of them into this ticket.

* [ ] The **single-channel** mode specifies only the **Target Node**, however, the **multi-channel** requires the **Target Endpoint** too. If I have a device that supports both of them (e.g. Qubino Flush 2 Relay) then the **Target Endpoint** will specify the communication mode of an assign. So...
  - singngle-channel mode: 
    **Root device (Endpoint 0)** / **Lifeline (Group 0)**
    => **Node 1** / `None` **Target Endpoint**
  - multi-channel mode:
    **Root device (Endpoint 0)** / **Lifeline (Group 0)** 
    => **Node 1** / `0` **Target Endpoint**

  that is so important because...

  - the modes have different value/switch/... identifiers
  - some values/switches/... are not reachable in single-channel mode
  - the multi-channel mode 1.5 - 2 times slower because it has an extra multi-channel packet-encapsulation

  _(Based on my 3 days long tests and this detailed [Sylab: Z-Wave Association Basics](https://www.silabs.com/documents/public/application-notes/APL12956-Z-Wave-Association-Basics.pdf)_ documentation.)
  
  ... So I think the channel mode should be highlight somehow instead of a simple `0` or `None`... because this is not so intuitive.

* [ ] The current .json device descriptors contain only one set of association groups. However, a multi-channel device can have different groups per endpoint, e.g. check the **Association Groups** in the manual of the [Flush 2 Relays](https://qubino.com/manuals/Flush_2_Relay.pdf) on **page 44**. This has a **Root Device** + 3 extra **Endpoints** with totally different(!) groups.

* [x] I would like to display the **Group Index** beside the **Group Names** to be able to easily identify the desired group based on user manuals.

* [ ] Some devices haven't multiple endpoints, however, the manual says it has several. E.g. the _Qubino's Flush Dimmer Plus_ has a root device (Endpoint 0)  + 4 extra endpoints... but the UI shows only one.

**Describe the solution you'd like**
- Please, highlight the single-multi channel mode
- The device descriptor files have to contain different association groups for Entrypoints, according to the user manuals.
- Show all endpoints and group indexes too.

Thank you!

andras-tim:
> @andras-tim What version are you using?

[...]

Issue #2257, opened 2021-04-07
Cleanup bad associations to controller set by OZW


blhoward2:
Reference #2159. OZW sets up some bad associations to the controller, which we should check for and cleanup.

Unsolicited report when lifeline association is set to Node 1 Endpoint 0:

![](https://community.inovelli.com/uploads/default/original/2X/a/a0aabd1d77fcce302d81ab0814f4413e5b1fa506.png)

Unsolicited report when lifeline association is set to Node 1 (with no endpoint):
![](https://community.inovelli.com/uploads/default/original/2X/a/afe7ece6d6f71c6b93b63c72b708c7e14ff55358.png)

Issue #1989, opened 2021-03-04
[bug] CT200X Thermostat Unreliable communication / not getting updates


srirams:
Before submitting a bug please read: <https://zwave-js.github.io/zwavejs2mqtt/#/troubleshooting/bug_report>

### Version

Build/Run method

-  [X] Docker
- [ ] PKG
- [ ] Manually built (git clone - npm install - npm run build )

zwavejs2mqtt version: 2.1.0
zwavejs version: 6.6.0

### Describe the bug

CT200X Thermostat status switches between Alive / Dead constantly. Setting setpoint from zwavejs seems to work, but changing values on thermostat does not update on zwavejs. Temperature values from thermostat are not updated.

### To Reproduce

1. Change temperature on thermostat
2. Value is not updated on zwavejs2mqtt
3. Change setpoint on zwavejs2mqtt
4. Value is updated on thermostat

### Expected behavior

1. Values updated on both thermostat and zwavejs2mqtt

### Additional context

Attached logs below. Node 2 is the thermostat I'm adjusting temperatures on. Node 5 is also a CT200X. Using OpenZWave, thermostat operation was more reliable. https://github.com/zwave-js/node-zwave-js/issues/1721 is related issue, but I don't believe it fixed the issue.

[zwavejs2mqtt.log](https://github.com/zwave-js/zwavejs2mqtt/files/6080800/zwavejs2mqtt.log)
[zwavejs_1.log](https://github.com/zwave-js/zwavejs2mqtt/files/6080801/zwavejs_1.log)
[fd661d38.json.txt](https://github.com/zwave-js/zwavejs2mqtt/files/6080802/fd661d38.json.txt)

srirams:
I think with version 7, this has stopped  working again. Attached log below
[zwavejs_1.log](https://github.com/zwave-js/node-zwave-js/files/6281498/zwavejs_1.log)

AlCalzone:
Can you tell me what I'm looking for? Which node? What did you control and when?

srirams:
Sorry, basically the same symptoms as the original report. There are no temperature updates from the thermostat. If I hit "Refresh Values" I get the updated values. It is nodes 10 and 11, which I've filtered by.

I've tried "Reinterview node" as well as deleting the cache files (only keeping the settings.json), and it doesn't make a difference. Unsure if maybe its the same issue as https://github.com/zwave-js/node-zwave-js/issues/2165?

AlCalzone:
[...]

PR #2997, opened 2021-07-10
feat: rework how endpoints and lifeline associations are handled

AlCalzone:
This PR solves a long-standing issue with devices that misuse endpoints and don't correctly use multi channel associations (at least according to how I understand the terrible specification around these). See #2286 for a detailed explanation.

**We expect that this will be breaking for some devices that need one or more of the below compat flags from now on.  If that is the case, please open issues and include a driver log of a re-interview.
At the very least, value IDs will change from endpoints 1+ to 0 for several devices.**

Here's a breakdown of the changes:

### Ignore unnecessary endpoints instead of the root device
Until now we used to hide the root device's application CC values if the same value was existing on an endpoint. It has been found that many devices mis-use endpoints when it does not actually make sense and don't correctly report their values. If all endpoints have different device classes, they are most likely unnecessary and will be ignored, because all the functionality will also be exposed on the root device. Those devices that only report via the root device should now work out of the box.

For the other devices that actually need their endpoints despite them being different, a new compat flag was added to restore the current behavior:
```jsonc
"preserveEndpoints": "*", // to preserve all endpoints and hide the root values instead
"preserveEndpoints": [2, 3], // to preserve endpoints 2 and 3, but ignore endpoint 1.
```
For example, there are some thermostats which expose an external sensor with battery through a seemingly unnecessary endpoint, but it needs to be preserved

If some endpoints share a device class, like multi-channel switches, the current behavior of hiding the root device's in favor of the endpoints' values will stay in place.

### Define which endpoint to map root reports to
Because the specs stated that the root device should mirror at least endpoint 1 (and possibly others), we used to map reports that came through the root device to the first supporting endpoint. This behavior has changed a couple of times and has never been perfect for everyone. With this PR, device configs need to opt-in instead to turn on the mapping and specify the target endpoint. Otherwise, reports to the root device get silently ignored if the root's values are hidden.

To do so, a new compat flag was added:
```jsonc
"mapRootReportsToEndpoint": 1
```
This also needs to be used for some devices, e.g. dual channel switches, that partially report un-encapsulated through the root device. 

### Prefer node associations on each endpoint over multi channel associations
Many manufacturers seem to get multi channel associations "wrong". We now prefer setting up node associations on each endpoint instead if possible. The default strategy for each endpoint is the following:
1. Try a node association on the current endpoint/root
2. If Association CC is not supported, try assigning a node association with the Multi Channel Association CC
3. If that did not work, fall back to a multi channel association (target endpoint 0)
4. If that did not work either, the endpoint index is `> 0` and the node supports Z-Wave+:  
Fall back to a multi channel association (target endpoint 0) on the root, if it doesn't have one yet.

Incorrect lifeline associations like the ones set up by OZW with target endpoint 1 will automatically get cleaned up during the process.

This strategy can be controlled with the `multiChannel` flag in the association config. `true` forces a multi channel association (if supported) and does not fall back to node associations. `false` forces a node association (if supported) and does not fall back to a node association.

fixes: #2286
fixes: #2257
fixes: #1989
fixes: #2714
fixes: #1860
fixes: #2910


blhoward2:
I assume this will potentially fix that series of devices where we preserved the root endpoint temporarily so that endpoint 1 didn't change with endpoint 2. I already have the Zooz ones and I emailed Inovelli to see if they'd either send me one, or test on their own, for the LZW36.

AlCalzone:
Yep, that specifically. We're going to have to remove the workaround then.

AlCalzone:
@blhoward2 in https://github.com/zwave-js/node-zwave-js/pull/2293 we added the root value workaround to the following devices:
* Zooz ZEN16
* Zooz ZEN17
* Zooz ZEN25

Comment: _This device improperly reports the state of R2 (endpoint 2) through the root endpoint in a way that also changes the state of R1 (endpoint 1)_

And

* LZW36

Comment: _This device improperly reports the state of endpoint 2 through the root endpoint in a way that also changes the state of endpoint 1_

Do you remember how exactly this problem manifested or do you still have some of those and can make a new re-interview log? My guess is that those are cases of devices that either don't need any endpoints or that benefit from the simplified Lifeline association setup.

blhoward2:
I have the ZEN16 and ZEN17. Upon local operation of the switch for endpoint 2 it would toggle endpoint 1 because it would map the report. I believe I still have logs in my email we sent Zooz and I'll send those to you. 

The LZW36 was a similar issue. It has two switches and operating endpoint 2 would send a report across the root endpoint that would map to endpoint 1. There is an entire thread about this over at Inovelli. https://community.inovelli.com/t/re-zwave-js-lzw36-driver/7103/82?u=blhoward2 We also had an issue with logs: https://github.com/zwave-js/node-zwave-js/issues/2155

blhoward2:
There are some logs of the ZEN17 here. https://github.com/zwave-js/node-zwave-js/issues/2379

AlCalzone:
> There are some logs of the ZEN17 here. #2379

Okay, that looks like the device uses the root as a "compound" state. Can you confirm that? Zen16 is probably similar. In that case we should keep the compat flag or that "compound" state will be hidden.

For the LZW36 I don't think we need the flag. The incorrect update was caused by a mapping of a multicast the device sent, probably as a result of the weird OZW association. In any case we're not mapping that anymore.

AlCalzone:
I'm going to merge this so I don't have to keep it up to date all the time. We can solve these device-specific issues afterwards.

PR #3018, opened 2021-07-14
fix(config): fingerprints for two Sunricher dimmers

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/discussions/3015


AlCalzone:
@zwave-js-bot automerge

PR #3023, opened 2021-07-15
fix(config): add additional product ID for Willis Electric NZW30T

Kirchoff:
This is a switch that I have, that want to add support for. It appears entirely identical to the other Willis Electric switches I have that report a device ID of 0x1e00. Thanks!


Kirchoff:
@zwave-js-bot - help with a semantic commit message?

AlCalzone:
@zwave-js-bot rename commit

PR #3030, opened 2021-07-15
feat(config): create config file for ZWA011

ScratMan:
ZWA011 based on ZWA008 but with Z-Wave Series 700 but without tilt detection and external switch input.

File tested working locally.

### DIFF
--- a/packages/config/config/devices/0x0371/zwa011.json
+++ b/packages/config/config/devices/0x0371/zwa011.json
@@ -20,15 +20,15 @@
 			"isLifeline": true
 		},
 		"2": {
-			"label": "Control devices when a magnet trips",
+			"label": "Control",
 			"maxNodes": 5
 		},
 		"3": {
-			"label": "Sends our alarm message when magnet controlled",
+			"label": "Alarm sensor",
 			"maxNodes": 5
 		},
 		"4": {
-			"label": "Sends alarm messages when tamper is tripped",
+			"label": "Tamper",
 			"maxNodes": 5
 		}
 	},
--- a/packages/config/config/devices/0x0371/zwa011.json
+++ b/packages/config/config/devices/0x0371/zwa011.json
@@ -58,11 +58,11 @@
 			"$import": "../0x0086/templates/aeotec_template.json#association_group_2_basic_set_on"
 		},
 		"9": {
-			"$import": "../0x0086/templates/aeotec_template.json#association_group_2_basic_set_off"
+			"$import": "../0x0086/templates/aeotec_template.json#association_group_2_basic_set_off",
 			"defaultValue": 0
 		},
 		"10": {
-			"$import": "../0x0086/templates/aeotec_template.json#time_delay_on"
+			"$import": "../0x0086/templates/aeotec_template.json#time_delay_on",
 			"defaultValue": 0
 		},
 		"11": {
@@ -70,7 +70,7 @@
 			"defaultValue": 0
 		},
 		"12": {
-			"$import": "../0x0086/templates/aeotec_template.json#time_delay_tamper"
+			"$import": "../0x0086/templates/aeotec_template.json#time_delay_tamper",
 			"defaultValue": 0
 		},
 		"13": {
--- a/packages/config/config/devices/0x0371/zwa011.json
+++ b/packages/config/config/devices/0x0371/zwa011.json
@@ -6,7 +6,7 @@
 	"devices": [
 		{
 			"productType": "0x0002",
-			"productId": "0x000B"
+			"productId": "0x000b"
 		}
 	],
 	"firmwareVersion": {
--- a/packages/config/config/devices/0x0371/zwa011.json
+++ b/packages/config/config/devices/0x0371/zwa011.json
@@ -34,7 +34,18 @@
 	},
 	"paramInformation": {
 		"3": {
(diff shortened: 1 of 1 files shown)


ScratMan:
OK, there is an issue with the parameters definition. I reused the parameters from zwa012.json, that seems to be based on zwa008.json ; but the zwa012 parameters definition is wrong in the json file :

- Parameter 2+ of zwa008 becomes parameter 3+ 
- Parameter 2 is an added setting for the external dry contact

So when trying to set some parameters of the ZWA011, it messes the device's settings.

I tried to modify the zwa011.json file as attached, but now I can't get it to be recognized
[zwa011.txt](https://github.com/zwave-js/node-zwave-js/files/6825796/zwa011.txt)

The control panel indicates my ZWA011 nodes as
<html>
<body>


Unknown manufacturer 0x0371 | 0x0002 | Unknown product 0x000b
-- | -- | --



</body>
</html>
Trying to reinterview the devices doesn't help.

AlCalzone:
> Trying to reinterview the devices doesn't help.

Just to be sure - did you restart zwavejs2mqtt (assuming you use that)?

ScratMan:
Yes, I restarted the docker container running zwavejs2mqtt.

I tried to copy directly the original JSON file in /use/SRC/app/store/.config-db/devices/0x0371/ folder where the container stores the config files, and it worked.
But when I tried to modify the file it started to fail to recognize the device.
I created the /use/src/app/store/config/ folder to try with placing the file inside, but it fails to parse the file due to incorrect template import paths.
So I made a new version of the file without templates, but the driver still doesn't recognize the device.

AlCalzone:
> I created the /use/src/app/store/config/ folder to try with placing the file inside, but it fails to parse the file due to incorrect template import paths.

Testing config files with templates is a bit trickier. You can try and replicate the structure in that directory:
```
templates/
-- master_template.json
0x0086/
-- templates/
   -- aeotec_template.json
0x0371/
-- zwa011.json
```

ScratMan:
This is something I need to try. But first I need to understand why the non-template config file is not used.
I activated the 2 logging in a file options in the interface to try to log the errors. I'll retry with the different files.

AlCalzone:
If you clone the repo and edit the file in the local clone, you can check it with `yarn run lint:config`. Maybe there's an error.

https://zwave-js.github.io/node-zwave-js/#/development/intro

ScratMan:
I checked the file, and I found two small typo errors (missing comma and unexpected tab). I fixed the file, excluded the ZWA011 devices from the network, stopped the ZWaveJS2MQTT container, removed custom config files from all paths, added the /usr/src/app/config folder and copied my fixed non-template JSON inside.
Then I started the ZWaveJS2MQTT container, included one ZWA011, and voilà, it is correctly detected. In the log I can see the attached custom config file is used.
[zwa011.txt](https://github.com/zwave-js/node-zwave-js/files/6829584/zwa011.txt)


I'll now try with the templates.

ScratMan:
After testing by recreating the templates folders in the /usr/src/app/store/config folder, it works correctly with the following json
[zwa011.json.txt](https://github.com/zwave-js/node-zwave-js/files/6829756/zwa011.json.txt)

ScratMan:
@zwave-js-bot fix lint

ScratMan:
By removing /usr/src/app/store/config/ folder and putting back the zwa011.json in the /usr/src/app/store/.config-db/devices/0x0371/ folder, the device is not recognized anymore.
I returned to the /usr/src/app/store/config folder with the correct structure, and it stil works.
I made an update to the definition of parameter 3 to align it with the sensor type

AlCalzone:
> By removing /usr/src/app/store/config/ folder and putting back the zwa011.json in the /usr/src/app/store/.config-db/devices/0x0371/ folder, the device is not recognized anymore.

If you change the files on disk, the driver needs to be restarted, because it only indexes the files on startup.

packages/config/config/devices/0x0371/zwa011.json
```diff
@@ -0,0 +1,98 @@
...
+			"maxNodes": 5
+		},
+		"3": {
+			"label": "Group 3",
+			"maxNodes": 5
+		},
+		"4": {
+			"label": "Group 4",
+			"maxNodes": 5
+		}
```
AlCalzone: Are there any better labels for these?

ScratMan: Labels added.

ScratMan:
> > By removing /usr/src/app/store/config/ folder and putting back the zwa011.json in the /usr/src/app/store/.config-db/devices/0x0371/ folder, the device is not recognized anymore.
> 
> If you change the files on disk, the driver needs to be restarted, because it only indexes the files on startup.

I restarted the docker container (using "zwavejs/zwavejs2mqtt:latest" image), I suppose it should be enough.

I tried to copy the file from my file system to the docker container with the docker cp command ; but the /usr/src/app/store folder from the container is mapped to the external folder /home/homeassistant/zwave-js/ on the host's HDD. Would the driver rewrite the config folder while stopping overwriting my mods ?
I don't remember if I tried to copy the zwa011.json file to the local /home/homeassistant/zwave-js/.config-db/devices/0x0371/ folder while container was stopped and before starting it. I think I did, but not sure.

PR #3031, opened 2021-07-16
fix(config): fix size of 0-65535 value for aeotec devices

kdknigga:


kdknigga:
@AlCalzone , @blhoward2 ,

I'm not sure how to resolve this lint issue.  The Aeotec engineering sheet shows config parameters 52, 53, and 54 as being 4 bytes with 2 bytes being used for battery voltage and 2 bytes being used for software version, but it appears that one bit in the voltage field is completely not used.

image

I guess maybe the mask for the battery voltage portion in the device file should be 0x7fff0000?

Unfortunately, I don't have a ZW162 to test.  I'm just trying to fix the ZWA023s I do have.

blhoward2:
There is a calculator in the documentation. You'd just mask the bottom two bytes for the one, then the top two bytes for the other. The min/max range will bound the value within that to the range they provide, though since it's read only that doesn't matter much. 

https://zwave-js.github.io/node-zwave-js/#/config-files/file-format. Look under Partial Parameters for the calculator. Set to four bytes, then just check all of them for the bottom two. E.g. 0xffff for Battery Software Version. 0xffff0000 for the top two. You only need to worry about individual bits where the individual bit does something different. Like where bit 1-2 are amperage report, 3-4 are voltage, etc. Otherwise just mask the whole byte.

kdknigga:
The problem with masking the whole byte is that the linter complains that it doesn't jibe with the max value provided by the engineering sheet (32767 vs 65535).

I'll set "52[0x7fff0000]" (and 53 and 54) in zw162.json (that calculator is handy!) and leave maxValue at 32767 in the template.  It seems most right.  I'm just a little concerned I don't have the hardware to test.

blhoward2:
Lint only cares if the value exceeds the maximum, not if it is under. Can you post your actual parameter definition?

blhoward2:
Your problem is that you are splitting the valueSize. It has to be four, the size of the full parameter. The mask then takes only part of it. Lint is throwing an error because you're masking something out of bounds.

blhoward2:
You also cannot change the template base. You can override it in your definition if need be by doing the import statement and then the valueSize there. What you have now would break most of the Aeotec devices. That's being imported all over the place. It's defaulted to four because that's the most common appearance in the Aeotec files. 

I'm not actually clear what you're trying to do... The size is determined by the firmware, not the value range. It may be a size of four and only allow values 0-1 as an example, but changing the size will make the set fail.

kdknigga:
Let me give the full rundown of what I'm trying to do.

I have a handful of ZWA023s.  I can't set config parameters 23, 24, 25, or 26 because zwave_js is trying to send 4 bytes for those parameters when the device expects 2.  

Here you can see the controller try to set the parameter with a size 4, but the device doesn't take it and reports the current value with size 2:

```
2021-07-16T14:05:14.450Z DRIVER » [Node 039] [REQ] [SendData]
                                  │ transmit options: 0x25
                                  │ callback id:      25
                                  └─[ConfigurationCCSet]
                                      parameter #:      23
                                      reset to default: false
                                      value size:       4
                                      value format:     UnsignedInteger
                                      value:            30
2021-07-16T14:05:14.476Z DRIVER « [RES] [SendData]
                                    was sent: true
2021-07-16T14:05:14.492Z DRIVER « [REQ] [SendData]
                                    callback id:     25
                                    transmit status: OK
2021-07-16T14:05:15.508Z DRIVER » [Node 039] [REQ] [SendData]
                                  │ transmit options: 0x25
                                  │ callback id:      26
                                  └─[ConfigurationCCGet]
                                      parameter #: 23
2021-07-16T14:05:15.524Z DRIVER « [RES] [SendData]
                                    was sent: true
2021-07-16T14:05:15.536Z DRIVER « [REQ] [SendData]
                                    callback id:     26
                                    transmit status: OK
2021-07-16T14:05:15.552Z DRIVER « [Node 039] [REQ] [ApplicationCommand]
                                  └─[ConfigurationCCReport]
                                      parameter #: 23
                                      value size:  2
                                      value:       600
```

The config for this device has those parameters using the base_0-65535 template.  It appeared to me that template has a bug where it has a valueSize of 4 instead of 2.  You only need 2 bytes to get 0-65535 after all.

I tried just changing that, but then the linter was complaining that all of the Partial Parameters that make up on real parameter have to be of the same valueSize for parameters 52, 53, and 54 for the ZW162.  Those parameters break down into two partial parameters: button_battery_voltage and button_version.  button_battery_voltage is set to valueSize 4 and button_version inherts its valueSize from base_0-65535.  So now I'm just trying to get all of these things lined up.

I was thinking the valueSize of a partial parameter was the size of the partial parameter itself, not the size of the parent parameter, so I was trying to change button_battery_voltage to 2 to match base_0-65535.  But, if that's not the case, if valueSize of a partial parameter is the size of the whole parent parameter, then I guess I need to make button_version *not* inherit from base_0-65535.

Does this make sense?

blhoward2:
Your problem makes sense, your solution does not. The template does not have a bug, that is intentional. The device file has an error. Just make them all like:

```
"23": {
			"$import": "../0x0086/templates/aeotec_template.json#auto_report_interval_watt",
                         "valueSize": 2
		},
```

That will override the template definition for the applicable base.

kdknigga:
Sounds good.  Thanks a lot for your help!

blhoward2:
No problem, thanks for fixing it. I know it's confusing. Anything after the import statement overrides what is there or fills in something missing from the template.

PR #3032, opened 2021-07-16
fix(config): config params 23, 24, 25, and 26 on aeotec zwa023s are size 2, not 4

kdknigga:


blhoward2 (approved):
These are fine, I was helping in another PR that is since closed.

kdknigga:
Thanks again for your help, @blhoward2 !

AlCalzone:
Thanks!

kdknigga:
@blhoward2 ,

Is there a way to dump the effective config for a device after all of the imports and templating have been applied?

AlCalzone:
Not "offline" (at least not super easy), but it is available on the `deviceConfig` property of a node, which should also be included in the node dumps HA and zwavejs2mqtt create.

kdknigga:
I do see a trimmed-down version of the config when I dump a node to json in zwavejs2mqtt, but it's not everything that's included in the config file.  valueSize is what I'm looking for.

Unfortunately, overriding the template-provided value of valueSize doesn't seem to be working.  It looks like I have the files in the right place on the filesystem, but even after a zwavejs2mqtt restart and a device reinterview the wrong value is being used.  So now I'm trying to figure out how to tell if the config isn't getting read, or if something is overriding the override, or if there's something else going on.

AlCalzone:
> It looks like I have the files in the right place on the filesystem

How and where did you place them?

kdknigga:
I'm running HA, so /mnt/data/supervisor/addons/data/a0d7b954_zwavejs2mqtt/db/devices/0x0371/zwa023.json.  In fact, I've since downloaded device database 7.12.1-20210717 through zwavejs2mqtt, and it also went to /mnt/data/supervisor/addons/data/a0d7b954_zwavejs2mqtt/db.  On zwavejs2mqtt startup, I see:

```
2021-07-19T13:24:37.536Z DRIVER   loading configuration...                                                                                                                                                                                    
2021-07-19T13:24:37.573Z CONFIG   version 7.12.1-20210717
```

AlCalzone:
Ok I have no clue what hoops you need to jump through with that setup 😅

kdknigga:
Yeah... this sort of thing is great for making deployments easy.  It's not great for making development and testing easy.  Unfortunately, I don't have a test/dev zwave network.  Maybe I should add some stuff to my Christmas list for that.

But, the fact that it's reporting loading the 20210717 version of the device database should suggest that it's getting the updated configs if by no other means, yeah?

kdknigga:
@AlCalzone , just FYI, but I updated to 8.0.5 and this change is working as expected now.  I'm guessing the change just wasn't getting loaded correctly from the online device database update.  /shrug

PR #3081, opened 2021-07-23
fix(config): add groups for Heatit Smoke Battery

geirra:
Didn't notice that the groups where also missing when i added the updated fingerprint yesterday, sorry about that.

Manual to verify that groups are used: https://media.heatit.com/1757


packages/config/config/devices/0x019b/4512688.json
```diff
@@ -18,6 +18,18 @@
...
 	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 1,
+			"isLifeline": true
+		},
+		"2": {
+			"label": "Smoke Sensor (Basic Set)",
+			"description": "When the smoke detector is active, it will send Basic Set (0xFF) to this group. When the smoke detector is deactived, it will send Basic Set (0x00).",
```
blhoward2: Remove the description from the groups. None of the UIs display it anyway. Defining the groups isn't actually required for most devices but doing so is fine.

blhoward2 (dismissed):

packages/config/config/devices/0x019b/4512688.json
geirra: removed description  from group 2

AlCalzone (approved):

PR #3082, opened 2021-07-23
fix(config): updates to FGS223 and FGD212

glyph-se:
After updating the FGD212 file in https://github.com/zwave-js/node-zwave-js/pull/2854 I have now continued with the FGS223. I might to more Fibaro devices in the future, but I'm taking it step-by-step and hopefully improving a little bit each time.

I have extracted the things I found to be common into a new `fibaro_template.json` file.

As I went through the FGS223 I also did several text updates to the FGD212. E.g. consistently using "Blink" instead of "Flash", which is the term used in the Aeotec files.

I am a bit unsure of a few things:

* When to define association groups. The [guide ](https://zwave-js.github.io/node-zwave-js/#/config-files/file-format?id=associations) says to only do this when necessary, one of the devices had, the other didn't. The default names are not terrible, they are: `On/Off (S2)`, `Dimmer (S2)` and `Lifeline`. I guess that means I should leave them out in both files? Even since they were already present for one device previously and I have only tried with the latest FW?
* What to do with values in templates that differ between the 2 devices. E.g. `reports_periodic` and `energy_report_threshold`. They have different default and max values. Currently I left that out in the template and defined the differences in both device files. As I only have done 2 devices I don't know if one is more common than the other. Is this ok?
* Can templates be used for other things than parameters? E.g. the metadata explaining how to do inclusion etc. I found no examples of this when I checked the existing files.

Happy for any feedback.

### DIFF
--- a/packages/config/config/devices/0x010f/fgd212_3.5.json
+++ b/packages/config/config/devices/0x010f/fgd212_3.5.json
@@ -29,6 +29,34 @@
 		"min": "3.5",
 		"max": "255.255"
 	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"description": "Reports the device status and allows for assigning single device only (main controller by default).",
+			"maxNodes": 1,
+			"isLifeline": true
+		},
+		"2": {
+			"label": "Key S1 - On/Off",
+			"description": "Assigned to Key S1 (uses basic command class)",
+			"maxNodes": 5
+		},
+		"3": {
+			"label": "Key S1 - Dimmer",
+			"description": "Assigned to Key S1 (uses multilevel switch command class)",
+			"maxNodes": 5
+		},
+		"4": {
+			"label": "Key S2 - On/Off",
+			"description": "Assigned to Key S2 (uses Basic command class)",
+			"maxNodes": 5
+		},
+		"5": {
+			"label": "Key S2 - Dimmer",
+			"description": "Assigned to Key S2 (uses multilevel switch command class)",
+			"maxNodes": 5
+		}
+	},
 	"paramInformation": {
 		"1": {
 			"label": "Minimum Brightness Level",
@@ -172,11 +200,11 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "No calibration or manual settings",
+					"label": "No calibration or manual override",
 					"value": 0
 				},
 				{
-					"label": "Dimmer operates on auto-calibration",
+					"label": "Using auto-calibration",
 					"value": 1
 				}
 			]
@@ -234,7 +262,7 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Mono-stable input (button)",
+					"label": "Mono-stable input (momentary button)",
 					"value": 0
 				},
 				{
@@ -242,7 +270,7 @@
 					"value": 1
 				},
 				{
-					"label": "Roller blind switch (UP / DOWN)",
+					"label": "Roller blind switch (up / down)",
 					"value": 2
 				}
 			]
@@ -280,7 +308,7 @@
 					"value": 0
 				},
 				{
-					"label": "Output synced with input switch position",
+					"label": "Output syncs with input switch position",
 					"value": 1
 				}
 			]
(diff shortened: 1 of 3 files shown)


AlCalzone:
>     * When to define association groups

Here is a little more detail - although in a different context: https://zwave-js.github.io/node-zwave-js/#/config-files/using-telemetry-data?id=supportsagi-associationgroups

> The default names are not terrible, they are: On/Off (S2), Dimmer (S2) and Lifeline.

I think they're okay. Unless you need to force a multi channel lifeline association to make reporting work, you should be good without them.

> Currently I left that out in the template and defined the differences in both device files. As I only have done 2 devices I don't know if one is more common than the other. Is this ok?

Seems okay to me - maybe put a comment into the template mentioning that. If you find more devices that share a value, it might make sense to move the default back into the template. The definitions after `$import` will override it where necessary.

> Can templates be used for other things than parameters?

Yes, use them where it seems to make sense. Everywhere a JSON property is expected, you can use `$import`.

packages/config/config/devices/0x010f/fgs223.json
```diff
@@ -232,478 +208,232 @@
...
+			"$import": "templates/fibaro_template.json#active_power_report_threshold",
 			"defaultValue": 20,
-			"unsigned": true
+			"label": "First Channel - Active Power Report: Threshold"
 		},
 		"51": {
-			"label": "First channel - minimal time between power reports",
-			"description": "This parameter determines minimum time that has to elapse before sending new power report to the main controller.",
-			"valueSize": 2,
+			"label": "First Channel - Active Power Report: Minimum Time Between Reports",
```
AlCalzone, line 374: Wouldn't this just be
```suggestion
			"label": "First Channel - Active Power Report: Interval",
```
in simpler words?

packages/config/config/devices/0x010f/templates/fibaro_template.json
```diff
@@ -0,0 +1,224 @@
...
+		"description": "Only active when the node is included securely"
+	},
+	"send_secure_commands_group3": {
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "Send Secure Z-wave Commands to 3rd Association Group",
+		"description": "Only active when the node is included securely"
+	},
+	"send_secure_commands_group4": {
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "Send Secure Z-wave Commands to 4tg Association Group",
```
AlCalzone:
```suggestion
		"label": "Send Secure Z-wave Commands to 4th Association Group",
```

packages/config/config/devices/0x010f/templates/fibaro_template.json
```diff
@@ -0,0 +1,224 @@
+{
+	"default_metadata_for_fibaro": {
+		"inclusion": "Turn the primary controller of Z-Wave network into inclusion mode, quickly press key S1 or B-button on product's housing three times",
+		"exclusion": "Turn the primary controller of Z-Wave network into exclusion mode, quickly press key S1 or B-button on product's housing three times",
+		"reset": "1. Power-cycle the device\n2. Press and hold the B-button to enter the menu\n3. Wait for the LED to turn yellow\n4. Quickly release and click B-button again\n5. Device will restart and LED will turn red"
+	},
+	"send_secure_commands_group2": {
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "Send Secure Z-wave Commands to 2nd Association Group",
```
AlCalzone:
```suggestion
		"label": "Send Secure Commands to 2nd Association Group",
```
Seems unnecessary. Please do the same change to the other variants of this param.

AlCalzone (changes requested):

packages/config/config/devices/0x010f/templates/fibaro_template.json
```diff
@@ -1,4 +1,9 @@
 {
+	"default_metadata_for_fibaro": {
```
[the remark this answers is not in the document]
glyph-se: Agree. Didn't know what to call this. But dropping the fibaro definitely makes it better.

packages/config/config/devices/0x010f/fgd212_3.5.json
```diff
@@ -445,11 +429,11 @@
...
 			"options": [
 				{
-					"label": "Dimming possible",
+					"label": "Enable (dimming possible",
 					"value": 0
 				},
 				{
-					"label": "Dimming not possible",
+					"label": "Disable (dimming not possible)",
 					"value": 1
```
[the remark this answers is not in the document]
glyph-se: I wanted to make it more look like the settable parameter 32. But I will revert to the old value if you believe that to be better.

packages/config/config/devices/0x010f/fgs223.json
glyph-se: I don't think minimum time is the same as interval. At least to me naming it interval would make the difference between this and parameter 58 `Periodic Active Power Reports` confusing

packages/config/config/devices/0x010f/templates/fibaro_template.json
glyph-se: Done

packages/config/config/devices/0x010f/fgs223.json
AlCalzone: Okay

AlCalzone (approved):
LGTM - I'll give @blhoward2 some time to look at this too

blhoward2:
Are we sure this will always work in the template file? It is different than we normally do it:

""~/templates/master_template.json#base_enable_disable""

AlCalzone:
If you mean the `~/`, I added that feature recently to avoid having to deal with multiple levels of `../../`
https://github.com/zwave-js/node-zwave-js/pull/3035

AlCalzone:
Thanks - I've discussed with @blhoward2 and while templating metadata isn't something we're going to do on a large scale, I think it is okay for Fibaro devices as they are mostly identical in that regard.

PR #3100, opened 2021-07-27
refactor(config): convert `paramInformation` to an array

AlCalzone:
`paramInformation` being an object has a few drawbacks:
* The source file order is not preserved, so we have no means to logically sort config parameters in large files
* Bot actions that modify the files where we don't follow the JS object ordering will mess up the hand-crafted ordering
* It is (almost) impossible to detect duplicate param numbers

This PR changes the definition of `paramInformation` to an array. Legacy files will still be parsed correctly, but now the enforced format is as follows:
```jsonc
"paramInformation": [
	{
		"#": "1",
		// parameter 1 definition
	},
	{
		"#": "2",
		// parameter 2 definition
	}
	// ... more parameters ...
]
```
To identify the individual param numbers (including the bitmask), the `#` property had to be added. There is no more additional indentation for conditional param variants, they are simply defined on the same level as normal parameters.


AlCalzone:
@blhoward2 this might interest you

AlCalzone:
@zwave-js-bot fix lint

blhoward2:
LGTM in concept. Did you mean for the params to be a single line where there is only an import statement, or is that a lint error?

blhoward2 (dismissed):

AlCalzone:
> Did you mean for the params to be a single line where there is only an import statement, or is that a lint error?

They are auto-formatted with Prettier and that would be a lint error otherwise. I can check if adding a line break after the opening bracket will change its desired format.

blhoward2 (approved):

packages/config/maintenance/importConfig.ts
```diff
@@ -1388,7 +1388,7 @@ async function parseZWAProduct(
 				}
 			}
 		}
-		newConfig.paramInformation[param.ParameterNumber] = parsedParam;
+		if (!found) newConfig.paramInformation.push(parsedParam);
```
blhoward2, line 1411: A few things. Is this right? Doesn't this make it so that we only push the parameter if there was not already a matching param in the original file, which means we can never update what is there? We've been purposely merging the zwa parameters over what we have otherwise there is no way to see errors in our current files. It is also easier to matchup for templating if you abandon our current description/label because the manufacturers tend to use similar labels and such for Levenshtein distance calcs. Also, since newConfig is what we're writing back, wouldn't we lose all parameters that presently exist?

I think we should always push the parameter if I understand what this is doing. Since we then never use the variable found, can't parsedParam just be an alternative declaration with the function? The point is to start with either what we have, or a blank object.

Secondly, if we don't already have the file or it is missing a parameter, we never setup the new parameter number with your # key since we're no longer calling the object like we used to? Somewhere we need to make # = param.ParameterNumber if the file didn't already exist or it was missing the parameter.

Finally, if a bitmask exists we should skip importing the full parameter. This is an improvement over the original.

I pushed two commits that I think fixes these things, but please double check me. I'm especially shooting in the dark with the second commit attempting to look for matching bitmasks.

blhoward2 (changes requested):

PR #3152, opened 2021-08-04
fix(config): update Fibaro FGT001 for v8 changes, preserve root and endpoint 2

kvaster:
Fibaro FGT001 may have additional external temperature sensor and we must preserve additional endpoints explicitely according to documentation.

### DIFF
--- a/packages/config/config/devices/0x010f/fgt001.json
+++ b/packages/config/config/devices/0x010f/fgt001.json
@@ -119,13 +119,6 @@
 		}
 	},
 	"compat": {
-		"commandClasses": {
-			"remove": {
-				// Root endpoint duplicates endpoint 1, causing duplicated Battery CC values
-				"0x80": {
-					"endpoints": [0]
-				}
-			}
-		}
+		"preserveEndpoints": [2]
 	}
 }


AlCalzone:
Makes sense, but these compat flags may need a complete overhaul now in v8. Can you please share a log of a re-interview with the config file you propose here?

AlCalzone:
Just read the issue you opened. I think the desired change would be to preserve only endpoint 2, since basically nothing but the battery from endpoint 2 actually needs endpoints.

kvaster:
Am I right saying that it's better to use root insteadof endpoint 1 when endpoint 1 mirrors root ? I will change PR.

kvaster:
I've updated my PR.

This log is with first version of patch:
[zwavejs-1.log](https://github.com/zwave-js/node-zwave-js/files/6930374/zwavejs-1.log)

And this log is with current version of patch:
[zwavejs-2.log](https://github.com/zwave-js/node-zwave-js/files/6930376/zwavejs-2.log)

AlCalzone:
> Am I right saying that it's better to use root insteadof endpoint 1 when endpoint 1 mirrors root ?

There's no simple answer to this. Without endpoints, each Command Class (CC) can only exist once on a device. Endpoints are meant to give devices the possibility to have duplicate functionality, like two-channel switches which need the Binary Switch CC twice (once for each channel).

The main problem is that many devices exist which have unnecessary endpoints, for example a switch on endpoint 1 and a sensor on endpoint 2. Endpoint 0 (the root) MUST also provide access to the main functionality, which is usually endpoint 1, but sometimes also endpoint 2 or higher, so there is a duplication by design.
What makes it worse: many of these devices with unnecessary endpoints don't send their reports from the endpoints but from the root. This means we cannot know which endpoint the reports came from, so working as much as possible with only the root makes things easier and more reliable.

AlCalzone:
> And this log is with current version of patch:

Can you change something on the device, so I can see what it sends?

kvaster:
I've tried to manually change temp by rotating knob and it seems that all changes are sent on endpoint 1. Log attached.

[zwavejs.log](https://github.com/zwave-js/node-zwave-js/files/6931764/zwavejs.log)

But also there is one more 'problem' with first version of patch - there is 'Basic' class command with strange value inside.

kvaster:
Also it seems that I need to make more experiments, cause in my current setup all thermostats are attached to external sensor.

kvaster:
I have now more answers:

* Air temperature is always reported on endpoint 2 with or without external sensor connected.
* Manual changes of thermostat setpoints (via knob) are reported on endpoint 1 on some of by fgt001 devices and on endpoint 0 on others....

It seems that it will be better to use endpoints 1 & 2 and to map endpoint 0 events to endpoint 1. This will also require to remove battery class name from endpoint 0... But I'm not sure already... :(

kvaster:
Argh.... Setting such way (using endpoint 1 insteadof root) is also breaking thermostat setpoint :/

kvaster:
I've managed to work all properly!

I've cleared all associations for devices which were reporting on endpoint 1 and re-added associations on endpoint 0 and 2. Now all devices are working OK and the same way!

AlCalzone (approved):
Nice! Thanks for the work!

PR #3155, opened 2021-08-04
fix(config): add option 11 to ZEN17 parameters 2 & 3

EnochPrime:
Adds value 11 to parameters 2 & 3 which was added in firmwareVersion 1.4.

Source:
https://www.support.getzooz.com/kb/article/698-zen17-universal-relay-advanced-settings/

### DIFF
--- a/packages/config/config/devices/0x027a/zen17.json
+++ b/packages/config/config/devices/0x027a/zen17.json
@@ -363,94 +363,46 @@
 			"defaultValue": 0,
 			"unsigned": true
 		},
-		"10": [
-			{
-				"$if": "firmwareVersion >= 1.4",
-				"label": "Control Relay 1 with S1 Input",
-				"description": "If disabled and Parameter 2 is values 4-11, a Z-Wave report will be sent but R1 will not be triggered",
-				"valueSize": 1,
-				"minValue": 0,
-				"maxValue": 1,
-				"defaultValue": 1,
-				"unsigned": true,
-				"allowManualEntry": false,
-				"options": [
-					{
-						"label": "Disable",
-						"value": 0
-					},
-					{
-						"label": "Enable",
-						"value": 1
-					}
-				]
-			},
-			{
-				"$if": "firmwareVersion < 1.4",
-				"label": "Control Relay 1 with S1 Input",
-				"description": "If disabled and Parameter 2 is values 4-10, a Z-Wave report will be sent but R1 will not be triggered",
-				"valueSize": 1,
-				"minValue": 0,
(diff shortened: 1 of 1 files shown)


EnochPrime:
@zwave-js-bot help?

packages/config/config/devices/0x027a/zen17.json
```diff
@@ -235,46 +363,94 @@
...
-					"label": "Enable",
-					"value": 1
-				}
-			]
-		},
+		"10": [
+			{
+				"$if": "firmwareVersion >= 1.4",
+				"label": "Control Relay 1 with S1 Input",
+				"description": "If disabled and Parameter 2 is values 4-11, a Z-Wave report will be sent but R1 will not be triggered",
```
AlCalzone: Since 11 is the new maximum, I think it would be better to change the wording here instead of introducing a conditional parameter just for one character.
```suggestion
				"description": "If disabled and Parameter 2 is >= 4, a Z-Wave report will be sent but R1 will not be triggered",
```

packages/config/config/devices/0x027a/zen17.json
```diff
@@ -235,46 +363,94 @@
...
+						"value": 1
+					}
+				]
+			}
+		],
+		"11": [
+			{
+				"$if": "firmwareVersion >= 1.4",
+				"label": "Control Relay 2 with S2 Input",
+				"description": "If disabled and Parameter 3 is values 4-11, a Z-Wave report will be sent but R2 will not be triggered",
```
AlCalzone: dito

AlCalzone (changes requested):

packages/config/config/devices/0x027a/zen17.json
EnochPrime: Agreed, that is a much better approach.

packages/config/config/devices/0x027a/zen17.json
```diff
@@ -60,118 +60,246 @@
...
-					"label": "On/off report (dry contact switch/sensor)",
-					"value": 10
-				}
-			]
-		},
+		"2": [
+			{
+				"$if": "firmwareVersion >= 1.4",
+				"label": "Input Type for S1 Terminal",
+				"description": "Note: The device must be excluded (without resetting it) and re-included after changing to/from values 4-11",
```
EnochPrime, line 67: Should we also use >= 4 in the description here and then only have a conditional extra value? The only apparent downside is the max would need to equal 11 even for 1.3 and below.

AlCalzone (approved):

PR #2425, opened 2021-04-22
fix(config): change scenes sent config and on/off mode - fibaro walli dimmer

Micka41:
fixe : #2424

PR #3169, opened 2021-08-10
fix(config): update and cleanup Fibaro Walli Double Switch

popescunsergiu:
feat(config): Update WALLI SWITCH FIBARO SWITCH to support multiple scenes configuration

### DIFF
--- a/maintenance/schemas/device-config.json
+++ b/maintenance/schemas/device-config.json
@@ -344,7 +344,7 @@
 		"import": {
 			"type": "string",
 			"description": "Used to import another config file or parts thereof",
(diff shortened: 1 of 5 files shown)


AlCalzone (changes requested):
See my comment in the other open PR for this device:
https://github.com/zwave-js/node-zwave-js/pull/2425/files/f6c4e6ef83df3b2a44e1912200cdb95acf76a748#diff-0a059d3986d31489542c5570a41a5bc5ed72c640afa6671787c2758381ff1a85

These should really be partial parameters, so you can select the individual combinations one by one.

AlCalzone:
Thanks for the help - I'm going to take it from here. That entire file is a mess and needs to be reworked completely.

AlCalzone (approved):

PR #3175, opened 2021-08-11
chore: update README.md modify terminology and mention hardware compability

Hedda:
Following confusion and reasoning in discussion at https://github.com/zwave-js/node-zwave-js/discussions/3087

Modifying terminology to read "device driver" which I understand as an industry-standard terminology instead of "driver" as per wikipedia explanation of the two https://en.wikipedia.org/wiki/Device_driver verses https://en.wikipedia.org/wiki/Driver_(software)

Also mention that zwave-js is compatible with all known Z-Wave controller hardware adapters in order to somewhat further clarify that it is a device driver for those..

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
 Z-Wave is an amazing piece of technology, but it is difficult to get started with and develop for.  
 **It shouldn't be!** We've got you covered.
 
-Built on Node.js and compatible with all known Z-Wave 500 series + Z-Wave 700 series controller hardware adapters, **Z-Wave JS** runs on almost anything with a little bit of computing power and a serial port. Check out the [documentation](https://zwave-js.github.io/node-zwave-js) to get started!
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,7 +20,7 @@
 Z-Wave is an amazing piece of technology, but it is difficult to get started with and develop for.  
 **It shouldn't be!** We've got you covered.
 
-Built on Node.js, **Z-Wave JS** runs on almost anything with a little bit of computing power and a serial port. Check out the [documentation](https://zwave-js.github.io/node-zwave-js) to get started!
+Built on Node.js and compatible with all known Z-Wave controller hardware adapters, **Z-Wave JS** runs on almost anything with a little bit of computing power and a serial port. Check out the [documentation](https://zwave-js.github.io/node-zwave-js) to get started!
```
AlCalzone: I think we need to relativize a bit here. 500-series and Z-Wave 700 (yeah, they changed the name) controllers are supported - I won't make such promises for the 300 series, although they probably work too.

AlCalzone (changes requested):

README.md
Hedda: OK changed to "compatible with all known Z-Wave 500 series + Z-Wave 700 series controller hardware adapters"

AlCalzone (approved):

PR #3176, opened 2021-08-11
feat(config): add support for FGWCEU

popescunsergiu:


AlCalzone (changes requested):
Many parameters are similar to the ones in https://github.com/zwave-js/node-zwave-js/pull/3169 - please align their definitions so they match.

popescunsergiu:
Sorry Al, this was a broken PR, I'll create a new one for FGWCEU-201

PR #3179, opened 2021-08-11
feat(config): add support for HE-TPS05

popescunsergiu:


packages/config/config/devices/0x0344/he-tps05.json
```diff
@@ -0,0 +1,970 @@
...
+					"value": 5
+				},
+				{
+					"label": "Turn off, then on after timer",
+					"value": 6
+				}
+			]
+		},
+		"61": {
+			"label": "Relay 1: Button number source",
```
AlCalzone: Parameter labels should be `Title Case` - please do the same for the other parameters.

packages/config/config/devices/0x0344/he-tps05.json
```diff
@@ -0,0 +1,970 @@
...
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 5,
+			"defaultValue": 1,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Controlled by gateway or associated device",
```
AlCalzone:
```suggestion
					"label": "Gateway or associated device",
```

Thinking of, doesn't this just mean the relay is controlled via Z-Wave?

packages/config/config/devices/0x0344/he-tps05.json
```diff
@@ -0,0 +1,970 @@
...
+					"value": 0
+				},
+				{
+					"label": "Controlled by touch button N1 (Top Left)",
```
AlCalzone:
```suggestion
					"label": "Touch button N1 (top left)",
```
Please change the others in a similar fashion.

packages/config/config/devices/0x0344/he-tps05.json
```diff
@@ -0,0 +1,970 @@
...
+					"label": "Controlled by touch button N4 (Bottom Right)",
+					"value": 4
+				},
+				{
+					"label": "Controlled by touch button N5 (Center)",
+					"value": 5
+				}
+			]
+		},
+		"62": {
```
AlCalzone, line 671: For parameters 62-65, please apply the changes from 61

packages/config/config/devices/0x0344/he-tps05.json
```diff
@@ -0,0 +1,970 @@
...
+					"value": 0
+				},
+				{
+					"label": "Enable",
+					"value": 1
+				}
+			]
+		},
+		"143": {
+			"label": "Sensors Consecutive Reporting Interval",
```
AlCalzone: I think this is better
```suggestion
			"label": "All Sensors: Reporting Interval",
```

AlCalzone (changes requested):
A few changes please.

Side note: At this rate of new HELTUN devices, it might make sense to add a template for them. They seem to share a LOT of parameters. But that's something for another PR.

packages/config/config/devices/0x0344/he-tps05.json
```diff
@@ -0,0 +1,970 @@
...
+					"value": 9
+				}
+			]
+		},
+		"3": {
+			"label": "Hardware and Software Versions",
+			"valueSize": 4,
+			"minValue": 0,
+			"maxValue": 0,
+			"readOnly": true
```
blhoward2, line 72: None of these can do anything with a min of 0 and a max of 0 even if readonly. There are a couple of them.

packages/config/config/devices/0x0344/he-tps05.json
```diff
@@ -0,0 +1,970 @@
...
+			"label": "All Sensors: Reporting Interval",
+			"unit": "minutes",
+			"valueSize": 1,
+			"minValue": 1,
+			"maxValue": 120,
+			"defaultValue": 10
+		},
+		"144": {
+			"label": "Air Temperature Sensor Report Threshold",
+			"unit": "0.1 °C",
```
blhoward2, line 947: Can you confirm that this only accepts celsius?

blhoward2 (changes requested):

packages/config/config/devices/0x0344/he-tps05.json
popescunsergiu: According to their documentation, yes. it's available on their website

packages/config/config/devices/0x0344/he-tps05.json
popescunsergiu: Please advise on what to put there. Since it's readonly, min and max does not have any sense, though it's requested.

blhoward2: Even if it's read only there is still a range of values possible. Most manufacturers still list the min and max. 

@AlCalzone we still need to provide a range for the UIs, right? This isn't a candidate to remove for read only params?

AlCalzone: I've thought about it... Applications may use the information to show a gauge or something but sometimes we really don't know the range. If this isn't an error, we could leave it at 0/0 IMO.

popescunsergiu:
@AlCalzone  is there anything else I have to do here? thx

AlCalzone:
blhoward will take a look shortly, thanks for your patience

blhoward2 (dismissed):

AlCalzone:
I'm not sure what caused the review to be dismissed - I hope it was "accept"

AlCalzone (approved):

PR #3189, opened 2021-08-13
fix: avoid writing into node_modules when updating external config

### LINKED ISSUE #3064: Feat: Skip installing device config in source when ZWAVEJS_EXTERNAL_CONFIG is set.
The device configs should be installed into the folder specified in `ZWAVEJS_EXTERNAL_CONFIG` when set, and skip installing into the source (e.g `node_modules/@zwave-js/config`).

Applications running in restricted enviroments (RO filesystems) such as the snap package of zwavejs2mqtt do not allow writing to this destination. Previously we solve this by mounting `node_modules/@zwave-js/config` as a RW filesystem within the tree, but this sometimes caused issues on restarts, updates etc where config files got unavailable and every device showed up as «unknown». A new restart typically solved this.

I guess that the new "update method" has changed and introduced a lock file, which is to be stored at `lib/node_modules/zwavejs2mqtt/package.json.lock` which also is within the RO filesystem. 

Snaps has the ability to map layouts for such files, but snaps will create files that doesn't exists. Since node-zwave-js treats this as an "ongoing update" when the file exists, it will always fail. 

Suggests that the lock file and config updates is placed under the `ZWAVEJS_EXTERNAL_CONFIG` directory directly, and skipping writing to the source tree.

AlCalzone:
fixes: #3064

### DIFF
--- a/packages/config/src/ConfigManager.ts
+++ b/packages/config/src/ConfigManager.ts
@@ -129,20 +129,23 @@ export class ConfigManager {
 	private fulltextIndex: FulltextDeviceConfigIndex | undefined;
 	private notifications: NotificationMap | undefined;
 
-	private useExternalConfig: boolean = false;
+	private _useExternalConfig: boolean = false;
+	public get useExternalConfig(): boolean {
+		return this._useExternalConfig;
+	}
 
 	public async loadAll(): Promise<void> {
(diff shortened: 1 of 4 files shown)


AlCalzone:
@zwave-js-bot pack this

PR #3205, opened 2021-08-17
feat(config): update hank electronics devices

nmummau:
I have a HKZW-SCN04. I noticed when reading its documentation that there was a missing parameter in the Z-Wave JS integration of Home Assistant. When making that change I noticed inconsistencies in other config files for the same manufacturer. I fixed those too. I also found/fixed a misspelling of "triggerd". I also found/removed an unused manufacturer.


packages/config/config/manufacturers.json
```diff
@@ -206,7 +206,6 @@
 	"0x00cc": "Beijing Sino-American Boyi Software Development Co. Ltd",
 	"0x00cd": "Shanghai Longchuang Eco-energy Systems",
 	"0x00cf": "Forest Group Nederland B.V",
-	"0x00d0": "HANK Electronics Ltd",
```
AlCalzone: Are you 100% sure about this? There are other manufacturers with multiple manufacturer IDs (Aeotec/Aeon Labs)

packages/config/config/devices/0x0208/hkzw_ms01.json
```diff
@@ -1,7 +1,7 @@
 {
 	"manufacturer": "HANK Electronics Ltd.",
 	"manufacturerId": "0x0208",
-	"label": "HKZW_MS01",
+	"label": "HKZW MS01",
```
AlCalzone: In the other files you used a dash, here it is a space. Intended?

packages/config/config/devices/0x0208/hkzw-scn04.json
```diff
@@ -37,6 +37,25 @@
...
 				}
 			]
+		},
+		"255": {
+			"label": "Reset/Remove",
+			"description": "Reset the sensor or remove from the Z-Wave network.",
+			"valueSize": 4,
+			"minValue": 0,
+			"maxValue": 2147483647,
+			"defaultValue": 1,
```
AlCalzone, line 47: There should probably be an option "Idle" for the default value of 1

AlCalzone (changes requested):
A few minor things

packages/config/config/manufacturers.json
nmummau: I'm _not_ sure about it. I noticed that the manufacturer ID of '0x00d0' was unused in the project. I assumed that meant it was no longer needed. Could you help me understand where the source of these manufacturer IDs come from? ie. How do I confirm that HANK does/doesn't have multiple manufacturer IDs?

packages/config/config/devices/0x0208/hkzw_ms01.json
nmummau: You're correct. This should have a dash for consistency! I will fix  this.

packages/config/config/devices/0x0208/hkzw-scn04.json
nmummau: I mimicked one of the other similar devices from this same manufacturer, where the same '255' parameter is used. I will confirm if the "Idle" is missing in this and the one I copied from.

packages/config/config/manufacturers.json
AlCalzone: https://z-wavealliance.org/z-wave-specifications/ -> Registries -> Assigned Manufacturer IDs

I've used these to update our json file over the years and never deleted something unless there is a new manufacturer with the same ID. It might just be that HANK used to have `0x00d0` and there might be devices out there that use this ID.

Honestly, I prefer having one too many entries here over missing one.

blhoward2 (dismissed):
LGTM

AlCalzone (approved):

nmummau:
@AlCalzone Thanks for the nits commit. Those strings were written verbatim from the manual of the device. But I'm fine with your changes.

blhoward2:
Manufacturers have terrible labels and such. We strive for much better.

nmummau:
I agree, since this was my first PR for the repo I figured I'd tread lightly on which strings I improved. Moving forward I'll do better. I appreciate what I have observed to be a high standard of detail with the contributors of this repository. It's great.

PR #3218, opened 2021-08-23
feat: add new ConfigManager properties to cover more config maps

raman325:
This is a follow up to the questions I was asking Al in Discord - basically taking any config data that is not already exposed as properties and exposing it as new properties


AlCalzone (approved):
Aside from the fact that applications probably don't want to deal with the mess that the notification defintions are - LGTM

PR #3251, opened 2021-08-27
feat: add context object to log messages

raman325:
I'm not sure that everything I've captured is strictly necessary so we can relax some of this based on feedback


packages/zwave-js/src/lib/log/Driver.ts
```diff
@@ -53,6 +54,7 @@ export class DriverLogger extends ZWaveLoggerBase {
 			level: actualLevel,
 			message,
 			direction: getDirectionPrefix("none"),
+			context: CONTEXT,
```
AlCalzone:
```suggestion
			context: { ...CONTEXT },
```
would avoid a situation where applications can change our `CONTEXT` object's contents.

packages/zwave-js/src/lib/log/Driver.ts
```diff
@@ -219,6 +222,7 @@ export class DriverLogger extends ZWaveLoggerBase {
 				queue.length === 1 ? "" : "s"
 			})`,
 			direction: getDirectionPrefix("none"),
+			context: CONTEXT,
```
AlCalzone: dito

AlCalzone (changes requested):
I'm wondering if `context` should have a `type` property or something that can be used to distinguish between several possible contexts an application might encounter.

Also, for type safety, we should probably define the different specific context types and export them.

raman325:
> I'm wondering if `context` should have a `type` property or something that can be used to distinguish between several possible contexts an application might encounter.
> 
> Also, for type safety, we should probably define the different specific context types and export them.

That's `source` the way I have currently implemented it, but we can switch to `type` since you use that elsewhere. And stronger typing is a good idea

raman325:
@AlCalzone when you say you want to type the different context types, do you mean for example typing https://github.com/zwave-js/node-zwave-js/pull/3251/files#diff-7076042951692d76ef597b7bae4634cb8110f9a95eb8708065fdf41cd0a4c919R98-R103 and https://github.com/zwave-js/node-zwave-js/pull/3251/files#diff-7076042951692d76ef597b7bae4634cb8110f9a95eb8708065fdf41cd0a4c919R144-R151 as distinct types?

AlCalzone:
> That's `source` the way I have currently implemented it

Yes, but it doesn't distinguish between the possible contexts, e.g. there are multiple `source: "controller"` contexts.


> @AlCalzone when you say you want to type the different context types, do you mean for example typing [#3251 (files)](https://github.com/zwave-js/node-zwave-js/pull/3251/files#diff-7076042951692d76ef597b7bae4634cb8110f9a95eb8708065fdf41cd0a4c919R98-R103) and [#3251 (files)](https://github.com/zwave-js/node-zwave-js/pull/3251/files#diff-7076042951692d76ef597b7bae4634cb8110f9a95eb8708065fdf41cd0a4c919R144-R151) as distinct types?

Yep

raman325:
@zwave-js-bot fix lint

raman325:
@zwave-js-bot fix lint

AlCalzone:
@raman325 let me know when you're done here :)

raman325:
I'm done, this is ready for review!

AlCalzone:
Sorry it took me so long. I've made a few changes:
* exported all log context types
* converted some of the types to generics to let TypeScript validate the contexts better
* made the `type` property to distinguish logs from the same logger optional
* removed the limitation of `source`, so applications which hook into the logging functionality (like Node Red) can still do it.

AlCalzone (approved):
OK with my changes - @raman325 can you give this another test?

raman325:
> OK with my changes - @raman325 can you give this another test?

I will test this in the next day or two and get back to you

AlCalzone:
Huh?

raman325:
uhh this was not intentional. Shit

raman325:
sorry for the delay @AlCalzone but I tested the changes and everything looks good! I think we are OK to merge

Issue #3266, opened 2021-09-02
Trane XR524 setting incorrect paremeters


tmjpugh:
### Is your problem within Home Assistant (Core or Z-Wave JS Integration)?

NO, my problem is NOT within Home Assistant or the ZWave JS integration

### Is your problem within ZWaveJS2MQTT?

NO, my problem is NOT within ZWaveJS2MQTT

### Checklist

- [X] I have checked the [troubleshooting section](https://zwave-js.github.io/node-zwave-js/#/troubleshooting/index) and my problem is **not** described there.

- [X] I have read the [changelog](https://github.com/zwave-js/node-zwave-js/blob/master/CHANGELOG.md) and my problem was **not** mentioned there.

### Describe the bug

What causes the bug?
- Configuration Parameters use illegal values
- Configuration Parameters do not match device settings
- When Configuration Parameters are changed, they revert back to illegal value

What do you observe?
device>>configuration>>112-0-1
Options in zwavejs are

    0 Gas/Electric (default)
    1 HeatPump

option in Trane 524 menu are

    none
    AC
    HP

if i select AC on Trane, [1]HeatPump is selected in zwaveJS
if i select [1]HeatPump in zwaveJS, AC is slected on Trane

ALSO
112-0-18 Cooling Delta Stage 1 ON equal 18
I change value to 3 and receive message in log that show value is change From 3 to 18
value is change back from 3 to 18 in zwavejs UI as well
8 is maxValue for this setting

Refresh does not change these result and appears to cause device setting on device to become 0 for value that exceed maxValue

What did you expect to happen?
I expect zwavejs value to change to similar option on device Gas/Electric to AC or Heatpump to HP
I expect value change in zwaveJS not to revert OR not to allow change to illegal value

  1. Go to 'node'
  2. Click on 'node >> Configuration's
  3. Scroll down to 'parameter 18'
  4. change data for parameter and watch it revert to 18, this will occur for many parameter

### Device information

Manufacturer: Trane
Model name: XR524
Version: 1.38 and 1.14
Node ID in your network: 93 and 58

### How are you using `node-zwave-js`?

- [ ] `zwavejs2mqtt` Docker image (latest)
- [ ] `zwavejs2mqtt` Docker image (dev)
- [ ] `zwavejs2mqtt` Docker manually built _(please specify branches)_
- [ ] `ioBroker.zwave2` adapter _(please specify version)_
- [ ] `HomeAssistant zwave_js` integration  _(please specify version)_
[...]

PR #3277, opened 2021-09-04
fix(config): correct Trane parameters

blhoward2:
Will fix part of #3266


blhoward2:
@AlCalzone Note the discussion in the issue. This device has been reverse engineered and the manufacturer doesn't say what values exist. Our current ranges are off according to his log, so I'm setting the max range permitted by the value size until someone can provide better information.

blhoward2:
@zwave-js-bot rename commit

AlCalzone (approved):
Fine with me - should we mark this as fixing the issue or is there something else?

xannor:
I want to suggest a change. I dont know if is something unique to my particular x542 or if it is a re-occurring oversite that was never corrected but parameter 76 is not the fan off timer but a toggle for run/hold on the schedule (0 = Hold, 1 = Run) and 77 is not the 2nd stage heat but a toggle for the ESM [economy mode],  (0 = Off, 2 = On) This has been "published" for some time (original reference here: https://community.home-assistant.io/t/trane-xr524-with-control-of-the-schedule-run-hold-esm-modes/94461) and I have mentioned this to the older python zwave project, but never been able to submit a change. Since this is already making changes, (and it seems you have the means to verify them) could those two parameters be corrected for this model as part of this?

blhoward2:
I can make that change. No way around it, this is going to be a community sourced file. I don't have one of these to try to test it so you all are going to have to report changes.

blhoward2:
@AlCalzone I wasn't sure if you needed more on the scales. I have one more change to make.

AlCalzone (changes requested):
Ok, marking this as not ready yet until you're done, so I don't accidentally merge

packages/config/config/devices/0x008b/trane_xr524.json
```diff
@@ -235,20 +235,30 @@
...
 		},
 		"76": {
-			"label": "Fan OFF Time",
-			"unit": "minutes",
+			"label": "Run/Hold Schedule",
 			"valueSize": 1,
 			"minValue": 10,
 			"maxValue": 120,
-			"defaultValue": 10
+			"defaultValue": 10,
```
xannor: I think this should be 1 not 10.

blhoward2: Which? The min value or the default?

xannor: I meant the default value but now that you mention it, since 0 and 1 are the only valid values, that propbably should be the min and max as well, unless they are not needed when there are option choices. Not overly famillar with the config format.

AlCalzone: Yeah, there's something off about this parameter definition

AlCalzone:
@blhoward2 is this ready?

blhoward2:
No, someone reported other problems in the issue.

AlCalzone:
#3100 was just merged, so I've gone ahead and updated the PR and your file to match the new config format

AlCalzone (approved):

AlCalzone:
@blhoward2 the merged PR should solve the remaining issues I think. Unless you have anything to add, I'm going to merge this.

PR #3312, opened 2021-09-09
fix(config): add product id 0x0103 to Aeotec ZW141

crxporter:
I've updated my nanoshutter to the newest firmware. Using custom config files I've determined this config file works for the device (used to be mfg 0x0086) after getting the updated firmware.

### DIFF
--- a/packages/config/config/devices/0x0371/zw141.json
+++ b/packages/config/config/devices/0x0371/zw141.json
@@ -8,6 +8,11 @@
 			"productType": "0x0003",
 			"productId": "0x008d",
 			"zwaveAllianceId": [3693, 3742]
+		},
+		{
+			"productType": "0x0103",
+			"productId": "0x008d",
+			"zwaveAllianceId": [3693, 3742]
 		}
 	],
 	"firmwareVersion": {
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0371/zw141.json
```diff
@@ -8,6 +8,11 @@
 			"productType": "0x0003",
 			"productId": "0x008d",
 			"zwaveAllianceId": [3693, 3742]
+		},
+		{
+			"productType": "0x0103",
+			"productId": "0x008d",
+			"zwaveAllianceId": [3693, 3742]
```
AlCalzone:
```suggestion
			"productId": "0x008d"
```
The product pages https://products.z-wavealliance.org/products/3693 and https://products.z-wavealliance.org/products/3742 both belong to the other ID

AlCalzone (changes requested):

packages/config/config/devices/0x0371/zw141.json
crxporter: Awesome. Good catch.

AlCalzone (approved):

PR #3313, opened 2021-09-09
feat(cc): add setting `dimmingDuration` for `Scene Controller Configuration CC` with `setValue` API

### LINKED ISSUE #3309: [bug] Unable to set Associated Scene ID option in Scene Controller Configuration via Control Panel
### Version

**Checklist:**

Build/Run method

- [ ] Docker
- [ ] PKG
- [ ] Snap package
- [x] Manually built (git clone - yarn install - yarn run build )

zwavejs2mqtt: 5.5.1.a937e5e
zwave-js: 8.2.1

### Describe the bug
Attempts to Set the SceneID on a Leviton VRCS4 scene controller appear to work through the interface but the stored value on the device is never actually updated.  The driver returns the error: 
`Writing 6 to 33-45-0-sceneId-1 2021-08-27T18:09:40.079Z DRIVER Unexpected error during SerialAPI command: ZWaveError: Set commands don't support unknown durationsat Duration.serializeSet`

Along a similar vein, attempts to set dimmingDuration also fail on this same page with error in debug interface:
`Driver: Scene Controller Configuration: "dimmingDuration" is not a supported property`

### To Reproduce

1. Go to Control Panel.
2. Expand VRCS4 node details.
3. Scroll down and click to expand options for Scene Controller Configuration
4. Enter value in Associated Scene ID and press Arrow to send command.
5. Successfully call api writeValue is briefly displayed as notification.
6. Refresh page and observe that the Scene ID value is not actually updated.

### Expected behavior

sceneID value is updated to match sane numeric value entered in Scene Controller Configuration Associated Scene ID field.

### Additional context

[zwavejs_2021-08-27.log](https://github.com/zwave-js/zwavejs2mqtt/files/7068809/zwavejs_2021-08-27.log)
[zwavejs2mqtt.log](https://github.com/zwave-js/zwavejs2mqtt/files/7068811/zwavejs2mqtt.log)

Ikcelaks:
This pull request should fix #3309 by adding support for setting the `dimmingDuration` for a group's scene configuration with the `SET_VALUE` API.

One thing I'm not 100% certain on is if the client (like zwavejs2mqtt) will send the `value` with type `number` corresponding to the z-wave encoding of a `Duration`.

I'm not really able to test this code, because I don't have any devices that support the Scene Controller Configuration CC.

Once this pull request goes through, I can do the equivalent changes for Scene Actuator Configuration CC.

### DIFF
--- a/packages/core/src/values/Duration.ts
+++ b/packages/core/src/values/Duration.ts
@@ -95,26 +95,6 @@ export class Duration {
 		}
 	}
 
-	/**
-	 * Guards that an arbitrary object is a valid Duration
-	 */
-	public static isDuration(input: any): input is Duration {
-		if (typeof input.value !== "number" || typeof input.unit !== "string") {
-			return false;
-		}
-		switch (input.unit) {
-			case "minutes":
-				return input.value >= 1 && input.value <= 127;
-			case "seconds":
-				return input.value <= 127;
-			case "unknown":
-			case "default":
-				return input.value === 0;
-			default:
-				return false;
-		}
-	}
-
 	/** Serializes a duration for a Set command */
 	public serializeSet(): number {
 		if (this.unit === "default") return 0xff;
--- a/packages/core/src/values/Duration.ts
+++ b/packages/core/src/values/Duration.ts
@@ -95,6 +95,26 @@ export class Duration {
 		}
 	}
 
+	/**
+	 * Guards that an arbitrary object is a valid Duration
+	 */
+	public static isDuration(input: any): input is Duration {
+		if (typeof input.value !== "number" || typeof input.unit !== "string") {
+			return false;
+		}
+		switch (input.unit) {
+			case "minutes":
+				return input.value >= 1 && input.value <= 127;
+			case "seconds":
+				return input.value <= 127;
+			case "unknown":
+			case "default":
+				return input.value === 0;
+			default:
+				return false;
+		}
+	}
+
 	/** Serializes a duration for a Set command */
 	public serializeSet(): number {
 		if (this.unit === "default") return 0xff;
--- a/packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
@@ -1,6 +1,7 @@
 import {
 	CommandClasses,
 	Duration,
+	getCCName,
 	Maybe,
 	MessageOrCCLogEntry,
 	validatePayload,
@@ -149,7 +150,7 @@ export class SceneControllerConfigurationCCAPI extends CCAPI {
 				await this.set(propertyKey, value, dimmingDuration);
 			}
 		} else if (property === "dimmingDuration") {
-			if (typeof value !== "string" && !Duration.isDuration(value)) {
+			if (typeof value !== "string" && !(value instanceof Duration)) {
 				throwWrongValueType(
 					this.ccId,
 					property,
@@ -161,7 +162,11 @@ export class SceneControllerConfigurationCCAPI extends CCAPI {
 			const dimmingDuration = Duration.from(value);
 			if (dimmingDuration == undefined) {
 				throw new ZWaveError(
-					`${CommandClasses["Scene Controller Configuration"]}: "${property}" could not be set. Value is not a valid duration.`,
+					`${getCCName(
+						this.ccId,
+					)}: "${property}" could not be set. ${JSON.stringify(
+						value,
+					)} is not a valid duration.`,
 					ZWaveErrorCodes.Argument_Invalid,
 				);
 			}
@@ -419,7 +424,7 @@ export class SceneControllerConfigurationCCSet extends SceneControllerConfigurat
 			const groupCount = this.getGroupCountCached();
 			this.groupId = options.groupId;
 			this.sceneId = options.sceneId;
-			// if dimmingDuration was missing, default to 0xff
+			// if dimmingDuration was missing, use default duration.
 			this.dimmingDuration =
 				options.dimmingDuration ?? new Duration(0, "default");
 
--- a/packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
@@ -178,14 +178,6 @@ export class SceneControllerConfigurationCCAPI extends CCAPI {
 					propertyKey,
 				);
 				const valueDB = node.valueDB;
-
-				if (!valueDB.hasMetadata(dimmingDurationValueId)) {
-					valueDB.setMetadata(dimmingDurationValueId, {
-						...ValueMetadata.Duration,
-						label: `Dimming duration (${propertyKey})`,
-					});
-				}
-
 				valueDB.setValue(dimmingDurationValueId, dimmingDuration);
 				return;
 			}
--- a/packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
@@ -149,7 +149,7 @@ export class SceneControllerConfigurationCCAPI extends CCAPI {
 				await this.set(propertyKey, value, dimmingDuration);
 			}
 		} else if (property === "dimmingDuration") {
-			if (typeof value !== "number") {
+			if (typeof value !== "string" && !Duration.isDuration(value)) {
 				throwWrongValueType(
 					this.ccId,
 					property,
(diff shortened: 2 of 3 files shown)


packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -148,9 +148,27 @@ export class SceneControllerConfigurationCCAPI extends CCAPI {
...
+			if (typeof value !== "number") {
+				throwWrongValueType(
+					this.ccId,
+					property,
+					"duration",
+					typeof value,
+				);
+			}
+
+			const dimmingDuration = Duration.parseSet(value);
```
Ikcelaks: Morally, I feel like we should throw some kind of error if `parseSet` is unable to parse the `value`, but I don't see an appropriate error to `throw`. I was looking for something like `throwInvalidValue`, but didn't see anything. So, instead I'm just failing through to setting the `dimmingDuration` to `0x00`.

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -386,7 +404,9 @@ export class SceneControllerConfigurationCCSet extends SceneControllerConfigurat
 			const groupCount = this.getGroupCountCached();
 			this.groupId = options.groupId;
 			this.sceneId = options.sceneId;
-			this.dimmingDuration = options.dimmingDuration;
+			// if dimmingDuration was missing, default to 0 seconds
+			this.dimmingDuration =
+				options.dimmingDuration ?? new Duration(0, "seconds");
```
AlCalzone: Shouldn't we use a `"default"` duration in this case?

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
AlCalzone: IMO we should use `Duration.from` here:
https://github.com/zwave-js/node-zwave-js/blob/17ae97650e93df81e606f47bd5b24fd993c3635b/packages/zwave-js/src/lib/commandclass/BinarySwitchCC.ts#L128
If the result is `undefined`, i.e. could not be parsed, you can do this
```ts
	throw new ZWaveError(
		`${CommandClasses[cc]}: "${property}" could not be set. ${value} is not a valid duration.`,
		ZWaveErrorCodes.Argument_Invalid,
	);
```

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -148,9 +148,27 @@ export class SceneControllerConfigurationCCAPI extends CCAPI {
...
+			const dimmingDuration = Duration.parseSet(value);
+
+			const node = this.endpoint.getNodeUnsafe()!;
+			// If SceneId missing, we need to set a default to 0,
+			// which disables the scene configuration for the association group
+			const sceneId =
+				node.getValue<number>(
+					getSceneIdValueID(this.endpoint.index, propertyKey),
+				) ?? 0;
+			await this.set(propertyKey, sceneId, dimmingDuration);
```
AlCalzone, line 198: I'm not sure if we should send a command here if the scene ID isn't set yet. If a user sets `dimmingDuration`, followed by `scene ID`, I think we could just save the duration value ID locally until it is used in the next command.

What do you think?

AlCalzone (changes requested):

AlCalzone:
One more thing:
`SceneActivationCC` defines a value change option `transitionDuration` on the scene ID value:
https://github.com/zwave-js/node-zwave-js/blob/ba43549005b4e08ef1a145c394cd88766b6091b4/packages/zwave-js/src/lib/commandclass/SceneActivationCC.ts#L135-L149
and uses it in the `setValue` call:
https://github.com/zwave-js/node-zwave-js/blob/ba43549005b4e08ef1a145c394cd88766b6091b4/packages/zwave-js/src/lib/commandclass/SceneActivationCC.ts#L64-L77

This same option is used in multiple locations and allows you to set a value together with a corresponding duration in a single call. HA uses that for dimming for example. We should probably use that too and fall back to the duration value when changing the scene ID.

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
Ikcelaks: Yes, setting it to `0xFF` makes more sense, so `Duration(0, "default")`

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
Ikcelaks: Okay, now I see why `Duration.from` should be used here. I was thinking that the clients would send a `number` representing the encoding of the duration, but it looks like Home-Assistant would send a `string` and zwavejs2mqtt would send a `Duration` object.

However, using `Duration.from` is awkward here, because it requires that `typeof value` is `string | Duration | undefined`, which means that I need to coerce `value` into a `Duration` if it isn't a `string`.

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
Ikcelaks: Looking closer at the specs, `dimmingDuration` SHOULD be `0` if `sceneId` is `0`, so my code here breaks the recommendations.

The two valid options are your suggestion of saving the value locally or throwing an error. I think throwing an error would surprise users who arbitrarily set the values in the wrong order, so I'll store it in the `valueDB` where it can get picked up by a subsequent `SET_VALUE` call that sets the `sceneId` property.

Ikcelaks:
I thought about using `valueChangeOptions` here, but I wasn't sure if it was appropriate.

Unlike `SceneActivationCC` and `MultilevelSwitchCC`, the `dimmingDuration` in `SceneControllerConfigurationCC` (and `SceneActuatorConfigurationCC`) is a real persistent value. Of course, that doesn't mean we can't use `valueChangeOptions` here as a loophole to allow both properties to be set simultaneously.

AlCalzone:
> I thought about using valueChangeOptions here, but I wasn't sure if it was appropriate.

It feels a bit off, I agree.

Regarding your other comment:
> However, using Duration.from is awkward here, because it requires that typeof value is string | Duration | undefined, which means that I need to coerce value into a Duration if it isn't a string.

I don't see the problem with that. if `typeof value !== "string" && !(value instanceof Duration)`, then `throwWrongValueType`. This should take care of the type issue so there should be no need for a type assertion.

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -149,7 +149,7 @@ export class SceneControllerConfigurationCCAPI extends CCAPI {
 				await this.set(propertyKey, value, dimmingDuration);
 			}
 		} else if (property === "dimmingDuration") {
-			if (typeof value !== "number") {
+			if (typeof value !== "string" && !Duration.isDuration(value)) {
```
AlCalzone:
```suggestion
			if (typeof value !== "string" && !(value instanceof Duration)) {
```
We currently don't accept arbitrary objects that look like a Duration. If we were to change that, it needs to be part of another PR that changes all locations. The `Duration.isDuration` method itself is also not necessary here.

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -158,15 +158,30 @@ export class SceneControllerConfigurationCCAPI extends CCAPI {
 				);
 			}
 
-			const dimmingDuration = Duration.parseSet(value);
+			const dimmingDuration = Duration.from(value);
+			if (dimmingDuration == undefined) {
+				throw new ZWaveError(
+					`${CommandClasses["Scene Controller Configuration"]}: "${property}" could not be set. Value is not a valid duration.`,
```
AlCalzone:
```suggestion
					`${getCCName(this.ccId)}: "${property}" could not be set. ${JSON.stringify(value)} is not a valid duration.`,
```

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -404,9 +419,9 @@ export class SceneControllerConfigurationCCSet extends SceneControllerConfigurat
 			const groupCount = this.getGroupCountCached();
 			this.groupId = options.groupId;
 			this.sceneId = options.sceneId;
-			// if dimmingDuration was missing, default to 0 seconds
+			// if dimmingDuration was missing, default to 0xff
```
AlCalzone:
```suggestion
			// if dimmingDuration was missing, use the device's default duration
```

packages/core/src/values/Duration.ts
```diff
@@ -95,6 +95,26 @@ export class Duration {
...
+			case "seconds":
+				return input.value <= 127;
+			case "unknown":
+			case "default":
+				return input.value === 0;
+			default:
+				return false;
+		}
+	}
+
```
AlCalzone: not necessary - see other comment

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/SceneControllerConfigurationCC.ts
```diff
@@ -84,6 +85,7 @@ function persistSceneConfig(
 		valueDB.setMetadata(sceneIdValueId, {
 			...ValueMetadata.UInt8,
 			label: `Associated Scene ID (${groupId})`,
+			valueChangeOptions: ["transitionDuration"],
```
Ikcelaks, line 88: I think that this is insufficient, because nodes that have already been interviewed prior to this change will not get updated Metadata. What else should be done for this?

AlCalzone: ATM we can't do much else than to tell users to re-interview.

AlCalzone (approved):

AlCalzone:
Thanks!

PR #3322, opened 2021-09-11
feat: add setValue options for Scene * Config CCs

raman325:


raman325:
this will get included in @Ikcelaks 's PR: https://github.com/Ikcelaks/node-zwave-js/pull/1 and https://github.com/zwave-js/node-zwave-js/pull/3313

PR #3325, opened 2021-09-12
feat(cc): add setting `dimmingDuration` for `Scene Actuator Configuration CC` with `setValue` API

Ikcelaks:
This is basically the same as #3313, except for Scene Actuator Configuration CC.


AlCalzone (approved):
LGTM, thanks!

PR #3332, opened 2021-09-13
feat(cc): support Wake Up On Demand

### LINKED ISSUE #1158: Support Wake Up CC V3 and Supervision CC V2
This combination adds support for requesting a node to wake up.
https://github.com/zwave-js/node-zwave-js/blob/master/specs/SDS13782%20Z-Wave%20Management%20Command%20Class%20Specification.pdf

Implementation is more complicated and might involve the driver and message flow

Ikcelaks:
This PR is the first step in resolving #1158.

This adds support for Wake Up On Demand.

`SupervisionCC` support is upgraded to `V2`.
`WakeUpCC` support is upgraded to `V3`.
The `Driver` and `Node` classes are updated to set the Wake Up On Demand bit when sending `SupervisionCCReport` replies to sleeping nodes that have important messages pending in the queue.

A Wake Up On Demand for a node is requested by setting `options.requestWakeUpOnDemand = true` in the `SendCommandOptions` when calling `driver.sendCommand`.

The code is currently untested, because I'd like feedback that this is a good way to implement this functionality before spending too much time on it.

fixes: #1158

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/SupervisionCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SupervisionCC.ts
@@ -187,7 +187,7 @@ export class SupervisionCC extends CommandClass {
 
 export type SupervisionCCReportOptions = {
 	moreUpdatesFollow: boolean;
-	requestWakeUp?: boolean;
+	requestWakeUpOnDemand?: boolean;
 	sessionId: number;
 } & (
 	| {
@@ -224,7 +224,7 @@ export class SupervisionCCReport extends SupervisionCC {
 			this.duration = Duration.parseReport(this.payload[2]);
 		} else {
 			this.moreUpdatesFollow = options.moreUpdatesFollow;
-			this.requestWakeUp = options.requestWakeUp ?? false;
+			this.requestWakeUp = options.requestWakeUpOnDemand ?? false;
 			this.sessionId = options.sessionId;
 			this.status = options.status;
 			if (options.status === SupervisionStatus.Working) {
--- a/packages/zwave-js/src/lib/commandclass/SupervisionCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SupervisionCC.ts
@@ -88,10 +88,17 @@ export class SupervisionCCAPI extends PhysicalCCAPI {
 	public async sendReport(
 		options: SupervisionCCReportOptions & { secure?: boolean },
 	): Promise<void> {
+		// Ask the node if it needs Wake Up On Demand,
+		// unless it's overriden in `options`
+		const requestWakeUpOnDemand =
+			options.requestWakeUpOnDemand ??
+			this.endpoint.getNodeUnsafe()?.shouldRequestWakeUpOnDemand;
+
 		const { secure = false, ...cmdOptions } = options;
 		const cc = new SupervisionCCReport(this.driver, {
 			nodeId: this.endpoint.nodeId,
 			...cmdOptions,
+			requestWakeUpOnDemand,
 		});
 
 		// The report should be sent back with security if the received command was secure
@@ -217,14 +224,13 @@ export class SupervisionCCReport extends SupervisionCC {
 		if (gotDeserializationOptions(options)) {
 			validatePayload(this.payload.length >= 3);
 			this.moreUpdatesFollow = !!(this.payload[0] & 0b1_0_000000);
-			// Ignore requestWakeUpOnDemand bit when receiving Report
-			this.requestWakeUpOnDemand = false;
+			this.requestWakeUpOnDemand = !!(this.payload[0] & 0b0_1_000000);
 			this.sessionId = this.payload[0] & 0b111111;
 			this.status = this.payload[1];
 			this.duration = Duration.parseReport(this.payload[2]);
 		} else {
 			this.moreUpdatesFollow = options.moreUpdatesFollow;
-			this.requestWakeUpOnDemand = options.requestWakeUpOnDemand ?? false;
+			this.requestWakeUpOnDemand = !!options.requestWakeUpOnDemand;
 			this.sessionId = options.sessionId;
 			this.status = options.status;
 			if (options.status === SupervisionStatus.Working) {
--- a/packages/zwave-js/src/lib/commandclass/SupervisionCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SupervisionCC.ts
@@ -217,14 +217,14 @@ export class SupervisionCCReport extends SupervisionCC {
 		if (gotDeserializationOptions(options)) {
 			validatePayload(this.payload.length >= 3);
 			this.moreUpdatesFollow = !!(this.payload[0] & 0b1_0_000000);
-			// Ignore requestWakeUp bit when receiving Report
-			this.requestWakeUp = false;
+			// Ignore requestWakeUpOnDemand bit when receiving Report
+			this.requestWakeUpOnDemand = false;
 			this.sessionId = this.payload[0] & 0b111111;
 			this.status = this.payload[1];
 			this.duration = Duration.parseReport(this.payload[2]);
 		} else {
 			this.moreUpdatesFollow = options.moreUpdatesFollow;
-			this.requestWakeUp = options.requestWakeUpOnDemand ?? false;
+			this.requestWakeUpOnDemand = options.requestWakeUpOnDemand ?? false;
 			this.sessionId = options.sessionId;
 			this.status = options.status;
 			if (options.status === SupervisionStatus.Working) {
@@ -236,7 +236,7 @@ export class SupervisionCCReport extends SupervisionCC {
 	}
 
 	public readonly moreUpdatesFollow: boolean;
-	public readonly requestWakeUp: boolean;
+	public readonly requestWakeUpOnDemand: boolean;
 	public readonly sessionId: number;
 	public readonly status: SupervisionStatus;
 	public readonly duration: Duration | undefined;
@@ -245,7 +245,7 @@ export class SupervisionCCReport extends SupervisionCC {
 		this.payload = Buffer.concat([
 			Buffer.from([
 				(this.moreUpdatesFollow ? 0b1_0_000000 : 0) |
-					(this.requestWakeUp ? 0b0_1_000000 : 0) |
+					(this.requestWakeUpOnDemand ? 0b0_1_000000 : 0) |
 					(this.sessionId & 0b111111),
 				this.status,
 			]),
--- a/packages/zwave-js/src/lib/commandclass/SupervisionCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/SupervisionCC.ts
@@ -130,7 +130,7 @@ export class SupervisionCCAPI extends PhysicalCCAPI {
 }
 
 @commandClass(CommandClasses.Supervision)
-@implementedVersion(1)
+@implementedVersion(2)
 export class SupervisionCC extends CommandClass {
 	declare ccCommand: SupervisionCommand;
 	// Force singlecast for the supervision CC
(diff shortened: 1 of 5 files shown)


packages/zwave-js/src/lib/commandclass/SupervisionCC.ts
```diff
@@ -216,11 +217,14 @@ export class SupervisionCCReport extends SupervisionCC {
 		if (gotDeserializationOptions(options)) {
 			validatePayload(this.payload.length >= 3);
 			this.moreUpdatesFollow = !!(this.payload[0] & 0b1_0_000000);
+			// Ignore requestWakeUpOnDemand bit when receiving Report
+			this.requestWakeUpOnDemand = false;
```
AlCalzone: Wouldn't hurt to parse it, even if we're not doing anything with it.

packages/zwave-js/src/lib/commandclass/SupervisionCC.ts
```diff
@@ -216,11 +217,14 @@ export class SupervisionCCReport extends SupervisionCC {
...
 			this.sessionId = this.payload[0] & 0b111111;
 			this.status = this.payload[1];
 			this.duration = Duration.parseReport(this.payload[2]);
 		} else {
 			this.moreUpdatesFollow = options.moreUpdatesFollow;
+			this.requestWakeUpOnDemand = options.requestWakeUpOnDemand ?? false;
```
AlCalzone:
```suggestion
			this.requestWakeUpOnDemand = !!options.requestWakeUpOnDemand;
```

packages/zwave-js/src/lib/commandclass/WakeUpCC.ts
```diff
@@ -237,7 +245,8 @@ export class WakeUpCC extends CommandClass {
 default wakeup interval: ${wakeupCaps.defaultWakeUpInterval} seconds
 minimum wakeup interval: ${wakeupCaps.minWakeUpInterval} seconds
 maximum wakeup interval: ${wakeupCaps.maxWakeUpInterval} seconds
-wakeup interval steps:   ${wakeupCaps.wakeUpIntervalSteps} seconds`;
+wakeup interval steps:   ${wakeupCaps.wakeUpIntervalSteps} seconds
+wake up on demand supported: ${wakeupCaps.wakeUpOnDemandSupported}`;
```
AlCalzone:
```suggestion
wakeup on demand supported: ${wakeupCaps.wakeUpOnDemandSupported}`;
```
for consistency. Or make it `wake up` everywhere else :)

packages/zwave-js/src/lib/commandclass/WakeUpCC.ts
```diff
@@ -455,6 +474,16 @@ export class WakeUpCCIntervalCapabilitiesReport extends WakeUpCC {
 		return this._wakeUpIntervalSteps;
 	}
 
+	private _wakeUpOnDemandSupported: boolean;
+	@ccValue({ minVersion: 3 })
```
AlCalzone, line 478: I think this should be an internal value ID

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -335,6 +335,8 @@ export interface SendMessageOptions {
 	 * @internal
 	 */
 	tag?: any;
+	/** If a Wake Up On Demand should be requested for the target node. */
+	requestWakeUpOnDemand?: boolean;
```
AlCalzone, line 339: I'm not sure I like the approach of putting this on a message/transaction. I'd add a writable property to `ZWaveNode` (like the `keepAwake` property) and check this when deciding whether to set the flag.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -1711,14 +1713,15 @@ export class Driver extends TypedEventEmitter<DriverEventCallbacks> {
...
-									status: SupervisionStatus.NoSupport,
-									secure: msg.command.secure,
-								});
+							const node = msg.getNodeUnsafe();
+							await node?.commandClasses.Supervision.sendReport({
+								sessionId: supervisionSessionId,
+								moreUpdatesFollow: false,
+								status: SupervisionStatus.NoSupport,
+								requestWakeUpOnDemand:
+									node.shouldRequestWakeUpOnDemand,
```
AlCalzone: IMO the `Supervision.sendReport` method itself could handle the `requestWakeUpOnDemand` flag instead. If the node's `requestWakeUp` property is true, it supports the feature and there is no other `Supervision CC Report` with the flag is queued, set the `requestWakeUpOnDemand` flag on the outgoing command.

And don't forget to reset the property after requesting wake up.

AlCalzone (changes requested):
I'd approach it a bit differently.

Ikcelaks:
I initially began writing the code using `requestWakeUpOnDemand` flag in the `ZwaveNode` class. However, I realized that logic for deciding when to unset the flag was not simple.

Here is the main scenario that caused me to drop the idea: What if there is a long queue of commands waiting for the node to wake up, and one of the commands the user really wants to go through (flagged for Wake On Demand) doesn't go through during the first wake up? (I know I read somewhere in the specs that you have to account for the possibility that a node powered by a watch battery won't be able to stay awake for very long at any one time, and we can't assume that we can keep it awake indefinitely).

 In this case, we'll need to ask it to wake up again as soon as it's able, so we would NOT want to turn off the `requestWakeUpOnDemand` flag. But how do we determine if any of the messages still left in the queue are critical messages that warrant a Wake Up On Demand? So I thought of tagging them. Then I realized that the presence of that tag on one or more message in the queue removes the need for manual book-keeping on the `requestWakeUpOnDemand` flag, because we can just query the state machine to find out if message with the tag is present in the queue.

I'm really happy with the elegance of this general solution, but I'm less confident that adding a `tag` to the message is the correct way to do it. In particular, if `sendCommand` with the `requestWakeUpOnDemand` flag set is sent as part of a node interview, the message can't have both the "requestWakeUpOnDemand" and "interview" flags set at the same time. Perhaps it's better to add a new `MessagePriority` for these commands, but I thought it best not to mess with the order of messages queued for the node. Another option could be to enrich the way `tag` is handled so that a message can be marked as both "interview" and "requestWakeUpOnDemand".

I briefly considered the performance impact of querying the state machine every time the driver receives a `SupervisionCCGet` command from a sleeping node (instead of only when the node goes to sleep, which would be possible if we manually updated a flag), but that has to be trivial.

If you have a better idea for how to handle the problem scenario I described while using a flag on the `ZwaveNode` class, I'm happy to reimplement it that way.

packages/zwave-js/src/lib/commandclass/WakeUpCC.ts
Ikcelaks: If this is internal, how will the client know if a node supports this feature? Even if the driver doesn't allow the client to have any control over whether or not a command will trigger a Wake Up On Demand, it's definitely useful for the user to know if the feature is supported.

packages/zwave-js/src/lib/commandclass/WakeUpCC.ts
Ikcelaks: I standardized on "wake up", because that's how it was written in the spec sheet. However, it's definitely more important to be consistent within zwave-js, and changing existing conventions without good reasons is dumb... so "wakeup" (one word) it is.

packages/zwave-js/src/lib/driver/Driver.ts
Ikcelaks: See my top level [comment](https://github.com/zwave-js/node-zwave-js/pull/3332#issuecomment-921349465) where I explain why I went away from having a manual flag.

packages/zwave-js/src/lib/driver/Driver.ts
Ikcelaks: > IMO the `Supervision.sendReport` method itself could handle the `requestWakeUpOnDemand` flag instead. If the node's `requestWakeUp` property is true, it supports the feature and there is no other `Supervision CC Report` with the flag is queued, set the `requestWakeUpOnDemand` flag on the outgoing command.

Yeah, I'll move the `shouldRequestWakeUpOnDemand` logic into the `Supervision.sendReport` function. I originally made it a parameter, because I thought there might be a scenario where we wanted to force it to `false`, but I know longer see the purpose of doing that, and I just forgot to simplify it down.

> And don't forget to reset the property after requesting wake up.

The difficulty of deciding whether or not the flag should be reset is why I opted away from the flag altogether.
(See my top level [comment](https://github.com/zwave-js/node-zwave-js/pull/3332#issuecomment-921349465) where I explain why I went away from having a manual flag.)

packages/zwave-js/src/lib/commandclass/WakeUpCC.ts
Ikcelaks: Just looked at the existing code, and it looks like "wake up" is spelled:

1. "wakeup" in the interview logs
2. "Wake Up" in the `@ccValueMetaData` label
3. "wake-up" in the `Set` logs.

🤣 

I'm still changing it to "wakeup" here for local consistency.

AlCalzone:
> But how do we determine if any of the messages still left in the queue are critical messages that warrant a Wake Up On Demand? So I thought of tagging them.

Good counter point - I didn't consider that scenario. I guess it would make sense to have separate wake up queues for each node (or a flag that doesn't interfere with the normal priorities), so we can have different priorities among the messages to sleeping nodes (which is not really possible at the moment). That way we could make sure the critical messages are the first ones to be sent out. But that's out of scope for this PR if you ask me.

> I'm less confident that adding a tag to the message is the correct way to do it. In particular, if sendCommand with the requestWakeUpOnDemand flag set is sent as part of a node interview, the message can't have both the "requestWakeUpOnDemand" and "interview" flags set at the same time

The original idea of the `tag` was to give some context to the transactions, specifically identify messages coming from an interview. I'm not opposed to having an extra `requestWakeUp` flag on the transactions, what do you think?

Ikcelaks:
Adding a `requestWakeUpOnDemand` flag eliminates the `tag` conflict. Sounds good.

Side note, I thought the 700-series sensor I bought a while back supported this, but as I went to test it, I realized that it only supports `WakeUpCC v2`. In fact, I couldn't find a single device that actually supports this! I suppose I should have looked closer, but it was such a natural extension of the `SupervisionCCGet` work, and there was an open issue, so I just dived in. Fortunately there wasn't too much work. Hope you didn't spend much time on reviewing it.

I'll finish the work here tomorrow after the scene controller configuration compat work. Not sure if you'll want to merge it or not. Shouldn't be much harm if you do, because all the operations will basically be no-ops.

AlCalzone:
Let me know when you're done with this. I'll still merge it and we'll have to wait for issues to pop up if it doesn't work entirely. :)

Ikcelaks:
I consider this done as of my commit from earlier today, which created a `requestWakeUpOnDemand` property in the `Transaction` class.

As it's currently setup, no command is ever sent with the `requestWakeUpOnDemand` flag defined, and no node will ever get pass the first check in `shouldRequestWakeUpOnDemand`, so absolutely nothing should change for existing environments.

When/if devices support it, we'll need to consider how which commands we give `requestWakeUpOnDemand = true`.

AlCalzone (approved):

AlCalzone:
@Ikcelaks I think this actually resolves #1158 entirely or is something missing in your opinion?

Ikcelaks:
> @Ikcelaks I think this actually resolves #1158 entirely or is something missing in your opinion?

All of the technical work to support `WakeUpCC v3` and `Supervision CC v2` is done, but no commands actually set the `requestWakeUpOnDemand` option. That's a broader policy decision that I considered outside the scope of this PR. Might make sense to close #1158 and open a new issue for that work.

AlCalzone:
Done => https://github.com/zwave-js/node-zwave-js/issues/3365

PR #3334, opened 2021-09-13
fix(config): add templating and remove unnecessary parameters

popescunsergiu:
Add templating to Heltun devices


packages/config/config/devices/0x0344/he-rs01.json
```diff
@@ -18,51 +18,8 @@
 			"label": "Frequency Region",
 			"valueSize": 1,
 			"minValue": 0,
```
blhoward2, line 20: Why are we removing the options? Even if read only the UI will display those for the corresponding value (or it should).

packages/config/config/devices/0x0344/templates/heltun_template.json
```diff
@@ -0,0 +1,169 @@
...
+			},
+			{
+				"label": "Normally Closed",
+				"value": 1
+			}
+		]
+	},
+	"relay_timer_mode_duration": {
+		"description": "Used to open/close garage doors, blinds, curtains, etc. or to turn attached devices like door lock or security OFF for a short time",
+		"unit": "seconds",
```
blhoward2, line 157: Just to verify, this is seconds and not 0.1 seconds or something?

blhoward2 (changes requested):
Looks pretty good. A couple of minor comments.

packages/config/config/devices/0x0344/he-rs01.json
popescunsergiu: I've brought them back but made them readonly because they are readonly and I removed the options as not needed. Was a mistake

packages/config/config/devices/0x0344/templates/heltun_template.json
popescunsergiu: Based on their official documentation, this is seconds. Please see: https://drive.google.com/file/d/1b5OU8L0FCmAWAqXjWrczILsKaEBW2g-Y/view
Parameters 71-75 as example

packages/config/config/devices/0x0344/he-rs01.json
AlCalzone: Just to give some context: UIs can use the options to show nice labels for the values, even if the parameters are readonly.

AlCalzone: I don't see the options anymore. We should keep them even when readonly, because UIs will display the label instead of numbers.

AlCalzone (changes requested):
see https://github.com/zwave-js/node-zwave-js/pull/3334#discussion_r709904155

packages/config/config/devices/0x0344/he-rs01.json
popescunsergiu: Done, missed one commit, sorry

blhoward2 (approved):

AlCalzone (approved):

PR #3339, opened 2021-09-14
fix(config): tidy up Vitrum devices

AlexCPU:
Updates to number of the first generation "Wireless" Vitrum devices.  

Documentation for the Vitrum devices is very light.  The official "manual" from the Vitrum website is very basic, see [example](https://www.vitrum.com/EN/download/wireless/BS/Vitrum-I-BS-Dimmer-Wireless.pdf), and the manuals provided with the switches are light on Z-Wave details.  Most of the devices aren't listed on the Z-Wave Alliance either.

- Changed labels and descriptions to be consistent.  We could swap the label to product codes like the first "Cloud" device already in this folder, if that is preferred, however we then have twice the files as the BS and EU sized parts are identical save for productType/ID.
- Removed references to EU where both an EU and BS sized part are available
- Added BS productType/ID to "II Switch"
- Added 1-channel satellite device (can't find the productType/ID for the EU variant, so may require future update)
- Added and cleaned up association data/labels where missing/appropriate.
- Added metadata.


packages/config/config/devices/0x010a/vitrum_i_dimmer.json
```diff
@@ -1,8 +1,8 @@
 {
 	"manufacturer": "VDA",
 	"manufacturerId": "0x010a",
-	"label": "VITRUM I",
-	"description": "Vitrum I dimmer wireless",
+	"label": "Vitrum I Dimmer",
+	"description": "Vitrum I Dimmer Wireless",
```
blhoward2: If you haven't already, please take a look at our style guide. We don't repeat the label in the description. https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide. Maybe a label of VITRUM I Dimmer, and a description of "Wireless 1-Channel Dimmer Switch"? Or something that makes sense.

There is also a set order for which the various elements appear. they may be messed up in the originals, but please fix them. Metadata goes at the bottom, etc.

blhoward2 (changes requested):

packages/config/config/devices/0x010a/vitrum_i_dimmer.json
AlexCPU: I'm tempted to change the manufacturer to Vitrum from VDA, as it's the brand under which the device is sold.  Which would remove Vitrum from the Label + Description as per the style guide.

The style guide says that the description should match the name under which the device is sold, so would be "I Dimmer Wireless".

So I think it's the label that is more up for debate, I can split out to SKU as the style guide suggests if you're happy with two identical files for similar products?  For example, the British I Dimmer is 01B010010 and the European I Dimmer is 01E010010.  It's always just the one letter different between the British and European.

Thoughts on all of those as proposals?

I'll sort the order of elements out.

blhoward2: You can change the manufacturer to the brand it's sold under. We do that elsewhere. RE the label, just do one that makes sense. It doesn't have to be exactly the manufacturer's if they use a stupid one. We just don't want the UI to say Vitrium Vitrium I Vitrium I Dimmer Wireless, all in a row. You can make it Wireless Dimmer Switch I or whatever looks better. 

What we do elsewhere is "01B010010 / 01E010010" for the label.

AlexCPU:
I've updated all the labels, descriptions, ans manufacturer fields.  Just two that I can't find details for which I've left as they were.

packages/config/config/devices/0x010a/01e06s263.json
```diff
@@ -1,10 +1,11 @@
 {
-	"manufacturer": "VDA",
+	"manufacturer": "Vitrum",
 	"manufacturerId": "0x010a",
 	"label": "01E06S263",
-	"description": "Vitrum VI EU on-off 6 triac 868.4MHz",
+	"description": "VI On-Off 6 Triac",
 	"devices": [
 		{
+			// EU
```
blhoward2: Don't comment the regions, it doesn't matter

packages/config/config/devices/0x010a/01e06s263.json
```diff
@@ -17,7 +18,7 @@
 	"paramInformation": {
 		"7": {
 			"label": "Button N.1 Off Status Led",
-			"description": "Led is off: 0: LED is off; 1: red; 2: green; 3: blue; 4: yellow; 5: magenta; 6: cyan; 7; white",
+			"description": "0: LED is off; 1: red; 2: green; 3: blue; 4: yellow; 5: magenta; 6: cyan; 7; white",
```
blhoward2: These should be defined as options, not listed int he description. I know that's how they were, but please clean them up.

blhoward2 (changes requested):
Let's also change the labels in the 01e file, to the format: "Button N.2: LED (Off)" and "Button N.2: Output Port Connected"

AlexCPU:
@zwave-js-bot fix lint

AlexCPU:
I've added a template and used it for 01E06S263, as the new 2021 "Cloud" range all uses the same configuration parameters.  Datasheet for new parts is [here](https://www.vitrum.com/IT/wp-content/uploads/2021/04/USER-MANUAL-29.04.21-1.pdf).

AlCalzone:
#3100 was just merged, so I've gone ahead and updated your file to match the new config format. As you'll see, paramInformation is now an array of objects.

AlexCPU:
@AlCalzone, thanks!  Do you need me to do anything to this before it can be merged?

AlCalzone:
I wanted to give @blhoward2 another chance to review first.

packages/config/config/devices/0x010a/templates/vitrum_template.json
```diff
@@ -0,0 +1,236 @@
...
+			{
+				"label": "White",
+				"value": 7
+			}
+		]
+	},
+	"button_1_off_color": {
+		"$import": "#led_color_options",
+		"label": "Button 1 Off",
+		"description": "LED color of the button when the output is in the Off state",
```
blhoward2: We want to avoid descriptions whenever possible as they clutter the UI. The description here can easily be included in the label. "Button 1: LED Color When Off", etc. Please carry throughout.

packages/config/config/devices/0x010a/templates/vitrum_template.json
```diff
@@ -0,0 +1,236 @@
+{
+	"led_color_options": {
+		"valueSize": 1,
+		"minValue": 0,
+		"maxValue": 7,
+		"unsigned": true,
+		"allowManualEntry": false,
+		"options": [
+			{
+				"label": "Off",
```
blhoward2: Please make this "Disable" instead of "Off" so its consistent with what we do elsewhere.

packages/config/config/devices/0x010a/templates/vitrum_template.json
```diff
@@ -0,0 +1,236 @@
...
+	},
+	"output_port_options_6ch": {
+		"valueSize": 1,
+		"minValue": 0,
+		"maxValue": 6,
+		"unsigned": true,
+		"allowManualEntry": false,
+		"options": [
+			{
+				"label": "Disconnected",
```
blhoward2: Again, make this "Disable"

packages/config/config/devices/0x010a/templates/vitrum_template.json
```diff
@@ -0,0 +1,236 @@
...
+			},
+			{
+				"label": "Output Port 6",
+				"value": 6
+			}
+		]
+	},
+	"button_1_output_6ch": {
+		"$import": "#output_port_options_6ch",
+		"label": "Button 1 connected to",
```
blhoward2: Parameter labels should be Title Case. Please reference our style guide in the docs. 

https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide

packages/config/config/devices/0x010a/templates/vitrum_template.json
```diff
@@ -0,0 +1,236 @@
...
+		"defaultValue": 1,
+		"unsigned": true,
+		"allowManualEntry": false,
+		"options": [
+			{
+				"label": "100% (0x63)",
+				"value": 0
+			},
+			{
+				"label": "Last Level (0xFF)",
```
blhoward2: Option labels should be Sentence case.

packages/config/config/devices/0x010a/vitrum_i_eu_boiler.json
```diff
@@ -1,8 +1,8 @@
 {
-	"manufacturer": "VDA",
+	"manufacturer": "Vitrum",
 	"manufacturerId": "0x010a",
-	"label": "Vitrum I EU Boiler",
-	"description": "Vitrum touch timer switch Dual Relay",
+	"label": "Boiler",
+	"description": "touch timer switch Dual Relay",
```
blhoward2: Device descriptions should be Title Case. See the style guide referenced above.

packages/config/config/devices/0x010a/vitrum_scenario.json
```diff
@@ -1,8 +1,8 @@
 {
-	"manufacturer": "VDA",
+	"manufacturer": "Vitrum",
 	"manufacturerId": "0x010a",
-	"label": "Vitrum Scenario",
-	"description": "Vitrum Scenario controller",
+	"label": "Scenario",
+	"description": "Scenario controller",
```
blhoward2: Again, Title Case.

blhoward2 (changes requested):
Sorry for the delay. We appreciate your efforts. There are some style changes yet to meet our style guide, but its looking good.

packages/config/config/devices/0x010a/vitrum_i_eu_boiler.json
AlexCPU: I've changed this now, I'm sure it has a better name, but I can't find any instruction manuals or other references to this product anywhere, so I've just changed the case rather than the actual description to something better.

packages/config/config/devices/0x010a/vitrum_scenario.json
AlexCPU: I've changed this now, I'm sure it has a better name, but I can't find any instruction manuals or other references to this product anywhere, so I've just changed the case rather than the actual description to something better.

blhoward2 (approved):

PR #3354, opened 2021-09-18
feat(cc): add `compat` for `Scene Controller Configuration CC`

### LINKED ISSUE #1640: Enerwave ZWN-SC7 Scene Controller
I have one of these scene controllers that I've been using with a Vera Plus for several years. It has never wanted to work with Home Assistant and, as it's on the compatibility list for zwaveJS, I figured things might work now. It does appear in the interface and it's correctly identified and everything, but I can't figure out how to get it to work. Anyone out there actually have one of these working in their system that could point me in the right direction? If it works on a 5 year old Vera, I can't see why it won't work here...

### LINKED ISSUE #3350: feat(cc): add compatibility flag(s) to Scene Controller Configuration CC
I'm opening this issue to consolidate planning for Compat flag strategies for the Scene Controller Configuration CC.

zwave-js/zwavejs2mqtt#1640 exposed two ways that devices fail to follow the specs:

1. Some devices support more `groupId`s in the `SceneControllerConfigurationCC` than in `AssociationCC`.

I propose this be handled with a `compat` flag that directly specifies the number of groups the `SceneControllerConfigurationCC` supports. Along with this, a value to track this separately will be added to the `valueDB`. For devices that follow the specs, this value will just be set equal to the group count reported by the `AssociationCC`.

2. Some devices are write-only and don't respond to `SceneControllerConfigurationCC.Get` commands.

No "compat" flag required, imo. We should just switch to the `UserCode` interview style of setting MetaData and not even trying `Get` commands. Most users won't even care about the existing values, and, if they do, they can refresh values and hope their device responds.

Ikcelaks:
Resolves #3350 
Should also resolve zwave-js/zwavejs2mqtt#1640

- `forceSceneControllerGroupCount` compat flag sets the `groupCount` for devices that don't adhere to the 1:1 corespondence between `Scene Controller Configuration CC` groups and `Association CC` groups. This is know to be needed by the "Leviton VRCZ4-M0Z".
- The `SceneControllerConfigurationCC.interview` no longer queries the node for the scene configuration of each group. Instead, it only presets the `ValueMetaData` for each group. This speeds up the interview and avoids problems with devices that break spec and don't support `SceneControllerConfigurationCCGet` (ex: "Enerwave ZWN-SC7"). No `compat` flag is required for this.

### DIFF
--- a/docs/config-files/file-format.md
+++ b/docs/config-files/file-format.md
@@ -355,7 +355,7 @@ Version 8 of the `Notification CC` added the requirement that devices must issue
 
 ### `forceSceneControllerGroupCount`
 
-The specifications mandate that each `Scene Controller Configuration CC` Group ID corresponds to that actual `Association CC` group. Some devices ignore this rule, and the count of `Scene Controller Configuration CC` groups cannot be known from the count of `Association CC` groups. `forceSceneControllerGroupCount` directly specifies the number of `Scene Controller Configuration CC` groups to support.
+The specifications mandate that each `Scene Controller Configuration CC` Group ID corresponds to exactly one association group. Some devices ignore this rule, and as a result not all scenes can be configured. Using the `forceSceneControllerGroupCount` flag, the actual number of scenes of these devices can be configured.
 
 ### `manualValueRefreshDelayMs`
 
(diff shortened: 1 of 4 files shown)


AlCalzone (commented):
mostly LGTM. I don't have any of those devices, but could it be necessary to share that flag with the `Scene Actuator Configuration CC`?

Ikcelaks:
> mostly LGTM. I don't have any of those devices, but could it be necessary to share that flag with the `Scene Actuator Configuration CC`?

That flag is not relevant for `Scene Actuator Configuration CC`, which always supports all 255 scenes.

However, we don't set the ValueMetaData at all in that interview, because there is no interview. So, the values won't show up in HA or the zwavejs2mqtt Control Panel. This was actually intentional, because I didn't want to spew out 512 values for every device when most would go un-used. BUT, I realize now that the `setValue` API is basically the only way anything actually interacts with the driver, so maybe that should be reconsidered? There feels like no correct answer there.

AlCalzone:
> I didn't want to spew out 512 values for every device when most would go un-used

I still agree with that decision. Some CCs aren't a good fit for the `setValue` API.

AlCalzone (approved):

PR #3361, opened 2021-09-19
fix: rename consumed/produced in meter label to consumption/production

blhoward2:
As proposed in Slack discussion

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/MeterCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/MeterCC.ts
@@ -148,7 +148,7 @@ function getValueLabel(
 	suffix?: string,
 ): string {
 	let ret = getMeterTypeName(configManager, meterType);
-	if (rateType !== RateType.Unspecified && rateType !== RateType.Consumed) {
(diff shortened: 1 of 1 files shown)


AlCalzone:
I'm not convinced. There's a reason the specs distinguish between unspecified and consumed/produced. For older devices, you just cannot know which it is. You might have a legacy device that measures produced energy, but doesn't specify the rate type. If you then have a non-legacy device that measures consumed energy and you leave out this distinction, the two very different measurements might be confused as being similar.

blhoward2:
But wouldn't the user know they have a device that produces energy (which is a very small number of devices anyway...if any), even if the device didn't follow the spec? Would they be more confused that their solar panel says it's consuming energy, or that their electric plug says consumed amps?

How about we change to the descriptive nouns "consumption" and "production", which is what they should have used to describe a user or producer of electricity instead of the past tense nouns that describe past use?

AlCalzone (changes requested):
As discussed, let's use `Consumption/Production` for the labels

AlCalzone (approved):

PR #3376, opened 2021-09-20
fix(config): correct units for Fibaro FGS-224 and FGS-214 params 154 and 155

Redominus:
As stated in the [documentation](https://manuals.fibaro.com/content/manuals/en/FGS-2x4/FGS-2x4-T-EN-1.0.pdf): Parameter 154 is an integer that represents increments of 0.1 seconds also known as deciseconds (ds) https://en.wikipedia.org/wiki/Second

Maybe is a better idea to remove unit and explain the step size in the description?

### DIFF
--- a/packages/config/config/devices/0x010f/fgs214.json
+++ b/packages/config/config/devices/0x010f/fgs214.json
@@ -175,11 +175,11 @@
 		},
 		"154": {
 			"label": "First channel – time parameter for delayed/auto OFF and flashing modes",
-			"description": "This parameter allows to set time parameter used in timed modes. (s)",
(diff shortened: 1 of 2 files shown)


AlCalzone (changes requested):
I've thought of introducing a multiplier to handle these kinds of values. For now, we use `0.1 s` as the `unit`.
Also, please remove the unit from the description.

Redominus:
Applied changes as suggested. Also updated param 155 in FGS-224 file

AlCalzone (approved):

AlCalzone:
Thanks

PR #3396, opened 2021-09-23
fix(config): update Philio PST02A and PST02B to use partial config parameters (5, 6, 7)

orandasoft:
- Remove supportsZWavePlus: true
- Update configuration parameters 5, 6, and 7 to use partial parameters
- Add meta section with description for include, exclude, and reset

### DIFF
--- a/packages/config/config/devices/0x013c/pst02a.json
+++ b/packages/config/config/devices/0x013c/pst02a.json
@@ -426,5 +426,4 @@
 		"exclusion": "Turn the primary controller of Z-Wave network into exclusion mode, triple click tamper switch within 1.5 seconds",
 		"reset": "Pressing tamper key four times within 1.5 seconds and do not release the tamper key in the 4th  pressed, and the LED will light on. After 3 seconds the LED will turn off, after that within 2 seconds, release the tamper key. If successful, the LED will light on one second. Otherwise, the LED will flash once. IDs are excluded and all settings will reset to factory default."
 	}
-
 }
--- a/packages/config/config/devices/0x013c/pst02a.json
+++ b/packages/config/config/devices/0x013c/pst02a.json
@@ -245,7 +245,7 @@
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 1,
-			"defaultValue": 1,
+			"defaultValue": 0,
 			"allowManualEntry": false,
 			"options": [
 				{
--- a/packages/config/config/devices/0x013c/pst02a.json
+++ b/packages/config/config/devices/0x013c/pst02a.json
@@ -13,7 +13,6 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
-	"supportsZWavePlus": true,
 	"paramInformation": {
 		"2": {
 			"label": "Basic Set Level",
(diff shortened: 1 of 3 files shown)


orandasoft:
error file=packages/config/config/devices/0x013c/pst02b.json::Parameter #6 has a single bit mask defined. Either add more, or delete the bit mask.

PST02A, PST02B, and PST02C all use the same configuration definition. PST02B just happens to support only one bit. Shouldn't we use the same code for consistency?

orandasoft:
@zwave-js-bot fix lint

AlCalzone:
> Shouldn't we use the same code for consistency?

Ideally yes, but this would require a much larger change and the linter doesn't know this bitmask is shared with other devices.

AlCalzone (changes requested):
Looks good, but all the new partials with just 2 options should have
```
"allowManualEntry": false
```

and that single bitmask for param 6 in the one device needs to be a non-partial.

orandasoft:
> ```
> "allowManualEntry": false
> ```

Where should I put this?

AlCalzone:
> Where should I put this?

Before each of these lines:
```
"options": [
```
(where the options define ALL the possibilities)

packages/config/config/devices/0x013c/pst02a.json
```diff
@@ -61,47 +60,309 @@
...
 		},
-		"7": {
-			"label": "Customer Function",
-			"description": "Enable or disable functions of the multisensor.",
+		"5[0x08]": {
+			"label": "Temperature Scale",
 			"valueSize": 1,
 			"minValue": 0,
-			"maxValue": 127,
-			"defaultValue": 20,
```
AlCalzone, line 96: The old parameter 7 had a default value of 20, which means that the default for 7[0x10] and 7[0x04] should be 1 if I'm not mistaken.

packages/config/config/devices/0x013c/pst02a.json
```diff
@@ -61,47 +60,309 @@
...
 		},
-		"6": {
-			"label": "Multi-Sensor Function Switch",
-			"description": "Enable or disable functions of the multisensor.",
+		"5[0x04]": {
+			"label": "Door/Window Function",
 			"valueSize": 1,
 			"minValue": 0,
-			"maxValue": 127,
-			"defaultValue": 4
```
AlCalzone, line 88: The old parameter 6 had a default value of 4, so 6[0x04] should have a default value of 1, the rest 0 (if I'm not mistaken).

packages/config/config/devices/0x013c/pst02b.json
```diff
@@ -38,37 +37,183 @@
...
 			"minValue": 0,
-			"maxValue": 127,
-			"defaultValue": 4
+			"maxValue": 2,
+			"defaultValue": 0,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Disable",
+					"value": 2
```
AlCalzone, line 122: This doesn't match the old default value of 4. Which one is correct?

packages/config/config/devices/0x013c/pst02b.json
```diff
@@ -38,37 +37,183 @@
...
+					"value": 1
+				}
+			]
+		},
+		"7[0x04]": {
+			"label": "PIR Super Sensitivity Mode",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 1,
+			"defaultValue": 1,
```
AlCalzone, line 153: This doesn't match the old default value of 0.

AlCalzone (commented):
I'm not sure about the new default values. They don't always match the old ones, but the old ones could also be wrong.

packages/config/config/devices/0x013c/pst02a.json
orandasoft: According to this document, the default value is 4: https://doc.eedomus.com/files/PST02_Manual-2014-09-17.pdf

packages/config/config/devices/0x013c/pst02a.json
orandasoft: Changed default value to 0 for 6[0x20]

packages/config/config/devices/0x013c/pst02b.json
orandasoft: PST02B only supports one bit with default value 0: https://doc.eedomus.com/files/PST02_Manual-2014-09-17.pdf

packages/config/config/devices/0x013c/pst02b.json
orandasoft: According to this document, the default value for parameter 7 is 4: https://doc.eedomus.com/files/PST02_Manual-2014-09-17.pdf

AlCalzone (approved):

AlCalzone:
Thanks!

Issue #3368, opened 2021-09-19
Discard duplicate segments in segmented ConfigurationCCNameReports


AlCalzone:

Ikcelaks:
I imagine this bug also lurks in `AssociationCCReport` or any other CCs that deal with segmented reports (except `TransportServiceCCSubsequentSegment`, which merges very differently).

I think the proper solution is to add a new member in `CommandClass` like:
`public addPartialTo(session: CommandClass[]): CommandClass[]`

It could default to the current `session.push(this);` but be overridden in the specific `CC` implementations to drop duplicate partials or do other error checking.

I've created a minimal test suite for `ConfigurationCC` for this bug and could make a PR if you'd like.

AlCalzone:
Sounds good!

Ikcelaks:
#3401 should is just a start. It should be expanded to fix the underlying issues with how partial CC sessions are handled.

PR #3401, opened 2021-09-23
fix(cc): discard duplicate partial CCs in ConfigurationCC reports

Ikcelaks:
This PR adds tests for and fixes the symptoms shown in #3368, but there are more gremlins to kill with regards to `assemblePartialCCs`.

There are a couple big issues with `assemblePartialCCs`

1. It always adds an incoming CC to an existing session if one exists. (This PR addresses that, although it needs to be expanded to other CCs).
2. It never deletes a partial cc session unless it successfully completes.

Number 2 is a big problem. I'm surprised it wasn't noticed earlier, but I guess that very few CCs that are used in normal operation string together partial CCs. `TransportCC` masks the issue, because its implementation of `mergePartialCCs` causes the last received partial CC segments to overwrite the segments of any partial CCs that were left over from an earlier failed session.

Properly fixing number 2 is not trivial, at least for the `Report` CCs that follow the `reportsRemaining` pattern. I think we need to have timeouts for each session and blocks preventing `Get` commands from being sent if they match with an active session, because there's no good way to know if an incoming `Report` is actually from a new session if the node is misbehaving or having communication issues.

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/AssociationCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/AssociationCC.ts
@@ -608,6 +608,25 @@ export class AssociationCCReport extends AssociationCC {
 		return this._reportsToFollow > 0;
 	}
 
+	public addToPartialCCSession(session: AssociationCCReport[]): boolean {
+		if (
+			session.length === 0 ||
+			this.reportsToFollow ===
+				session[session.length - 1].reportsToFollow - 1
+		) {
+			session.push(this);
+			return true;
+		}
+		if (
+			this.reportsToFollow === session[session.length - 1].reportsToFollow
+		) {
+			// Assume this report is a duplicate. Keep the session and discard this cc.
+			return true;
+		}
+		// This cc doesn't fit into the session. A new session must be created.
+		return false;
+	}
+
 	public mergePartialCCs(partials: AssociationCCReport[]): void {
 		// Concat the list of nodes
 		this._nodeIds = [...partials, this]
--- a/packages/zwave-js/src/lib/commandclass/CommandClass.ts
+++ b/packages/zwave-js/src/lib/commandclass/CommandClass.ts
@@ -856,6 +856,17 @@ export class CommandClass {
 		return false; // By default, all CCs are monolithic
 	}
 
+	/**
+	 * Attempt to add this CC to the partial CC session
+	 * @param session The previously received set of messages received in this partial CC session
+	 * @returns `true` if Partial CC is successfully added to the session
+	 */
+	public addToPartialCCSession(session: CommandClass[]): boolean {
+		// Default to the old behavior of pushing the CC onto the session without any sanity checks
+		session.push(this);
+		return true;
+	}
+
 	/** Include previously received partial responses into a final CC */
 	/* istanbul ignore next */
 	// eslint-disable-next-line @typescript-eslint/no-unused-vars
--- a/packages/zwave-js/src/lib/commandclass/CommandClass.ts
+++ b/packages/zwave-js/src/lib/commandclass/CommandClass.ts
@@ -859,7 +859,7 @@ export class CommandClass {
 	/**
 	 * Attempt to add this CC to the partial CC session
 	 * @param session The previously received set of messages received in this partial CC session
-	 * @returns `true` if Partial CC is successfully added to the session
+	 * @returns `true` if session is valid for this CC, false if it needs to be reset
 	 */
 	public addToPartialCCSession(session: CommandClass[]): boolean {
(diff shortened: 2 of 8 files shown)


AlCalzone:
IMO, the PR looks good. Regarding the expired sessions, I don't think the fix is that complicated. Each of the sessions is identified by the session id, which is a combination of CC, CC command and sometimes a command specific identifier. I can't think of any situation where multiple of these sessions can (should) be active for a given node, so we could use the start of a new session to clean up all other partial sessions of the same node.

There are two exceptions:
* Transport Service, which needs special handling and does not allow cleaning up previous sessions immediately (see comment in `handleTransportServiceCommand`). We should be able to clean up the partial sessions in `handleTransportServiceCommand` though.
* Those commands where we have nothing to distinguish sessions. But those only have a single entry, so I think the problem is limited.

New CCs won't use this reportsRemaining pattern and will use TransportService instead. So once we've cleaned up what we have, we should be good.

Ikcelaks:
The non-trivial part that I was thinking about was determining the start of a new session for the `reportsToFollow` CCs.

It's not possible to look at an individual report and determine if it's the first report in the session or a continuation of an existing session. Usually we'll be able to infer that we're starting a new session by checking if the new partial CC advertises more `reportsToFollow` than the last partial CC stored in the session, but it _could_ be that an old session failed a long time ago after one report, subsequent actions caused the the data that the node needed to send to shrink by one report, and a fresh partial CC comes in that looks like it fits with the really old session (causing corrupted data). If you want to say that such a possibility is remote enough to ignore... I'm fine with that.

For this PR, I will use the following logic:
- If the `reportsRemaining` is exactly one less than the last partial CC in the session, assume it's a continuation and push it on the session as usual.
- If the `reportsRemaining` is the same as in the last partial CC in the session, assume it's a duplicate and drop it but keep the session.
- Otherwise, assume it's starting a new session. Clear the existing partial CCs and set this new partial CC as the only item in the session.

Still, I think it _is_ worthwhile to prevent new `Get` commands that expect `reportsToFollow` style `Report` responses from being retried while there is an existing session. Adding that functionality would necessitate clearing stale partial CC sessions anyways.

Finally, I think using the driver function `assemblePartialCCs` to handle both Transport Service and `reportsToFollow` -style CCs is complicating things. `TransportServiceCC` segments end up being tracked in two separate sessions: `nodeSessions.TransportService` and the generic `partialCCSessions`. When a Transport Service session fails, it's cleaned up in `nodeSessions.TransportService`, but not `partialCCSessions`.

Ikcelaks:
I've updated the PR to clear out sessions when they don't make sense with the newest partial CC. I also added the logic to all CCs with `reportsToFollow`. I think they should all be identical. Ideally this would be abstracted out into a shared function, but I still think that most of this work should ultimately be scrapped to completely rework the `assemblePartialCCs` function in the driver.

This does not address bug with non-`reportsToFollow` style partial CCs (like Transport Service) not clearing out their failed sessions in `partialCCSessions`.

FYI, I'm going to be on vacation and unavailable to code for the next week. I'll still see messages here though. The code is completely untested outside of unit tests. Just wanted to toss something up here.

AlCalzone:
> It's not possible to look at an individual report and determine if it's the first report in the session or a continuation of an existing session

For most CCs, this doesn't matter. Config params (which started this issue) identify their session by the param number, so if there is a report coming in for an old session with the same ID, it will fit. If it doesn't match, there are bigger problems than corrupted param names. So I'm fine with ignoring that edge case.

> For this PR, I will use the following logic:
> * If the reportsRemaining is exactly one less than the last partial CC in the session, assume it's a continuation and push it on the session as usual.
> * If the reportsRemaining is the same as in the last partial CC in the session, assume it's a duplicate and drop it but keep the session.
> * Otherwise, assume it's starting a new session. Clear the existing partial CCs and set this new partial CC as the only item in the session.

I think that's too strict. I've seen lots of cases where single reports out of a sequence got lost, e.g. 17, 16, 14, 13, ... Zooz devices, which send unnecessarily small 4-byte chunks are particularly prone to losing segments.
Though you could argue that those aren't complete and need to be retried anyways.

I agree with your last two paragraphs about the Get commands and separating reportsToFollow and TransportService. `TransportService` required the more complicated logic, so I made it a separate state machine. `reportsToFollow` seems to need more sophisticated logic too - maybe it's worth making it into its own state machine. 
Also, what's currently missing completely is a refresh for the timeout in `SendThreadMachine` in the `waitForUpdate` state. As long as we're receiving **new** parts of an ongoing session, that timeout shouldn't elapse.

> I'm going to be on vacation and unavailable to code for the next week

No stress - I'll be away the week after.

PR #3406, opened 2021-09-24
Update for correct name of button and switch

pepsonEL:
0	Mono-stable input (button)
1	Bi-stable input (switch)

### DIFF
--- a/packages/config/config/devices/0x010f/fgs221_2.1-2.3.json
+++ b/packages/config/config/devices/0x010f/fgs221_2.1-2.3.json
@@ -213,11 +213,11 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Momentary switch",
+					"label": "Mono-stable input (button)",
 					"value": 0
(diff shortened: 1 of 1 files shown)


pepsonEL:
Why this can not be merged ?

AlCalzone:
I don't have time to review all config file changes if they match the style guide. I've asked @blhoward2 to take a look, but he has a life too :)

pepsonEL:
But it is name type from manual Fibaro.

AlCalzone:
The texts and descriptions manufacturer manuals are often not good and every manufacturer calls things differently, so we often use different labels to be consistent.

blhoward2:
Yes, and actually we're not going to merge this. We're aiming for consistency across devices and the labels you're replacing are what we generally use. We want human readable labels. "Mono-stable input" is not very user friendly. 

We very much welcome suggestions for improvements to the files, but this isn't an improvement just because it matches the manual.

PR #3412, opened 2021-09-25
feat(config): add fingerprint to Popp & Co POPE700342

mundschenk-at:
New device comes with FW 3.5 and `productID` set to `0x0010`. I've written an email to Popp support asking if there are any additional changes (their published manual has not been updated and still refers to `0x000d` and FW 3.05 instead).

### DIFF
--- a/packages/config/config/devices/0x0154/pope700342.json
+++ b/packages/config/config/devices/0x0154/pope700342.json
@@ -7,7 +7,8 @@
 		{
 			"productType": "0x0004",
 			"productId": "0x000d"
-		},{
+		},
+		{
 			"productType": "0x0004",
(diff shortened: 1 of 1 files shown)


mundschenk-at:
@zwave-js-bot fix lint

AlCalzone:
> FW 3.05

FYI, that **is** 3.5. Some manufacturers just use leading zeroes for their sub-versions.

AlCalzone (approved):

mundschenk-at:
According to the mail I just got from Popp support, `0x000d` was a documentation error that has now been fixed in their [manuals](https://manuals.zwave.eu/backend/make.php?lang=en&sku=pope700342&type=popp).

PR #3417, opened 2021-09-27
fix(config): update ZCOMBO-G units/metadata

Gunsmithy:
I noticed that the SuperVision report timeout was erroneously using seconds instead of milliseconds, so changed to ms as per the style guide, and changed s for the SuperVision wait time to seconds as well, also per the style guide.

Added a wakeup field to metadata as well just to clarify that a short press of the test button is not sufficient to wake the device.


AlCalzone (approved):
Thanks!

PR #3422, opened 2021-09-28
fix(config): change NAS-PD07Z parameters to match actual device configuration

luzik:
Based on few tests, and real default values I have read from brand new 4 devices. There is no param no (4 or 5) and 13, all params except 99 have numbers -1 and > 13 -2.
Documentation have an error, is outdated or there is newer firmware but I do not have access to it.

### DIFF
--- a/packages/config/config/devices/0x0258/nas-pd07z.json
+++ b/packages/config/config/devices/0x0258/nas-pd07z.json
@@ -39,24 +39,6 @@
 			"$import": "templates/shenzhen_neo_template.json#motion_prevent_retrigger"
 		},
 		"5": {
-			"label": "Binary Sensor Report",
-			"valueSize": 1,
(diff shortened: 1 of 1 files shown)


AlCalzone:
I wonder if something changed with a firmware update, but I don't have high hopes that we'll ever find documentation on that.

AlCalzone:
@zwave-js-bot rename commit

AlCalzone (approved):

AlCalzone:
Just for completeness sake: which firmware versions do your devices have?

luzik:
It is 2.6. I will ask someone for test it with other device

AlCalzone:
We can merge this for now, I just wanted to have the version somewhere in case another user shows up with different config params. Then we could divide the file by firmware version.

PR #3475, opened 2021-10-06
Correctly label Parameters 1-4 with Manual/Z-wave

bpottle1:
Changed the labels on parameters 1-4 to have the correct label of manual or z-wave in parentheses. These new labels match the data found in the user manual parameters table at Inovelli's website.
https://support.inovelli.com/portal/api/kbArticles/407409000001873397/attachments/1abnfa5822ce11f1e4dfe9a7306effa38a8cf/content?portalId=1ce1db55e4825f16146555b24f0f1753cbf5f7d83ca0a3d8b31013b2a2d69e1f&inline=true

Parameters 1 and 2 had Manual and Z-wave descriptions in reverse and Parameters 3 and 4 had the same issue.


AlCalzone:
In this PR the labels were swapped because they didn't match the device behavior:
https://github.com/zwave-js/node-zwave-js/pull/2399

Did you confirm with your devices that your changes are correct or did you simply align them to the manual?

AlCalzone:
#3100 was just merged, so I've gone ahead and updated your file to match the new config format. As you'll see, paramInformation is now an array of objects.

Also, marking this as draft until the question is resolved.

bpottle1:
Oh good point. I just tested with my switch and it looks like you're right. The manual is wrong for some reason. I'll go ahead and close this PR.

PR #3476, opened 2021-10-07
fix(config): add fingerprint to BeNext Energy switch

q1x:
I have a slightly different BeNext Energy Switch that uses a different productID and is not currently recognised as such.
I've added it to the list of devices.

Let me know if any other info is needed to add it to the device database.


AlCalzone (dismissed):

AlCalzone:
@zwave-js-bot rename commit

AlCalzone (approved):

PR #3512, opened 2021-10-12
fix(config): auto-assign `Binary Sensor Report` association group for FortrezZ MIMOLite

edalquist:
See: https://community.home-assistant.io/t/zwavejs2mqtt-fotrezz-mimolite-configuration/346084

The current configuration does not setup the Binary Sensor report association and as a result the input contact does not update unless polled.


AlCalzone:
@zwave-js-bot rename commit

AlCalzone (approved):

PR #3522, opened 2021-10-13
feat(config): add support for Heltun panels

popescunsergiu:
Add support for HE-TPS01, HE-TPS02, HE-TPS03, HE-TPS04


AlCalzone (approved):
LGTM. I think it would make sense to use a template for `Frequency Region` and `Weekday Override`, but we can do that separately.

popescunsergiu:
I'll do it in a next PR, I have few more Heltun devices that I'd like to add support for.  Thx

PR #3262, opened 2021-09-01
feat: implement soft reset, use it on startup

AlCalzone:
This PR adds support for the SoftReset command that can be used to restart a hanging controller. We now also use it during the startup procedure as recommended by the specs. Hopefully this fixes the problem where the controller seems stuck after a container reboot.

**Note:** USB modules will reconnect when this is called, meaning that they might get a new address. This shouldn't be a problem when using the `/dev/serial-by-id/` approach on Unix.

fixes: #3202

PR #3525, opened 2021-10-14
feat: implement 700-series variant of NVM backup/restore, including soft reset

### LINKED ISSUE #3202: Soft-reset controller on driver startup

### LINKED ISSUE #3482: Use Firmware Update API to backup/restore NVM on 700 series controllers
GetNVMId does not work here

AlCalzone:
This PR implements the Firmware Update API which is required to do NVM backups/restore on 700 series controllers. This includes the soft reset PR (https://github.com/zwave-js/node-zwave-js/pull/3262) which is necessary to activate a restored backup.

fixes: #3482
fixes: #3202
closes: https://github.com/zwave-js/node-zwave-js/pull/3262


docs/api/driver.md
```diff
@@ -585,6 +599,13 @@ interface ZWaveOptions {
 	 */
 	disableOptimisticValueUpdate?: boolean;
 
+	/**
+	 * Soft Reset is required after some commands like changing the RF region or restoring an NVM backup.
+	 * Because it may be problematic in certain environments like Docker, the functionality must be opted into.
+	 * Default: `false` in Docker, `true` when ZWAVEJS_ENABLE_SOFT_RESET env variable is set or outside Docker.
+	 */
+	enableSoftReset?: boolean;
+
```
AlCalzone, line 608: @robertsLando you might want to expose this in zwavejs2mqtt

robertsLando: @AlCalzone Sure! Same for soft reset I think, right?

BTW When a user calls the nvm backup from a 700 series should I use a different method to backup it?

AlCalzone: This PR makes it so you don't need to care whether the user has a 500 or 700 series stick.

Regarding softReset: I'd add an option that users have to enable/disable. Disabled by default in Docker, enabled by default outside Docker. If that is enabled, expose the `softReset` action.

robertsLando: Hmmm cool 😆

darkbasic:
Damn that was fast, thank you so much! I've bought not one but three UZB7 (they're cheap as hell): I don't want to be stuck with broken undowngradable firmware ever again.

EDIT: even too fast, shipping couldn't keep up with your pace :)

AlCalzone:
@zwave-js-bot pack this

PR #3528, opened 2021-10-15
refactor: throw if soft reset is called when not enabled

AlCalzone:
With this PR, `driver.softReset()` will throw an error with code `Driver_FeatureDisabled` if the soft reset feature is not enabled.

To avoid this, use `maybeSoftReset`, which only performs a soft-reset when enabled.

### DIFF
--- a/docs/api/driver.md
+++ b/docs/api/driver.md
@@ -102,7 +102,7 @@ Since it might be necessary to control a node **before** its supported CC versio
 
 ```ts
 async softReset(): Promise<void>
-async maybeSoftReset(): Promise<void>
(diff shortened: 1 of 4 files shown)


docs/api/driver.md
```diff
@@ -102,13 +102,17 @@ Since it might be necessary to control a node **before** its supported CC versio
 
 ```ts
 async softReset(): Promise<void>
+async maybeSoftReset(): Promise<void>
```
robertsLando: I would call this `trySoftReset`

robertsLando (approved):

PR #3533, opened 2021-10-16
refactor: reverse functionality of soft reset env var, enable by default

blhoward2:
After further testing, it appears the docker issue is largely a non-issue. This PR enables soft reset on startup by default and reverses the functionality of the environment variable.

@robertslando this will require a change to the UI too.

### DIFF
--- a/docs/api/driver.md
+++ b/docs/api/driver.md
@@ -107,8 +107,7 @@ async trySoftReset(): Promise<void>
 
 Instruct the controller to soft-reset (restart). The returned Promise will resolve after the controller has restarted and can be used again.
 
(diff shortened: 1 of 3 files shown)


blhoward2:
@robertsLando @AlCalzone asked if you could test this since you'd had some issues before. One thing to note, enableSoftReset should be set to true by default now, but zwavejs2mqtt still says its disabled if I have the setting set to Auto. It works if I set it to enabled. Is there a bug in the UI?

AlCalzone:
This seems to work fine for me, so if the feature stays disabled, the bug must be in z2m.

blhoward2:
I think it's because Auto sets null, but I'm not sure.

AlCalzone:
That might be. @robertsLando `null` is never a good default choice for zwave-js's config options. It only overrides `undefined` with the default.

AlCalzone (approved):

robertsLando:
Sorry missed this

PR #3545, opened 2021-10-18
fix: avoid force-adding Supervision support, remove encapsulation CCs from mandatory

### LINKED ISSUE #3536: TechniSat / Flush Roller Shutter Switch does not support Supervision
https://devices.zwave-js.io/?jumpTo=0x0299:0x0005:0x1a93:0.0

### LINKED ISSUE #3543: [question] HomeSeer WD200+ report state inconsistent
Greetings! I have most everything working now with one issue... I have three HomeSeer WD200+ dimmers that are no reporting state consistently.

They SEEM to work correctly immediately after inclusion / interview (changing states reports state change back immediately). Then as soon as I operate the switch manually, things change. Past that point, any manual manipulation sends state changes perfectly fine. But any commands sent to change state (via targetValue) no longer receive an update on currentValue..

I've had issues with HomeSeer being... selective... of how they interpret spec before, but any ideas here on work arounds?

AlCalzone:
With this PR, we no longer force-add Supervision support when receiving a `Supervision CC::Get` because of some devices that only actually support Supervision when sending.

In addition, we remove the non-essential encapsulation CCs from the list of mandatory CCs to avoid running into situations where a device doesn't list the CC in its NIF, we force-add it and then are unable to communicate.

fixes: #3536
fixes: #3543

### DIFF
--- a/packages/config/config/deviceClasses.json
+++ b/packages/config/config/deviceClasses.json
@@ -80,16 +80,16 @@
 					"supportedCCs": [
 						"Association", // V2+
 						"Association Group Information",
-						"CRC-16 Encapsulation",
+						// "CRC-16 Encapsulation", (trust the NIF! Assuming support would break communication)
 						// "Device Reset Locally", (we can't know if the device can be reset)
 						"Inclusion Controller",
 						"Powerlevel",
 						"Manufacturer Specific",
 						// "Security", (some devices only support S2)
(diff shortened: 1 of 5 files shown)


AlCalzone:
Note: We could remove the compat flag from devices, but I'd prefer not breaking things further.

PR #3566, opened 2021-10-20
fix(config): correct manufactuerID for Zooz ZAC36

blhoward2:
Corrects discussion RE ZAC36
fixes: #3414


shbatm:
Confirmed locally, this fixes #3414.

AlCalzone (approved):

PR #3578, opened 2021-10-22
Update iblindsv3.json

aLTeReGo-SWI:
Corrected required firmware version for Remote Calibration. There is no 3.6 version of iBlinds firmware. Correct version is 3.06.

 https://support.myiblinds.com/knowledge-base/iblinds-v3-firmware/

PR description here


AlCalzone:
A firmware version in Z-Wave does not have leading zeroes, as they are just two numbers (major and minor). Some manufacturers still display them, leading to confusion.

PR #3532, opened 2021-10-16
fix: show status for FGS223

jbenc:
Fibaro FGS223 needs the same workaround as FGS222. Tested successfully with
the device.

PR #3597, opened 2021-10-26
fix(config): fix reporting for Fibaro FGS-223

AlCalzone:
closes: #3532

### DIFF
--- a/packages/config/config/devices/0x010f/fgs223.json
+++ b/packages/config/config/devices/0x010f/fgs223.json
@@ -29,6 +29,16 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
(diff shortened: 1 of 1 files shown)


kpine:
Any clue if this same change would fix problems with other FGS devices, reported at https://github.com/home-assistant/core/issues/57395?

AlCalzone:
Would be worth a try. You can test manually by removing the lifeline association(s) and adding one to
Endpoint 0, group 1 -> node 1, endpoint 0

Some devices might also need associations on each endpoint, with target node 1, no endpoint. Or target endpoint 0.

Basically we need to figure out which it is by looking if the reports come multi channel encapsulated.

PR #3610, opened 2021-10-28
feat(config): add Aeotec aerq humidity sensor v2

blhoward2:
Per Aeotec, they have released v2 of their sensor, which is distinguishable only by the firmware version. The model number changes, and we do not support conditions there, so I had to split the file. 

They also have a v11.1 custom firmware file which fixes some issues with the reporting intervals that are corrected with v2.


blhoward2:
@zwave-js-bot fix lint

AlCalzone (approved):
I've changed the ranges to be a bit more permissive

PR #3623, opened 2021-10-31
fix(config): update Inovelli LZW31-SN device config to match official documentation.

GaTechThomas:
Updating LZW31-SN device config to include detailed parameter descriptions and updated values from newer firmware updates. Pulled information directly from Inovelli's documentation. 

Confirmation info: see attachments at this page: https://support.inovelli.com/portal/en/kb/articles/products-switches-dimmer-lzw31-sn-spec-sheet as well as firmware update info at https://support.inovelli.com/portal/en/kb/articles/firmware-change-log-lzw31-sn-dimmer-switch-red-series

### DIFF
--- a/packages/config/config/devices/0x031e/lzw31-sn.json
+++ b/packages/config/config/devices/0x031e/lzw31-sn.json
@@ -451,7 +451,7 @@
 		{
 			"#": "16[0xff]",
 			"label": "LED Bar Effect Color",
-			"description": "This will allow you to add some sweet effects to your LED bar (e.g., pulse, chase, solid, etc.). This is Byte 1 (of 4): Choose color using the scaled hue color circle (realHue / 360 * 255)."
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
This isn't going to get merged in its current state.

If you read our [style guide](https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide), you'll see that we strive for short and comprehensible descriptions and labels, not perfect alignment with manufacturer-provided manuals. These are often pretty bad, especially when it comes to chinese manufacturers. Also, we avoid repeating information in descriptions that's available via other means (unit, options, max/min, ...).

Furthermore, the manuals are wrong for params 1-4, see https://github.com/zwave-js/node-zwave-js/pull/2399 for reference.

We can keep the firmware changes, but please undo the descriptions unless absolutely necessary.

AlCalzone:
@GaTechThomas Are you planning to finish this?

GaTechThomas:
Yes.

PR #3640, opened 2021-11-02
feat(config): add HomeSeer HS-WX300 Switch

jberry:
New HomeSeer HS-WX300 switch

closes: #3612

### DIFF
--- a/packages/config/config/devices/0x000c/hs-wx300.json
+++ b/packages/config/config/devices/0x000c/hs-wx300.json
@@ -45,7 +45,7 @@
 		},
 		{
 			"#": "6",
-			"label": "Central Scene Control and Multi-Tap",
+			"label": "Central Scene, Multi-Tap, Press and Hold",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 1,
@@ -53,11 +53,11 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Central Scene Enabled, controls load with delay. Enables Multi-tap and press and hold",
+					"label": "Enable (control load with delay)",
 					"value": 0
 				},
 				{
-					"label": "Central Scene Disabled, controls load instantly. Disables multi-tap, central scene, and press and hold",
+					"label": "Disable (control load instantly)",
 					"value": 1
 				}
 			]
@@ -168,8 +168,7 @@
 		},
 		{
 			"#": "32",
-			"label": "Wire mode",
-			"description": "Sets the Wire mode / no neutral mod of the switch",
+			"label": "Wire Mode",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 1,
@@ -177,7 +176,7 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "3 wire mode (Neutral, Line, & Load)",
+					"label": "3 wire mode (Neutral, Line & Load)",
 					"value": 0
 				},
 				{
--- a/packages/config/config/devices/0x000c/hs-wx300.json
+++ b/packages/config/config/devices/0x000c/hs-wx300.json
@@ -132,28 +132,8 @@
 			"$import": "templates/homeseer_template.json#blink_frequency"
 		},
 		{
-			"#": "31[0x40]",
-			"label": "LED 7 Blink Status",
-			"$import": "templates/homeseer_template.json#led_blink_status"
-		},
-		{
-			"#": "31[0x20]",
-			"label": "LED 6 Blink Status",
-			"$import": "templates/homeseer_template.json#led_blink_status"
-		},
-		{
-			"#": "31[0x10]",
-			"label": "LED 5 Blink Status",
-			"$import": "templates/homeseer_template.json#led_blink_status"
-		},
-		{
-			"#": "31[0x08]",
-			"label": "LED 4 Blink Status",
-			"$import": "templates/homeseer_template.json#led_blink_status"
-		},
-		{
-			"#": "31[0x04]",
-			"label": "LED 3 Blink Status",
+			"#": "31[0x01]",
+			"label": "LED 1 Blink Status",
 			"$import": "templates/homeseer_template.json#led_blink_status"
 		},
 		{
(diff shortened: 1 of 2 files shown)


packages/config/config/devices/0x000c/hs-wx300.json
```diff
@@ -0,0 +1,633 @@
+{
+	"revision": 1,
```
AlCalzone:
```suggestion
```
Where did you get this from? The import script?

packages/config/config/devices/0x000c/hs-wx300.json
```diff
@@ -0,0 +1,633 @@
...
+		}
+	],
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"description": "Lifeline association only supports the \"manual reset\" event.",
```
AlCalzone:
```suggestion
```

packages/config/config/devices/0x000c/hs-wx300.json
```diff
@@ -0,0 +1,633 @@
...
+			"maxNodes": 5,
+			"isLifeline": true
+		},
+		"2": {
+			"label": "Companion Sync",
+			"description": "Any device added to group 2 will be controlled. This allows other WX300 devices to used as companion switches. LEDS will be kept in sync.",
```
AlCalzone:
```suggestion
```

packages/config/config/devices/0x000c/hs-wx300.json
```diff
@@ -0,0 +1,633 @@
...
+			"maxNodes": 5
+		}
+	},
+	"paramInformation": [
+		{
+			"#": "3",
```
AlCalzone, line 29: A couple of these parameters are already templated, see other HomeSeer files for reference, e.g.
https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x000c/hs-ws200.json

We should use the available templates here for consistency.

packages/config/config/devices/0x000c/hs-wx300.json
```diff
@@ -0,0 +1,633 @@
...
+			"#": "30",
+			"label": "Blink Frequency",
+			"description": "Set blink frequency when displaying custom status",
+			"valueSize": 3,
+			"minValue": 0,
+			"maxValue": 255,
+			"defaultValue": 0
+		},
+		{
+			"#": "31[0x3f]",
```
AlCalzone: These bitmasks seem off (they overlap). You can use https://zwave-js.github.io/node-zwave-js/#/config-files/file-format?id=bitmask-calculator to figure out the correct ones

AlCalzone (changes requested):
We can keep this one and close the other, but there are a couple of things that need to be cleaned up.

packages/config/config/devices/0x000c/hs-wx300.json
jberry: Yes, that was from the import script pulling from the OpenSmartHouse database via this command/device ID;  This is also where it auto-pulled the bitmasks from:
`npm run config -- import -s oh -Dd --ids 1432`

packages/config/config/devices/0x000c/hs-wx300.json
jberry: I've flipped everything I could find that matches to templates, as well as added a new homeseer template stanza for individual LED blink status (led_blink_status) since it was repeated 7 times in this device-config

packages/config/config/devices/0x000c/hs-wx300.json
jberry: I believe I fixed these using the calculator and matching the checkbox'd bit with the device manual's parameters (https://docs.homeseer.com/display/HSPRODKB/HS-WX300+User+Guide)

packages/config/config/devices/0x000c/hs-wx300.json
AlCalzone: Ok we need to remove it then. That was something we tried a while back.

packages/config/config/devices/0x000c/hs-wx300.json
```diff
@@ -0,0 +1,207 @@
...
+				{
+					"label": "Top of Paddle turns load ON",
+					"value": 0
+				},
+				{
+					"label": "Bottom of Paddle turns load ON",
+					"value": 1
+				}
+			]
+		},
```
AlCalzone: I believe this parameter is the same as that one:
https://github.com/zwave-js/node-zwave-js/blob/b23ee76e6bb654ef76580fb5518f9e2e0a14dff6/packages/config/config/devices/0x000c/hs-ws200.json#L22-L24

packages/config/config/devices/0x000c/hs-wx300.json
```diff
@@ -0,0 +1,207 @@
...
+			"description": "Sets the lowest dimming threshold",
+			"valueSize": 1,
+			"minValue": 1,
+			"maxValue": 14,
+			"defaultValue": 1
+		},
+		{
+			"#": "6",
+			"label": "Central Scene Control",
+			"description": "Enables/Disables Central Scene and multiple tap functions",
```
AlCalzone: This can likely be merged into only a label

packages/config/config/devices/0x000c/hs-wx300.json
```diff
@@ -0,0 +1,207 @@
...
+			"maxValue": 1,
+			"defaultValue": 0,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Central Scene Enabled, controls load with delay. Enables Multi-tap and",
+					"value": 0
+				},
+				{
+					"label": "Central Scene Disabled, controls load instantly. Disables multi-tap, c",
```
AlCalzone: These labels are cut off

packages/config/config/devices/0x000c/hs-wx300.json
```diff
@@ -0,0 +1,207 @@
...
+		},
+		{
+			"#": "12",
+			"label": "Local Ramp Rate",
+			"description": "Set dimmer ramp rate for local control",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 90,
+			"defaultValue": 3
+		},
```
AlCalzone: Align to https://github.com/zwave-js/node-zwave-js/blob/b23ee76e6bb654ef76580fb5518f9e2e0a14dff6/packages/config/config/devices/0x000c/hs-wd200.json#L26-L43 for consistency please

AlCalzone (changes requested):
Thanks, a few more things

packages/config/config/devices/0x000c/hs-wx300.json
jberry: It does appear to be the same parameter; Looking at that master orientation template, it imports "base_enable_disable", which just has the options "enable" and "disable"; Those options don't really clarify this setting as well as the current-labels do.  I'll switch it if you'd like, it just didn't make good/clear sense when I was comparing objects/labels.  Or maybe the current orientation setting should be reworded, or a new cleaner/clearer template for "Which side of the paddle turns stuff on"?

AlCalzone: @blhoward2 what do you think? Knowing what is inverted and what not kinda makes sense.

blhoward2: Ordinarily we just do enable/disable with inverted. It's pretty well understood that top = on is default, so this seems like added complexity. (And if you install it upside down, it's no longer the top, which is what these settings are generally meant for. Is it the top before inversion, or after?)

jberry: Alrighty, I went ahead and just set this to the normal orientation inverted template for now

AlCalzone: Did you commit this change already? I'm not seeing it.

jberry: That was odd, my cloud-vscode didn't want to add-to-commit my last changes; it's there now

AlCalzone (dismissed):
OK with my changes

AlCalzone (approved):

PR #3648, opened 2021-11-03
fix(config): add missing parameters for Telldus TZWP-102

norrland:
Enable additional reporting configurations.

* 17 - Report on power consumption change %
* 101 - Power (W) report interval
* 102 - Energy (kWh) report interval
* 103 - Voltage (V) report interval
* 104 - Current (A) report interval


packages/config/config/devices/0x0176/tzwp-102.json
```diff
@@ -93,6 +93,76 @@
...
+				{
+					"label": "Disabled",
+					"value": 0
+				}
+			]
+		},
+		{
+			"#": "101",
+			"label": "Power (W) report interval",
+			"valueSize": 4,
```
AlCalzone, line 117:
```suggestion
			"valueSize": 4,
			"unit": "seconds",
```
Do the same for the following parameters please.

packages/config/config/devices/0x0176/tzwp-102.json
```diff
@@ -93,6 +93,76 @@
 					"value": 0
 				}
 			]
+		},
+		{
+			"#": "17",
+			"label": "Report on change of power consumption (%)",
```
AlCalzone: We should distingish between param 16 and this one better. I suggest:
Param 16:
- change label to "Power Consumption Reports: Change Threshold"
- add unit `W`

Param 17:
- change label to "Power Consumption Reports: Relative Change Threshold"
- add unit `%`

AlCalzone (changes requested):

packages/config/config/devices/0x0176/tzwp-102.json
```diff
@@ -93,6 +94,81 @@
...
+					"value": 0
+				}
+			]
+		},
+		{
+			"#": "101",
+			"label": "Power (W) report interval",
+			"valueSize": 4,
+			"unit": "seconds",
+			"minValue": 5,
```
AlCalzone: Oh... one more thing I missed. `minValue` needs to be 0 if there is an option for 0. We generally explain the gap in the valid range with a description, e.g. "Allowable range: 5-2678400"

Same for the following parameters

AlCalzone (changes requested):

packages/config/config/devices/0x0176/tzwp-102.json
norrland: Gotcha, was wondering about that one. If it accepted an array of [0, 5-n] or something.
Fixing coming up.

AlCalzone (approved):

AlCalzone:
Thanks!

PR #3653, opened 2021-11-03
fix: gate SmartStart behind SDK 6.81+, add method to test support

### LINKED ISSUE #2320: Whitelist commands by Serial API version
The more and more "advanced" commands we're trying to implement, the more aren't actually supported, even if the supported list says so. We'll need a whitelist per Serial API version that is checked first before asking the stick if it supports the command.

AlCalzone:
New investigations have shown that SmartStart is only supported in SDK 6.81+ and using it will prevent inclusion of new nodes on older sticks. Therefore it is now gated by the SDK version and trying to provision a node on an older stick will throw. Already provisioned nodes will stay on the list but won't be auto-included.

Before using SmartStart, checking for support is now mandatory with `controller.supportsFeature(ZWaveFeature.SmartStart)`.

fixes: #2320

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -4,6 +4,24 @@ The controller instance contains information about the controller and a list of 
 
 ## Controller methods
 
+### `supportsFeature`
+
+```ts
+supportsFeature(feature: ZWaveFeature): boolean | undefined
+```
+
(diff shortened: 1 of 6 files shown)


AlCalzone:
@zwave-js-bot pack this

AlCalzone:
@zwave-js-bot pack this

PR #3660, opened 2021-11-03
fix(config): add Ring 4AR1SZ-0EN0 Range Extender Gen2

Botched1:
Add config file for Ring Alarm Extender Gen2

### DIFF
--- a/packages/config/config/devices/0x0346/range_extender_gen2.json
+++ b/packages/config/config/devices/0x0346/range_extender_gen2.json
@@ -1,0 +1,68 @@
+{
+	"manufacturer": "Ring",
+	"manufacturerId": "0x0346",
+	"label": "4AR1SZ-0EN0",
+	"description": "Range Extender (2nd generation)",
+	"devices": [
+		{
+			"productType": "0x0401",
+			"productId": "0x0301"
+		}
+	],
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"paramInformation": [
+		{
+			"#": "1",
+			"label": "Battery Report Interval",
+			"valueSize": 2,
+			"unit": "minutes",
+			"minValue": 4,
+			"maxValue": 70,
+			"defaultValue": 70
+		},
+		{
+			"#": "2",
+			"label": "Application Retries",
+			"description": "Number of application level retries attempted for messages either not ACKed or messages encapsulated via supervision get that did not receive a report.",
+			"valueSize": 1,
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
The parameters look similar to https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x0346/contact_sensor_gen2.json where I reviewed the labels, options, etc. for consistency.

Please align the parameters to the linked file where possible.

Botched1:
I made the changes to make it like the contact sensor, because you are right that the parameters are largely the same,.

Hopefully I integrated them correctly (still learning git).

Thanks!

AlCalzone (approved):

truvec:
Coincidentally I was changing over to HA and have a 4AR1SZ-0EN0 in my Zwave setup. 

Home Assistant 2021.11.1
ZWAVE:
Driver Version: 8.7.5
Server Version: 1.10.8

I see that 4AR1SZ-0EN0 was just added. I am getting this error:

2021-11-08T20:45:45.241Z 
SERIAL « 0x012400a814ff001bece228b70127d39c010f015e598580705a7a87728e71739f6 (38 bytes) c558600daa5
2021-11-08T20:45:45.242Z
DRIVER   Dropping message because it could not be deserialized: The command class unknown (0xec) is not implemented (ZW0303)

Is this having an issue at the config file or further downstream?

Thank You.

AlCalzone:
That seems to be a message that should have never been received by the application. I've seen it happen on 700 series sticks.

It contains the homeid in the location where the CC is normally defined. Just ignore it.

truvec:
Im not able to get the 4AR1SZ-0EN0 to join the network. While in inclusion the log just repeats the error.

Botched1:
I'm not sure how you are pairing it - smartstart, other.


But what I would do:
1. Initiate normal pairing in zwave js
2. Start pairing on the extender (plug it in, reset it, whichever)
3. Once the llight starts flashing on the extender press and hold the front button for 1-2s. That takes it out of smart pairing mode, and puts it in standard pairing mode. Then it should continue pairing in zwave js.

I paired 5x of them that way.

Or you could try it with smartstart I guess. I'm not a fan of smartstart, so never bother. But as zwave js now supports smartstart that should theoretically work too.

truvec:
@Botched1  Thanks for that.  I wasn't holding it long enough. RTFM.

I guess it really was downstream. 
Thanks for the help.

PR #3670, opened 2021-11-05
fix(config): add Zooz ZEN23 config file for firmware v20.15-20.16

cybergrimes:
There was an original batch of V1 ZEN23 devices that had firmware versions 20.15 and 20.16. They are not capable of OTA upgrades and only have a single configuration parameter. It's the same as the 'zen23_0.0-2.9.json' config file.

https://thesmartesthouse.happyfox.com/kb/article/301-zen23-toggle-switch-change-log/

I am uncertain how to proceed with the 'zen23_4.0.json' file which covers firmware 4.0 - 255.255, does it need to be edited to cover 4.0 - 20.14 and a second file for 20.17 - 255.255? I doubt such a situation would ever occur but I want to make sure I am adhering to your standards for submission. Please let me know! 

Thank you.


AlCalzone:
Yes, this sounds like we need to split the 4.0+ file into the two ranges you mentioned.
/cc @blhoward2

cybergrimes:
I split the 4.0+ file but wanted to mention I did notice for ZEN22 the opposite seems to be in place. That device also had early hardware running firmware 20.15-20.16, not capable of OTA updates. For that it seems the "current" firmware, 4.3 as of Feb. 2021, would pull file 'zen22_3.7-19.0.json'. There is no split file for after the 20.15-20.16 firmware versions.

blhoward2:
Theee are actually in our old format (different files per fw version). Rather than create new split files, if only the parameters change we should probably just use one file and use conditional parameters. Is there a reason not to do that here?

cybergrimes:
I don't see any reason not to use conditional parameters. I did a quick review of the existing files and the Zooz changelog-- it should be feasible. It looks like there are some corrections that could be made at the same time.

blhoward2:
We appreciate it. We haven't gotten to the Zooz folder yet due to life events on my end.

cybergrimes:
I'll resubmit with a fresh commit later. Thanks!

PR #3671, opened 2021-11-06
fix(config): add missing units to Fibaro FGBS-222

Masterz69:
Added couple of missing units definitions for parameters.
Units source : [Fibaro FGBS-222 manual](https://manuals.fibaro.com/content/manuals/en/FGBS-222/FGBS-222-EN-T-v1.0.pdf).


AlCalzone (changes requested):
Thanks! However we avoid these nonstandard units and add the multiplier in the unit field.
For example `"0.1 s"`

This will allow us to factor these out into actual multipliers in the future.

AlCalzone (approved):

PR #3681, opened 2021-11-08
fix(config): update ZEN23 config file with conditional parameters

cybergrimes:
Previously discussed on #3670.
ZEN23 change log: https://thesmartesthouse.happyfox.com/kb/article/301-zen23-toggle-switch-change-log/

ZEN23 switches have some older hardware that could not do OTA updates. Those early models were limited to a single configuration setting. This PR adds consolidates the three ZEN23 config files currently in place to a single config file using conditional parameters that will limit configuration settings for appropriate firmware versions.

Parameter 1 with 2 values is available beginning with firmware 1.5 and also on firmware 20.15. A third value was added beginning with firmware 3.0. Change log indicates firmware 1.5 and 20.15 only ever received parameter 1 configuration setting. **It states that a firmware 20.16 is not capable of any configuration settings.**

I could not find any examples of using conditional parameter where so many ranges had occurred. I thought it was correct to use both AND / OR operators for this to work. Please let me know if this is not the case.

thank you!


cybergrimes:
@zwave-js-bot fix lint

packages/config/config/devices/0x027a/zen23.json
```diff
@@ -28,6 +32,28 @@
 	"paramInformation": [
 		{
 			"#": "1",
+			"$if": "firmwareVersion >= 1.5 && firmwareVersion < 3.0 || firmwareVersion === 20.15",
+			"label": "Switch behavior",
+			"description": "Determines whether up is on and down is off; or up is off, down is on",
```
AlCalzone: This seems unnecessary, the option labels mention this already.

AlCalzone (changes requested):
The rules look ok, I'd add a space behind `>=` and `<=` though

packages/config/config/devices/0x027a/zen23.json
cybergrimes: I removed the description since as you said it's redundant. It was a copy/paste of the existing config, I had not intended to any review of the labels while adding these conditional parameters. 

I'm sure you're aware but this parameter label (and others) are inconsistent throughout the Zooz files. It sometimes called 'Toggle control' or 'Orientation' also with different labels for the values. From what I've seen it's probably all taken from how Zooz has changed the language over time, different descriptors each product generation.

I notice from your style guide that you're big on consistency-- and are using templates for config files! So I had assumed there was probably a larger scope cleanup of the Zooz files planned for the future.

cybergrimes: I also added the spaces, thank you!

AlCalzone: > there was probably a larger scope cleanup of the Zooz files planned for the future

There is, but we still try to clean up the files a bit if they are touched anyways.

AlCalzone (approved):

PR #3686, opened 2021-11-09
feat: add support for Door Lock Logging CC

dawnho:
# Add Door Lock Logging CC

- Successfully interviews the supported records number:

Screen Shot 2021-11-22 at 11 51 22 PM

- Successfully interviews the latest record here:

Screen Shot 2021-11-22 at 11 39 51 PM

Todo in the future:
- My door locks don't return the `User Code` attribute on the logging record, so I don't have a good idea of what that data looks like. For now, I set the `userCode` to `undefined`.

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
@@ -162,7 +162,7 @@ export function getDoorLockLoggingRecordTimestampID(endpoint: number): ValueID {
 	return {
 		commandClass: CommandClasses["Door Lock Logging"],
 		endpoint,
-		property: "record timestamp",
+		property: "recordTimestamp",
 	};
 }
 
@@ -170,7 +170,7 @@ export function getDoorLockLoggingRecordEventTypeID(endpoint: number): ValueID {
 	return {
 		commandClass: CommandClasses["Door Lock Logging"],
 		endpoint,
-		property: "record event type",
+		property: "recordEventType",
 	};
 }
 
@@ -178,7 +178,7 @@ export function getDoorLockLoggingRecordUserIDID(endpoint: number): ValueID {
 	return {
 		commandClass: CommandClasses["Door Lock Logging"],
 		endpoint,
-		property: "record user id",
+		property: "recordUserID",
 	};
 }
 
@@ -186,7 +186,7 @@ export function getDoorLockLoggingRecordUserCodeID(endpoint: number): ValueID {
 	return {
 		commandClass: CommandClasses["Door Lock Logging"],
 		endpoint,
-		property: "record user code",
+		property: "recordUserCode",
 	};
 }
 
@@ -516,6 +516,7 @@ export class DoorLockLoggingCCRecordReport extends DoorLockLoggingCC {
 			};
 
 			const eventType = this.payload[9];
+			this.recordUserID = this.payload[9];
 			const userCodeLength = this.payload[10];
 			const userCodeBuffer = this.payload[11];
 
@@ -532,7 +533,6 @@ export class DoorLockLoggingCCRecordReport extends DoorLockLoggingCC {
 			};
 			this.recordTimestamp = segmentsToDate(dateSegments).toISOString();
 			this.recordEventType = eventType;
-			this.recordUserID = this.payload[9];
 			this.recordUserCode = parseUserCodeFromPayload(
 				userCodeLength,
 				userCodeBuffer,
--- a/packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
@@ -44,7 +44,7 @@ interface DateSegments {
 function segmentsToDate(segments: DateSegments): Date {
 	return new Date(
 		segments.year,
-		segments.month,
+		segments.month - 1, // JS months are 0-based.
 		segments.day,
 		segments.hour,
 		segments.minute,
@@ -54,8 +54,8 @@ function segmentsToDate(segments: DateSegments): Date {
 
 // All the supported commands
 export enum DoorLockLoggingCommand {
-	RecordsSupportedGet = 0x01,
-	RecordsSupportedReport = 0x02,
+	RecordsCountGet = 0x01,
+	RecordsCountReport = 0x02,
 	RecordGet = 0x03,
 	RecordReport = 0x04,
 }
@@ -66,8 +66,8 @@ export enum EventType {
 	UnlockCode = 0x02,
 	LockButton = 0x03,
 	UnlockButton = 0x04,
-	LockCodeOOSchedule = 0x05,
-	UnlockCodeOOSchedule = 0x06,
+	LockCodeOutOfSchedule = 0x05,
+	UnlockCodeOutOfSchedule = 0x06,
 	IllegalCode = 0x07,
 	LockManual = 0x08,
 	UnlockManual = 0x09,
@@ -77,8 +77,8 @@ export enum EventType {
 	UnlockRemoteCode = 0x0d,
 	LockRemote = 0x0e,
 	UnlockRemote = 0x0f,
-	LockRemoteCodeOOSchedule = 0x10,
-	UnlockRemoteCodeOOSchedule = 0x11,
+	LockRemoteCodeOutOfSchedule = 0x10,
+	UnlockRemoteCodeOutOfSchedule = 0x11,
 	RemoteIllegalCode = 0x12,
 	LockManual2 = 0x13,
 	UnlockManual2 = 0x14,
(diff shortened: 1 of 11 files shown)


packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+	day: number;
+	hour: number;
+	minute: number;
+	second: number;
+}
+
+function segmentsToDate(segments: DateSegments): Date {
+	return new Date(
+		segments.year,
+		segments.month,
```
AlCalzone: Someone had a brainfart 26 years ago and as a result, months in JS are 0-based. This code will create a date that's one month ahead of the logging record.

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+}
+
+// @publicAPI
+export enum EventType {
+	LockCode = 0x01,
+	UnlockCode = 0x02,
+	LockButton = 0x03,
+	UnlockButton = 0x04,
+	LockCodeOOSchedule = 0x05,
+	UnlockCodeOOSchedule = 0x06,
```
AlCalzone: These are already so long, let's write out the "out of":
```suggestion
	LockCodeOutOfSchedule = 0x05,
	UnlockCodeOutOfSchedule = 0x06,
```

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+	LockManual = 0x08,
+	UnlockManual = 0x09,
+	LockAuto = 0x0a,
+	UnlockAuto = 0x0b,
+	LockRemoteCode = 0x0c,
+	UnlockRemoteCode = 0x0d,
+	LockRemote = 0x0e,
+	UnlockRemote = 0x0f,
+	LockRemoteCodeOOSchedule = 0x10,
+	UnlockRemoteCodeOOSchedule = 0x11,
```
AlCalzone: dito

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+	Unknown = 0x20,
+}
+
+const eventTypeLabel: { [key in keyof typeof EventType]: string } = {
+	LockCode: "Locked via Access Code",
+	UnlockCode: "Unlocked via Access Code",
+	LockButton: "Locked via Lock Button",
+	UnlockButton: "Unlocked via Unlock Button",
+	LockCodeOOSchedule: "Lock Attempt via Out of Schedule Access Code",
+	UnlockCodeOOSchedule: "Unlock Attempt via Out of Schedule Access Code",
```
AlCalzone:
```suggestion
	LockCodeOOSchedule: "Out of Schedule Lock Attempt via Access Code",
	UnlockCodeOOSchedule: "Out of Schedule Unlock Attempt via Access Code",
```
I think this rolls off the tongue a bit better. Do the same for the other ones that are out of schedule

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+	IllegalCode: "Illegal Access Code Entered",
+	LockManual: "Manually Locked",
+	UnlockManual: "Manually Unlocked",
+	LockAuto: "Auto Locked",
+	UnlockAuto: "Auto Unlocked",
+	LockRemoteCode: "Locked via Remote Out of Schedule Access Code",
+	UnlockRemoteCode: "Unlocked via Remote Out of Schedule Access Code",
```
AlCalzone: These two aren't out of schedule, are they?

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+		EventType.UserCodeDeleted,
+		EventType.AllUserCodesDeleted,
+		EventType.MasterCodeChanged,
+		EventType.UserCodeChanged,
+		EventType.LockReset,
+		EventType.ConfigurationChanged,
+		EventType.LowBattery,
+		EventType.NewBattery,
+		EventType.Unknown,
+	];
```
AlCalzone: You don't need to specify this. Leave out the 2nd argument to `enumValuesToMetadataStates` and `states` will contain all possible values.

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+				// Fetches latest record by default
+				return (await this.getRecord(LATEST_RECORD_NUMBER_KEY))?.[
+					property
+				];
+			default:
+				throwUnsupportedProperty(this.ccId, property);
+		}
+	};
+
+	public async getSupportedRecordsNumber(): Promise<any> {
```
AlCalzone: Let's call this `getRecordCount` or something like that. The "official" Z-Wave commands often have weird names and I like to simplify them for the CCAPI methods.

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+		);
+		const recordTimestampID = getDoorLockLoggingRecordTimestampID(
+			this.endpointIndex,
+		);
+		const recordUserIDID = getDoorLockLoggingRecordUserIDID(
+			this.endpointIndex,
+		);
+		const recordUserCodeID = getDoorLockLoggingRecordUserCodeID(
+			this.endpointIndex,
+		);
```
AlCalzone: This is going to cause problems when you request a record that is not the last one. The constructor (and by extension persistValues) is called for every report, including the ones that are returned in response to a GET. This would mean the values hold the last **received** record, but not necessarily the most recent one.

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+	public constructor(
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+		validatePayload(this.payload.length >= 11);
+
+		this.recordNumber = this.payload[0];
+		this.recordStatus = this.payload[5] >>> 5;
+		if (this.recordStatus === RecordStatus.Empty) {
```
AlCalzone: This condition (`recordStatus === 0xfe`) will never be true if you shift the value.

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+				this.commandOptions,
+			);
+		if (response) {
+			return pick(response, ["supportedRecordsNumber"]);
+		}
+	}
+
+	public async getRecord(
+		recordNumber: number = LATEST_RECORD_NUMBER_KEY,
+	): Promise<any> {
```
AlCalzone:
```suggestion
	): Promise<Record | undefined> {
```

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
AlCalzone:
```suggestion
	public async getSupportedRecordsNumber(): Promise<number | undefined> {
```

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+		validatePayload(this.payload.length >= 1);
+
+		this.supportedRecordsNumber = this.payload[0];
+		this.persistValues();
+	}
+
+	@ccValue()
```
AlCalzone:
```suggestion
	@ccValue({internal: true})
```

AlCalzone: This doesn't need to be visible in UIs that work off the exposed values

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+	public persistValues(): boolean {
+		if (!super.persistValues()) return false;
+
+		this.getValueDB().setValue(
+			getDoorLockLoggingRecordsSupportedValueID(this.endpointIndex),
+			this.supportedRecordsNumber,
+		);
+
+		return true;
+	}
```
AlCalzone: Remove this, values marked with @ccValue and @ccValueMetadata are persisted automatically

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
AlCalzone: I still think it is somewhat tricky to work with persisted values here as opposed to events. If I see it correctly, there's no way to know if a record is the last one, except when we know we are requesting the one with number 0. Or when handling an unsolicited report - but that's not yet known here.

In my head it makes sense to align this CC more with the Entry Control CC. When requesting a specific record via `commandClasses["Door Lock Logging"].get(n)` you get the correct one in response. And when the lock sends an unsolicited report, it would be emitted as an event the application can handle.

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
```diff
@@ -0,0 +1,662 @@
...
+		this.persistValues();
+	}
+
+	public readonly record: Record | undefined;
+	public readonly recordNumber: number;
+	public readonly recordStatus: RecordStatus;
+	public readonly recordTimestamp: string | undefined;
+	public readonly recordEventType: EventType | undefined;
+	public readonly recordUserID: number | undefined;
+	public readonly recordUserCode: string | Buffer | undefined;
```
AlCalzone: I don't really like duplicating the data here, but let's solve the other issues first.

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
dawnho: no wonder I see `month - 1` elsewhere in the code. Will address

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
dawnho: It's checking against `0x00`, which should work.
I've changed the `HoldsLegalData` payload to reflect your comment.

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
dawnho: I tried that before, and the house of cards falls down when I do that. I tried tinkering with the types all about, and I can't seem to get it working

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
dawnho: ditto here

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
dawnho: Yeah, this was a very hacky way to do this, let me try another approach

packages/zwave-js/src/lib/commandclass/DoorLockLoggingCC.ts
AlCalzone: The problem was that `pick` returns an object with the picked property. You have to access the desired property directly.

AlCalzone (approved):
Approved with my changes

PR #3691, opened 2021-11-10
refactor: rework SendThreadMachine to eliminate nonce and update steps and handle sequenced messages

### LINKED ISSUE #1826: Rework hardcoded handshake step in SendThreadMachine
Currently, we send a handshake message before sending the actual transaction if necessary (S0 nonce). This makes the SendThreadMachine awkward because it has separate stages and triggers for that.

Ideally, we should introduce a concept of `message generators`. These would be functions that can be called multiple times by the state machine and return a new message instance every time. These generators could check if a message needs a nonce request and return that nonce request first. If sending fails somewhere in the chain, the generator could be reset to start from the beginning for the re-send attempt.

Use cases include:
* Security S0 nonce requests
* Security S2 nonce synchronization
* Setting partial config parameters with up-to-date base values
* Sequenced Security S0 messages (currently unsupported)
* Transport Service CC

AlCalzone:
This PR reworks how the SendThreadMachine handles messages, more specifically waiting for a node response and the "handshake" steps before secure (S0) transmissions. With the separate stages for handshake transactions, triggering them and later waiting for the response while further handshakes might be necessary, the machine was quite unwieldy and not obvious to follow through.

This is now changed to use the concept of message generators as described in #1826. A transaction now works off a message generator (an async generator function) which dynamically yields the required messages in sequence. Each sequence is handled entirely before the next one is started.
The SendThreadMachine waits before each send attempt until the message generator has yielded a message, which also includes waiting for a response. In the simplest case, such a generator just yields a message and then returns. More complicated use cases are S0 nonce handling and the other cases descibed in #1826.

fixes: #1826


AlCalzone:
@zwave-js-bot pack this

PR #3731, opened 2021-11-16
feat(config): add Ecolink TBZ500 Z-Wave Smart Thermostat

GrizzlyAK:
This is my first PR, so please be gentle. I've created this Config file for this new device. It is an RCS-based device used by both Ecolink and Nortek. I found a GoControl device (gocontrol_gc-tbz48.json) that contained many similar parameters, and built on that to build this file based upon the details in the Ecolink manual. I used Lint in the GitPod editor with no errors. This is about as far as I can take it, if someone can take over, verify and merge.


GrizzlyAK:
yay

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+			"#": "7",
+			"label": "Temperature Unit",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 1,
+			"defaultValue": 1,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Celcius",
```
AlCalzone:
```suggestion
					"label": "Celsius",
```

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+			"unit": "minutes",
+			"valueSize": 1,
+			"minValue": 3,
+			"maxValue": 9,
+			"defaultValue": 3
+		},
+		{
+			"#": "10",
+			"label": "Setpoint H/C Delta",
+			"unit": "°F",
```
AlCalzone, line 132: Is this always °F or depending on the chosen unit?

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+			"options": [
+				{
+					"label": "w/Cool",
+					"value": 0
+				},
+				{
+					"label": "w/Heat",
+					"value": 1
+				}
+			]
```
AlCalzone: These labels seem weird. Can you turn them into something more natural sounding?
I'm also not sure what this parameter does. Maybe it needs a better combination of parameter and option labels.

In general we think it is okay to modify the manufacturer's label to something clearer. They are often pretty bad.

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+			"#": "5",
+			"label": "Aux Heat Enable",
+			"$import": "~/templates/master_template.json#base_enable_disable",
+			"defaultValue": 1
+		},
+		{
+			"#": "6",
+			"label": "2nd Stage Cool Enable",
+			"$import": "~/templates/master_template.json#base_enable_disable"
+		},
```
AlCalzone: We avoid "Enable/Disable" in the labels. The available options (here Enable/Disable) should be enough.

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+			"label": "Minimum Run Time",
+			"unit": "minutes",
+			"valueSize": 1,
+			"minValue": 3,
+			"maxValue": 9,
+			"defaultValue": 3
+		},
+		{
+			"#": "10",
+			"label": "Setpoint H/C Delta",
```
AlCalzone: This and the following labels seem cryptic. Can you make them sound more naturally?

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+				},
+				{
+					"label": "Unoccupied Mode",
+					"value": 2
+				}
+			]
+		},
+		{
+			"#": "41",
+			"label": "Unoccupied HSP",
```
AlCalzone: a bit cryptic

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+			"unit": "°F",
+			"valueSize": 1,
+			"minValue": 30,
+			"maxValue": 109,
+			"defaultValue": 62
+		},
+		{
+			"#": "42",
+			"label": "Unoccupied CSP",
```
AlCalzone: dito

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+			"unit": "°F",
+			"valueSize": 1,
+			"minValue": 33,
+			"maxValue": 112,
+			"defaultValue": 80
+		},
+		{
+			"#": "43",
+			"label": "Remote Sensor 1 Node Number",
```
AlCalzone: Not sure what this one does, maybe it needs a short description.

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 252,
```
AlCalzone: Z-Wave only supports 232, not 252 nodes in a network

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+			"minValue": 0,
+			"maxValue": 120,
+			"defaultValue": 0,
+			"options": [
+				{
+					"label": "Disable",
+					"value": 0
+				}
+			]
+		},
```
AlCalzone: We usually call these parameters something like _Temperature Report Threshold_ and _Temperature Report Interval_

AlCalzone (changes requested):
Looks pretty good! Some of the labels could be improved to be more natural/user friendly.

If/where it makes sense, you can reorder the parameters to group related ones together.

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: I'll make that change. Note that this was carried over from `gocontrol_gc-tbz48.json`, so it will need to be changed there as well. If you agree, I will try my hand at it.

EDIT: Although, this range is what is shown in the manual, both read/write. Hmm...

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Yes, I will add one. The devices allows use of a secondary temperature sensor (this specifies the ZW node) that can be used instead of (or in addition to), the one built into the unit. Haven't tried it yet.

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: I left these as they appear in the manual. Having read the manual, and after setting up the device using its screen, they become clear. I can add a description, but it would be basically spelling out Heat and Cool and maybe explaining that the delta is between two. I favor verbosity as a fault, but was trying _hard_ to adhere to the style guide by just sticking with the manual's verbiage. I'd be happy to expand and add a description to these if you'd like. Let me know.

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: HSP = Heat Setpoint

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: CSP = Cold Setpoint

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Got it. Will change. Similar issue with the one I used as a base: `gocontrol_gc-tbz48.json`

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Has to do with Heat Pumps and how the device is wired. Per the manual: 

**• Changeover type (For Heat Pump HVAC systems only)**
The changeover (or reversing) valve is used to change from
heating to cooling operation. It is either a changeover with
cooling type (Orange wire) or changeover with heating
type (Brown wire). Most are changeover with cooling, which
is the default setting.
- For Changeover with Cooling systems (Orange wire),
select “WITH COOL”. This is the default setting.
- For Changeover with Heating systems (Brown wire),
use the Up/Dn arrows to change to “WITH HEAT”.

I can add the last two bullets as option labels, and add a description that includes bits of the stuff following the first bullet if that makes sense.

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: I chose to do this because of the following statement in the manual:

Note all temperature related parameters are in degrees F.

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Not sure how I missed that!

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Done

packages/config/config/devices/0x014a/tstat-tbz500.json
AlCalzone: Ok so the unit is strictly for the sent reports then?

packages/config/config/devices/0x014a/tstat-tbz500.json
AlCalzone: I think the following is enough, what do you think?

label: "Changeover Type"
description: "For Heat Pump HVAC systems only"
options: "Changeover with Cooling systems (Orange wire)" and "Changeover with Heating systems (Brown wire)".

I don't mind repeating the "Changeover" in the option labels here.

packages/config/config/devices/0x014a/tstat-tbz500.json
AlCalzone: > spelling out Heat and Cool

Ahh 💡 That wasn't immediately clear to me :D
That would be enough for this param.

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+			"label": "Setpoint H/C Delta",
+			"unit": "°F",
+			"valueSize": 1,
+			"minValue": 3,
+			"maxValue": 15,
+			"defaultValue": 3
+		},
+		{
+			"#": "11",
+			"label": "H Delta Stage 1 ON",
```
AlCalzone: So this would be
```suggestion
			"label": "Heat Delta Stage 1 On",
```
then, right?
Maybe a **short** description would be beneficial, not sure.

AlCalzone: same for the next ones

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+			"label": "Mechanical Status",
+			"valueSize": 2,
+			"minValue": 1,
+			"maxValue": 256,
+			"unsigned": true,
+			"readOnly": true
+		},
+		{
+			"#": "22",
+			"label": "SCP Status",
```
AlCalzone: I don't think it is immediately clear what SCP is.

packages/config/config/devices/0x014a/tstat-tbz500.json
AlCalzone: Ah, I'd write it out then.

packages/config/config/devices/0x014a/tstat-tbz500.json
AlCalzone: It's not impossible that the manufacturer made a typo.

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,625 @@
...
+			"label": "Filter Timer",
+			"unit": "hours",
+			"valueSize": 2,
+			"minValue": 0,
+			"maxValue": 4000,
+			"defaultValue": 0
+		},
+		{
+			"#": "53",
+			"label": "Filter Timer Nax",
```
AlCalzone: Was this supposed to be
```suggestion
			"label": "Filter Timer Max",
```
?

AlCalzone (changes requested):

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Yes

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Done

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Will include this with my other changes.

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Didn't realize you could do Options like that. I think what you have is fine. Will change that.

GrizzlyAK: I like it. Will change.

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: I read this to mean that all values used in temperature-related parameters (like Deltas and setpoints) are assumed to be in deg F, which would be consistent with the units listed on those parameters here. Most of these are writeable params. 

There is a user preference in Setup that I believe this parameter refers to, which is the DISPLAYED unit, whereas internally, it appears everything is F. Operation thus far seems consistent with all of that.

GrizzlyAK:
@AlCalzone I committed the changes. Didn't know if you'll get a notice or if I need to `Re-request Review`?

GrizzlyAK:
@zwave-js-bot fix lint

GrizzlyAK (commented):
I think I've addressed all of these

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,712 @@
...
+			"unit": "°F",
+			"valueSize": 1,
+			"minValue": 1,
+			"maxValue": 6,
+			"defaultValue": 1
+		},
+		{
+			"#": "12",
+			"label": "Heating Stage 1 Off Threshold",
+			"description": "Sets the delta from setpoint that stage 1 heating stops. Stage 1 turns off at setpoint + Delta Stage 1.",
```
AlCalzone: I'm a bit confused by this description. Does "Delta Stage 1" refer to this parameter or no. 10?

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,712 @@
...
+					"value": 32
+				},
+				{
+					"label": "MECH_F",
+					"value": 64
+				},
+				{
+					"label": "MANUAL_F",
+					"value": 128
+				},
```
AlCalzone: Looking at those values, those terribly look like they are supposed to be used in combination as [partial parameters](https://zwave-js.github.io/node-zwave-js/#/config-files/partial-parameters). Not sure if it makes sense though, maybe just one of these states can be active at any given time.

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,712 @@
...
+				{
+					"label": "Reserved",
+					"value": 256
+				}
```
AlCalzone: remove

packages/config/config/devices/0x014a/tstat-tbz500.json
```diff
@@ -0,0 +1,712 @@
...
+				},
+				{
+					"label": "STATE_MOT",
+					"value": 64
+				},
+				{
+					"label": "STATE_MRT",
+					"value": 128
+				}
+			]
```
AlCalzone, line 322: These also look like they are supposed to be partials, but again I'm not sure if that makes sense here.
Are you able to try if there is ever a value other than the listed ones?

AlCalzone (changes requested):
Just a few more nits :)

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Neither. It actually refers to No. 11. 

No. 11 is a delta from the Setpoint that heat turns ON.  No. 12 is a delta from No. 11 that the Heat turns OFF (= Setpoint + No. 11). I changed the text to match the verbiage in the User Guide in the Setup table. This is what MOST users will use to configure the unit on the device itself. It is more user friendly than the Parameter verbiage for the same settings (which is what I had).

I guess the larger question for me is, are these supposed to be stand alone descriptions that can be understandable in the absence of a user manual? I think that would require a lot more information, unless one is an HVAC expert. I approached it as as "ok, you've read the manual, set up your device on its screen, and now you want to do some more zwave-based fiddling" type of language, and "here's some hints as to what you're changing".

I guess we could replace "Delta Stage 1" with "the Delta from Parameter 11"?

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Hmm... there's nothing in the manual about these parameters other than the above, in a table. Best I can tell, these are states the device can be in and, for the most part, I believe, only one at a time. For example, you can be in Heating Stage 1, or 2 or 3, or Cooling STage 1 or 2, but not any of those at the same time. I'm assuming that "_F" is Fan, since it can be manual. But I don't know what MECH_F and PHANTOM_F mean, but I would think you could have the fan running when either Heating or Cooling is going on.  But I don't think the lot of them fits the partial parameters pattern.

I have a series of questions for my contact at Ecolink who provided this device to me, but would prefer to get these parameters incorporated so I can experiment with them a bit to see how they work. And maybe refine them later with additional knowledge. Right now, I can't change any of them in the CP (although, I perhaps I could do so through zwave.services?).

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: I contemplated not including this parameter, but since it is readOnly, I thought that the unit _might_ actually return it and would appear as "unknown" in the UI if I didn't, right? That's why I left it in. Let me know if you still think it is unnecessary.

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: This one, unlike No. 23, which is clearly meant to be a enable mask for various autosends, appears to be like the previous: just a byte that can contain a value that represents a particular STATE. All of these are mutually exclusive, with the possible exception of STATE_LAST. I don't know what it represents. Both 21 and 22 are triggered by setting the two MSBs in No. 23, so we know these values are reported. Again, I didn't think these fit the partials profile.

GrizzlyAK:
I think I posted this elsewhere, but if you want to see what I'm seeing, you can find the manual [here](https://discoverecolink.com/wp-content/uploads/TBZ500_manual_8_5x5_5_Webversion_RDN1280521.pdf).

packages/config/config/devices/0x014a/tstat-tbz500.json
AlCalzone: I know that many users rely on the defined parameters alone without checking the manual, except for more complicated cases (which this one might be).

I checked the manual and I understand this a bit differently than you just described it. For me it seems like:
Stage 1 starts heating at **Setpoint + Param 11** and stop heating at **Setpoint + Param 12** (which is lower)
Stage 2 starts heating at **Setpoint + Param 13** and stop heating at **Setpoint + Param 14**
etc. (negative for the cooling stages).

If that makes sense, I would recommend this:
**Param 11**:
label: "Heating Stage 1: On Delta"
description: "Stage 1 starts heating at _\<Setpoint> + \<Heating Stage 1: On Delta>_"

**Param 12**:
label: "Heating Stage 1: Off Delta"
description: "Stage 1 stops heating at _\<Setpoint> + \<Heating Stage 1: Off Delta>_"

and so on. What do you think? This way, the params would be self-explanatory without having to refer to the manual or paramter numbers.

packages/config/config/devices/0x014a/tstat-tbz500.json
AlCalzone: The zwavejs2mqtt control panel has a field where you provide the param number, value size and value to set parameters. Your explanation makes sense though, we can always change this later if we find that its necessary.

packages/config/config/devices/0x014a/tstat-tbz500.json
AlCalzone: From my experience, "reserved" usually means not used. Let's remove it, worst case it shows 256 instead of Reserved.

packages/config/config/devices/0x014a/tstat-tbz500.json
AlCalzone: fine with me 👍🏻

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: You are correct! I was wrong in my description (and understanding). After further reading of other thermostat manuals, in particular the RCS TZ45 (which is basically THIS one), I found the following statement:

> Note on Delta Settings : The Delta T Setting is the delta, or difference between, the Setpoint and
current temp for determining when a heat or cool call comes on. The “delta” is the number of
degrees away from Setpoint .

I'm not sure what you mean when you say: "Param 12 (which is lower)". It is truly a 'delta' in that it has no +/-, i.e., direction is inferred by the mode. To restate what I think it means (which I believe is what you said):

Heat:
Stage 1 starts when temp is No. 11 deg below Setpoint and runs until temp is EITHER No. 12 deg below Setpoint (goes off) OR temp is No. 13 deg below Setpoint , at which point Stage 2 starts and continues until temp is EITHER No. 14 deg below Setpoint (goes off) OR temp is No. 15 deg below Setpoint , at which point Stage 3 starts and continues until temp is No 16 deg below Setpoint . 

Cool:
Stage 1 starts when temp is No. 17 above Setpoint and runs until temp is EITHER No. 18 deg above Setpoint  (goes off) OR temp is No. 19 deg above Setpoint, at which point Stage 2 starts and continues until temp is No. 20 deg above Setpoint.

What about this?
Param 11:
label: "Heating Stage 1: On Delta"
description: "Stage 1 starts heating when current temp is <Heating Stage 1: On Delta> deg below \<Setpoint>"

Param 12:
label: "Heating Stage 1: Off Delta"
description: "Stage 1 stops heating when current temp is <Heating Stage 1: Off Delta> deg below \<Setpoint>" 

Param 13:
label: "Heating Stage 2: On Delta"
description: "Stage 2 starts heating when current temp is <Heating Stage 2: On Delta> deg below \<Setpoint>"

Param 14:
label: "Heating Stage 2: Off Delta"
description: "Stage 2 stops heating when current temp is <Heating Stage 2: Off Delta> deg below \<Setpoint>" 

Param 15:
label: "Heating Stage 3: On Delta"
description: "Stage 3 starts heating when current temp is <Heating Stage 3: On Delta> deg below \<Setpoint>"

Param 16:
label: "Heating Stage 3: Off Delta"
description: "Stage 3 stops heating when current temp is <Heating Stage 3: Off Delta> deg below \<Setpoint>" 

Param 17:
label: "Cooling Stage 1: On Delta"
description: "Stage 1 starts cooling when current temp is <Cooling Stage 1: On Delta> deg above \<Setpoint>"

Param 18:
label: "Cooling Stage 1: Off Delta"
description: "Stage 1 stops cooling when current temp is <Cooling Stage 1: Off Delta> deg above \<Setpoint>" 

Param 18:
label: "Cooling Stage 2: On Delta"
description: "Stage 2 starts cooling when current temp is <Cooling Stage 2: On Delta> deg above \<Setpoint>"

Param 20:
label: "Cooling Stage 2: Off Delta"
description: "Stage 2 stops cooling when current temp is <Cooling Stage 2: Off Delta> deg above \<Setpoint>" 

Is that your understanding too? We could simplify some by replacing the <X Stage X: On Delta> blocks, such as:

description: "Stage 2 stops cooling when current temperature is this parameter degrees above \<Setpoint>" 

Thoughts? The "+" sign used in the manual has always been confusing to me.

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Interestingly, for this device I DO NOT have such a field. I have no Config block at all! That's why I can't test any of this via zwave and why i just remembered the services in HA. I haven't tried it yet.

image

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Sounds good. Will remove once we settle the other issues.

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Just tried using the zwave_js.set_config_parameter service in HA Dev Tools and it failed. I tried targeting both the `sensor.node` and the `climate.xxx` entity. ` Failed to call service: unknown error.`

AlCalzone: > Interestingly, for this device I DO NOT have such a field

@robertsLando what could cause this? Do you display the Configuration CC tab only when there are values?

packages/config/config/devices/0x014a/tstat-tbz500.json
AlCalzone: > I'm not sure what you mean when you say: "Param 12 (which is lower)"

I meant the range (0-6) is smaller than for the "ON" part (1-7), so the OFF point is closer to the setpoint than the ON point.

Doesn't matter though, I agree with your parameter descriptions. Lets use the versions that mention the specific labels.
The manual seems to explain the deltas the exact opposite way but that makes no sense IMO.

packages/config/config/devices/0x014a/tstat-tbz500.json
robertsLando: > Do you display the Configuration CC tab only when there are values?

If it's not shown it means there is no value for that CC

packages/config/config/devices/0x014a/tstat-tbz500.json
GrizzlyAK: Will do. Thanks for your patience!

GrizzlyAK:
@zwave-js-bot fix lint

AlCalzone (approved):

AlCalzone:
Thanks for all the work!
From your POV, is this ready to merge?

GrizzlyAK:
Yes!

GrizzlyAK:
I should delete this fork, right? 

I've got VSCode up and running and working to figure out to use it next time and what the right procedures are. Should get easier.

AlCalzone:
The fork is your copy of the repo, you should leave that if you want to continue working on it.
The branch can be removed now

PR #3744, opened 2021-11-17
feat: implement node/route health check

AlCalzone:
This PR adds two methods to the `ZWaveNode` class to test the connection quality between the controller and the node or the node and another node.

#### `checkLifelineHealth` method - test the connection quality between the controller
In multiple rounds, it will test whether re-routing attempts are necessary, how many packets are dropped and (if supported by the node) what the minimum powerlevel for a reliable connection is. These results are combined into a health rating similar to how Silabs' PC controller does it, where 10 is a perfect connection and 0 means _not working at all_:

| Rating | Failed pings | Latency       | No. of neighbors | min. powerlevel | SNR margin |
| -----: | -----------: | ------------: | ---------------: | --------------: | ---------: |
| ✅  10 |            0 |      ≤  50 ms |              > 2 |        ≤ −6 dBm |   ≥ 17 dBm |
| 🟢   9 |            0 |      ≤ 100 ms |              > 2 |        ≤ −6 dBm |   ≥ 17 dBm |
| 🟢   8 |            0 |      ≤ 100 ms |              ≤ 2 |        ≤ −6 dBm |   ≥ 17 dBm |
| 🟢   7 |            0 |      ≤ 100 ms |              > 2 |               - |          - |
| 🟢   6 |            0 |      ≤ 100 ms |              ≤ 2 |               - |          - |
|        |              |               |                  |                 |            |
| 🟡   5 |            0 |      ≤ 250 ms |                - |               - |          - |
| 🟡   4 |            0 |      ≤ 500 ms |                - |               - |          - |
|        |              |               |                  |                 |            |
| 🔴   3 |            1 |     ≤ 1000 ms |                - |               - |          - |
| 🔴   2 |          ≤ 2 |     > 1000 ms |                - |               - |          - |
| 🔴   1 |          ≤ 9 |             - |                - |               - |          - |
|        |              |               |                  |                 |            |
| ❌   0 |           10 |             - |                - |               - |          - |

This check is done in multiple rounds (configurable) where the final health rating is equal to the one of the worst check round.

#### `checkRouteHealth` method - test the connection quality between this node and the target node
In multiple rounds, it will test how many packets are dropped and what the minimum powerlevel for a reliable connection is. This will be done on one or both of the nodes, depending on which ones support `Powerlevel CC`.
These results are combined into a health rating similar to the one above:

| Rating | Failed pings | No. of neighbors | min. powerlevel |
| -----: | -----------: | ---------------: | --------------: |
|  ✅ 10 |            0 |              > 2 |        ≤ −6 dBm |
|   🟢 8 |            0 |              ≤ 2 |        ≤ −6 dBm |
|   🟢 7 |            0 |              > 2 |               - |
|   🟢 6 |            0 |              ≤ 2 |               - |
|        |              |                  |                 |
|   🔴 3 |            1 |                - |               - |
|   🔴 2 |            2 |                - |               - |
|   🔴 1 |          ≤ 9 |                - |               - |
|        |              |                  |                 |
|   ❌ 0 |           10 |                - |               - |

Due to missing insight into re-routing attempts between two nodes or the round-trip-times of their tests, some of the values for the lifeline health rating don't exist here. Furthermore, it is not guaranteed that a route between two nodes and lifeline with the same health rating have the same quality.

### DIFF
--- a/docs/api/node.md
+++ b/docs/api/node.md
@@ -338,6 +338,89 @@ Depending on the number of test frames and involved hops, this may take a while.
 
 > [!ATTENTION] This will throw when the target node is a FLiRS node or a sleeping node that is not awake.
 
+### `checkHealth`
+
+```ts
+checkHealth(
+	rounds?: number,
+	onProgress?: (round: number, totalRounds: number, lastRating: number) => void,
+): Promise<HealthCheckSummary>
+```
+
+Checks the health of connection between the controller and this node and returns the results. The test is done in multiple rounds (1...10, default: 5), which can be configured using the first parameter. To monitor the progress, the optional `onProgress` callback can be used.
+
+The returned object contains the measurements of each round as well as a final rating (which is the worst of all round ratings):
+
+```ts
+interface HealthCheckSummary {
+	/** The check results of each round */
+	results: HealthCheckResult[];
+	/** The health rating expressed as a number from 0 (not working at all) to 10 (perfect connectivity). */
+	rating: number;
+}
+```
+
+where each result looks as follows:
+
+<!-- #import HealthCheckResult from "zwave-js" -->
+
+```ts
+interface HealthCheckResult {
+	/**
+	 * How many route changes were needed. Lower = better, ideally 0.
+	 *
+	 * Only available if the controller supports TX reports.
+	 */
+	routeChanges?: number;
+	/** How many routing neighbors this node has. Higher = better, ideally > 2. */
+	numNeighbors: number;
+	/** How many pings were not ACKed by the node. Lower = better, ideally 0. */
+	failedPingsNode: number;
+	/**
+	 * The minimum powerlevel where all pings from the node were ACKed by the controller. Higher = better, ideally 6dBm or more.
+	 *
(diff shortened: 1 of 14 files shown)


Botched1:
Just a comment - the SiLabs health calculation often provides bad results (low health scores) due to excessive route changes, even when all devices are working correctly. In large networks devices with a large # of neighbors with essentially equal routing weight/quality will sometimes change routes relatively frequently (certainly more often than one would expect), causing the "route change" scoring penalty. SiLabs has previously commented on their forums that some semi-frequent neighbor switching in larger networks is not unexpected.

AlCalzone:
Thanks for the info. I can see though why they included the rerouting in the scoring.
If that happens often, those transmission have a (sometimes much) larger delay than normal.

Botched1:
Agreed - it "should" be a good indicator or routing/delivery issues. I'm definitely not saying that you should change anything in the PR, it is just something to be aware of and watch for on larger systems for those using that health check.

AlCalzone:
In fact I do think directly looking at the latency is a much better indicator of route health.

PR #3746, opened 2021-11-18
fix(config): change some parameter units to seconds for Minoston MP20Z

oblivionratula:
I know the manual says the timer values are in minutes, but I have two of these devices and can confirm the units are seconds.

Pretty new to pull requests, hope this is done properly. Thanks for your work on this.

PR description here


AlCalzone:
@zwave-js-bot rename commit

AlCalzone (approved):
Makes sense, thanks!

PR #3757, opened 2021-11-19
fix: disable soft reset before destroying the driver on failed reconnect

### LINKED ISSUE #3687: Investigate unexpected error handling behavior during soft-reset
Regarding HA addon issue https://github.com/home-assistant/addons/issues/2252.

In a VM, the soft-reset is causing the USB device to go missing and zwave-js fails to re-open the device path. Based on the logs and code inspection, I _think_ there are some problems with how it's handled. The driver is supposed to remember that a soft-reset failed and disable it for the future, but that doesn't seem to happen (there's no log to indicate it). Here's a thought dump on why that may be the case.

Here's the log snippet with line numbers.

```text
1. 2021-11-05T20:07:50.014Z CNTRLR   Performing soft reset...
2. 2021-11-05T20:07:50.033Z CNTRLR   Waiting for the controller to reconnect...
3. 2021-11-05T20:07:51.535Z CNTRLR   Re-opening serial port...
4. 2021-11-05T20:08:00.554Z DRIVER   Failed to open the serial port: Error: No such file or directory, cannot open 
5.                                   /dev/ttyACM0
6. Error in driver ZWaveError: Failed to open the serial port: Error: No such file or directory, cannot open /dev/ttyACM0 (ZW0100)
7.     at Driver.tryOpenSerialport (/usr/src/node_modules/zwave-js/src/lib/driver/Driver.ts:874:17)
8.     at Driver.ensureSerialAPI (/usr/src/node_modules/zwave-js/src/lib/driver/Driver.ts:1846:4)
9.     at Driver.softResetInternal (/usr/src/node_modules/zwave-js/src/lib/driver/Driver.ts:1810:9)
10.     at Driver.initializeControllerAndNodes (/usr/src/node_modules/zwave-js/src/lib/driver/Driver.ts:959:6)
11.     at Immediate.<anonymous> (/usr/src/node_modules/zwave-js/src/lib/driver/Driver.ts:817:5) {
12.   code: 100,
13.   context: undefined,
14.   transactionSource: undefined
15. }
16. Shutting down
17. 2021-11-05T20:08:00.562Z CNTRLR   Waiting for the Serial API to start...
```

Line 16 is coming from the server, when it [handles](https://github.com/zwave-js/zwave-js-server/blob/c1c8b1453763d2344b6179ec257ebacca1943cb1/src/bin/server.ts#L140) a shutdown. That's either a SIGINT/TERM signal, or the zwave-js `"exit"` event. I'm betting on the latter, I don't think any signals are involved.

Line 3 is from the driver, and it means `this.tryOpenSerialport()` was called (no promise). Line 6 is an error condition, and it points to the `"exit"` event being emitted with that message, as the `openPromise` was not provided.

https://github.com/zwave-js/node-zwave-js/blob/c2122b8e99841f4f3c527a2a8ba63a336b71ffd0/packages/zwave-js/src/lib/driver/Driver.ts#L874-L882

The driver is also being destroyed, and the function returns a `false` value. However, the caller, `tryOpenSerialport()` does not check for the return value and proceeds to log the message at line 17.

https://github.com/zwave-js/node-zwave-js/blob/c2122b8e99841f4f3c527a2a8ba63a336b71ffd0/packages/zwave-js/src/lib/driver/Driver.ts#L1843-L1854

It ignores the failure to open the serial port path again, and wants to wait for the API start, which obviously won't happen at this point. Because the server exit handler is running at the same time, it exits right after this log message.

_If_ the server was ignoring the error like it used to, the controller polling probably would have failed and I'd guess that would throw the exception that is expected at the point the failure is remembered.

If any of that is correct, then the only thing I'm not sure about is why there isn't a log message from the driver destroy function (`"destroying driver instance..."`)? The server is definitely calling destroy when it shuts down, and so is `tryOpenSerialport`.

AlCalzone:
fixes: #3687

### DIFF
--- a/packages/zwave-js/src/lib/driver/Driver.ts
+++ b/packages/zwave-js/src/lib/driver/Driver.ts
@@ -798,7 +798,14 @@ export class Driver extends TypedEventEmitter<DriverEventCallbacks> {
 
 		// asynchronously open the serial port
 		setImmediate(async () => {
-			if (!(await this.tryOpenSerialport(spOpenPromise))) return;
+			try {
+				await this.openSerialport();
+			} catch (e) {
+				this.emit("error", e as Error);
+				void this.destroy();
+				return;
+			}
+
 			this.driverLog.print("serial port opened");
 			this._isOpen = true;
 			spOpenPromise.resolve();
@@ -882,9 +889,7 @@ export class Driver extends TypedEventEmitter<DriverEventCallbacks> {
 	private _nodesReady = new Set<number>();
 	private _nodesReadyEventEmitted: boolean = false;
 
-	private async tryOpenSerialport(
-		openPromise?: DeferredPromise<void>,
-	): Promise<boolean> {
+	private async openSerialport(): Promise<void> {
 		let lastError: unknown;
 		// After a reset, the serial port may need a few seconds until we can open it - try a few times
 		for (
@@ -894,7 +899,7 @@ export class Driver extends TypedEventEmitter<DriverEventCallbacks> {
 		) {
 			try {
 				await this.serial!.open();
-				return true;
+				return;
 			} catch (e) {
 				lastError = e;
 			}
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot pack this

AlCalzone:
@kpine can you test this fix?

AlCalzone:
Nevermind, the build failed. Let me fix it.

AlCalzone:
Here we go
@zwave-js-bot pack this

kpine:
Sure, I'll give it a try tonight or tomorrow.

kpine:
Unfortunately I wasn't able to reproduce the original issue. Whatever causes the serial port to be re-opened doesn't happen for me. My test setup is a VirtualBox VM and Docker. Maybe it has something to do with the combination of a VM, and HA Supervisor.

I did try this PR, it successfully detected the inability to do soft-reset. Then when I reloaded it worked again. I do have some weird problem with the VM where it will refuse to work after the soft-reset, until I pull the stick and re-insert it, even with a filter place. 🤷🏻‍♂️

PR #3768, opened 2021-11-21
fix(config): update Zooz ZEN15 Config to FW 1.6+

GrizzlyAK:
This PR updates Parameter #27 to include additional options for later FW versions. Note that the changes to this parameter were introduced in FW v1.3, not v1.6 as suggested in the Issue (per the device's [changelog](https://www.support.getzooz.com/kb/article/309-zen15%C2%A0power-switch-change-log/)).

For reference, the manufacturer's parameter guide is [here](https://www.support.getzooz.com/kb/article/616-zen15-power-switch-ver-1-06-advanced-settings/).

Also fixed an incorrect FW version used in an existing conditional for Parameters #33 and #34. v1.6  (v1.06) is the latest FW, whereas v2.4 refers to HW Version 2.04 (which introduced FW 1.6).

fixes: #3745

PR #3778, opened 2021-11-22
fix(config): clarify ZEN15 param 27 options

AlCalzone:
in response to https://github.com/zwave-js/node-zwave-js/pull/3768#issuecomment-974891169

### DIFF
--- a/packages/config/config/devices/0x027a/zen15.json
+++ b/packages/config/config/devices/0x027a/zen15.json
@@ -137,15 +137,15 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "When device is plugged in, off otherwise",
+					"label": "Always On",
 					"value": 0
 				},
 				{
-					"label": "When switch is on, off when switch is off",
+					"label": "On when switch is on, off when switch is off",
 					"value": 1
 				},
 				{
-					"label": "5 seconds on toggle, off otherwise",
+					"label": "5 seconds following switch activation, then off",
 					"value": 2
 				},
 				{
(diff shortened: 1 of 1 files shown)


AlCalzone:
@GrizzlyAK I think this is better than the extra verbose labels and still capture the gist of whats going on.
What do you think?

packages/config/config/devices/0x027a/zen15.json
```diff
@@ -136,15 +137,15 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Always On",
+					"label": "When device is plugged in, off otherwise",
```
GrizzlyAK: I think some might take "device" to mean either the ZEN15 itself, or something plugged into it, given we are talking about power consumption  of what is plugged _into_ it. I actually think "Always On" here is OK.

packages/config/config/devices/0x027a/zen15.json
```diff
@@ -136,15 +137,15 @@
...
 					"value": 0
 				},
 				{
-					"label": "On when switch is on; Off when switch is off",
+					"label": "When switch is on, off when switch is off",
```
GrizzlyAK: I personally prefer the "on" being included here, as it was.

packages/config/config/devices/0x027a/zen15.json
```diff
@@ -136,15 +137,15 @@
...
 					"value": 1
 				},
 				{
-					"label": "Momentary (at status change)",
+					"label": "5 seconds on toggle, off otherwise",
 					"value": 2
```
GrizzlyAK, line 149: This is OK. You might also consider "5 seconds following switch activation, then off."

GrizzlyAK (commented):
These are my thoughts. Adding the main description helped to clarify things greatly, so that the shorter options didn't feel so truncated.

PR #3792, opened 2021-11-25
fix: don't cancel scheduled polls for value updates caused by the driver

AlCalzone:
#3751 broke the value verification by always canceling it. Now we jump through a few hoops to re-enable it.


AlCalzone:
@zwave-js-bot pack this

PR #3798, opened 2021-11-26
fix(config): several units and re-trigger parameter 5 default on ZP3111-5

luckyjay:
I recently ran into a problem with a ZP3111-5. The device was included and showed a trigger refresh of 3 seconds and so I changed it to 180. Well, then it was only re-triggering once every 3 hours. Digging through what appears to be the manufacturer's manual, they indicate the default is 3 minutes and that parameter 5 is supposed to be minutes.

https://products.z-wavealliance.org/ProductManual/File?folder=&filename=MarketCertificationFiles/2479/ZP3111-5_R2_20170316.pdf


AlCalzone (commented):
I noticed two other odd units, so I hijacked this to update them too

AlCalzone (approved):

PR #3801, opened 2021-11-27
feat(config): additional Product ID for HS-WX300

CWSpear:
I read in the changelog that the HomeSeer HS-WX300 had added support, and I got some new dimmer/switch combos for finishing the basement.

When I added it (via zwavejs2mqtt), it very much didn't look like it was supported. A little digging and it seemed my device had a different ID code. Not 100% sure this is how you add support for this, but thought I'd try to help.

### DIFF
--- a/packages/config/config/devices/0x000c/hs-wx300.json
+++ b/packages/config/config/devices/0x000c/hs-wx300.json
@@ -7,6 +7,10 @@
 		{
 			"productType": "0x4447",
 			"productId": "0x4037"
+		},
+		{
+			"productType": "0x4447",
+			"productId: "0x4036"
 		}
 	],
 	"firmwareVersion": {
--- a/packages/config/config/devices/0x000c/hs-wx300.json
+++ b/packages/config/config/devices/0x000c/hs-wx300.json
@@ -10,7 +10,7 @@
 		},
 		{
 			"productType": "0x4447",
-			"productId: "0x4036"
+			"productId": "0x4036"
 		}
 	],
 	"firmwareVersion": {


CWSpear:
@zwave-js-bot fix lint

AlCalzone (approved):

mick1988:
After these changes my HS-WX300 do not automatically register in Home Assistant with zwavejs2mqtt. Is it a formatting error that it is picking up the new device ID code you added (4037) and it no longer sees the device ID of 4036 which is still the device ID of the HS-WX300's that I received in the mail today? When you view the page on devices.zwave-js.io, it only displays the 4037 ID.

mick1988:
I made a fork where I updated the zwave allaince ID and inclusion/exclusion details. I removed the 4037 device ID, but the correct fix would be to make sure both register. Can you have two device ID's or will it require two devices in the database?

CWSpear:
`0x4036` was the one that was added. It was only just merged a little over a day ago, so make sure your zwavejs2mqtt has the latest database.

I tested before the PR was merged and at least for me, my `0x4036` were able register with both entries there.

mick1988:
Sure enough, once I restarted the service, I got the notification for an update. Thanks for getting the 0x4036 in there.

CWSpear:
Yep! No idea what's up with the different product IDs, but yay for open source and open code. I could get a local fix in a matter of hours and only another day or two before it was distributed to everyone else.

mick1988:
Since I have your attention... Did the 0.31.1 zwavejs2mqtt update break your HS-WX300? It says my database version is still 8.8.3, but they all show as unknown. Restoring 0.31.0 fixes it. Just seeing if you had the same issue or if it's with my install.

PR #3806, opened 2021-11-29
fix(config): update Keemple Smart Radiator

zollek:
Discription, product type and Wake Up command update.

PR description here

### DIFF
--- a/packages/config/config/devices/0x0002/keemple_smart_radiator.json
+++ b/packages/config/config/devices/0x0002/keemple_smart_radiator.json
@@ -5,6 +5,10 @@
 	"description": "Electronic radiator thermostat, intended for use with water based room radiators.",
 	"devices": [
 		{
+			"productType": "0x5fff",
+			"productId": "0xa010"
+		},
+		{
 			"productType": "0x6fff",
 			"productId": "0xa010"
 		}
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0002/keemple_smart_radiator.json
```diff
@@ -12,5 +16,21 @@
 	"firmwareVersion": {
 		"min": "0.0",
 		"max": "255.255"
+	},
+	"paramInformation": [
+		{
+			"#": "132",
+			"label": "Wake Up",
+			"description": "Wake Up Interval",
+			"valueSize": 60,
```
AlCalzone, line 25: That's not a valid value size, must be 1...4 bytes.
Also, please add a unit.

AlCalzone (changes requested):

packages/config/config/devices/0x0002/keemple_smart_radiator.json
```diff
@@ -12,5 +16,21 @@
...
+			"minValue": 60,
+			"maxValue": 1800,
+			"defaultValue": 300
+		}
+	],
```
AlCalzone: This doesn't seem correct. 132 is the ID of the `Wake Up CC`, which has nothing to do with the configuration parameters that are defined through `paramInformation`.

AlCalzone (changes requested):

packages/config/config/devices/0x0002/keemple_smart_radiator.json
zollek: I've hoped that updating this command I can change Wake Up Interval in my battery thermostat.
I don't have any z-wave device that has parameters to configure, so I didn't realize I can't update Wake Up time it this way.
Deprecated OpenZWave integration had an option to set it through UI, do You know how can I change it in Z-Wave JS?

AlCalzone (approved):

packages/config/config/devices/0x0002/keemple_smart_radiator.json
AlCalzone: See [my other comment](https://github.com/zwave-js/node-zwave-js/pull/3807#discussion_r759589266)

PR #3807, opened 2021-11-29
fix(config): add metadata to POPP Smart Radiator

zollek:
Discription and Wake Up command update

PR description here

### DIFF
--- a/packages/config/config/devices/0x0002/010101.json
+++ b/packages/config/config/devices/0x0002/010101.json
@@ -29,17 +29,6 @@
 			["Thermostat Setpoint", "get", 1]
 		]
 	},
-	"paramInformation": [
-		{
-			"#": "132",
-			"label": "Wake Up",
-			"description": "Wake Up Interval",
-			"valueSize": 60,
-			"minValue": 60,
-			"maxValue": 1800,
-			"defaultValue": 300
-		}
-	],
 	"metadata": {
 		"inclusion": "• Ensure smart radiator valve is factory reset. \n• Activate “Inclusion” on the Z-Wave controller. \n• Quickly press and release menu button (one with a dot) on smart radiator valve. \n• Observe both controller and smart radiator valve for status of the process.",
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0002/010101.json
```diff
@@ -28,5 +28,21 @@
...
+		{
+			"#": "132",
+			"label": "Wake Up",
+			"description": "Wake Up Interval",
+			"valueSize": 60,
+			"minValue": 60,
+			"maxValue": 1800,
+			"defaultValue": 300
+		}
+	],
```
AlCalzone: This doesn't seem correct. 132 is the ID of the `Wake Up CC`, which has nothing to do with the configuration parameters that are defined through `paramInformation`.

AlCalzone (changes requested):

AlCalzone (approved):

packages/config/config/devices/0x0002/010101.json
zollek: I've hoped that updating this command I can change Wake Up Interval in my battery thermostat.
I don't have any z-wave device that has parameters to configure, so I didn't realize I can't update Wake Up time it this way.
Deprecated OpenZWave integration had an option to set it through UI, do You know how can I change it in Z-Wave JS?

AlCalzone: If you're using zwavejs2mqtt, it's here: 
grafik

zollek: I'm using Z-Wave JS Supervisor and theres no option like this.
Is there a big difference? Is it worth using zwavejs2mqtt?

AlCalzone: zwavejs2mqtt has a control panel which gives you more access, especially the stuff that's not implemented in the official integration yet. You can also install both (AFAIK) and switch between them, e.g. to do these settings.
You'll have to refer to HA's documentation though because I don't know how.

zollek: Ok, I'll check it out.
Thank You for your help.

PR #3814, opened 2021-11-30
feat(config): add Simon Tech Master Blind Button

craco:
Hi,
New config file for centralized roller blinds buttons by Simon Tech.
Device info: https://www.simonelectric.com/10002081-130-tecla-master-persianas-io-blanco-simon-100.html
And installation instructions: https://static.simonelectric.com/estaticos/extras/ficheros/contenidos/00/00/01/15/15//10002081-13x%20Tecla%20Master%20persianas%20IO_2.pdf

CrAcO

### DIFF
--- a/packages/config/config/devices/0x0267/10002081-13x.json
+++ b/packages/config/config/devices/0x0267/10002081-13x.json
@@ -1,0 +1,259 @@
+{
+	"manufacturer": "SimonTech S.L.U",
+	"manufacturerId": "0x0267",
+	"label": "10002081-13x",
+	"description": "Simon IO: Master Roller Blind",
+	"devices": [
+		{
+			"productType": "0x0005",
+			"productId": "0x01c3"
+		},
+		{
+			"productType": "0x0005",
+			"productId": "0x00f4"
+		}
+	],
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"paramInformation": [
+		{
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0267/10002081-13x.json
```diff
@@ -0,0 +1,259 @@
...
+			"options": [
+				{
+					"label": "0x00 Load is OFF. 0x01 Load is ON.",
+					"value": 176
+				},
+				{
+					"label": "0x00 Load is OFF. 0x01 Load is ON.",
+					"value": 177
+				}
+			]
```
AlCalzone, line 178: I don't understand this. Two options with the same label that seemingly have nothing to do with the values.

AlCalzone (changes requested):
Please read the [style guide](https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide) and apply it to the parameters. There are a bunch of things that don't conform, like:
- Unnecessary descriptions that are very similar to the label
- Typos, missing spaces, etc..
- Labels that are not in *Title Case*
- Ultra long option labels. Try to keep them short while still making clear what they mean.

AlCalzone:
Also, there is an error:
> Parameter #15 is invalid: min/maxValue is incompatible with valueSize 2 (min = -32768, max = 32767).
> Consider converting this parameter to unsigned using "unsigned": true!

PR #3831, opened 2021-12-02
fix(config): 14287-Fan-Controller-Update

rockosocko:
Added parameter for GE fan controller 14287 to set LED light status.


kpine:
I have a 14287 and have never been able to control the LED. This configuration parameter is working for you? What version is your firmware?

AlCalzone (changes requested):
Unless there's a good reason not to, param no. 3 should come before param no. 4

And knowing what kpine asked would be interesting too

kpine:
My fan controller has FW 5.20. I tried this config file but it doesn't work for me. The LED doesn't change state, and the value refreshes are all returning 255 instead of the value that was set. I'm curious if this is just my switch, or if you have different and successful results. There are numerous forum threads that talk about the problem of being unable to change the LED for this model. (I would love to be able to disable it!)

[fan.log](https://github.com/zwave-js/node-zwave-js/files/7644131/fan.log)

rockosocko:
The requested change to put parameter 3 before 4 has been commited.

rockosocko:
I have several of these fan switches, Parameter 3 to turn the LED off appears to work correctly on FW 5.24 but does not appear to work on FW 5.22. I have been looking for a while for a place to download a firmware update.

kpine:
Thanks for confirming. I think that means this parameter should be a conditional based on the firmware version. Unfortunately, I have never heard of Jasco releasing any firmware updates. It's very frustrating.

kpine:
One other question, does your 14287 with 5.24 support S2?

rockosocko:
> One other question, does your 14287 with 5.24 support S2?

I didn't perform a complete re-add for node, but it doesn't look like it used the S2 security.

AlCalzone:
> but it doesn't look like it used the S2 security

The manual and/or node dump should be able to answer that question.

kpine:
It's not a big deal. I just saw a post supposedly [quoting](https://community.smartthings.com/t/ge-jasco-z-wave-fan-speed-control-14287-notification-light-wont-turn-off/117950/22) a Jasco rep.

> This controller does not have an LED Always Off setting.
>
>From Jasco -
>“**Currently we only offer the manual button press method to invert the LED on the Fan Control.** Originally, we did not want to allow any modification of the LED because it was used to determine the fan speed. Many customers requested a feature to add an option to adjust the LED. Instead of adding a parameter we added a button press method because most hubs do not support configuration adjustments. **We will probably add a parameter whenever we move our line to S2 and Smart Start which will be sometime next year.**”

Was just curious if that's what happened or not. It doesn't really matter. The S2 devices are probably a different model number.

AlCalzone:
Before we introduce conflicts here, I think this should become a part of https://github.com/zwave-js/node-zwave-js/pull/3855
/cc @blhoward2

shackrat:
I see there is some ambiguity here.  For the record I have 3 of these fan controls..  Two are newer and are running Firmware v5.24, the last one is a little older and has v5.22.  I can confirm from testing the OP's code that Parameter 3 is working for those which are at 5.24, but not for the one at 5.22.

It might be prudent to publish as a new driver and set firmware min/max as appropriate for both.

kpine:
> It might be prudent to publish as a new driver and set firmware min/max as appropriate for both.

That has already been done. The change contains a conditional statement to only include the param for fw >= 5.24.

packages/config/config/devices/0x0063/14287.json
```diff
@@ -31,6 +31,32 @@
 		}
 	},
 	"paramInformation": [
+		{
+			"$if": "firmwareVersion >= 5.24",
+			"#": "3",
+			"label": "LED Light",
```
blhoward2, line 37: Please make the label "LED Indicator"

packages/config/config/devices/0x0063/14287.json
```diff
@@ -31,6 +31,32 @@
...
+			"description": "Sets when the LED on the switch is lit",
```
blhoward2, line 38: This description is unnecessary. Please delete.

packages/config/config/devices/0x0063/14287.json
```diff
@@ -31,6 +31,32 @@
...
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 2,
+			"defaultValue": 0,
+			"unsigned": true,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "LED On When Light On",
```
blhoward2, line 47: Option labels should be Sentence case. LED should remain fully capitalized.

blhoward2 (changes requested):

PR #3839, opened 2021-12-03
fix(config): Add Heatit Z-Push Button 4 variant product ID

larstobi:
Missing variant. This Product ID is listed at:
https://products.z-wavealliance.org/products/3828?selectedFrequencyId=1

### DIFF
--- a/packages/config/config/devices/0x019b/heatit_z_push_button_4.json
+++ b/packages/config/config/devices/0x019b/heatit_z_push_button_4.json
@@ -6,6 +6,10 @@
 	"devices": [
 		{
 			"productType": "0x0300",
+			"productId": "0xa305"
+		},
+		{
+			"productType": "0x0300",
 			"productId": "0xa306"
 		}
 	],


AlCalzone:
?

larstobi:
@AlCalzone Sorry, I was trying to help a friend add support for his 4 button branded Future Home variant, when I noticed there is som kind of bug in the product ID. So, I thought I would double check what was going on, and so I closed the PR meanwhile.

https://products.z-wavealliance.org/products/3595?selectedFrequencyId=-1
https://products.z-wavealliance.org/products/3828?selectedFrequencyId=-1

As you can see, both 4-button and 8-button have type 0x0300 productID 0xa305 in the catalogue... 

But in the node-zwave-js config, they are both added:

https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x019b/heatit_z_push_button_4.json
https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x019b/heatit_z_push_button_8.json

And it looks like 4-button is 0xa306. Anyway, I'll investigate further and reopen if I find something wrong.

PR #3882, opened 2021-12-11
fix(config): add keemple power plug support

zollek:
Add support for keemple power plug.

I hope this time I don't have so many mistakes.

### DIFF
--- a/packages/config/config/devices/0x041a/keemple_power_plug.json
+++ b/packages/config/config/devices/0x041a/keemple_power_plug.json
@@ -179,6 +179,6 @@
 		"info": "Input voltage: 110-230V AC 50/60Hz \nMax current: 13A \nRated input current:\n   10A, 110-230V, 50/60Hz Continous load current\n   16A, 110-230V AC, 50-60Hz Instantaneous load current ",
 		"inclusion": "• Power up the device.\n• Set Z-Wave Controller into inclusion mode.\n• Press the button 3 times within 1,5s.",
 		"exclusion": "• Power up the device.\n• Set Z-Wave Controller into exnclusion mode.\n• Press the button 3 times within 1,5s.",
-		"reset":"• Power up the device.\n• Press and hold the button for 10s until led lights is on, then relese the button.\nReset successfully, led will blink 5 times."
+		"reset": "• Power up the device.\n• Press and hold the button for 10s until led lights is on, then relese the button.\nReset successfully, led will blink 5 times."
 	}
 }
--- a/packages/config/config/devices/0x041a/keemple_power_plug.json
+++ b/packages/config/config/devices/0x041a/keemple_power_plug.json
@@ -2,7 +2,7 @@
 	"manufacturer": "keemple",
 	"manufacturerId": "0x041a",
 	"label": "KP-SO-02",
-	"description": "keemple Power Plug",
+	"description": "keemple Power Plug\n\nInput voltage: 110-230V AC 50/60Hz \nMax current: 13A \nRated input current:\n   10A, 110-230V, 50/60Hz Continous load current\n   16A, 110-230V AC, 50-60Hz Instantaneous load current",
 	"devices": [
 		{
 			"productType": "0x0200",
@@ -176,7 +176,6 @@
 		}
 	],
 	"metadata": {
-		"info": "Input voltage: 110-230V AC 50/60Hz \nMax current: 13A \nRated input current:\n   10A, 110-230V, 50/60Hz Continous load current\n   16A, 110-230V AC, 50-60Hz Instantaneous load current ",
 		"inclusion": "• Power up the device.\n• Set Z-Wave Controller into inclusion mode.\n• Press the button 3 times within 1,5s.",
 		"exclusion": "• Power up the device.\n• Set Z-Wave Controller into exnclusion mode.\n• Press the button 3 times within 1,5s.",
 		"reset": "• Power up the device.\n• Press and hold the button for 10s until led lights is on, then relese the button.\nReset successfully, led will blink 5 times."
--- a/packages/config/config/devices/0x041a/keemple_power_plug.json
+++ b/packages/config/config/devices/0x041a/keemple_power_plug.json
@@ -2,7 +2,7 @@
 	"manufacturer": "keemple",
 	"manufacturerId": "0x041a",
 	"label": "KP-SO-02",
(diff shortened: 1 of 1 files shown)


zollek:
@zwave-js-bot fix lint

AlCalzone (changes requested):
Please take a good look at our [style guide](https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide) and apply it.

There are a bunch of things wrong here, e.g.
- unnecessary / too long descriptions
- Labels should be in **Title Case**
- Typos
- etc.

knthm:
The Keemple KP-SO-02 is a branded Shenzhen Neo NAS-WR01ZE.

~~I suggest copying that [device's config](https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x0258/nas-wr01ze.json) and replacing the manufacturerId, productType, and productId with the Keemple-specific values.~~

I've a batch of unbranded NAS-WR01ZE's from a chinese distributor on which they accidentally applied the Keemple-brand firmware configuration.

**Update:** I think it's better to import the NAS-WR01ZE device configuration with the template engine and override the branding-specific values. See referenced PR below.

zollek:
I've shortened it using Style guide and @KloudJack  advice.

Sorry for earlier, didn't know that such a guide exist. Simply copied the manual.

packages/config/config/devices/0x041a/keemple_power_plug.json
```diff
@@ -0,0 +1,177 @@
...
+			"unit": "second",
+			"valueSize": 2,
+			"minValue": 30,
+			"maxValue": 32767,
+			"defaultValue": 300
+		},
+		{
+			"#": "8",
+			"label": "Maximum Current",
+			"description": "Maximum output current limit",
```
AlCalzone, line 144: This description seems unnecessary

packages/config/config/devices/0x041a/keemple_power_plug.json
```diff
@@ -0,0 +1,177 @@
...
+			"unit": "A",
+			"valueSize": 1,
+			"minValue": 1,
+			"maxValue": 16,
+			"defaultValue": 16
+		},
+		{
+			"#": "9",
+			"label": "Current Report Treshold",
+			"description": "Current treshold that triggers a report",
```
AlCalzone, line 154: dito

packages/config/config/devices/0x041a/keemple_power_plug.json
```diff
@@ -0,0 +1,177 @@
...
+			"options": [
+				{
+					"label": "Always off",
+					"value": 0
+				},
+				{
+					"label": "Previous state",
+					"value": 1
+				}
+			]
```
AlCalzone, line 44:
```suggestion
			"$import": "~/templates/master_template.json#state_after_power_failure_off_prev"
```

packages/config/config/devices/0x041a/keemple_power_plug.json
```diff
@@ -0,0 +1,177 @@
...
+			"maxValue": 1,
+			"defaultValue": 1,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Disabled",
+					"value": 0
+				},
+				{
+					"label": "Active",
```
AlCalzone, line 60:
```suggestion
					"label": "Enable",
```

packages/config/config/devices/0x041a/keemple_power_plug.json
```diff
@@ -0,0 +1,177 @@
+{
+	"manufacturer": "keemple",
```
AlCalzone, line 2:
```suggestion
	"manufacturer": "Keemple",
```

AlCalzone (changes requested):
Thanks, this is better. Just a few minor changes.

Also, please rename the file to `kp-so-02.json`

AlCalzone:
Oh, I just saw the other PR https://github.com/zwave-js/node-zwave-js/pull/4125/
Let me think about what's the better approach here.

knthm:
> Oh, I just saw the other PR #4125 Let me think about what's the better approach here.

@AlCalzone any update on this?

If importing the whole device config feels too hackish, I could move most of the paramInformation into `~/0x0258/templates/shenzhen_neo_template.json` and import it from there (updating both Shenzhen and Keemple configs).

Almost half of the parameters are `base_enable_disable` which can be imported from the master_template as well.

This way both the NAS-WR01ZE and KP-SO-O2 configs would be structured as follows: https://github.com/zwave-js/node-zwave-js/blob/aff1bc6f1f431e250eb48bf344214a457b0ca290/packages/config/config/devices/0x0371/zw162.json#L2-L35

AlCalzone:
Eh, sorry this slipped through the cracks. We've decided not to import the entire file - from our experience, it sometimes happens that OEM devices deviate from the original firmwares.

I think we have the most flexibility by having both import from a common template like you proposed. Could you do that?

knthm:
@AlCalzone 

After a quick look at the Shenzhen Neo template and configs, I decided to cleanup the device configs: #4186

At first I'd also moved some of the power plug-specific parameters into the Shenzhen Neo manufacturer template, but after noticing that the NAS-WR01Z and NAS-WR01ZE both have functionally-identical parameters all with (annoyingly) different configurations, I removed them again as it felt out of place to have those in the manufacturer template when they only applied to one device from said manufacturer.

See the following parameter as an example:
https://github.com/zwave-js/node-zwave-js/blob/5827bc98fc9577a6586eb3fd58d62f1d38eb71f2/packages/config/config/devices/0x0258/nas-wr01z.json#L68-L77
https://github.com/zwave-js/node-zwave-js/blob/5827bc98fc9577a6586eb3fd58d62f1d38eb71f2/packages/config/config/devices/0x0258/nas-wr01ze.json#L146-L155

or

https://github.com/zwave-js/node-zwave-js/blob/5827bc98fc9577a6586eb3fd58d62f1d38eb71f2/packages/config/config/devices/0x0258/nas-wr01z.json#L48-L57
https://github.com/zwave-js/node-zwave-js/blob/5827bc98fc9577a6586eb3fd58d62f1d38eb71f2/packages/config/config/devices/0x0258/nas-wr01ze.json#L116-L125

:neutral_face:

I've updated #4125 which uses some master template parameters but no Shenzhen Neo template parameters.

PR #3886, opened 2021-12-11
fix(config): correct params for fibaro FGWP102 firmware 3.2

ualex73:
PR description here
The Fibaro FGWP102 configuration parameters does not cover firmware 3.2 correctly, it has too many of them in there. This PR makes those parameters valid for firmware <= 3.1, but exclude them for firmware 3.2.


AlCalzone:
@zwave-js-bot rename commit

AlCalzone (approved):

PR #3903, opened 2021-12-15
fix(config): add fingerprint for AU/NZ model of Aeotec ZW141

sammcewan:
Product type 0x0203 fails to map correctly to this config.

Have included this product type variant so it maps correctly to this config.


AlCalzone:
@zwave-js-bot rename commit

AlCalzone (approved):

PR #3926, opened 2021-12-18
fix(config): correct parameter 40 & 42, Qubino ZMNHTD

ppetru:
The new value size is noted in the device manual and has been tested on
my home network.


AlCalzone:
While you're at it with the manual, the unit for the next parameter seems wrong. Could you check that too?

ppetru:
Done, PTAL.

AlCalzone (approved):

AlCalzone:
Thanks!

PR #3945, opened 2021-12-22
fix: update the SUC node id after self-promotion

dfederm:
This is a small fix to update the controller's `_sucNodeId` with its own node id immediately after promoting itself. Currently the behavior is that the `subNodeId` property remains 0 in this case.


AlCalzone (approved):

AlCalzone:
Thanks!

PR #3950, opened 2021-12-22
fix(config): Add Parameter 17 to Zen71 Switch

Hilo4321:
Adds Parameter 17 to configs for FW version 10.0+.

Toggles local control of Inclusion, Exclusion, and LED settings.


Hilo4321:
fail

PR #3952, opened 2021-12-22
feat(config): add Parameter 17 to Zen71 Switch (firmware dependent)

Hilo4321:
Adds Parameter 17 to configs for FW version 10.0+.

Toggles local control of Inclusion, Exclusion, and LED settings.


Hilo4321:
I can't read what the Semantic issue is, not sure what I need to change here. The title is ok, as far as I can tell.

AlCalzone:
@zwave-js-bot rename commit

AlCalzone (approved):

PR #3972, opened 2021-12-28
refactor: convert node.isControllerNode from method to property

AlCalzone:
While there is no hard style guide governing this, we tend to use readonly properties for static things like a node's IDs and methods for (potentially) dynamic things like endpoint counts. Since the controller node ID doesn't change during usage, this feels more "correct".

/cc @raman325

### DIFF
--- a/docs/api/node.md
+++ b/docs/api/node.md
@@ -150,8 +150,6 @@ getAllEndpoints(): Endpoint[]
 
 This method returns an array of all endpoints on this node. At each index `i` the returned array contains the endpoint instance that would be returned by `getEndpoint(i)`.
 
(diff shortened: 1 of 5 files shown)


docs/api/node.md
```diff
@@ -150,12 +150,6 @@ getAllEndpoints(): Endpoint[]
 
 This method returns an array of all endpoints on this node. At each index `i` the returned array contains the endpoint instance that would be returned by `getEndpoint(i)`.
 
-### `isControllerNode`
-
-```ts
-isControllerNode(): boolean
-```
-
 This is a little utility function to check if this node is the controller.
```
OXINARF: Left-over?

PR #3977, opened 2021-12-28
feat: create symlink to current logfile

### LINKED ISSUE #3796: Create symlink to latest logfile

It's useful to be able to tail the driver log file and watch it for awhile, or grep for certain terms and leave it running overnight, in order to do troubleshooting. The problem is that the log file becomes unavailable when the rotation occurs. Then you need to start over with the new file name.

Create a symlink that always points to the latest logfile. This would make it easier to analyze the latest log messages. Programs that watch the file via the symlink should be able to automatically start watching the new file.

My understanding is that the log file rotation is based on [winston-daily-rotate-file](https://github.com/winstonjs/winston-daily-rotate-file), which has these optional configuration options:

> - **createSymlink:** Create a tailable symlink to the current active log file. (default: false)
> - **symlinkName**: The name of the tailable symlink. (default: 'current.log')

So would it be as simple as turning those options on? (Not sure how this workings in Windows)

None

**Additional context**

N/A

AlCalzone:
fixes:  #3796

### DIFF
--- a/packages/core/src/log/shared.ts
+++ b/packages/core/src/log/shared.ts
@@ -138,7 +138,7 @@ export class ZWaveLogContainer extends winston.Container {
 					path.dirname(require.main.filename),
 					`zwavejs_%DATE%.log`,
 			  )
-			: path.join(__dirname, "../../..", `zwave_%DATE%.log`),
+			: path.join(__dirname, "../../..", `zwavejs_%DATE%.log`),
 		forceConsole: false,
 	};
 
@@ -300,6 +300,8 @@ export class ZWaveLogContainer extends winston.Container {
 		const ret = new DailyRotateFile({
 			filename: this.logConfig.filename,
 			datePattern: "YYYY-MM-DD",
+			createSymlink: true,
+			symlinkName: this.logConfig.filename.replace(`_%DATE%`, "_current"),
 			zippedArchive: true,
 			maxFiles: "7d",
 			format: createDefaultTransportFormat(false, false),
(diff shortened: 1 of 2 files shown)


packages/core/src/log/shared.ts
```diff
@@ -300,6 +300,8 @@ export class ZWaveLogContainer extends winston.Container {
 		const ret = new DailyRotateFile({
 			filename: this.logConfig.filename,
 			datePattern: "YYYY-MM-DD",
+			createSymlink: true,
+			symlinkName: this.logConfig.filename.replace(`_%DATE%`, "_current"),
```
kpine, line 304: What happens if `_%DATE%` is missing from the filename, as if someone customized the filename via the driver options?

kpine:
Poking around the docs, this note about the default filename is out of date, isn't it, as it's now date-base?

```md
-   `filename`: When `logToFile` is `true`, this is the path to the log file. The default file is called `zwave-${process.pid}.log` and located in the same directory as the main executable.
```

https://github.com/zwave-js/node-zwave-js/blob/7469a8c8830fea28228dd9c1b6b1db8817079844/docs/api/driver.md?plain=1#L397

packages/core/src/log/shared.ts
kpine: N/M, I think it's appended to the filename elsewhere.

AlCalzone: yup

PR #3983, opened 2021-12-30
feat(config): add Inteset Door/Window Sensor

cdnninja:
Added Inteset Door/Window Sensor.  I can't find a real device name for the life of me.

### DIFF
--- a/packages/config/config/devices/0x039a/doorWindow.json
+++ b/packages/config/config/devices/0x039a/doorWindow.json
@@ -1,0 +1,115 @@
+{
+	"manufacturer": "Inteset",
+	"manufacturerId": "0x039a",
+	"label": "Inteset Door/Window Sensor",
+	"description": "Door/Window Sensor",
+	"devices": [
+		{
+			"productType": "0x0003",
+			"productId": "0x0101"
+		}
+	],
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"associations": {
+		"1": {
+			"label": "Lifeline", 
+			"maxNodes": 1 
+			"isLifeline": true,
+		},
+		"2": {
+			"label": "Control",
+			"description": "Basic Set", 
+			"isLifeline": true, 
+		},
+
+	}
+	"paramInformation": [
+		{
+			"#": "1",
+			"label": "Basic Set off Delay Time",
+			"unit": "seconds",
+			"valueSize": 2,
+			"minValue": 0,
+			"maxValue": 32767,
+			"defaultValue": 0
+		},
+		{
+			"#": "2",
+			"label": "Basic Set Level",
+			"unit": "%",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 100,
+			"defaultValue": 100
+		},
+		{
+			"#": "3",
+			"label": "Led Indicator",
+			"description": "Indicator on open close event",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 1,
+			"defaultValue": 1
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Disable",
+					"value": 0
+				},
+				{
+					"label": "Enable",
+					"value": 1
+				}
+			]
+		},
+		{
+			"#": "4",
+			"label": " Binary Sensor Report",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 1,
+			"defaultValue": 0,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Disable",
+					"value": 0
+				},
+				{
+					"label": "Enable",
+					"value": 1
+				}
+			]
+
+		},
+		{
+			"#": "5",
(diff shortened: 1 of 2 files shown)


cdnninja:
@zwave-js-bot fix lint

packages/config/config/devices/0x039a/doorWindow.json
```diff
@@ -0,0 +1,91 @@
+{
+	"manufacturer": "Inteset",
+	"manufacturerId": "0x039a",
+	"label": "Inteset Door/Window Sensor",
+	"description": "Door/Window Sensor",
```
AlCalzone: @blhoward2 what should we do about these?

packages/config/config/devices/0x039a/doorWindow.json
```diff
@@ -0,0 +1,91 @@
...
+		}
+	],
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"paramInformation": [
+		{
+			"#": "1",
+			"label": "Basic Set off Delay Time",
```
AlCalzone:
```suggestion
			"label": "Basic Set: Off Delay Time",
```

packages/config/config/devices/0x039a/doorWindow.json
```diff
@@ -0,0 +1,91 @@
...
+			"unit": "seconds",
+			"valueSize": 2,
+			"minValue": 0,
+			"maxValue": 32767,
+			"defaultValue": 0
+		},
+		{
+			"#": "2",
+			"label": "Basic Set Level",
```
AlCalzone:
```suggestion
			"label": "Basic Set: Level",
```

packages/config/config/devices/0x039a/doorWindow.json
```diff
@@ -0,0 +1,91 @@
...
+			"unit": "%",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 100,
+			"defaultValue": 100
+		},
+		{
+			"#": "3",
+			"label": "Led Indicator",
+			"description": "Indicator on open close event",
```
AlCalzone:
```suggestion
			"label": "LED Indicator: Open/Close",
```

packages/config/config/devices/0x039a/doorWindow.json
```diff
@@ -0,0 +1,91 @@
...
+			]
+		},
+		{
+			"#": "5",
+			"label": "Battery Report Interval",
+			"unit": "seconds",
+			"valueSize": 2,
+			"minValue": 1,
+			"maxValue": 1080,
+			"defaultValue": 480
```
AlCalzone: Are these really seconds? Battery reports every 8 minutes sounds like a sure way to drain the battery quickly.

AlCalzone: Edit: This seems to be minutes, not seconds. See the manual you linked.

packages/config/config/devices/0x039a/doorWindow.json
```diff
@@ -0,0 +1,91 @@
...
+			"defaultValue": 0
+		},
+		{
+			"#": "2",
+			"label": "Basic Set Level",
+			"unit": "%",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 100,
+			"defaultValue": 100
```
AlCalzone: 1. This isn't %. Basic Set commands go from 0 to 99.
2. The linked manual states that 100 maps to an "ON" command (0xff). This should be represented as an option

AlCalzone (changes requested):

packages/config/config/devices/0x039a/doorWindow.json
cdnninja: Done.

cdnninja: Looks like the checker isn't happy having options in addition to manual entry.   I don't think this is a partial parameter though rather the value it passes to the second association group?

blhoward2: That's not the issue. You can have an option with manual entry. You just only include allowManualEntry if it's false. It defaults to true. 

```
Parameter #2: allowManualEntry must be false or omitted! (ZW0340)
10
```

packages/config/config/devices/0x039a/doorWindow.json
blhoward2: I've had a few like this and I searched for the actual model number or part number. It's there is truly no model number, we'd just have to repeat it.

packages/config/config/devices/0x039a/doorWindow.json
cdnninja: Gotcha!  I will correct.

packages/config/config/devices/0x039a/doorWindow.json
cdnninja: Found the model number from what I can tell.  I have since included it.

AlCalzone (approved):

Issue #2692, opened 2021-05-24
[bug] NAS-WR01ZE: kWh meter value displayed as negative number


andre-richter:
### Version

**Checklist:**

- [x] I am **not** using HomeAssistant. **Or:** a developer has told me to come here.
- [x] I have checked the troubleshooting section and my problem is **not** described there.
- [x] I have read the changelog problem was **not** mentioned there.

Build/Run method

- [x] Docker
- [ ] PKG
- [ ] Snap package
- [ ] Manually built (git clone - yarn install - yarn run build )

zwavejs2mqtt version: 4.3.1
zwave-js version: 7.5.1

### Describe the bug

The `kWh` meter of the `NAS-WR01ZE` (https://devices.zwave-js.io/?jumpTo=0x0258:0x0200:0x1027:2.32) smart plug is reported/displayed as a negative number.

After resetting the accumulated meter values, sane positive values are reported for sometime, but at some point it flips to negative.

zwavejs2mqtt.log:
```
2021-05-24 20:46:26.806 DEBUG SOCKET: Event ZWAVE_API emitted to FHb1fJMDSyU3lbGJAAAB
2021-05-24 20:46:26.808 INFO ZWAVE: Calling api refreshCCValues with args: [ 42, 50, [length]: 2 ]
2021-05-24 20:46:26.971 INFO ZWAVE: Node 42: value updated: 50-0-value-65537 -21474836.13 => -21474836.13
2021-05-24 20:46:27.120 INFO ZWAVE: Node 42: value updated: 50-0-value-66049 27.85 => 27.49
2021-05-24 20:46:27.266 INFO ZWAVE: Node 42: value updated: 50-0-value-66561 234.62 => 234.93
2021-05-24 20:46:27.410 INFO ZWAVE: Node 42: value updated: 50-0-value-66817 0.25 => 0.25
```
zwavejs.log:
```
2021-05-24T20:46:26.811Z CNTRLR » [Node 042] querying meter value (type = Electric, scale = kWh, rate type = Con
                                  sumed)...
2021-05-24T20:46:26.975Z CNTRLR » [Node 042] querying meter value (type = Electric, scale = W, rate type = Consu
                                  med)...
2021-05-24T20:46:27.123Z CNTRLR » [Node 042] querying meter value (type = Electric, scale = V, rate type = Consu
                                  med)...
[...]

PR #3986, opened 2021-12-30
fix(config): add warning to NAS-WR01ZE about huge negative values

AlCalzone:


AlCalzone:
@zwave-js-bot fix lint

mhulsman:
Hi, I was just encountering this same issue. It was fixed/worked-around before here: 
https://github.com/OpenZWave/open-zwave/pull/2598/commits/8a193383c86a73164ac326018b368527070050a3

Problem is in most sig. bit of the kWh measurement (which is the sign bit, and when set makes the value equal to -2^31 + kWh measurement).  When ignored, things were reported to be ok.  Maybe this work-around can also be used here? I'll open an issue on this.

AlCalzone:
Not necessary, see https://github.com/zwave-js/node-zwave-js/issues/2692

PR #3993, opened 2021-12-31
fix(config): add fingerprint 0x0300:0xa10b to Sunricher SR-ZV9001T4-DIM

grotan1:
I have 2 of these devices and they have productId 0xa10b instead of 0xa10d.


grotan1:
@zwave-js-bot fix lint

AlCalzone (approved):

PR #4011, opened 2022-01-02
feat: handle SmartStart inclusion requests from included nodes

AlCalzone:
This is not the correct way to handle it. According to the specs:

> On receipt of this command, a controller node trying to perform a SmartStart inclusion of a node whose S2 DSK matches the NWI HomeID field of this command should indicate to the application layer that the node to be included is currently included in another network and needs to be removed from the foreign network before it can be included.

grafik


AlCalzone:
@zwave-js-bot pack this

PR #4021, opened 2022-01-03
fix product ID for EU version of Ring Contact Sensor v2

r2xud:
According to the z-wave alliance product DB, the Product ID for the EU version of the Ring Contact Sendor v2 needs to be 0x0301 and not 0x0401.

See: [https://products.z-wavealliance.org/products/4147?selectedFrequencyId=-1](https://products.z-wavealliance.org/products/4147?selectedFrequencyId=-1)


AlCalzone:
0x0301 is the entry above the one you edited. Maybe the ZWA ID doesn't match, but this change isn't correct.

PR #4010, opened 2022-01-01
feat(config): add Shenzen Neo DS07Z door/window switch

fierland:

PR #4033, opened 2022-01-04
feat(config): add Shenzen Neo DS07Z door/window switch

AlCalzone:
Supersedes #4010, but without the specs submodule changes.
Thanks @fierland


AlCalzone:
Or maybe not, wtf

PR #4046, opened 2022-01-05
fix(config): correct lifeline association and param 3 for ZWN-RSM2-PLUS

### LINKED ISSUE #2280: Incorrect device config for ZWN-RSM2-PLUS
https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x011a/zwn-rsm2.json
https://github.com/zwave-js/zwavejs2mqtt/discussions/985#discussioncomment-568172

- 0x011a:0x0111:0x0606 is called ZWN-RSM2-PLUS
- Association groups 2/3 (Switch Binary Report EP1 / 2) need to be auto-assigned

ronytomen:
Fixes: #2280

- Force Endpoint 0 for Group 1 Associations Lifeline so that CurrentValue
  status reports are sent and processed.
- Define the one configuration parameters options as per mfg data

---

Did some experimentation to get this device to report manual local change (someone turns the switch off or on) and found that setting the Endpoint Target to 0 for Group 1 Association (Lifeline) did the trick.  I only found one other example of Endpoint usage in device configurations, so I hope I got this correct.

### DIFF
--- a/packages/config/config/devices/0x011a/zwn-rsm2.json
+++ b/packages/config/config/devices/0x011a/zwn-rsm2.json
@@ -18,45 +18,20 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
-	"endpoints": {
-		"0": {
-			"associations": {
-				"1": {
-					"label": "Lifeline",
-					"maxNodes": 1,
-					"isLifeline": true
-				},
-				"2": {
-					"label": "Switch Binary Report EP1",
-					"maxNodes": 3
-				},
-				"3": {
-					"label": "Switch Binary Report EP2",
-					"maxNodes": 3
-				}
-			}
-		},
+	"associations": {
 		"1": {
-			"associations": {
-				"1": {
-					"$import": "#endpoints/0/associations/1",
-					"isLifeline": false
-				},
-				"2": {
-					"$import": "#endpoints/0/associations/2"
-				}
-			}
+			"label": "Lifeline",
+			"maxNodes": 1,
+			"isLifeline": true,
+			"multiChannel": true
 		},
 		"2": {
-			"associations": {
-				"1": {
-					"$import": "#endpoints/0/associations/1",
-					"isLifeline": false
-				},
-				"2": {
-					"$import": "#endpoints/0/associations/3"
-				}
-			}
+			"label": "Switch Binary Report EP1",
+			"maxNodes": 3
+		},
+		"3": {
+			"label": "Switch Binary Report EP2",
+			"maxNodes": 3
 		}
 	},
 	"paramInformation": [
--- a/packages/config/config/devices/0x011a/zwn-rsm2.json
+++ b/packages/config/config/devices/0x011a/zwn-rsm2.json
@@ -18,19 +18,45 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
-	"associations": {
+	"endpoints": {
+		"0": {
+			"associations": {
+				"1": {
+					"label": "Lifeline",
+					"maxNodes": 1,
+					"isLifeline": true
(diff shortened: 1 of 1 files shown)


ronytomen:
Screen Shot 2022-01-05 at 4 32 53 PM

For the record, attached screenshot of working group association with endpoint target.

AlCalzone:
I think the documentation of that config setting isn't up to date. You should be able to achieve the same result by setting `"multiChannel": true` on the first association group without dabbling with endpoint associations.

AlCalzone (changes requested):
Please try the suggestion, re-interview your device and post a driver log of that. I'm fine with keeping it this way if it won't work, but ideally it should.

ronytomen:
> I think the documentation of that config setting isn't up to date. You should be able to achieve the same result by setting `"multiChannel": true` on the first association group without dabbling with endpoint associations.

Thank you!  I was trying to figure this out, but decided to post the PR for some sage advise.  I will revise, test and update in the next few days.

ronytomen:
Attached is the Interview log for node 38 with the suggested change in place.

[node_38.log](https://github.com/zwave-js/node-zwave-js/files/7834439/node_38.log)

AlCalzone:
Looks OK, does it work for you?

ronytomen:
Just did some more testing.

1.  Cleared the current associations
2. Re-interviewed

No group associations were created and the configuration parameter does not show up.  
Screen Shot 2022-01-09 at 11 35 37 AM
Screen Shot 2022-01-09 at 11 35 15 AM

It is using the modified device configuration, debug for node 38 attached.
[node_38_debug.txt](https://github.com/zwave-js/node-zwave-js/files/7835474/node_38_debug.txt)

packages/config/config/devices/0x011a/zwn-rsm2.json
```diff
@@ -18,19 +18,45 @@
...
-			"maxNodes": 1,
-			"isLifeline": true
+			"associations": {
+				"1": {
+					"$import": "#endpoints/0/associations/1",
+					"isLifeline": false
+				},
+				"2": {
+					"$import": "#endpoints/0/associations/2"
+				}
```
1085986063a: packages/config/config/devices/0x011a/zwn-rsm2.json

AlCalzone:
Weird, your log did show an existing multi channel association

ronytomen:
I don't think that interview had the associations removed.  Let me get the interview with the associations removed before starting.

ronytomen:
I wanted to note that either the Lifeline or per switch association works, but only if Target Endpoint 0 is specified.   Otherwise the status of each of the switches does not update. 

lifeline_association_working
switch_association_working

ronytomen:
Here is the log for Node 38 Interview with associations cleared.  Result, no associations created.
[node_38_interview_20210109_1.log](https://github.com/zwave-js/node-zwave-js/files/7835515/node_38_interview_20210109_1.log)

AlCalzone:
Heh, the device claims it supports Multi Channel Association V2. However, being able to use 0 as the target endpoint was only added in V3 of that CC, so the driver won't do it.

We need to overwrite the reported version aswell, please add that to your config file (before metadata):
```jsonc
"compat": {
	"commandClasses": {
		"add": {
			// The device needs a multi channel lifeline association (target endpoint 0),
			// but only reports support for V2, which doesn't allow this.
			"0x8e": {
				"version": 3
			}
		}
	}
}
```

ronytomen:
Updated device file and tested.  Works as expected creating the correct Lifeline Association with a EndPoint Target 0.

😀

AlCalzone (approved):

PR #4048, opened 2022-01-06
feat(cc): add support for Humidity Control CCs

EiNSTeiN-:
Adds the following CCs:
* Humidity Control Mode
* Humidity Control Operating State
* Humidity Control Setpoint

They are almost identical to the various `Thermostat` CCs, so a lot of the code in this PR is modelled around those.

### DIFF
--- a/packages/config/config/scales.json
+++ b/packages/config/config/scales.json
@@ -10,6 +10,16 @@
 			"unit": "°F"
 		}
 	},
+	"humidity": {
+		"0x00": {
+			"label": "Percentage value",
+			"unit": "%"
+		},
+		"0x01": {
+			"label": "Absolute humidity",
+			"unit": "g/m³"
+		}
+	},
 	"mass": {
 		"0x00": {
 			"label": "Kilogram",
--- a/packages/zwave-js/src/lib/commandclass/HumidityControlModeCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/HumidityControlModeCC.ts
@@ -1,0 +1,383 @@
+import type { ValueID } from "@zwave-js/core";
+import {
+	CommandClasses,
+	enumValuesToMetadataStates,
+	Maybe,
+	MessageOrCCLogEntry,
+	MessageRecord,
+	parseBitMask,
+	validatePayload,
+	ValueMetadata,
+	ZWaveError,
+	ZWaveErrorCodes,
+} from "@zwave-js/core";
+import { getEnumMemberName, pick } from "@zwave-js/shared";
+import type { Driver } from "../driver/Driver";
+import { MessagePriority } from "../message/Constants";
+import {
+	CCAPI,
+	PollValueImplementation,
+	POLL_VALUE,
+	SetValueImplementation,
+	SET_VALUE,
+	throwUnsupportedProperty,
+	throwWrongValueType,
+} from "./API";
+import {
+	API,
+	CCCommand,
+	CCCommandOptions,
+	ccValue,
+	ccValueMetadata,
+	CommandClass,
+	commandClass,
+	CommandClassDeserializationOptions,
+	expectedCCResponse,
+	gotDeserializationOptions,
+	implementedVersion,
+} from "./CommandClass";
+
+// All the supported commands
+export enum HumidityControlModeCommand {
+	Set = 0x01,
+	Get = 0x02,
+	Report = 0x03,
+	SupportedGet = 0x04,
+	SupportedReport = 0x05,
+}
+
+export enum HumidityControlMode {
+	"Off" = 0x00,
+	"Humidify" = 0x01,
+	"De-humidify" = 0x02,
+	"Auto" = 0x03,
+}
+
+@API(CommandClasses["Humidity Control Mode"])
+export class HumidityControlModeCCAPI extends CCAPI {
+	public supportsCommand(cmd: HumidityControlModeCommand): Maybe<boolean> {
+		switch (cmd) {
+			case HumidityControlModeCommand.Get:
+			case HumidityControlModeCommand.SupportedGet:
+				return this.isSinglecast();
+			case HumidityControlModeCommand.Set:
+				return true; // This is mandatory
+		}
+		return super.supportsCommand(cmd);
+	}
+
+	protected [SET_VALUE]: SetValueImplementation = async (
+		{ property },
+		value,
+	): Promise<void> => {
+		const valueDB = this.endpoint.getNodeUnsafe()!.valueDB;
+		if (property === "mode") {
+			if (typeof value !== "number") {
+				throwWrongValueType(
+					this.ccId,
+					property,
+					"number",
+					typeof value,
+				);
+			}
+		} else {
+			throwUnsupportedProperty(this.ccId, property);
+		}
+
+		if (this.isSinglecast()) {
+			this.schedulePoll({ property });
+		}
+	};
+
+	protected [POLL_VALUE]: PollValueImplementation = async ({
+		property,
+	}): Promise<unknown> => {
+		switch (property) {
+			case "mode":
+				return (await this.get())?.[property];
+
+			default:
+				throwUnsupportedProperty(this.ccId, property);
+		}
+	};
+
+	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+	public async get() {
+		this.assertSupportsCommand(
+			HumidityControlModeCommand,
+			HumidityControlModeCommand.Get,
+		);
+
+		const cc = new HumidityControlModeCCGet(this.driver, {
+			nodeId: this.endpoint.nodeId,
+			endpoint: this.endpoint.index,
+		});
+		const response =
+			await this.driver.sendCommand<HumidityControlModeCCReport>(
+				cc,
+				this.commandOptions,
+			);
+		if (response) {
+			return pick(response, ["mode"]);
+		}
+	}
+
+	public async set(mode: HumidityControlMode): Promise<void> {
+		this.assertSupportsCommand(
+			HumidityControlModeCommand,
+			HumidityControlModeCommand.Set,
+		);
+
+		const cc = new HumidityControlModeCCSet(this.driver, {
+			nodeId: this.endpoint.nodeId,
+			endpoint: this.endpoint.index,
+			mode,
+		});
+		await this.driver.sendCommand(cc, this.commandOptions);
+	}
+
+	public async getSupportedModes(): Promise<
+		readonly HumidityControlMode[] | undefined
+	> {
+		this.assertSupportsCommand(
+			HumidityControlModeCommand,
+			HumidityControlModeCommand.SupportedGet,
+		);
+
+		const cc = new HumidityControlModeCCSupportedGet(this.driver, {
+			nodeId: this.endpoint.nodeId,
+			endpoint: this.endpoint.index,
+		});
+		const response =
+			await this.driver.sendCommand<HumidityControlModeCCSupportedReport>(
+				cc,
+				this.commandOptions,
+			);
+		return response?.supportedModes;
+	}
+}
+
+@commandClass(CommandClasses["Humidity Control Mode"])
+@implementedVersion(2)
+export class HumidityControlModeCC extends CommandClass {
+	declare ccCommand: HumidityControlModeCommand;
+
+	public async interview(): Promise<void> {
+		const node = this.getNode()!;
+		const endpoint = this.getEndpoint()!;
+		const api = endpoint.commandClasses[
+			"Humidity Control Mode"
+		].withOptions({
+			priority: MessagePriority.NodeQuery,
+		});
+
+		this.driver.controllerLog.logNode(node.id, {
+			endpoint: this.endpointIndex,
+			message: `Interviewing ${this.ccName}...`,
+			direction: "none",
+		});
+
+		// First query the possible modes to set the metadata
+		this.driver.controllerLog.logNode(node.id, {
+			endpoint: this.endpointIndex,
+			message: "querying supported humidity control modes...",
+			direction: "outbound",
+		});
+
+		const supportedModes = await api.getSupportedModes();
+		if (supportedModes) {
+			const logMessage = `received supported humidity control modes:${supportedModes
+				.map(
+					(mode) =>
+						`\n· ${getEnumMemberName(HumidityControlMode, mode)}`,
+				)
+				.join("")}`;
+			this.driver.controllerLog.logNode(node.id, {
+				endpoint: this.endpointIndex,
+				message: logMessage,
+				direction: "inbound",
+			});
+		} else {
+			this.driver.controllerLog.logNode(node.id, {
+				endpoint: this.endpointIndex,
+				message:
+					"Querying supported humidity control modes timed out, skipping interview...",
+				level: "warn",
+			});
+			return;
+		}
+
+		await this.refreshValues();
+
+		// Remember that the interview is complete
+		this.interviewComplete = true;
+	}
+
+	public async refreshValues(): Promise<void> {
+		const node = this.getNode()!;
+		const endpoint = this.getEndpoint()!;
+		const api = endpoint.commandClasses[
+			"Humidity Control Mode"
+		].withOptions({
+			priority: MessagePriority.NodeQuery,
+		});
+
+		// Query the current status
+		this.driver.controllerLog.logNode(node.id, {
+			endpoint: this.endpointIndex,
+			message: "querying current humidity control mode...",
+			direction: "outbound",
+		});
+		const currentStatus = await api.get();
+		if (currentStatus) {
+			this.driver.controllerLog.logNode(node.id, {
+				endpoint: this.endpointIndex,
+				message:
+					"received current humidity control mode: " +
+					getEnumMemberName(HumidityControlMode, currentStatus.mode),
+				direction: "inbound",
+			});
+		}
+	}
+}
+
+interface HumidityControlModeCCSetOptions extends CCCommandOptions {
+	mode: HumidityControlMode;
+}
+
+@CCCommand(HumidityControlModeCommand.Set)
+export class HumidityControlModeCCSet extends HumidityControlModeCC {
+	public constructor(
+		driver: Driver,
+		options:
+			| CommandClassDeserializationOptions
+			| HumidityControlModeCCSetOptions,
+	) {
+		super(driver, options);
+		if (gotDeserializationOptions(options)) {
+			// TODO: Deserialize payload
+			throw new ZWaveError(
+				`${this.constructor.name}: deserialization not implemented`,
+				ZWaveErrorCodes.Deserialization_NotImplemented,
+			);
+		} else {
+			this.mode = options.mode;
+		}
+	}
+
+	public mode: HumidityControlMode;
+
+	public serialize(): Buffer {
+		this.payload = Buffer.from([this.mode & 0b1111]);
+		return super.serialize();
+	}
+
+	public toLogEntry(): MessageOrCCLogEntry {
+		const message: MessageRecord = {
+			mode: getEnumMemberName(HumidityControlMode, this.mode),
+		};
+		return {
+			...super.toLogEntry(),
+			message,
+		};
+	}
+}
+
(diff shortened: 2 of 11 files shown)


packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
```diff
@@ -0,0 +1,937 @@
...
+export class HumidityControlSetpointCCScaleSupportedReport extends HumidityControlSetpointCC {
+	public constructor(
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+
+		validatePayload(this.payload.length >= 1);
+
+		const supportedScaleIndices = parseBitMask(this.payload, 0);
```
EiNSTeiN-: just a note on this command: most of the commands that pertain to a single setpoint will include the setpoint type in the response and I noticed there are checks to ensure the response is for the same setpoint (e.g. [`testResponseForThermostatSetpointGet`](https://github.com/zwave-js/node-zwave-js/blob/1a8f11a61cfc0fc94ad2fa1fc98e5e740d506b36/packages/zwave-js/src/lib/commandclass/ThermostatSetpointCC.ts#L743)). This response doesn't include the `setpointType` per the spec, and that comes with a few caveats:
1) not possible to ensure the response is for the setpoint type that was requested
2) not possible to save the supported scales for the given setpoint as `states` metadata on the `setpointScale` value, so I do it in `interview()`
3) log message won't mention the setpoint type like the other ones do

packages/config/config/scales.json
```diff
@@ -10,6 +10,16 @@
 			"unit": "°F"
 		}
 	},
+	"humidity": {
```
AlCalzone, line 13: This should be referenced in `sensorTypes.json` aswell, like the temperature scales.

packages/zwave-js/src/lib/commandclass/HumidityControlModeCC.ts
```diff
@@ -0,0 +1,383 @@
...
+					property,
+					"number",
+					typeof value,
+				);
+			}
+		} else {
+			throwUnsupportedProperty(this.ccId, property);
+		}
+
+		if (this.isSinglecast()) {
```
AlCalzone, line 86: You forgot to actually call the API method that sets the mode.

packages/zwave-js/src/lib/commandclass/HumidityControlModeCC.ts
```diff
@@ -0,0 +1,383 @@
...
+		switch (property) {
+			case "mode":
+				return (await this.get())?.[property];
+
+			default:
+				throwUnsupportedProperty(this.ccId, property);
+		}
+	};
+
+	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
AlCalzone: This return type should be simple enough that it can be specified instead of disabling the warning.

packages/zwave-js/src/lib/commandclass/HumidityControlModeCC.ts
```diff
@@ -0,0 +1,383 @@
...
+
+	public serialize(): Buffer {
+		this.payload = Buffer.from([this.mode & 0b1111]);
+		return super.serialize();
+	}
+
+	public toLogEntry(): MessageOrCCLogEntry {
+		const message: MessageRecord = {
+			mode: getEnumMemberName(HumidityControlMode, this.mode),
+		};
```
AlCalzone: nit: I tend to only use the separate `message` const if there are conditional values. You could just put the mode property in the returned object here.

packages/zwave-js/src/lib/commandclass/HumidityControlModeCC.ts
```diff
@@ -0,0 +1,383 @@
...
+	public constructor(
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+
+		this._supportedModes = parseBitMask(
+			this.payload,
+			HumidityControlMode["Off"],
+		);
```
AlCalzone: According to the specs, the bitmask should never contain a bit for the `Off` mode:
grafik

If one should be able to set the mode to `Off`, that mode should be inserted at the start of the array after parsing the bitmask. Please make sure it's not accidentally set before doing that - you never know...

packages/zwave-js/src/lib/commandclass/HumidityControlOperatingStateCC.ts
```diff
@@ -0,0 +1,174 @@
...
+		switch (property) {
+			case "state":
+				return this.get();
+
+			default:
+				throwUnsupportedProperty(this.ccId, property);
+		}
+	};
+
+	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
AlCalzone: Same as for the Mode CC, please specify the return type

packages/zwave-js/src/lib/commandclass/HumidityControlModeCC.ts
```diff
@@ -0,0 +1,383 @@
...
+
+@CCCommand(HumidityControlModeCommand.Report)
+export class HumidityControlModeCCReport extends HumidityControlModeCC {
+	public constructor(
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+
+		validatePayload(this.payload.length == 1);
```
AlCalzone: Always assume that a future CC version will add additional fields, therefore this should be `>=`.
(Side note: use `===` in JS/TS instead of `==`).

packages/zwave-js/src/lib/commandclass/HumidityControlOperatingStateCC.ts
```diff
@@ -0,0 +1,174 @@
...
+
+@CCCommand(HumidityControlOperatingStateCommand.Report)
+export class HumidityControlOperatingStateCCReport extends HumidityControlOperatingStateCC {
+	public constructor(
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+
+		validatePayload(this.payload.length == 1);
```
AlCalzone: Always assume that a future CC version will add additional fields, therefore this should be `>=`.
(Side note: use `===` in JS/TS instead of `==`).

packages/zwave-js/src/lib/commandclass/HumidityControlModeCC.ts
```diff
@@ -0,0 +1,383 @@
...
+
+@CCCommand(HumidityControlModeCommand.SupportedReport)
+export class HumidityControlModeCCSupportedReport extends HumidityControlModeCC {
+	public constructor(
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+
+		this._supportedModes = parseBitMask(
```
AlCalzone, line 331: This is missing a payload length validation

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
```diff
@@ -0,0 +1,939 @@
...
+		cmd: HumidityControlSetpointCommand,
+	): Maybe<boolean> {
+		switch (cmd) {
+			case HumidityControlSetpointCommand.Get:
+			case HumidityControlSetpointCommand.SupportedGet:
+				return this.isSinglecast();
+			case HumidityControlSetpointCommand.Set:
+				return true; // This is mandatory
+			case HumidityControlSetpointCommand.CapabilitiesGet:
+				return this.isSinglecast();
```
AlCalzone: I'd merge this with the Get/SupportedGet case.

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
```diff
@@ -0,0 +1,939 @@
...
+	SupportedGet = 0x04,
+	SupportedReport = 0x05,
+	ScaleSupportedGet = 0x06,
+	ScaleSupportedReport = 0x07,
+	CapabilitiesGet = 0x08,
+	CapabilitiesReport = 0x09,
+}
+
+export enum HumidityControlSetpointType {
+	"N/A" = 0x00,
```
AlCalzone, line 54: Where's this coming from? I couldn't find a reference in the specs.

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
```diff
@@ -0,0 +1,939 @@
...
+		});
+		const response =
+			await this.driver.sendCommand<HumidityControlSetpointCCReport>(
+				cc,
+				this.commandOptions,
+			);
+		if (!response) return;
+		return response.type === HumidityControlSetpointType["N/A"]
+			? // not supported
+			  undefined
```
AlCalzone, line 211: Is this setpoint type actually returned? If so, it could use a comment explaining it. The specs don't seem to mention it.

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
```diff
@@ -0,0 +1,939 @@
...
+		});
+		const response =
+			await this.driver.sendCommand<HumidityControlSetpointCCSupportedReport>(
+				cc,
+				this.commandOptions,
+			);
+		return response?.supportedSetpointTypes;
+	}
+
+	public async getScaleSupported(
```
AlCalzone:
```suggestion
	public async getSupportedScales(
```
I prefer using more "natural" method names for the CC API.

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
```diff
@@ -0,0 +1,939 @@
...
+			endpoint: this.endpoint.index,
+			setpointType,
+		});
+		const response =
+			await this.driver.sendCommand<HumidityControlSetpointCCScaleSupportedReport>(
+				cc,
+				this.commandOptions,
+			);
+
+		return response?.scaleSupported;
```
AlCalzone: Similarly, the property should be called `supportedScales`.

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
```diff
@@ -0,0 +1,939 @@
...
+			);
+			// Find out the capabilities of this setpoint
+			this.driver.controllerLog.logNode(node.id, {
+				endpoint: this.endpointIndex,
+				message: `retrieving capabilities for setpoint ${setpointName}...`,
+				direction: "outbound",
+			});
+			const setpointScaleSupported = await api.getScaleSupported(type);
+			if (setpointScaleSupported) {
+				const logMessage = `received scale supported for setpoint ${setpointName}:
```
AlCalzone:
```suggestion
				const logMessage = `received supported scales for setpoint ${setpointName}: 
```

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
```diff
@@ -0,0 +1,939 @@
...
+		}
+	}
+
+	public setpointType: HumidityControlSetpointType;
+	public value: number;
+	public scale: number;
+
+	public serialize(): Buffer {
+		const override =
+			this.getNodeUnsafe()?.deviceConfig?.compat?.overrideFloatEncoding;
```
AlCalzone: I'd prefer to leave this out for now unless there are devices which need this. This is only used for some Heatit thermostats so far.

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
```diff
@@ -0,0 +1,939 @@
...
+		super(driver, options);
+
+		validatePayload(this.payload.length >= 1);
+		this._type = this.payload[0] & 0b1111;
+		if (this._type === 0) {
+			// Not supported
+			this._value = 0;
+			this._scale = getScale(this.driver.configManager, 0);
+			return;
+		}
```
AlCalzone, line 556: I'm not sure about this. Does this happen?

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
```diff
@@ -0,0 +1,939 @@
...
+			},
+		};
+	}
+}
+
+function testResponseForHumidityControlSetpointGet(
+	sent: HumidityControlSetpointCCGet,
+	received: HumidityControlSetpointCCReport,
+) {
+	// We expect a Thermostat Setpoint Report that matches the requested setpoint type
```
AlCalzone: Not thermostat :)

EiNSTeiN-:
I pushed fixes for the failing test.

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
AlCalzone: Yay, consistent specs 🎉

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
```diff
@@ -0,0 +1,939 @@
...
+			getScale(this.driver.configManager, scale),
+		);
+
+		this.persistValues();
+	}
+
+	private _scaleSupported: Scale[];
+	public get scaleSupported(): Scale[] {
+		return this._scaleSupported;
+	}
```
AlCalzone:
```suggestion
	public readonly supportedScales: Scale[];
```

AlCalzone: (like I mentioned above)

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
```diff
@@ -0,0 +1,939 @@
...
+export class HumidityControlSetpointCCScaleSupportedReport extends HumidityControlSetpointCC {
+	public constructor(
+		driver: Driver,
+		options: CommandClassDeserializationOptions,
+	) {
+		super(driver, options);
+
+		validatePayload(this.payload.length >= 1);
+
+		const supportedScaleIndices = parseBitMask(this.payload, 0);
```
AlCalzone: This should make sure not to parse any unrelated bits, e.g. like this:
```suggestion
		const supportedScaleIndices = parseBitMask(Buffer.from([this.payload[0] & 0b1111]), 0);
```

AlCalzone (changes requested):

packages/zwave-js/src/lib/commandclass/HumidityControlModeCC.ts
EiNSTeiN-: I changed this one to return `response?.mode` rather than use `pick()`, similar to the Operating State. With quick search through the code I see that most methods that `return pick()` also disable the warning. What other fix did you have in mind? Perhaps adding an interface woult work, I can do that for the Setpoint `get()` api?

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
EiNSTeiN-: Copied from `ThermostatSetpointType` which is done the same way. It's used in a few places in this CC as well as ThermostatSetpointCC where it's copied from, like for providing the starting value for `parseBitMask` and the `get()` api method.

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
EiNSTeiN-: For the same reason `N/A` is set in the enum at index 0... There's no type 0 defined in the spec so if the command returns a response with type 0 for any reason, we respond with undefined... Copied from the Thermostat Setpoint CC which works the same way https://github.com/zwave-js/node-zwave-js/blob/583645c6618ae1abb2666eb81806f8d73c97d719/packages/zwave-js/src/lib/commandclass/ThermostatSetpointCC.ts#L233-L241

Screenshot from 2022-01-09 17-18-37

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
EiNSTeiN-: There's no setpoint type 0 defined in the spec, I guess this is just a guard to ensure we have a sane state if a device returns type 0. Like with the other places this is handled, this comes from Thermostat Setpoint CC which works the same (no type 0). Happy to change if you want, but how should this be handled?

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
EiNSTeiN-: removed from 2 places

EiNSTeiN-:
Updated!

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
AlCalzone: Guess it doesn't hurt to keep this in for now. Maybe add a comment that this is a fallback for behavior that's not defined in the specs.

EiNSTeiN-:
```suggestion
		// Setpoint type 0 is not defined in the spec, prevent devices from using it.
		if (this._type === 0) {
			// Not supported
			this._value = 0;
			this._scale = getScale(this.driver.configManager, 0);
			return;
		}
```

EiNSTeiN-:
Anything else I can do for this one?

packages/zwave-js/src/lib/commandclass/HumidityControlOperatingStateCC.ts
AlCalzone: This didn't change, did it? Or is Github showing me nonsense?

Edit: Looks like Github is displaying nonsense.

packages/zwave-js/src/lib/commandclass/ThermostatSetpointCC.ts
```diff
@@ -629,11 +629,9 @@ export class ThermostatSetpointCCSet extends ThermostatSetpointCC {
 
 	public serialize(): Buffer {
 		// If a config file overwrites how the float should be encoded, use that information
-		const override =
-			this.getNodeUnsafe()?.deviceConfig?.compat?.overrideFloatEncoding;
 		this.payload = Buffer.concat([
 			Buffer.from([this.setpointType & 0b1111]),
-			encodeFloatWithScale(this.value, this.scale, override),
+			encodeFloatWithScale(this.value, this.scale),
```
AlCalzone: This was supposed to be here, but not in the Humidity CCs

packages/zwave-js/src/lib/commandclass/HumidityControlSetpointCC.ts
AlCalzone: AFAIK the Thermostat CCs actually define a setpoint type N/A, but I'm okay with this.

AlCalzone (changes requested):
Sorry, got caught up in other stuff. Just undo the change in ThermostatSetpointCC and this should be good to merge.

packages/zwave-js/src/lib/commandclass/ThermostatSetpointCC.ts
EiNSTeiN-: oops! I didn't mean to edit this one

packages/zwave-js/src/lib/commandclass/HumidityControlOperatingStateCC.ts
EiNSTeiN-: FYI, `ThermostatFanStateCCReport` uses `this.payload.length == 1`, likely this used to be a copy/paste from that other file. There are a few more instances of this around the code base.

AlCalzone (approved):

AlCalzone:
Big thanks for doing this!

_Side note:_ You might wanna work in a branch next time. This will be squash-merged, after which you'll have to hard-reset your master branch to mine, if you want to avoid a bunch of unnecessary merge commits.

PR #4049, opened 2022-01-06
feat(config): add Ring Outdoor Contact Sensor, correct Contact Sensor Gen2

### LINKED ISSUE #4035: Missing device configuration: Ring Outdoor Contact Sensor
### Checklist

### Which device is missing?

Ring Outdoor Contact Sensor

### Manufacturer ID

0x0346

### Product Type

0x0c01

### Product ID

0x0301

### Firmware Version

1.24

No

### Do you have a manual?

* [Ring Alarm Outdoor Contact Sensor Zwave.pdf](https://github.com/zwave-js/node-zwave-js/files/7811393/Ring.Alarm.Outdoor.Contact.Sensor.Zwave.pdf) (found [here](https://support.ring.com/hc/en-us/articles/4414669191572))
* [Product Page](https://ring.com/products/alarm-outdoor-contact-sensor)

ahutchings:
Fix #4035

### DIFF
--- a/packages/config/config/devices/0x0346/outdoor_contact_sensor.json
+++ b/packages/config/config/devices/0x0346/outdoor_contact_sensor.json
@@ -76,7 +76,7 @@
 		},
 		{
 			"#": "5",
-			"label": "Force Wakeup After Delay",
+			"label": "One-Time Wakeup Delay",
 			"description": "Device will wakeup once after the configured delay",
 			"valueSize": 2,
 			"unit": "seconds",
--- a/packages/config/config/devices/0x0346/outdoor_contact_sensor.json
+++ b/packages/config/config/devices/0x0346/outdoor_contact_sensor.json
@@ -1,0 +1,106 @@
+{
+	"manufacturer": "Ring",
+	"manufacturerId": "0x0346",
+	"label": "B0923BK77S",
+	"description": "Outdoor Contact Sensor",
+	"devices": [
+		{
+			"productType": "0x0c01",
+			"productId": "0x0301"
+		}
+	],
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 5,
+			"isLifeline": true
+		}
+	},
+	"paramInformation": [
+		{
+			"#": "1",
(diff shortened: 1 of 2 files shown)


packages/config/config/devices/0x0346/outdoor_contact_sensor.json
```diff
@@ -0,0 +1,106 @@
...
+				{
+					"label": "On when activated or cleared",
+					"value": 2
+				}
+			]
+		},
+		{
+			"#": "5",
+			"label": "Force Wakeup After Delay",
+			"description": "Device will wakeup once after the configured delay",
```
AlCalzone: Did you copy this from a different device config? If not, I'd call this "Wakeup Interval" or something like that.

AlCalzone (changes requested):
Thanks, one question though

packages/config/config/devices/0x0346/outdoor_contact_sensor.json
ahutchings: Yes, I reused all of the parameter labels from the [Ring Contact Sensor Gen2](https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x0346/contact_sensor_gen2.json#L85).

The manual for both has this as the description:

> One shot timer:
> Writing to this parameter prompts the sensor to send a wakeup notification one time after this parameter’s number of seconds. After which it is reset back to 0.

If you'd like I can update both labels in this PR.

AlCalzone: Ohh that is completely different than what I understood. Please do, you can include it in this PR.

ahutchings: What should it be updated to? "Force Wakeup After Delay" sounds like a boolean, so it would make sense to me to change it. Is the text from the manual ("One Shot Timer") clear?

AlCalzone: Something like _One-Time Wakeup Delay_ maybe?

ahutchings: Sounds good. Pushed a commit with updated text for both devices.

AlCalzone (approved):

PR #4051, opened 2022-01-07
fix(config): fix configurations for Aeotec ZW111 and FT111

binzhou:
For ZW111, change param 81 and 82 option to Multilevel Switch CC Set (was incorrectly Report - see https://aeotec.freshdesk.com/helpdesk/attachments/6102435346) and fixed label for control destination s2.
For FT111, change param 81 and 82 to none/basic set only (see https://manuals-backend.z-wave.info/make.php?lang=en&sku=FT111-C&cert=ZC10-17035487)

BREAKING_CHANGE: FT111 now has one option less for config param 81/82 compared to before, however this matches all available documentation.

### DIFF
--- a/packages/config/config/devices/0x0086/templates/aeotec_template.json
+++ b/packages/config/config/devices/0x0086/templates/aeotec_template.json
@@ -300,7 +300,29 @@
 			}
 		]
 	},
-	"notification_report_type": {
+	"multilevel_report_type": {
+		"valueSize": 1,
+		"minValue": 0,
+		"maxValue": 2,
+		"defaultValue": 1,
+		"unsigned": true,
+		"allowManualEntry": false,
+		"options": [
+			{
+				"label": "Disable",
+				"value": 0
+			},
+			{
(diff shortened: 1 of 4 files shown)


packages/config/config/devices/0x0086/templates/aeotec_template.json
```diff
@@ -300,7 +300,7 @@
 			}
 		]
 	},
-	"notification_report_type": {
+	"multilevel_report_type": {
```
AlCalzone, line 303: Did you check that this isn't used anywhere else?

AlCalzone:
@blhoward2

packages/config/config/devices/0x0086/templates/aeotec_template.json
binzhou: Yes - I touched all the files from the search result https://github.com/zwave-js/node-zwave-js/search?q=notification_report_type

AlCalzone (approved):

PR #4052, opened 2022-01-07
feat(config): add ecolink tilt-zwave5

tmjpugh:
Add new ecolink tilt sensor 
changed from other sensors.  verified device configuration values match actual unit and manual


tmjpugh:
@zwave-js-bot fix lint

AlCalzone (changes requested):
Please make a new branch where you only add the new file and choose `master` as the target branch. This is unrelated to the Trane PR.

PR #4059, opened 2022-01-09
feat(serial): log discarded bytes when reading invalid/unexpected data

fallenpegasus:
SerialAPIParser silently drops unexpected bytes when reading a frame.
This makes it not silent, and logs it when it happens instead.


packages/serial/src/SerialAPIParser.ts
```diff
@@ -53,6 +53,10 @@ export class SerialAPIParser extends Transform {
 						// INS12350: A host or a Z-Wave chip waiting for new traffic MUST ignore all other
 						// byte values than 0x06 (ACK), 0x15 (NAK), 0x18 (CAN) or 0x01 (Data frame).
 						// Just skip this byte
+
+						// we should log this
+						// because corrupted message, a firmware bug, or a logic problem
+						this.logger.whut("inbound", this.receiveBuffer);
```
AlCalzone: I prefer not to create a log line for each invalid byte. If we do this, we should skip ahead to the end of the buffer or the next valid byte (whichever comes first) and log the entire buffer slice.

packages/serial/src/Logger.ts
```diff
@@ -53,6 +53,23 @@ export class SerialLogger extends ZWaveLoggerBase<SerialLogContext> {
...
 	}
 
+	/**
+	 * Logs receipt of an unexpected byte, instead of a ACK, NAK, CAN, or data frame
+	 */
+        public whut(direction: DataDirection, data: Buffer): void {
+		if (this.isVisible())
+			this.logger.log({
+				level: SERIAL_LOGLEVEL,
+				message: `garble. unexpected byte 0x${data[0].toString("hex")}`,
```
AlCalzone: This doesn't build. You can use the `buffer2hex` utility to print the entire buffer as hex with leading `0x`. Requires the input data to be sliced correctly though.

AlCalzone (changes requested):

AlCalzone (approved):

PR #2175, opened 2021-03-27
fix(config): correct Evolve LRM-AS

cdnninja:
LRM-AS is the same as WD500Z.  Updating to reflect same functionality.

PR #4081, opened 2022-01-14
fix(config): update Z-Uno definition for custom configuration parameters

robin-thoni:
Add custom parameters for the Z-Uno.

The Z-Uno is an Arduino/ESP-like custom development board. It features 32 2-bytes user-configurable Z-Wave configuration slots. Unfortunately, the Home Assistant [Z-Wave JS](https://www.home-assistant.io/integrations/zwave_js#service-zwave_jsset_config_parameter) integration won't allow to set a parameter that is not defined in the backend:

image

```
Logger: homeassistant.helpers.script.websocket_api_script
Source: components/zwave_js/services.py:354
First occurred: 2:36:15 PM (2 occurrences)
Last logged: 3:47:59 PM

websocket_api script: Error executing script. Unexpected error for call_service at pos 1: Configuration parameter with value ID 32-112-0-64 could not be found
websocket_api script: Error executing script. Unexpected error for call_service at pos 1: Configuration parameter with value ID 33-112-0-64 could not be found
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 381, in _async_step
    await getattr(self, handler)()
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 584, in _async_call_service_step
    await service_task
  File "/usr/src/homeassistant/homeassistant/core.py", line 1495, in async_call
    task.result()
  File "/usr/src/homeassistant/homeassistant/core.py", line 1530, in _execute_service
    await handler.job.target(service_call)
  File "/usr/src/homeassistant/homeassistant/components/zwave_js/services.py", line 354, in async_set_config_parameter
    zwave_value, cmd_status = await async_set_config_parameter(
  File "/usr/local/lib/python3.9/site-packages/zwave_js_server/util/node.py", line 65, in async_set_config_parameter
    raise NotFoundError(
zwave_js_server.exceptions.NotFoundError: Configuration parameter with value ID 32-112-0-64 could not be found
```

Not only this allows to set the parameters through HA's service, but this also allows to quickly display all of them, rather than having to go through the `Custom Configuration` input (which, BTW, is currently broken, as #2175 hasn't made it yet to HA).

Manufacturer doc: https://z-uno.z-wave.me/z-wave/configuration-parameters/ and https://z-uno.z-wave.me/Reference/ZUNO_SETUP_CFGPARAMETER_HANDLER/

Note: as opposed to what's stated in the doc:
> Parameters on the Z-Wave controller side are numbered from 64 to 96

Only `32` parameters are available, thus, only going up to `95` included.

### DIFF
--- a/packages/config/config/devices/0x0115/v3.json
+++ b/packages/config/config/devices/0x0115/v3.json
@@ -204,6 +204,262 @@
 					"value": 1
 				}
 			]
+		},
+		{
+			"#": "64",
+			"label": "Custom parameter 64",
+			"valueSize": 2,
+			"minValue": -32768,
+			"maxValue": 32767,
+			"defaultValue": 0
+		},
+		{
+			"#": "65",
+			"label": "Custom parameter 65",
+			"valueSize": 2,
+			"minValue": -32768,
+			"maxValue": 32767,
+			"defaultValue": 0
+		},
+		{
+			"#": "66",
+			"label": "Custom parameter 66",
+			"valueSize": 2,
+			"minValue": -32768,
+			"maxValue": 32767,
+			"defaultValue": 0
+		},
+		{
+			"#": "67",
+			"label": "Custom parameter 67",
+			"valueSize": 2,
+			"minValue": -32768,
+			"maxValue": 32767,
+			"defaultValue": 0
+		},
+		{
+			"#": "68",
+			"label": "Custom parameter 68",
+			"valueSize": 2,
+			"minValue": -32768,
+			"maxValue": 32767,
+			"defaultValue": 0
+		},
+		{
+			"#": "69",
+			"label": "Custom parameter 69",
+			"valueSize": 2,
+			"minValue": -32768,
+			"maxValue": 32767,
+			"defaultValue": 0
+		},
+		{
+			"#": "70",
+			"label": "Custom parameter 70",
+			"valueSize": 2,
+			"minValue": -32768,
+			"maxValue": 32767,
+			"defaultValue": 0
+		},
+		{
+			"#": "71",
+			"label": "Custom parameter 71",
+			"valueSize": 2,
+			"minValue": -32768,
+			"maxValue": 32767,
+			"defaultValue": 0
+		},
+		{
+			"#": "72",
+			"label": "Custom parameter 72",
+			"valueSize": 2,
+			"minValue": -32768,
+			"maxValue": 32767,
+			"defaultValue": 0
+		},
+		{
+			"#": "73",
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0115/v3.json
```diff
@@ -204,6 +204,262 @@
 					"value": 1
 				}
 			]
+		},
+		{
+			"#": "64",
```
marcus-j-davies, line 209: These labels are going to need to differentiate from each other.
In various UI's they will be provided - and having them all the same won't make for a nice user experience.

You could also maybe template these params - given how similar they are, will make it much easier for you. 👍 

```json
{
  "#": "64",
  "label": "Custom parameter 1",
  "valueSize": 2,
  "minValue": -32768,
  "maxValue": 32767,
  "defaultValue": 0
},
{
  "#": "65",
  "label": "Custom parameter 2",
  "valueSize": 2,
  "minValue": -32768,
  "maxValue": 32767,
  "defaultValue": 0
}
```

marcus-j-davies (changes requested):

AlCalzone (changes requested):
What @marcus-j-davies said :)

AlexCPU:
@marcus-j-davies & @AlCalzone, the issue here is the Z-Uno device is a little unique in that it's an Arduino platform that you can program to do whatever you want, so the code that I run on my Z-Uno might use custom parameter 1 to calibrate a temperature sensor, but someone else with the same device might run different code that uses the same parameter for something completely different (e.g. setting a voltage threshold for an alarm).
Having one of these devices in use, I think what @robin-thoni has proposed is both necessary, and correct (whilst admitting it's weird!)
See the config parameters section of the guide: https://z-uno.z-wave.me/z-wave/configuration-parameters/

marcus-j-davies:
Hi,

The issue is not so much what the param does or how it is used -  its more to make them unique.
We cant have 30+ params all labelled 'Custom parameter' 

They need to be different in their labels i.e **Custom Parameter 1**, **Custom Parameter 2**
what they do (in this special case) is irrelevant.

The change required is cosmetic only.

Example:

This - is not good.
image

This is.
image

AlexCPU:
Sorry, didn't read enough, ignore me!

robin-thoni:
Ah! I didn't realized they could be displayed that way. In Z-Wave JS 2 MQTT, the param number is displayed anyway. Should be good now.

Thanks for the review!

AlCalzone (approved):

PR #4090, opened 2022-01-15
feat(config): added roller blind switch type to Fibaro FGD211 FW2.2+

mmillmor:
feat: Added the roller blind configuration option

Parameter 14 has an option for Roller Blind Switch for devices with firmware 2.2 and later


packages/config/config/devices/0x010f/fgd211.json
```diff
@@ -230,6 +231,32 @@
...
 			]
 		},
+		{
+			"#": "14",
+			"$if": "firmwareVersion >= 2.2",
+			"label": "Inputs Button/Switch configuration",
+			"description": "Binary inputs type configuration",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 255,
```
AlCalzone: Please change this to 2 and for the <=2.1 variant to 1. `maxValue` should match the options.

AlCalzone (changes requested):
Just one little detail

packages/config/config/devices/0x010f/fgd211.json
mmillmor: Thanks for the suggestion. I have updated it

AlCalzone (approved):

PR #4107, opened 2022-01-18
fix(config): add association to Leviton DZ15S switch

EiNSTeiN-:
I have many of these switches at home and it seems like the association group is missing, I only have a Lifeline group which doesn't seem to work for grouping switches together. I checked other configuration files and this seems to be the way to define an association group (please confirm!). This is what the [DZ15S manual](https://www.leviton.com/en/docs/DI-000-DZ15S-02A-W.pdf) says:

Screenshot from 2022-01-18 13-16-36

### DIFF
--- a/packages/config/config/devices/0x001d/dz15s.json
+++ b/packages/config/config/devices/0x001d/dz15s.json
@@ -13,6 +13,13 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
+	"associations": {
+		"1": {
+			"label": "Group 1",
+			"maxNodes": 5,
+			"isLifeline": true
+		}
+	},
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x001d/dz15s.json
```diff
@@ -13,6 +13,13 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
+	"associations": {
+		"1": {
+			"label": "Group 1",
```
AlCalzone:
```suggestion
			"label": "Lifeline",
```

AlCalzone (changes requested):

AlCalzone (approved):

PR #4117, opened 2022-01-21
fix(config): add iblinds V3/V3.1 parameters (8-10)

chancehiblinds:
Add new parameters for the latest firmware V3.07/V3.12 of iblinds V3/V3.1 (respectively)
Parameters:
8: Minimum Tilt Level
9: Maximum Tilt Level
10: Remap 99

Chance H.
**Lead Developer** at _HAB Home Intelligence_


packages/config/config/devices/0x0287/iblindsv3.json
```diff
@@ -98,14 +100,78 @@
 				}
 			]
 		},
-		"6": {
-			"label": "Speed",
+		"6":{
+			"label": "Movement Duration",
 			"description": "How fast iblinds will close in seconds",
```
AlCalzone: This description can be removed

packages/config/config/devices/0x0287/iblindsv3.json
```diff
@@ -98,14 +100,78 @@
...
+			"$if": "firmwareVersion >= 3.7",
+			"label": "Maximum Tilt Level",
+			"valueSize": 1,
+			"minValue": 75,
+			"maxValue": 99,
+			"defaultValue": 99,
+			"allowManualEntry": true,
+			"writeOnly": true,
+			"unit":"Tilt %"
+		},
```
AlCalzone: I'd remove the units here.

packages/config/config/devices/0x0287/iblindsv3.json
```diff
@@ -98,14 +100,78 @@
...
+		"10":{
+			"$if": "firmwareVersion >= 3.7",
+			"label": "Remap 99",
+			"description":"Remap ON to 0x63 instead of 0xFF. Use if ON sends tilt all the way up instead of 50% or default ON.",
```
AlCalzone: Which command(s) does this apply to? I'd like to make this label/description a bit more concise.

AlCalzone (changes requested):

AlCalzone:
Also this seems to have merge conflicts. If you can, please reset your branch to our master branch and then apply the changes in a single commit.

packages/config/config/devices/0x0287/iblindsv3.json
chancehiblinds: This applies to the Basic Switch (ON/OFF), and remaps 0X63 (99%) to Default On (Parameter 4) Value  (Default 50%). Some Z-Wave controllers send (0x63, instead of 0XFF for ON and this causes blind when commanded to open/ON to tilt to the maximum end point 99%. 
The description I wrote is actually backwards,
Remap Default ON to 0xFF instead of 0x63

AlCalzone (approved):
Thanks for the clarification. I've update the label/description to something that feels a bit more approachable to us.

AlCalzone:
Also, your fork seems to be a bit out of date (still had the old config format). I suggest you re-create it after this is merged and sync it before each PR to prevent conflicts.

chancehiblinds:
Thanks @AlCalzone , always a pleasure

-Chance

PR #4125, opened 2022-01-23
feat(config): add config for Keemple KP-SO-02 Smart Socket

knthm:
This device is an exact match with the Shenzhen Neo NAS-WR01ZE.~~I've imported its config from the respective device config file and overridden only the brand-specific fields.~~

*Source: I've several NAS-WR01ZE devices from a Chinese dealer that accidentally provided me unbranded units with firmware configured for Keemple resale*

~~I'm not sure if this is the recommended way of handling rebranded devices that report a different manufacturerId, but if the maintainers agree, maybe we can add a config-files documentation entry describing this use-case.~~

Device configuration parameters are identical to the updated Shenzhen Neo NAS-WR01ZE in #4186.

Closes: #3882

### DIFF
--- a/packages/config/config/devices/0x041a/kp-so-02.json
+++ b/packages/config/config/devices/0x041a/kp-so-02.json
@@ -1,0 +1,101 @@
+{
+	// Device is rebranded Shenzhen Neo NAS-WR01ZE
+	"manufacturer": "Keemple",
+	"manufacturerId": "0x041a",
+	"label": "KP-SO-02",
+	"description": "Smart Socket",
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
I know that the Neo NAS-WR01ZE file doesn't do this correctly either, but parameter labels should be in `Title Case`. Please update them to match. Ideally also in the original file 🙏🏻

AlCalzone (approved):
Did a little change for consistency with the other PR

PR #4131, opened 2022-01-26
feat(config): add Alarm.com Smart Thermostat ADC-T3000

EiNSTeiN-:
This adds a configuration for the ADC-T3000 z-wave thermostat, the configuration being taken from [this manual](https://products.z-wavealliance.org/ProductManual/File?folder=&filename=Manuals/3190/T3000_user_guide_190630.pdf).

### DIFF
--- a/packages/config/config/devices/0x0190/adc-t_3000.json
+++ b/packages/config/config/devices/0x0190/adc-t_3000.json
@@ -98,23 +98,12 @@
 		},
 		{
 			"#": "5[0x18000000]",
+			"$import": "~/templates/master_template.json#temperature_scale_celsius_fahrenheit",
 			"label": "Calibration Temperature Scale",
 			"description": "The 'scale' part of a z-wave float representing this parameter",
 			"valueSize": 4,
-			"minValue": 0,
-			"maxValue": 1,
 			"defaultValue": 1,
-			"readOnly": true,
-			"options": [
-				{
-					"label": "Celsius",
-					"value": 0
-				},
-				{
-					"label": "Fahrenheit",
-					"value": 1
-				}
-			]
+			"readOnly": true
 		},
 		{
 			"#": "5[0x07000000]",
@@ -135,7 +124,7 @@
 			"valueSize": 4,
 			"minValue": -1,
 			"maxValue": 30,
-			"defaultValue": 50,
+			"defaultValue": 5,
 			"options": [
 				{
 					"label": "Disabled",
@@ -156,23 +145,12 @@
 		},
 		{
 			"#": "6[0x18000000]",
+			"$import": "~/templates/master_template.json#temperature_scale_celsius_fahrenheit",
 			"label": "Swing Scale",
 			"description": "The 'scale' part of a z-wave float representing this parameter",
 			"valueSize": 4,
-			"minValue": 0,
-			"maxValue": 1,
 			"defaultValue": 1,
-			"readOnly": true,
-			"options": [
-				{
-					"label": "Celsius",
-					"value": 0
-				},
-				{
-					"label": "Fahrenheit",
-					"value": 1
-				}
-			]
+			"readOnly": true
 		},
 		{
 			"#": "6[0x07000000]",
@@ -214,23 +192,12 @@
 		},
 		{
 			"#": "7[0x18000000]",
+			"$import": "~/templates/master_template.json#temperature_scale_celsius_fahrenheit",
 			"label": "Overshoot Scale",
 			"description": "The 'scale' part of a z-wave float representing this parameter",
 			"valueSize": 4,
-			"minValue": 0,
-			"maxValue": 1,
 			"defaultValue": 1,
-			"readOnly": true,
-			"options": [
-				{
-					"label": "Celsius",
-					"value": 0
-				},
-				{
-					"label": "Fahrenheit",
-					"value": 1
-				}
-			]
+			"readOnly": true
 		},
 		{
 			"#": "7[0x07000000]",
@@ -290,23 +257,12 @@
 		},
 		{
 			"#": "10[0x18000000]",
+			"$import": "~/templates/master_template.json#temperature_scale_celsius_fahrenheit",
 			"label": "Balance Setpoint Scale",
 			"description": "The 'scale' part of a z-wave float representing this parameter",
 			"valueSize": 4,
-			"minValue": 0,
-			"maxValue": 1,
 			"defaultValue": 1,
-			"readOnly": true,
-			"options": [
-				{
-					"label": "Celsius",
-					"value": 0
-				},
-				{
-					"label": "Fahrenheit",
-					"value": 1
-				}
-			]
+			"readOnly": true
 		},
 		{
 			"#": "10[0x07000000]",
@@ -376,23 +332,12 @@
 		},
 		{
 			"#": "15[0x18000000]",
+			"$import": "~/templates/master_template.json#temperature_scale_celsius_fahrenheit",
 			"label": "Maximum Heat Setpoint Scale",
 			"description": "The 'scale' part of a z-wave float representing this parameter",
 			"valueSize": 4,
-			"minValue": 0,
-			"maxValue": 1,
 			"defaultValue": 1,
-			"readOnly": true,
-			"options": [
-				{
-					"label": "Celsius",
-					"value": 0
-				},
-				{
-					"label": "Fahrenheit",
-					"value": 1
-				}
-			]
+			"readOnly": true
 		},
 		{
 			"#": "15[0x07000000]",
@@ -434,23 +379,12 @@
 		},
 		{
 			"#": "16[0x18000000]",
+			"$import": "~/templates/master_template.json#temperature_scale_celsius_fahrenheit",
 			"label": "Minimum Heat Setpoint Scale",
 			"description": "The 'scale' part of a z-wave float representing this parameter",
 			"valueSize": 4,
-			"minValue": 0,
-			"maxValue": 1,
 			"defaultValue": 1,
-			"readOnly": true,
-			"options": [
-				{
-					"label": "Celsius",
-					"value": 0
-				},
-				{
-					"label": "Fahrenheit",
-					"value": 1
-				}
-			]
+			"readOnly": true
 		},
 		{
 			"#": "16[0x07000000]",
@@ -492,23 +426,12 @@
 		},
 		{
 			"#": "17[0x18000000]",
+			"$import": "~/templates/master_template.json#temperature_scale_celsius_fahrenheit",
 			"label": "Maximum Cool Setpoint Scale",
 			"description": "The 'scale' part of a z-wave float representing this parameter",
 			"valueSize": 4,
-			"minValue": 0,
-			"maxValue": 1,
 			"defaultValue": 1,
-			"readOnly": true,
-			"options": [
-				{
-					"label": "Celsius",
-					"value": 0
-				},
-				{
-					"label": "Fahrenheit",
-					"value": 1
-				}
-			]
+			"readOnly": true
 		},
 		{
 			"#": "17[0x07000000]",
@@ -550,23 +473,12 @@
 		},
 		{
 			"#": "18[0x18000000]",
+			"$import": "~/templates/master_template.json#temperature_scale_celsius_fahrenheit",
 			"label": "Minimum Cool Setpoint Scale",
 			"description": "The 'scale' part of a z-wave float representing this parameter",
 			"valueSize": 4,
-			"minValue": 0,
-			"maxValue": 1,
 			"defaultValue": 1,
-			"readOnly": true,
-			"options": [
-				{
-					"label": "Celsius",
-					"value": 0
-				},
-				{
-					"label": "Fahrenheit",
-					"value": 1
-				}
-			]
+			"readOnly": true
 		},
 		{
 			"#": "18[0x07000000]",
@@ -614,27 +526,13 @@
 		},
 		{
 			"#": "23",
+			"$import": "~/templates/master_template.json#temperature_scale_celsius_fahrenheit",
 			"label": "Temperature Display Units",
-			"valueSize": 1,
-			"minValue": 0,
-			"maxValue": 1,
-			"defaultValue": 1,
-			"allowManualEntry": false,
-			"options": [
-				{
-					"label": "Celsius",
-					"value": 0
-				},
-				{
-					"label": "Farenheit",
-					"value": 1
-				}
-			]
+			"defaultValue": 1
 		},
 		{
 			"#": "24",
 			"label": "HVAC Modes Enabled",
-			"description": "Which heating/cooling modes are available.",
 			"valueSize": 1,
 			"minValue": 3,
 			"maxValue": 31,
@@ -674,7 +572,6 @@
 		{
 			"#": "25[0x00ff]",
 			"label": "Configurable Terminal Setting Z2",
-			"description": "Changes control of configurable terminal",
 			"valueSize": 2,
 			"minValue": 0,
 			"maxValue": 4,
@@ -706,7 +603,6 @@
 		{
 			"#": "25[0xff00]",
 			"label": "Configurable Terminal Setting Z1",
-			"description": "Changes control of configurable terminal",
 			"valueSize": 2,
 			"minValue": 0,
 			"maxValue": 4,
@@ -737,8 +633,7 @@
 		},
 		{
 			"#": "26",
-			"label": "Power Source",
-			"description": "Which source of power is utilized.",
+			"label": "Active Power Source",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 1,
@@ -1232,27 +1127,14 @@
 		},
 		{
 			"#": "30[0x00ff]",
-			"label": "Remote Temperature Enable",
-			"description": "Enables remote temperature sensor instead of built-in.",
-			"valueSize": 2,
-			"minValue": 0,
-			"maxValue": 4,
-			"defaultValue": 0,
-			"allowManualEntry": false,
-			"options": [
-				{
-					"label": "Disabled",
-					"value": 0
-				},
-				{
-					"label": "Enabled",
-					"value": 1
-				}
-			]
+			"$import": "~/templates/master_template.json#base_enable_disable",
+			"label": "Remote Temperature Sensor",
+			"description": "Use remote temperature sensor instead of built-in.",
+			"valueSize": 2
 		},
 		{
 			"#": "30[0xff00]",
-			"label": "Remote Temperature Status",
+			"label": "Remote Temperature Sensor Status",
 			"description": "Status of the remote temperature sensor.",
 			"valueSize": 2,
 			"minValue": 0,
@@ -1312,23 +1194,12 @@
 		},
 		{
 			"#": "31[0x18000000]",
+			"$import": "~/templates/master_template.json#temperature_scale_celsius_fahrenheit",
 			"label": "Heat Differential Scale",
 			"description": "The 'scale' part of a z-wave float representing this parameter",
 			"valueSize": 4,
-			"minValue": 0,
-			"maxValue": 1,
 			"defaultValue": 1,
-			"readOnly": true,
-			"options": [
-				{
-					"label": "Celsius",
-					"value": 0
-				},
-				{
-					"label": "Fahrenheit",
-					"value": 1
-				}
-			]
+			"readOnly": true
 		},
 		{
 			"#": "31[0x07000000]",
@@ -1370,23 +1241,12 @@
 		},
 		{
 			"#": "32[0x18000000]",
+			"$import": "~/templates/master_template.json#temperature_scale_celsius_fahrenheit",
 			"label": "Cool Differential Scale",
 			"description": "The 'scale' part of a z-wave float representing this parameter",
 			"valueSize": 4,
-			"minValue": 0,
-			"maxValue": 1,
 			"defaultValue": 1,
-			"readOnly": true,
-			"options": [
-				{
-					"label": "Celsius",
-					"value": 0
-				},
-				{
-					"label": "Fahrenheit",
-					"value": 1
-				}
-			]
+			"readOnly": true
 		},
 		{
 			"#": "32[0x07000000]",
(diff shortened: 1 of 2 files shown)


packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,801 @@
...
+					"label": "Off, Heat, Cool, Auto, Emergency Heat",
+					"value": 31
+				}
+			]
+		},
+		{
+			"#": "25",
+			"label": "Configurable Terminal Setting",
+			"description": "Changes control of configurable terminal",
+			"valueSize": 2,
```
EiNSTeiN-, line 606: Is it possible to specify configuration fields where each byte is a separate thing?

This thermostat for example has 2 configurable terminals, Z1 and Z2, and I believe (though it is not specified in the manual) that each byte will configure a different terminal. There are a few other cases like this, for example parameter 30 has a second byte that reports a status (readonly), whereas the first byte is writable.

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,801 @@
...
+			]
+		},
+		{
+			"#": "31",
+			"label": "Heat Differential",
+			"unit": "0.5°F",
+			"valueSize": 4,
+			"minValue": 2,
+			"maxValue": 20,
+			"defaultValue": 6
```
EiNSTeiN-: I hope this is correct, the manual says this field is specified in 0.5 deg. F increments, so the default value is 3 degrees, specified here as 3 / 0.5 = 6

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,801 @@
...
+			]
+		},
+		{
+			"#": "38",
+			"label": "Z-Wave Association Report Bit Mask",
+			"description": "Bitmask to selectively enable non-required Z-wave association reports.",
+			"valueSize": 4,
+			"minValue": 0,
+			"maxValue": 4294967295,
+			"defaultValue": 4294967295,
```
EiNSTeiN-: Is there a better way to represent bitmasks? I am not specifying the values for each bit here, it would be convenient if there was a way to do that, but didn't find one.

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,804 @@
...
+			"defaultValue": 1
+		},
+		{
+			"#": "7",
+			"label": "Overshoot",
+			"unit": "0.5°F",
+			"valueSize": 4,
+			"minValue": 0,
+			"maxValue": 6,
+			"defaultValue": 0
```
EiNSTeiN-: Something odd with all the parameters that are temperatures, here's how they show up:
image

This is what the manual says:
> Temperature parameters take the form of 4-byte Z-Wave temperature commands where byte 1
is the (precision|scale|size) and bytes 2 and 3 are the 16-bit signed temperature value. Byte 4
is unused. The thermostat will format temperature reports with the same scale as the
thermostat display temperature units. If the temperature value is 0xFFFF (e.g. 0x2AFFFF00)
the parameter is disabled.

This is consistent with how z-wave encodes other values that contain a precision and scale (from what I saw). Is there a way to specify these configuration parameters so they show up correctly in the UI?

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,804 @@
...
+			"defaultValue": 99
+		},
+		{
+			"#": "56",
+			"label": "Vent Minimum Outdoor Temperature",
+			"unit": "°F",
+			"valueSize": 4,
+			"minValue": 0,
+			"maxValue": 99,
+			"defaultValue": 0
```
EiNSTeiN-: the default for 55 and 56 is technically 0xFFFF which disables this functionality. But since this is outside the min/max range, I didn't think this would work. Is it possible to specify an `options` array with values that are outside the allowed range?

packages/config/config/devices/0x0190/adc-t_3000.json
AlCalzone: > Is it possible to specify configuration fields where each byte is a separate thing?

Yeah, [partial parameter](https://zwave-js.github.io/node-zwave-js/#/config-files/file-format?id=partial-parameters). There's also a [bitmask calculator](https://zwave-js.github.io/node-zwave-js/#/config-files/file-format?id=bitmask-calculator) to help you figure them out. And a [guide with examples](https://zwave-js.github.io/node-zwave-js/#/config-files/partial-parameters).

packages/config/config/devices/0x0190/adc-t_3000.json
AlCalzone: seems correct.

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,801 @@
...
+					"value": 19
+				},
+				{
+					"label": "Off, Heat, Cool, Emergency Heat",
+					"value": 23
+				},
+				{
+					"label": "Off, Heat, Cool, Auto, Emergency Heat",
+					"value": 31
+				}
```
AlCalzone: Are these the only possible combinations or should this be split into partial parameters instead for free combination of each possibility?

packages/config/config/devices/0x0190/adc-t_3000.json
AlCalzone: See my other [comment](https://github.com/zwave-js/node-zwave-js/pull/4131/files/cbba4e4731a4f9709617fa7e10428eb6aeeb07e4#r792375463).

packages/config/config/devices/0x0190/adc-t_3000.json
AlCalzone: We currently don't have anything in place to say "this is a Z-Wave float". This is also the very first time I've seen it, so it doesn't seem to be common. You can split this parameter into 4 seperate partials for Precision/Scale/Size/Temperature with appropriate descriptions to make it a bit easier to enter this for humans.

packages/config/config/devices/0x0190/adc-t_3000.json
AlCalzone: We typically mention it in the description in this case and just widen the min/max range:
```
"description": "Allowable range: 0-99.",
```
Then you can add an option for "Disable".

packages/config/config/devices/0x0190/adc-t_3000.json
EiNSTeiN-: Only these combinations are possible per the manual

packages/config/maintenance/lintConfigFiles.ts
```diff
@@ -604,50 +608,51 @@ Consider converting this parameter to unsigned using ${white(
...
-					) {
+					const minValue =
+						param.unsigned || bitMaskWidth == 1
+							? 0
+							: -(1 << bitMaskWidth) / 2;
+					const maxValue =
+						param.unsigned || bitMaskWidth == 1
+							? (1 << bitMaskWidth) - 1
+							: (1 << bitMaskWidth) / 2 - 1;
+					if (param.minValue < minValue) {
```
EiNSTeiN-: this is calculating the min/max value correctly for signed and unsigned params. This will touch a _lot_ of unrelated config files so I'd suggest merging separately.

packages/config/config/devices/0x0190/adc-t_3000.json
AlCalzone: @EiNSTeiN- Are these really 4 bytes wide but use 0xffff (2 bytes) for disabling?
If they were two bytes, you could make the parameter signed and set the range -1..99 where -1 is disable.

But on the other hand, manufacturer choices don't always make sense...

EiNSTeiN-: per the manual, all the temperature values are 4 bytes, where the first byte is precision/scale/size. In practice, precision is always 1, scale is always Fahrenheit, and size is always 2. Byte 2-3 is the temperature, which is why 0xFFFF (2 bytes) is the disabled magic value. Also since precision is 1, a range of 0-99 F is actually represented as 0-990. Some temperature offsets can be negative as well.

I could tighten the ranges if that's what you are asking. I initially thought the precision could be changed from 1, which would change the range, but from what I tested, that's not possible.

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2839 @@
...
+				{
+					"label": "Heat Pump",
+					"value": 1
+				}
+			]
+		},
+		{
+			"#": "2",
+			"label": "Number of Heat Stages",
+			"description": "Heat Stages 0-3 Default is 2.",
```
AlCalzone: Unnecessary description

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2839 @@
...
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 3,
+			"defaultValue": 2
+		},
+		{
+			"#": "3",
+			"label": "Number of Cool Stages",
+			"description": "Cool Stages 0-2 Default is 2.",
```
AlCalzone: Unnecessary description

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2839 @@
...
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 2,
+			"defaultValue": 2
+		},
+		{
+			"#": "4",
+			"label": "Heat Fuel Type",
+			"description": "Choose type of fuel. Reality - whether unit is boiler vs forced air.",
```
AlCalzone: The first part should be removed
```suggestion
			"description": "Reality - whether unit is boiler vs forced air.",
```
but even then the description is a bit weird. Maybe these options should be merged with the existing ones, like "Fossil Fuel / Boiler", "Electric / Forced air"?

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2839 @@
...
+					"label": "Electric",
+					"value": 1
+				}
+			]
+		},
+		{
+			"#": "5[0xffff00]",
+			"label": "Calibration Temperature",
+			"description": "Allowable range: -10 to 10 in 1 °F increments.",
+			"unit": "0.1°F",
```
AlCalzone, line 76: I'm not really happy with the separation into the 3 partials, especially since precision and scale are readonly.
Its confusing me, so this is definitely going to confuse users.

I propose this approach:
1. Keep the other parts (size, precision, scale) of the partial, but comment them out (`// ...`) so we can easily add them back if we ever decide to do so. Mention the reason in another comment above the first time we do this.
2. Where it isn't too much (like for `6[0xffff00]` where it would be 8 options, including "Disable"), prefer specific options with `allowManualEntry: false` instead of trying to explain.
3. For the parameters with a bigger range (like here), I'd actually mention some legal values in the description, e.g. `Allowable range: -10 to 10 in 1 °F increments. Example: -100, -90, ..., 0, 10, ... 100` (did I get this right?)

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2839 @@
...
+			"valueSize": 4,
+			"minValue": -32768,
+			"maxValue": 32767,
```
AlCalzone: I guess this should at least match the allowable range, so -100..100 (?)

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2839 @@
...
+			"valueSize": 1,
+			"unit": "minutes",
+			"minValue": 0,
+			"maxValue": 60,
+			"defaultValue": 5
+		},
+		{
+			"#": "23",
+			"label": "Temperature Display Units",
+			"description": "Celsius or Farenheit for temperature display.",
```
AlCalzone: Unnecessary description

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2839 @@
...
+				{
+					"label": "Enabled",
+					"value": 1
+				}
+			]
+		},
+		{
+			"#": "38[0x00000001]",
+			"label": "Z-Wave Association Report: Thermostat Mode",
+			"description": "Bitmask to selectively enable non-required Z-wave association reports.",
```
AlCalzone: Unnecessary description

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2839 @@
...
+			"options": [
+				{
+					"label": "Disabled",
+					"value": 0
+				},
+				{
+					"label": "Enabled",
+					"value": 1
+				}
+			]
```
AlCalzone: We should be able to immensely cut down on file size by importing the `base_enable_disable` template for these.

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2839 @@
...
+		},
+		{
+			"#": "50[0x60000000]",
+			"label": "System configuration: Cool Stages",
+			"description": "Summarized report of system configuration",
```
AlCalzone: Unnecessary description

AlCalzone (changes requested):
Sorry for taking so long to review this. This file is massive, so I'm almost sure I overlooked something.

In general, try removing unnecessary descriptions, of which there are a bunch. I highlighted a few.
I'd also get rid of the readonly floaty partials (sorry), since they are probably just confusing the hell out of everyone.
With that and a few `$import`s (see comment), this file isn't going to look as intimidating anymore.

packages/config/config/devices/0x0190/adc-t_3000.json
EiNSTeiN-: @AlCalzone just one thing regarding this plan: there is a linter that prevents having a single partial value for any given config parameter. E.g. if I leave only the value part of the float with  the `0xffff00` mask and remove the other 3 partials, then the linter will complain that the partial parameter is unnecessary and advise to remove the mask. A few options:
1) Change the linter to only complain when the mask is not shifted, so a single partial parameter defined as `6[0xff]` would fail the linter whereas `6[0xff00]` would pass.
2) Remove the linter completely as it may be useful to ignore part of a field no matter if it's bitshifted or not.

I'll let you decide but I think 2) makes more sense?

AlCalzone: Lets do a mix of both. If the single bitmask spans the entire parameter width (e.g. `0xffff` for a 2-byte value), complain.

packages/config/config/devices/0x0190/adc-t_3000.json
EiNSTeiN-: this description is actually copied from the T2000 config, since a good chunk of the configuration overlaps I used it as a base. There is no description in the manual so I'll leave it blank.

packages/config/config/devices/0x0190/adc-t_3000.json
EiNSTeiN-: Done in f31e3aec7531db506d9ede7a902694d43b3219a0

EiNSTeiN-:
Changes should be all done for this one, and I rebased

EiNSTeiN-:
> This file is massive, so I'm almost sure I overlooked something.

FWIW I've been running my ADC-T3000 for the past week with this config (the version before today's changes)

packages/config/maintenance/lintConfigFiles.ts
```diff
@@ -861,20 +862,31 @@ Consider converting this parameter to unsigned using ${white(
...
+			for (const [key, paramInfos] of partialParams) {
+				if (partialParamCounts.get(key.parameter) == 1) {
+					for (const param of paramInfos) {
+						const bitMask = key.valueBitMask!;
+						const shiftAmount = getMinimumShiftForBitMask(bitMask);
+						const bitMaskWidth = getBitMaskWidth(bitMask);
+
+						if (
+							shiftAmount == 0 &&
+							param.valueSize >= bitMaskWidth / 8
```
AlCalzone: That's not correct IMO. You want to catch bitmasks that span the entire value, not part of it starting with the LSB.
```suggestion
							shiftAmount === 0 &&
							param.valueSize === bitMaskWidth / 8
```

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2343 @@
...
+		},*/
+		{
+			"#": "6[0xffff00]",
+			"label": "Swing",
+			"description": "Allowable range: 0 to 3 in 0.5 °F increments.",
+			"unit": "0.1°F",
+			"valueSize": 4,
+			"minValue": -1,
+			"maxValue": 30,
+			"defaultValue": 50,
```
AlCalzone: That can't be. Did we break the lint check for min <= default <= max?

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2343 @@
...
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Celsius",
+					"value": 0
+				},
+				{
+					"label": "Farenheit",
+					"value": 1
+				}
```
AlCalzone: We have a template for Celsius/Fahrenheit, please use it.

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2343 @@
...
+			]
+		},
+		{
+			"#": "24",
+			"label": "HVAC Modes Enabled",
+			"description": "Which heating/cooling modes are available.",
```
AlCalzone: Unnecessary description

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2343 @@
...
+				{
+					"label": "Off, Heat, Cool, Auto, Emergency Heat",
+					"value": 31
+				}
+			]
+		},
+		{
+			"#": "25[0x00ff]",
+			"label": "Configurable Terminal Setting Z2",
+			"description": "Changes control of configurable terminal",
```
AlCalzone: Unnecessary description

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2343 @@
...
+				{
+					"label": "External Air Baffle or Vent",
+					"value": 4
+				}
+			]
+		},
+		{
+			"#": "25[0xff00]",
+			"label": "Configurable Terminal Setting Z1",
+			"description": "Changes control of configurable terminal",
```
AlCalzone: Unnecessary description

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2343 @@
...
+				{
+					"label": "External Air Baffle or Vent",
+					"value": 4
+				}
+			]
+		},
+		{
+			"#": "26",
+			"label": "Power Source",
+			"description": "Which source of power is utilized.",
```
AlCalzone:
```suggestion
			"label": "Active Power Source",
```

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2343 @@
...
+				{
+					"label": "Closed",
+					"value": 1
+				}
+			]
+		},
+		{
+			"#": "30[0x00ff]",
+			"label": "Remote Temperature Enable",
+			"description": "Enables remote temperature sensor instead of built-in.",
```
AlCalzone:
```suggestion
			"label": "Remote Temperature Sensor",
			"description": "Use remote temperature sensor instead of built-in.",
```

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2343 @@
...
+				{
+					"label": "Enabled",
+					"value": 1
+				}
+			]
+		},
+		{
+			"#": "30[0xff00]",
+			"label": "Remote Temperature Status",
+			"description": "Status of the remote temperature sensor.",
```
AlCalzone:
```suggestion
			"label": "Remote Temperature Sensor Status",
```

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2343 @@
...
+			"minValue": 0,
+			"maxValue": 7,
+			"defaultValue": 2,
+			"unsigned": true,
+			"readOnly": true
+		},*/
+		{
+			"#": "35",
+			"label": "Z-Wave Echo Association Reports",
+			"description": "Enable/Disabled Echo Assoc. Reports.",
```
AlCalzone: Unnecessary description

packages/config/config/devices/0x0190/adc-t_3000.json
```diff
@@ -0,0 +1,2343 @@
...
+			"minValue": 0,
+			"maxValue": 7,
+			"defaultValue": 2,
+			"unsigned": true,
+			"readOnly": true
+		},*/
+		{
+			"#": "37",
+			"label": "Run Fan With Auxiliary Heat",
+			"$import": "~/templates/master_template.json#base_enable_disable"
```
AlCalzone: Always $import before label please.

AlCalzone (changes requested):
Thanks, caught a few more things this time (sorry...)

packages/config/config/devices/0x0190/adc-t_3000.json
EiNSTeiN-: > Did we break the lint check for min <= default <= max?

Looks like the only thing being checked by the linter for `defaultValue` is that it fits the bitmask (even before my changes).

AlCalzone (approved):

AlCalzone:
Thanks!

PR #4137, opened 2022-01-27
fix: correctly check min/max/default values for signed partial parameters

EiNSTeiN-:
This is extracted from https://github.com/zwave-js/node-zwave-js/pull/4131/files since these changes are unrelated to the new config I'm adding.

This PR fixes the calculation for allowable min and max values when it comes to partial parameters. Currently the calculation is done with a bitmask, which does not work for signed values since node will represent these as two's complement on 4 or 8 bytes which won't fit any smaller bitmasks. For example -1 with a 2-byte bitmask should be represented as 0xFFFF but it does not pass current linter validation:
image

I changed the linter logic to calculate the allowable min and max value given the size of the bitmask and the expected signedness.  I had to modify all device configs that did not specify `unsigned: true` but specified values out of the signed range.

### DIFF
--- a/packages/config/maintenance/lintConfigFiles.ts
+++ b/packages/config/maintenance/lintConfigFiles.ts
@@ -1,4 +1,8 @@
-import { getIntegerLimits, getMinimumShiftForBitMask } from "@zwave-js/core";
+import {
+	getBitMaskWidth,
+	getIntegerLimits,
+	getMinimumShiftForBitMask,
+} from "@zwave-js/core";
 import { reportProblem } from "@zwave-js/maintenance";
 import { formatId, getErrorMessage, num2hex } from "@zwave-js/shared";
 import { distinct } from "alcalzone-shared/arrays";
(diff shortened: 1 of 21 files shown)


packages/config/maintenance/lintConfigFiles.ts
```diff
@@ -604,50 +608,51 @@ Consider converting this parameter to unsigned using ${white(
...
-						(param.minValue & shiftedBitMask) !== param.minValue
-					) {
+					const minValue =
+						param.unsigned || bitMaskWidth == 1
+							? 0
+							: -(1 << bitMaskWidth) / 2;
+					const maxValue =
+						param.unsigned || bitMaskWidth == 1
+							? (1 << bitMaskWidth) - 1
+							: (1 << bitMaskWidth) / 2 - 1;
```
EiNSTeiN-: with `0xFF`, `bitMaskWidth` will be 8, `-(1 << 8) / 2` gives us `minValue = -128` and `(1 << 8) / 2 - 1` gives us `maxValue = 127` for the signed params.

AlCalzone:
I've refactored this a bit, mainly because in Node.js bitshifts are limited to 32 bit values, but you need 33 to compute `(1 << 32) - 1` without overflow. This shouldn't ever happen, but better be safe than sorry.

Looks good otherwise.

AlCalzone (approved):

PR #4176, opened 2022-02-01
fix(config): add param 38 to Aeotec Siren Gen5, FW 3.25+

Spilarix:
Parameter described in the documentation : https://aeotec.freshdesk.com/helpdesk/attachments/6102433780

### DIFF
--- a/packages/config/config/devices/0x0086/dsd31.json
+++ b/packages/config/config/devices/0x0086/dsd31.json
@@ -52,6 +52,7 @@
 		},
 		{
 			"#": "38",
+			"$if": "firmwareVersion >= 3.25",
 			"$import": "templates/aeotec_template.json#stop_action_button"
 		},
 		{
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
Looking at the changelog in the linked manual, it seems that this parameter was added in a later firmware revision. Can you please contact Aeotec and verify in which firmware this was added?

That way we can avoid users with an older firmware getting confused by a non-existant param.

Spilarix:
> Looking at the changelog in the linked manual, it seems that this parameter was added in a later firmware revision. Can you please contact Aeotec and verify in which firmware this was added?
> 
> That way we can avoid users with an older firmware getting confused by a non-existant param.

Done : 3.25 or later.

AlCalzone (approved):

AlCalzone:
Great, thanks!

PR #4186, opened 2022-02-04
fix(config): cleanup Shenzhen Neo device configs

knthm:
Replace hardcoded parameters in Shenzhen Neo configs with their master template equivalent where applicable.

Relates to: #3882
Relates to: #4125


AlCalzone (changes requested):
Thanks, I'd like to change the labels a bit for consistency though.
https://github.com/zwave-js/node-zwave-js/blob/fcef2b877dfef9d36b06ab587a6babd183295172/packages/config/config/devices/0x0258/nas-wr01ze.json
has some "cleaner" labels, which seem to match at least some of the ones here. (except that they are not `Title Case`).

Can you please check which of those can be applied here and do it?

AlCalzone (approved):
Thanks, fixed some remaining nits myself

PR #4110, opened 2022-01-19
fix(config): various abandoned PRs and issues

AlexCPU:
### Abandoned PRs
Resolves #3303 - please check the 4-byte partial parameters, that was hard for a first partial parameter!
Resolves #3302 
Resolves #3623 
Resolves #3689 
Resolves #3831 - I believe looking at the comments only the changes requested were needed.
Resolves #3692 
Resolves #3814 
### Issues
Fixes #3701 
Fixes #3540

Issue #1581, opened 2021-02-03
Should valueSize for parameter 13 of LZW31-SN be 2?


bburan:
This line suggests that parameter 13 should be 1 byte:

https://github.com/zwave-js/node-zwave-js/blob/910155c88ffb2f8ab6ddce6adc2c0dbb57a135f6/packages/config/config/devices/0x031e/lzw31-sn.json#L151

However, if you look at the Inovelli config file, it says it should be 2 bytes:

https://github.com/InovelliUSA/OpenZWave/blob/master/config/inovelli/lzw31-sn.xml#L119

Setting values for all parameters except parameter 13 seem to work.  Parameter 13 just refuses to change. When I attempt to set parameter 13, I get the following from the logfile:

```
2021-02-02 04:28:34.620 INFO APP: Zwave api call: writeValue [ { nodeId: 13, commandClass: 112, property: 13 }, 179, [length]: 2 ]
2021-02-02 04:28:34.620 INFO ZWAVE: Writing 179 to 13-112-0-13
2021-02-02 04:28:34.687 INFO ZWAVE: Success zwave api call writeValue
2021-02-02 04:28:34.840 INFO ZWAVE: Node 13: value updated: 112-0-13 170 => 170
```

I'm currently using the dev Docker image of zwavejs/zwavejs2mqtt. I see this file is listed in two places on the Docker image (relative to /usr/src/app):

./node_modules/@zwave-js/config/config/devices/0x031e/lzw31-sn.json                                                                                  
./node_modules/@zwave-js/server/node_modules/@zwave-js/config/config/devices/0x031e/lzw31-sn.json                  

If I edit these files on the Docker image, will this allow me to test whether it should be 1 byte or 2? Or, do I need to recompile the Docker image to test this?

sneelco:
I noticed this didn't work either on the lzw31-bsd too but didn't have a chance to dig into it until now.  I've confirm @bburan suggestion works by overriding those files to set the size to 2 like the openzwave config has, thanks!  I'll put a PR together for both switches once i get my setup configured.

PR #4212, opened 2022-02-09
fix(config): regression of bug #1581 (wrong valueSize, param 13 of LZW31-SN)

raineth:
PR #4110 inadvertently re-introduced an old Inovelli documentation
error (bug #1581), which described parameter 13 (LED indicator color)
as 1 byte when it is actually 2.  This PR leaves the other changes from
PR #4110 but fixes the regression by correcting valueSize to 2.


AlCalzone (approved):

AlCalzone:
Thanks, this one must have slipped through the cracks when I reviewed it.

Issue #2589, opened 2021-05-10
Automatic Battery CC refresh for non-listening devices


kpine:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
Refresh the Battery CC values automatically for _non-listening_ devices based on a time-based schedule (at once a month or more often since last update).

Not all Battery devices send battery reports unsolicited, or maybe the controller was down when a device did send an update, so this would be useful as default functionality. _Listening_ devices already have this behavior. 

**Describe alternatives you've considered**
- Enable polling or automated updates in automation software. This works fine but requires manual setup.
- Use undocumented (?) `queryOnWakeup` compat flag with battery CC. Probably a bit excessive to query on every wake-up. This is enabled on a few battery thermostat devices, for what looks like other purposes.
- Add a new compat flag specifically for time-based refreshes, instead of per-wake up. Could individually opt-in those devices that don't send unsolicited battery reports.

**Additional context**
Some devices may never wake up on their own, if opt-in by default, would it be necessary to disable for some devices? At most it would keep a pending Get message in the queue.

AlCalzone:
I'm thinking of just removing the check for sleeping/non-sleeping and scheduling the refreshes for every node. When they do send unsolicited updates for a CC, those scheduled refreshes get canceled again, so it shouldn't hurt.

raintonr:
With the original ioBroker Z-Wave adapter our battery levels seemed to be updated so went  looking at the https://github.com/ioBroker/ioBroker.zwave / https://github.com/jperkin/node-openzwave code for clues as to why that was. Sadly it's not very obvious :disappointed: 

If anyone knows for sure (seems to be a bit of speculation on this issue) why the old adapter & library battery update works versus why the new one doesn't... well... I'm happy to mess about with some tests and have a stab at a PR for this but need pointing in the right direction to get going is all. Thx.

raintonr:
Poking around more... found issue #203, closed with PR #558, which introduced code that *appears* to request battery status periodically.

Comment in the code says once a month spec but that's not frequent enough and actually shows every 7 days. Perhaps we should add a debug thingo to make this do every wake-up or something just in test?

That aside, I looked in `BatteryCC` and there's the `refreshValues` method that should be triggered by the above.

So what gives - why isn't this working?

kpine:
Battery refresh is only done for devices like locks (FLIRS). That's why this feature request was created.

[...]

Issue #1608, opened 2021-02-05
Store timestamp along with the value


AlCalzone:
as requested in https://github.com/zwave-js/zwavejs2mqtt/issues/391

chrish987:
any potential timeline on this feature?

AlCalzone:
Tbh, there are more pressing issues on my end and the changes will be pretty invasive since the value storage is a core part of the library and used **everywhere**.

chrish987:
does that mean it is unlikely to be ever implemented? or weeks or months away?

AlCalzone:
I do see some value in this (e.g. auto-idling notifications across a restart could benefit). But looking at my issue pile, it is rather months away.

raintonr:
I had a few moments so took a quick look at the code today. @AlCalzone, how would you like to see this implemented as with anything there are many options?

From what I can tell there is a DB of key/value pairs that store metadata, values, etc. in an underlying `JsonlDB` database.

If we are just interested in timestamps of values then one could modify the `setValue` method (in `ValueDB.ts`) to have the value stored in the underlying DB to itself be a combination of timestamp+value. Then in `getValue` extract those parts out and only return the value part. Then one would need a `getTimestamp` method to do the same and only return the timestamp (or have a method that returns both in a single object for those callers who are aware of this functionality).

Would that work for you? Bit messy so I'm guessing you had something more elegant in mind :clown_face:

AlCalzone:
@raintonr I'd probably do it in the ValueDB, but I haven't thought it through completely. This will also need to consider edge cases like updates that don't actually come from the device.

In any case, a PR for this likely won't land before 2023 -> https://github.com/zwave-js/node-zwave-js/issues/5129

PR #4213, opened 2022-02-09
Issue/2589/battery update

raintonr:
For #2589 

Not gonna say this fixes this issue because I'm slightly concerned about the 7 day timer. If the process restarts regularly (like, every day or even every 3 days) then dunno if that timer will ever trigger. Sorry, didn't look if it is stored in any persistent way :clown_face: :wink:


AlCalzone:
I fear this change is too broad and at the same time not broad enough. `refreshValues` is meant for devices that don't report on their own and polls every application CC value of that device, which can lead to a shitload of traffic. Also this misses devices that do report on their own but that don't report everything.

What we need is a check for specific CCs if their values have been updated recently and if not, poll these values specifically. To do that, we also need https://github.com/zwave-js/node-zwave-js/issues/1608

PR #4233, opened 2022-02-13
fix: limit NVM page size

### LINKED ISSUE #4093: Node info incorrectly extracted from backup after reaching 232 nodes
First thank you very much for this information and conversion tool.

I was doing some testing between an Aeotec Z-Stick 500 plus and a SiLabs UZB-7.     Restoring to the 700 series stick.

I have added over 232 nodes to the Aeotec stick and I have looped back around and I am up to node 40 as my last added node. 

When converting the Aeotec NVM backup I noticed that only 40 nodes were restored to the 700 series stick.   When I converted the source NVM to json I see that the lastNodeId is set to 40 and only 40 nodes of information are present.   I assume that nvmedit is using the lastNodeId to determine how many nodes to include in the output and does not take into account this situation.

[NVM.zip](https://github.com/zwave-js/node-zwave-js/files/7875546/NVM.zip)

_Originally posted by @dwaleke in https://github.com/zwave-js/node-zwave-js/issues/3906#issuecomment-1013729260_

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/issues/4093

### DIFF
--- a/packages/nvmedit/src/nvm3/nvm.ts
+++ b/packages/nvmedit/src/nvm3/nvm.ts
@@ -117,11 +117,14 @@ export function encodeNVM(
 	options?: EncodeNVMOptions,
 ): Buffer {
 	const {
-		pageSize = FLASH_MAX_PAGE_SIZE,
 		deviceFamily = 2047,
 		writeSize = PageWriteSize.WRITE_SIZE_16,
 		memoryMapped = true,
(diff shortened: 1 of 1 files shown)


Madj42:
@AlCalzone I just wanted to say I was able to successfully convert and restore the backup with this change.  Thank you!

PR #4235, opened 2022-02-14
fix(config): align Trane XL624 config with XR524

cweakland:
The xl624 should match the xr524, the only difference is that the xl624 has a humidity sensor.

PR description here


AlCalzone:
@zwave-js-bot rename commit

AlCalzone (approved):

PR #4236, opened 2022-02-14
fix(config): add AU/NZ fingerprint to Aeotec Home Energy Meter - Gen5

keevanm:
"productType": "0x0202" added as Australia New Zealand version

not sure what the "zwaveAllianceId":  should be?

PR description here


keevanm (commented):
0x0202 added for Australia New Zealand version

AlCalzone (approved):
Thanks - no need for the alliance ID if you don't have it. It helps us track new files from the "official" DB.

PR #4259, opened 2022-02-18
fix: increase default controller response timeout to 10s

### LINKED ISSUE #4258: Increase timeout for SendDataResponse to 10s
This should normally never happen, as this message comes back in milliseconds, but in some circumstances, certain sticks take abnormally long (~8s +/-) to send the response when attempting communication with a dead node. This causes us to break the REQ->RES->Callback cycle and probably cause chaos.

AlCalzone:
fixes: #4258


AlCalzone:
@zwave-js-bot pack this

rossnelson (dismissed):

PR #4270, opened 2022-02-21
feat(cc): implement Irrigation CC

AlCalzone:
Progress:
- [x] SystemInfoGet
- [x] SystemInfoReport
- [x] SystemStatusGet
- [x] SystemStatusReport
- [x] SystemConfigSet
- [x] SystemConfigGet
- [x] SystemConfigReport
- [x] ValveInfoGet
- [x] ValveInfoReport
- [x] ValveConfigSet
- [x] ValveConfigGet
- [x] ValveConfigReport
- [x] ValveRun
- [x] ValveTableSet
- [x] ValveTableGet
- [x] ValveTableReport
- [x] ValveTableRun
- [x] SystemShutoff
- [x] CC API
- [x] SET_VALUE API
- [x] interview
- [x] refreshValues
- [x] logging


AlCalzone:
@jim-edwards may I ask how exactly you were using the Irrigation CC before? All commands will be available via the CC API, but only few of them (like the configuration) easily translate to the values concept that's heavily relied on by applications/UIs.
--> https://zwave-js.github.io/node-zwave-js/#/api/overview

System and valve statuses probably need to be queried on demand by the application.
Defining valve tables also doesn't translate to simple values.
Starting a single valve run could be a writeonly "button", same as stopping (without the shutoff duration specified).

jim-edwards:
I have seen 1 other implementation.  The way it worked was they exposed a single control endpoint, and in that endpoint you could set "All Off", "Zone 1, 5 minutes", "Zone 1, 10 minutes", "Zone 2, 5 minutes, Zone 2, 10 minutes", etc.  It was done as a "drop down" list in the UI, you pick one, it ran that zone for the the time specified.  It wasn't the best experience setup that way as the run time was not very flexible.

A simple binary switch for each "Zone" I think would be ideal.  I believe the device API has a time you specify in the zone for its run-time, I used that as a "Safety" time in case something went wrong it would run at most that, then the automation would turn the zone off when needed.  Not certain how that time would be specified, maybe just a configuration variable?  So in an ideal setup, I would have a home assistant automation script that looked something like:
     1.  Zone 1 on for 10 minutes (max I would ever want to run)
     2.  Delay for 7 minutes
     3.  Zone 1 off
     4.  Zone 2 on for 10 minutes (max I would ever want to run)
     5.  Delay for 7 minutes
     6.  Zone 2 off

It would be a bonus if the switch could report the Irrigation Valve info to update the on/off state, report current draw and errors that occur, but I consider that a nice to have, a write-only switch gets the job done.  I am not sure, but it may have been polling for this when any valve was on?  Or would the device just send these reports as the values change on its own?
     
There is also the concept of the "Master" zone, which I think can only be the first zone.  The idea for the master zone is it must be on for the other zones to flow, since they would be in-line in most cases.  While in most systems, there is not enough water flow to turn on more than 1 or 2 valves at a time (and usually a limit on the power-supply that drives the solenoids as well).  I think this is just set on the configuration screen, but it does control how switches behave on the device (I believe, have not been able to test this).  A master zone can (and must be) be on while the other zones are on.  While the other zones I think are mutually exclusive, only 1 at a time.  So if you turn 1 on, it turns all the other non-master valves off.

There are a few sensors, which I don't personally use, but some users might that are reported via the system info report command.  They can report the current flow per valve, but likely more used is the flow, rain, moisture, and pressure sensors.  In addition, I think these sensors can prevent valves from being commanded on via the configuration.  If those showed up as endpoints similar to how a weather station would, that would be also a nice to have.

Valve tables is not something I would personally use, it is more like a "internal macro" to me, so you can pre-program a sequence that plays back.  I would use home assistant to drive those instead and just turn the zones on / off in a script there using the single zone controls. I don't think I have seen this implemented in a UI anywhere else yet, so if that was API only, I don't think many users would be using that function.

AlCalzone:
Ok my question was targeted more at this first part:
> The way it worked was they exposed a single control endpoint, and in that endpoint you could set "All Off", "Zone 1, 5 minutes", "Zone 1, 10 minutes", "Zone 2, 5 minutes, Zone 2, 10 minutes", etc. It was done as a "drop down" list in the UI, you pick one, it ran that zone for the the time specified. It wasn't the best experience setup that way as the run time was not very flexible.

I'm not a fan of hardcoding these. What could work is having a configurable duration per valve and a switch to start/stop it - this won't be able to report the current status though without relying on the configured time, because there isn't a command to report the remaining runtime or current status AFAIK. So this would be a bare minimum implementation UI-wise. Everything else would need to be done in the HomeAssistant integration.

Not sure about your current setup, but there's also the possibility of using zwavejs2mqtt, and invoking the [`sendCommand` API](https://zwave-js.github.io/zwavejs2mqtt/#/guide/mqtt?id=send-command) through MQTT. Again no support for reading information though.
You can get crazy though and write JS code to control everything and invoke it as a driver funtion via MQTT: https://zwave-js.github.io/zwavejs2mqtt/#/guide/mqtt?id=apis (at the bottom of this section)

jim-edwards:
A configurable duration per valve, with a switch to start/stop it was what I was thinking would be good enough for most use-cases.  Users can turn the valves on and off with automation that way and the configured timer is really just more of a "Maximum it should run", in case the automation breaks in the middle because the restart or something.  That prevents the valve from flooding the area if home assistant stops running for some reason.

AlCalzone:
@jim-edwards I **think** this can be tested now. If I didn't miss anything, only logging is left to be implemented. Besides configuration, each valve now has two values `duration`  and `startStop` to control. In addition, there is a `shutoff` value to shut off the whole system.

If you're using zwavejs2mqtt in Docker, you can build a custom container: https://zwave-js.github.io/zwavejs2mqtt/#/development/custom-docker?id=building-a-container-using-dockerfilecontrib - use the `ZWJ_BRANCH=irrigation-cc`.

jim-edwards:
@AlCalzone you are a machine :) initial testing is great, I was able to turn on all zones after setting a runtime for each.  I used the zwavejs2mqtt UI specifically to trigger things for testing the first round.  Here is a screenshot of what the device looks like for reference and the debug output from the device itself:

[irrigation_cc_device.txt](https://github.com/zwave-js/node-zwave-js/files/8174816/irrigation_cc_device.txt)

image

Let me know if there is anything else you would like to see for debug output when testing, so far so good.  I will test the remainder of the parameters next just to try and see if they are all working as I would think.  Then I will add them to home assistant and see what it does there.

jim-edwards:
One thing I have noted from the log is I see the following message:
2022-03-03T06:03:50.469Z SERIAL « 0x0113000400020d6b04120002000001000011000088                        (21 bytes)
2022-03-03T06:03:50.470Z DRIVER   Dropping message with invalid payload
2022-03-03T06:03:50.471Z DRIVER « [Node 002] [REQ] [ApplicationCommand]
                                  └─[IrrigationCCSystemStatusReport] [INVALID]
2022-03-03T06:03:50.471Z SERIAL » [ACK]                                                                   (0x06)
2022-03-03T06:03:50.584Z SERIAL « 0x0113000400020d6b04120002000001000011000088                        (21 bytes)
2022-03-03T06:03:50.585Z DRIVER   Dropping message with invalid payload
2022-03-03T06:03:50.585Z DRIVER « [Node 002] [REQ] [ApplicationCommand]
                                  └─[IrrigationCCSystemStatusReport] [INVALID]
2022-03-03T06:03:50.585Z SERIAL » [ACK]                                                                   (0x06)
2022-03-03T06:03:50.676Z SERIAL « 0x0113000400020d6b04120002000001000011000088                        (21 bytes)
2022-03-03T06:03:50.677Z DRIVER   Dropping message with invalid payload
2022-03-03T06:03:50.677Z DRIVER « [Node 002] [REQ] [ApplicationCommand]
                                  └─[IrrigationCCSystemStatusReport] [INVALID]
2022-03-03T06:03:50.678Z SERIAL » [ACK]                                                                   (0x06)
2022-03-03T06:03:50.755Z SERIAL « 0x0113000400020d6b04120002000001000011000088                        (21 bytes)
2022-03-03T06:03:50.756Z DRIVER   Dropping message with invalid payload
2022-03-03T06:03:50.757Z DRIVER « [Node 002] [REQ] [ApplicationCommand]
                                  └─[IrrigationCCSystemStatusReport] [INVALID]

AlCalzone:
> One thing I have noted from the log is I see the following message:

Nice... the only (?) device that implements a CC and it doesn't follow the specs:

grafik

The reports have this field set to **2** 🙈

jim-edwards:
LOL, I suspect they wrote the spec for this device too, funny that its the only device that implements it I know of, but its out of its own spec :P.  

The fix for both issues seems to work, I can now use Home assistant API to control the valve and no longer get the reports.  Attaching a log with a re-inspect along with controlling a single valve 1 so you can see the full details.  But it is working great:

[zwavejs_2022-03-03.log](https://github.com/zwave-js/node-zwave-js/files/8179698/zwavejs_2022-03-03.log)

I will continue to test a few more things, but basic control is working and looking great.  Let me know if you have any specific areas you would like me to test out.

AlCalzone:
Ok found another one where the device goes against the Z-Wave specs. The values for maximum flow and high flow threshold (low too, but that's 0) are floats:
grafik

In Z-Wave, floats MUST use signed representation, but they seem to default to 255 (unsigned), which gets interpreted as -1 (signed). More specifically, the specs say that these must be encoded like the Multilevel Sensor CC, which in turn mentions this requirement.

AlCalzone:
@jim-edwards I think this is it for now. I don't think I'm going to work around that unsigned issue for now, unless that becomes a larger problem.

jim-edwards:
I have finished testing what I need as well, so I am very happy with the results :)  Thanks!

PR #4275, opened 2022-02-23
feat(config): add support for paddle programming on Zooz 700 Dimmers

halomakes:
This adds a new parameter that was added to Zooz dimmer switches that lets you disable the triple-tap up/down to enter inclusion/exclusion mode in firmware v10.0.  Some users will want to disable this so that those triple-tap gestures can be used for scene control instead.  I have many ZEN74 switches in my home but according to Zooz docs it should be the same on the ZEN72 and ZEN77 as well so I went ahead and included those. 

https://www.support.getzooz.com/kb/article/850-zen77-700-series-s2-dimmer-change-log/

### DIFF
--- a/packages/config/config/devices/0x027a/templates/zooz_template.json
+++ b/packages/config/config/devices/0x027a/templates/zooz_template.json
@@ -251,5 +251,24 @@
 				"value": 0
 			}
 		]
+	},
+	"paddle_programming": {
+		"label": "Paddle Programming",
+		"description": "Disables inclusion/exclusion commands on paddle for use with scene controls.",
+		"valueSize": 1,
+		"minValue": 0,
+		"maxValue": 1,
+		"defaultValue": 0,
+		"allowManualEntry": false,
+		"options": [
+			{
+				"label": "Enable",
+				"value": 0
+			},
+			{
+				"label": "Disable",
+				"value": 1
+			}
+		]
 	}
 }
--- a/packages/config/config/devices/0x027a/zen72.json
+++ b/packages/config/config/devices/0x027a/zen72.json
@@ -147,6 +147,11 @@
 		{
 			"#": "22",
 			"$import": "templates/zooz_template.json#night_light_mode"
+		},
+		{
+			"#": "26",
+			"$if": "firmwareVersion >= 10.0",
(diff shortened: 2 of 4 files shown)


packages/config/config/devices/0x027a/templates/zooz_template.json
```diff
@@ -251,5 +251,24 @@
 				"value": 0
 			}
 		]
+	},
+	"paddle_programming": {
+		"label": "Paddle Programming",
+		"description": "Disables inclusion/exclusion commands on paddle for use with scene controls.",
```
AlCalzone, line 257: Can we merge the description and the label into a somewhat short label? "Paddle Programming" doesn't sound intuitive at first glance. Maybe even invert the meaning like this?
```suggestion
		"label": "Scene Controls",
		"description": "When enabled, inclusion/exclusion via the paddle is disabled.",
```
(with inverted Enable/Disable options)

AlCalzone (changes requested):

packages/config/config/devices/0x027a/templates/zooz_template.json
halomakes: I'm all for finding a clearer way to put it but labelling it scene control would probably be more confusing as these switches already have a setting called scene control.  That other setting configures the switch so that they trigger a scene event any time the paddle is pressed (up to 5x up or down).  This is a separate attribute that just disables the programming shortcuts (3x up, 3x down) which normally override the 3x up/down scene controls.  Without changing attrib 26 scene controls still work for all the button presses except 3x down which will put your switch in exclusion mode.

halomakes: Here's that other param called scene control according to the manufacturer:
```
Enable/Disable Scene Control
Parameter 13: Enable or disable scene control functionality for quick double tap triggers.

Values: 0 – scene control disabled (default). 1 – scene control enabled.

Size: 1 byte dec
```

AlCalzone: I see - then I have no better idea for the label at the moment.

AlCalzone (approved):

PR #4339, opened 2022-03-08
fix(config): add metadata to Ecolink DWZWAVE25

aderusha:
fix(config): add metadata to Ecolink DWZWAVE25

### DIFF
--- a/packages/config/config/devices/0x014a/dwzwave25.json
+++ b/packages/config/config/devices/0x014a/dwzwave25.json
@@ -56,6 +56,13 @@
 			]
 		}
 	],
+	"metadata": {
+		"wakeup": "Removing the top case from the sensor will put in device into a tampered mode in which the sensor will stay awake and able to receive messages.",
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x014a/dwzwave25.json
```diff
@@ -56,6 +56,13 @@
...
+		"exclusion": "1. Start by placing the controller into exclusion mode.\n2. Remove the cover of the sensor, remove the battery for 10 seconds, reinsert the battery, and replace the sensor cover.",
+		"reset": "To restore this sensor to factory default settings, follow the instructions to exclude this sensor from the Z-Wave network. Upon completion of removal from the network the sensor will restore itself to factory default settings automatically.",
+		"manual": "https://products.z-wavealliance.org/ProductManual/File?folder=&filename=MarketCertificationFiles/1498/D.W.%20Zwave%20Manual%20R1.04.docx"
```
AlCalzone, line 67: wow... docx? Do they really not have a PDF? 🙈

AlCalzone (changes requested):
Please move the metadata section below the compat section. I prefer to have metadata last.

packages/config/config/devices/0x014a/dwzwave25.json
aderusha: Alright, thanks for the guidance on the metadata!  New commit moves that to the end.

Any guidance on that (admittedly godawful) docx link?  The vendor has no official doc hosted, there are PDFs scattered about the internet but none that are "canonical" in the way that a z-wavealliance.org link probably is.

AlCalzone: Its fine, we can leave the link as is - after all its from the "official" source. I just didn't expect that any sane manufacturer would provide manuals as something else than PDF.

AlCalzone (approved):

PR #4343, opened 2022-03-09
fix(config): add metadata to AEON Labs DSC26

aderusha:
I couldn't find a download link in a canonical source for the manual.  The link in the JSON goes to the manufacturer support page.  Z-Wave Alliance product page does not have a manual available: https://products.z-wavealliance.org/products/709?selectedFrequencyId=-1

### DIFF
--- a/packages/config/config/devices/0x0086/dsc26.json
+++ b/packages/config/config/devices/0x0086/dsc26.json
@@ -58,5 +58,12 @@
 			"#": "255",
 			"$import": "templates/aeotec_template.json#factory_reset"
 		}
-	]
+	],
+	"metadata": {
+		"wakeup": "Press and release the button which is on Micro Smart Energy Switch G2.",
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
If there is no manual, just remove the entry. Better than linking to a manual for a different device :)

aderusha:
Oh dear - that was my fault.  So, what I do have is this link: https://help.aeotec.com/support/solutions/articles/6000165733-micro-switch-g2-user-guide-

That's the correct manual from the vendor's site, but it's not a downloadable page.  Should I include that one?  Or just leave the manual tag out.

Sorry about the wrong link :D

AlCalzone:
@blhoward2 what do you think?

AlCalzone (approved):

AlCalzone:
Thanks again!

PR #4345, opened 2022-03-09
fix(config): add parameters and correct default values for Fibaro Smart Module FGS214 and Double Smart Module FGS224

cyrilst:
Added missing parameters (35, 40, 41, 159 and 161).
Changed incorrect default values (for parameters 20 and 21).
Improved descriptions according to Fibaro documentation.
Changed On and Off case for consistency.

### DIFF
--- a/packages/config/config/devices/0x010f/fgs214.json
+++ b/packages/config/config/devices/0x010f/fgs214.json
@@ -1,5 +1,5 @@
 {
-	"manufacturer": "FIBARO System",
+	"manufacturer": "Fibargroup",
 	"manufacturerId": "0x010f",
 	"label": "FGS214",
 	"description": "Smart Module",
@@ -35,38 +35,34 @@
 	"paramInformation": [
 		{
 			"#": "1",
-			"label": "Remember relays state",
-			"description": "This parameter determines the state of relays power supply failure (e.g. power outage). auto OFF and flashing modes the parameter not relevant and the relay will always remain off.",
+			"label": "Relays State After Power Failure",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 2,
 			"defaultValue": 1,
-			"unsigned": true,
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Relays remain switched off after restoring power",
+					"label": "Always off",
 					"value": 0
 				},
 				{
-					"label": "Restore remembered state of relays after restoring power",
+					"label": "Previous state",
 					"value": 1
 				},
 				{
-					"label": "Restore remembered state of relays after restoring power, but for toggle switches (parameter 20/21 set to 1) set the same state as the current state of the switches",
+					"label": "Toggle switch (synchronize output): current switch state / Other switch types: previous state",
 					"value": 2
 				}
 			]
 		},
 		{
 			"#": "20",
-			"label": "S1 input – switch type",
-			"description": "S1 - Inputs type configuration",
+			"label": "S1 Input – Switch Type",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 2,
-			"defaultValue": 2,
-			"unsigned": true,
+			"defaultValue": 0,
 			"allowManualEntry": false,
 			"options": [
 				{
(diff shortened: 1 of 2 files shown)


AlCalzone (changes requested):
Thanks, but your changes do not conform with our config file [style guide](https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide?id=configuration-parameters). Please read through the section on config parameters and apply it.

Note that we already have some predefined templates, e.g. for the [Scenes Sent](https://github.com/zwave-js/node-zwave-js/blob/93aa398d8df97f35e17cfa0d000ee785a2015cd6/packages/config/config/devices/0x010f/templates/fibaro_template.json#L67-L86), which should be used here. You can take a look at files for other similar Fibaro devices to see how they are used.

cyrilst:
Thanks for the pointers @AlCalzone 
Just updated the patch, don't hesitate to tell me if there is anything to improve.
When this patch will be ok, I'll update the FGS224 the same way.

packages/config/config/devices/0x010f/fgs214.json
```diff
@@ -35,121 +35,297 @@
...
 					"value": 0
 				},
 				{
-					"label": "Toggle switch (contact closed - On, contact opened - OFF)",
+					"label": "Toggle switch synchronized",
 					"value": 1
 				},
 				{
-					"label": "Toggle switch (device changes status when switch changes status)",
+					"label": "Toggle switch with memory",
```
AlCalzone: I'm not 100% happy with these, the original is a bit verbose, but the shortened version is not self-explanatory. What about something along these lines?

value 1 -> `"Toggle switch (synchronize output)"`
value 2 -> `"Toggle switch (toggle output)"`

packages/config/config/devices/0x010f/fgs214.json
```diff
@@ -35,121 +35,297 @@
...
```
AlCalzone: dito

packages/config/config/devices/0x010f/fgs214.json
```diff
@@ -35,121 +35,297 @@
...
 					"value": 0
 				},
 				{
-					"label": "Restore remembered state of relays after restoring power",
+					"label": "Previous state",
 					"value": 1
 				},
 				{
-					"label": "Restore remembered state of relays after restoring power, but for toggle switches (parameter 20/21 set to 1) set the same state as the current state of the switches",
+					"label": "Same as switches",
```
AlCalzone: I think this needs to mention the behavior when toggle switch (synchronized) is selected. Also "same as switches" is only true when param 20/21 is not set to 1.

AlCalzone (changes requested):
Thanks, much better. Just a few nits.

packages/config/config/devices/0x010f/fgs214.json
cyrilst: Indeed, that's way better.

packages/config/config/devices/0x010f/fgs214.json
cyrilst: Something like "Toggle switch (synchronize output): same as switches – other switch types: previous state"? I'm not convinced it's clear enough.

AlCalzone: Lets do "Toggle switch (synchronize output): current switch state / Other switch types: previous state"

cyrilst: Ok, I just pushed the updated version.

AlCalzone:
LGTM. Do you want to do the FGS224 as part of this PR too?

AlCalzone (dismissed):

cyrilst:
Why not, the file will really look similar. Should I do a new commit for FGS224 in the same PR?

AlCalzone:
Sure, go for it.

cyrilst:
Done. Maybe you should have a look at the last parameter (164). I changed the wording according to the style guide, but I'm not sure if it's self-explanatory.

AlCalzone (approved):

Issue #1995, opened 2021-03-05
🚨 PSA: DO NOT walk around with portable sticks to include devices! 🚨


AlCalzone:
This method is a recipe for disaster and reliably causes problems. This also includes moving the entire host along with the controller stick plugged in.

**Some background:**
`zwave-js` (and other gateway solutions) use the **Static Controller API**. This means "The controller doesn't physically move around".

However, after the walk-around inclusion, the controller thinks it's in direct range of everything. Effectively, its network map is as wrong as you can possibly get it - and the return routes it is assigning to all the devices won't work.

**To fix:**
Best case, do multiple network heals with the controller in multiple intermediate locations and its final location.
Worst case: restart from scratch.

If the problem is limited to single devices, the following workaround should help: 

* Shut down zwave-js
* Hard-reset device
* Start zwave-js, wait for device to be detected as dead (or asleep)
* Replace failed node -> re-include previously reset device under the same node ID

However, just starting over might take less time.

PR #4348, opened 2022-03-11
fix(config): add metadata to ZooZ ZEN26

aderusha:
fix(config): add metadata to ZooZ ZEN26

### DIFF
--- a/packages/config/config/devices/0x027a/zen26.json
+++ b/packages/config/config/devices/0x027a/zen26.json
@@ -316,9 +316,9 @@
 		}
 	],
 	"metadata": {
-		"inclusion": "1. Initiate inclusion (pairing) in the app (or web interface).\n2. Finalize inclusion at the switch. Tap the upper paddle of the dimmer 3 times quickly. The LED indicators will blink to signal communication and remain on for 2 seconds to confirm inclusion",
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x027a/zen26.json
```diff
@@ -314,5 +314,11 @@
 				}
 			]
 		}
-	]
+	],
+	"metadata": {
+		"inclusion": "1. Initiate inclusion (pairing) in the app (or web interface).\n2. Finalize inclusion at the switch. Tap the upper paddle of the dimmer 3 times quickly. The LED indicators will blink to signal communication and remain on for 2 seconds to confirm inclusion",
+		"exclusion": "1. Bring your Z-Wave gateway (hub) close to the switch if possible\n2. Put the Z-Wave hub into exclusion mode \n3. Tap the lower paddle on the dimmer 3 times quickly\n4. Your hub will confirm exclusion and the device will disappear from your\ncontroller's device list",
```
AlCalzone: Uhhhh. Even if they have this in their manual, I don't think we should ever advertise moving the controller.
https://github.com/zwave-js/node-zwave-js/issues/1995

AlCalzone (changes requested):

packages/config/config/devices/0x027a/zen26.json
aderusha: Agreed :D

AlCalzone (approved):

Issue #4075, opened 2022-01-12
Improvements for route/lifeline health checks


marcus-j-davies:
As Per slack discussion

> They can abort  already - but it only detaches the event listeners and closes the window - I couldn't see in the docs a way to abort it officially - So it is left to run in the background without the user getting feedback or the results.

> Might be a good idea to add support for aborting actually... 

and maybe `checkRouteHealth` also.

robertsLando:
Would also make sense to return last finished result in healtcheck progress callback instead of just the last rating

AlCalzone:
Additionally: 
- Automatically abort early if the route health is really bad
- Avoid infinite loop when nodes don't progress on their Powerlevel test https://github.com/zwave-js/node-zwave-js/pull/4372
- Give intermediate status reports (WHAT is happening right now?) https://github.com/zwave-js/node-zwave-js/pull/6086
- Wait for sleeping source nodes to wake up to prevent false positive delays https://github.com/zwave-js/node-zwave-js/pull/4376
- Don't allow testing the route health between two nodes when the target node is a sleeping node https://github.com/zwave-js/node-zwave-js/pull/4373

robertsLando:
> ah the failed pings is only included at the highest powerlevel. anything below implies 0/10

* Set failed pings to 0 instead of undefined on checkRouteHealth results https://github.com/zwave-js/node-zwave-js/pull/4377

mundschenk-at:
I've just triggered #4130 in my network after testing the Health Check functionality. I was wondering why some sensors missed reports today, so I looked at today's logfile: It's about 500 MB for 17 hours 😬

Incidentally, this looks like a good way to stress-test the resolution of #3906.

AlCalzone:
All points should be solved now.

PR #4372, opened 2022-03-15
fix: prevent infinite loop in health check when there is no progress

AlCalzone:
for https://github.com/zwave-js/node-zwave-js/issues/4075#issuecomment-1020898388


AlCalzone:
@zwave-js-bot pack this

PR #4378, opened 2022-03-16
feat: export LogContext types

raman325:
I want to add log filtering to the websocket log transport in `zwave-js-server` but I need to access more of the LogContext types


AlCalzone:
@raman325 I think my change captures the different types for the controller log context more accurately. What do you think? Other than that it looks good.

raman325:
yeah, this is better, thanks!

AlCalzone (approved):

PR #4400, opened 2022-03-21
fix(config): add values for Alarm and Doorbell Sound Selection to Shenzhen Neo Siren Alarm

jaym25:
Also added option "Always on" to "Alarm Duration"

### DIFF
--- a/packages/config/config/devices/0x0258/nas-ab01z.json
+++ b/packages/config/config/devices/0x0258/nas-ab01z.json
@@ -93,6 +93,8 @@
 			"#": "3",
 			"label": "Doorbell Duration",
 			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 255,
 			"defaultValue": 1,
 			"unsigned": true,
 			"options": [
--- a/packages/config/config/devices/0x0258/nas-ab01z.json
+++ b/packages/config/config/devices/0x0258/nas-ab01z.json
@@ -142,7 +142,7 @@
 					"value": 1
 				},
 				{
-					"label": "Fur Elise",
+					"label": "Für Elise",
 					"value": 2
 				},
 				{
@@ -191,7 +191,7 @@
 					"value": 1
 				},
 				{
-					"label": "Fur Elise",
+					"label": "Für Elise",
 					"value": 2
 				},
 				{
--- a/packages/config/config/devices/0x0258/nas-ab01z.json
+++ b/packages/config/config/devices/0x0258/nas-ab01z.json
@@ -63,9 +63,8 @@
 			"#": "2",
 			"label": "Alarm Duration",
 			"valueSize": 1,
-			"minValue": 0,
-			"maxValue": 3,
 			"defaultValue": 2,
+			"allowManualEntry": false,
 			"options": [
 				{
 					"label": "Off",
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0258/nas-ab01z.json
```diff
@@ -82,6 +82,10 @@
 				{
 					"label": "5 minutes",
 					"value": 3
+				},
+				{
+					"label": "Always on",
+					"value": 255
```
AlCalzone, line 88: please update the `maxValue` to match or (even better) remove min/maxValue and add `"allowManualEntry": false` before the options.

packages/config/config/devices/0x0258/nas-ab01z.json
```diff
@@ -132,15 +136,99 @@
 			"valueSize": 1,
 			"minValue": 1,
 			"maxValue": 10,
-			"defaultValue": 9
+			"defaultValue": 9,
+			"options": [
```
AlCalzone, line 139: Please add `"allowManualEntry": false` before the options.

packages/config/config/devices/0x0258/nas-ab01z.json
```diff
@@ -132,15 +136,99 @@
...
 		},
 		{
 			"#": "6",
 			"label": "Doorbell Sound Selection",
 			"valueSize": 1,
 			"minValue": 1,
 			"maxValue": 10,
-			"defaultValue": 10
+			"defaultValue": 10,
+			"options": [
```
AlCalzone, line 188: dito

AlCalzone (changes requested):
Now, this is better :) A few nits though.
Also, do you know if this was added in a specific firmware version or do these additional values exist for all firmwares?

jaym25:
This is a device I have been using in Home Assistant.  The firmware version is 2.102.  The additions come from the open ZWave configuration (with minor changes) that I have been using before switching to Z-Wave JS.  Some changes come from the user manual.  I did listen to the all the sounds and the descriptions are correct.  I cannot assure you the sounds are not different in different firmware revisions.

I also cannot figure out how to edit the revisions using Gitpod...

For parameter 2, the available settings are actually 0-3,255 so minValue and maxValue alone will not work.
I am concerned about using ("allowManualEntry": false) as it may prevent the values being set from Home Assistant automation code.  This would definitely cause problems.
If ("allowManualEntry": false) actually prevents changes from automation code, what option would be available to allow 255 as an additional setting in parameter 2?

If this is too much trouble for you, I will go ahead and close this request per your instruction.

jaym25:
Also, from the user manual:
Parameter 5 - Alarm Sound default is 10
Parameter 6 - Doorbell Sound default is 9
The existing config file has them reversed...

AlCalzone:
> For parameter 2, the available settings are actually 0-3,255 so minValue and maxValue alone will not work.

That's precisely what allowManualEntry does. You can still set it from automations, but UIs that respect the flag will show you a dropdown instead of a free text field with auto-complete. Having maxValue=3 won't let you set 255 though.

I know that it's hard/nearly impossible to track down which firmware changed which parameters for some of these asian manufacturers, so I'm fine with merging the changes.

> I will go ahead and close this request per your instruction.

Nonono, I just want to make sure everything is correct, as we put a lot of effort in having the best (and correct) config files. Just continue editing the file in your fork, adding `"allowManualEntry"`. That will update this PR and I can merge.

jaym25:
I think I got it this time...  I hope!

AlCalzone (approved):

PR #4409, opened 2022-03-29
fix(config): correct value for Brightness After Power Failure for LZW31-SN

amcfague:
It turns out that 101 is not a valid value for `Brightness After Power Failure`, based on the discussion here (https://community.inovelli.com/t/cant-choose-option-101-for-parameter-11-7-presses-power-on-state/1392/8). 101 is rejected by the switch.


AlCalzone:
Thanks

amcfague:
Wow, fast turn-around, @AlCalzone, thank you so much!

PR #4419, opened 2022-03-31
fix: query CC versions from endpoints if the node doesn't support the CC

### LINKED ISSUE #4413: Query CC versions for CCs that only exist on endpoints
See https://github.com/zwave-js/node-zwave-js/discussions/4402 for context

Essentially the device doesn't expose the `Multilevel Sensor CC` on the root device, so the driver never queries the CC version.
As a result, the endpoints are interviewed with `Multilevel Sensor CC V1`, which only queries the default sensor reading.

AlCalzone:
fixes: #4413

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/VersionCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/VersionCC.ts
@@ -217,7 +217,9 @@ export class VersionCC extends CommandClass {
 		return [CommandClasses["Manufacturer Specific"]];
 	}
 
-	public skipEndpointInterview(): boolean {
+	public async interview(): Promise<void> {
+		const node = this.getNode()!;
+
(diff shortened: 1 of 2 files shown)


espenthingstad:
Cool that it was solved so quickly :) When is v9 available for the HA version? I see that the current version (I have triggered auto-update) in my HA instance is 0.1.55 (0.1.55 -> Bump Z-Wave JS to 8.11.9)

Thanks!

AlCalzone:
This might take a while. AFAIK the devs are quite busy finalizing the HA beta and need to do some testing with v9 before it can be bumped. @raman325 might know a bit better than me, but I think it's scheduled for May.

raman325:
we'll have the v9 addon out before May, we just won't have support for the new features until May

PR #4451, opened 2022-04-08
fix: remove option 4 of LED Indicator Mode for Zooz ZSE40, FW 32.32+

blhoward2:
Confirmed with Zooz support. Starting with version 32.32 they have removed mode 4.


AlCalzone:
@zwave-js-bot fix lint

AlCalzone (approved):

PR #4469, opened 2022-04-15
feat: expose "safe" entrypoint that can be used from non-Node.js contexts

AlCalzone:
This PR adds an entry point to each package which can be used from non-Node.js contexts, like browser frontends. These `*/safe` entry points expose a safe subset of the main entry points of each module.

This is ensured with a custom lint script that recursively checks all imports that are done in these entry points and forbids all exports that are from outside the `zwave-js` monorepo, except some whitelisted modules.

- [x] `@zwave-js/config/safe`
- [x] `@zwave-js/core/safe`
- [x] `@zwave-js/nvmedit/safe`
- [x] `@zwave-js/serial/safe`
- [x] `@zwave-js/shared/safe`
- [x] `zwave-js/safe`

First step towards: #4323


AlCalzone:
@zwave-js-bot pack this

AlCalzone:
@zwave-js-bot pack this

PR #4470, opened 2022-04-15
feat(config): add bulk config params to inovelli switches

firstof9:
Add bulk configuration parameters to Inovelli devices: lzw30-sn, lzw31-sn, and lzw36 for LED notifications.


firstof9:
@zwave-js-bot fix lint

AlCalzone:
That doesn't work if there are partial parameters

PR #4472, opened 2022-04-16
feat(config): add param 92 (Battery Calibration Check) to aërQ v2.01

blhoward2:
Adding param 92 per Aeotec support:

In V2.01 of the new aerQ Sensor, there is a new battery calibration check parameter setting. There are only 3 different setting inputs:

 

1. Parameter 92 [4 byte] = 0x05010000 (83951616)

Example checked value returned (0x0B5B0B54 = 190516052 )

 

2. Parameter 92 [4 byte] = 0x05020000 (84017152)

Example checked value returned (0x0AF70AEC = 183962348 )

 

3. Parameter 92 [4 byte] = 0x05030000 (84082688)

Example checked value returned (0x0A2F0A28 = 170854952 )

### DIFF
--- a/packages/config/config/devices/0x0371/zwa039.json
+++ b/packages/config/config/devices/0x0371/zwa039.json
@@ -278,18 +278,12 @@
 			"#": "92",
 			"$if": "firmwareVersion >= 2.1",
 			"label": "Battery Calibration Check",
-			"description": "To determine the battery calibration value, you must set each step in order",
+			"description": "To determine the battery calibration value, set each step in order and wait for the value to update. If the returned values are NOT 0, the battery is still calibrated.",
 			"unsigned": true,
 			"valueSize": 4,
-			"minValue": 0,
-			"maxValue": 4294967295,
-			"allowManualEntry": false,
 			"defaultValue": 83951616,
+			"allowManualEntry": false,
 			"options": [
-				{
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0371/zwa039.json
```diff
@@ -274,6 +274,30 @@
...
 		},
+		{
+			"#": "92",
+			"$if": "firmwareVersion >= 2.1",
+			"label": "Battery Calibration Check",
+			"description": "To determine the battery calibration value, you must set each step in order",
+			"unsigned": true,
+			"valueSize": 4,
+			"allowManualEntry": false,
+			"defaultValue": 83951616,
```
AlCalzone, line 284: I'm not sure if the default value is correct. Maybe it should just be 0 with a corresponding option "Idle"? Also if I understood your email correctly, this can return values outside the specified range. So we should definitely add
```jsonc
			"minValue": 0,
			"maxValue": 4294967295
```
since the values are inferred from the options now.

AlCalzone (changes requested):

AlCalzone:
Ah shit, that min/max thing doesn't work with `allowManualEntry: false`.

AlCalzone (approved):

PR #4476, opened 2022-04-17
docs: use relative paths in CC index 

wahmedswl:
Hi, Command Classes links result in 404 due to wrong path structure.

image


AlCalzone:
That's not correct. If you open the rendered version of the documentation, those links are working:
https://zwave-js.github.io/node-zwave-js/#/api/CCs/index

AlCalzone:
I've update the docs generator so it now uses relative paths prefixed with `./`, which should work on both Github and in the rendered docs.

AlCalzone (approved):

AlCalzone:
Yeah, this isn't working

PR #4496, opened 2022-04-21
fix(config): add missing parameters to Zooz ZEN32

kaechele:
This adds two parameters:
1. ability to disable programming function from the button (Parameter 22)
2. control LED indicator flashing during a setting change (Parameter 23)

The [official documentation](https://www.support.getzooz.com/kb/article/608-zen32-scene-controller-advanced-settings/) on Parameter 23 is incorrect. It accepts values 0 (Enabled, default) and 1 (Disabled).

Verified this on my ZEN32 devices.

### DIFF
--- a/packages/config/config/devices/0x027a/zen32.json
+++ b/packages/config/config/devices/0x027a/zen32.json
@@ -519,6 +519,46 @@
 					"value": 1
 				}
 			]
+		},
+		{
+			"#": "22",
+			"$if": "firmwareVersion >= 10.0",
+			"label": "Programming through Relay Button",
+			"description": "Controls whether certain features (inclusion, exclusion, LED indicator modes) can be programmed through the relay button. Disabling this frees up the triple tap action for scene control. Factory reset will always work regardless of this setting.",
+			"valueSize": 1,
+			"defaultValue": 0,
+			"unsigned": true,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Enable",
+					"value": 0
+				},
+				{
+					"label": "Disable",
+					"value": 1
+				}
+			]
+		},
+		{
+			"#": "23",
+			"$if": "firmwareVersion >= 10.20",
+			"label": "LED Settings Indicator",
+			"description": "When enabled, the relay button LED indicator will flash to confirm the receipt of settings.",
+			"valueSize": 1,
+			"defaultValue": 0,
+			"unsigned": true,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Enable",
+					"value": 0
+				},
+				{
+					"label": "Disable",
+					"value": 1
+				}
+			]
 		}
 	],
 	"metadata": {


packages/config/config/devices/0x027a/zen32.json
```diff
@@ -519,6 +519,46 @@
...
+				{
+					"label": "Enable",
+					"value": 0
+				},
+				{
+					"label": "Disable",
+					"value": 1
+				}
+			]
+		},
```
AlCalzone, line 542: This seems to be the same as parameter 17 on the ZEN71. Please update the definition here to be the same. Ideally this should probably be moved into the `zooz_template.json` since they're using it on more than one device now.

packages/config/config/devices/0x027a/zen32.json
```diff
@@ -519,6 +519,46 @@
...
+		{
+			"#": "23",
+			"$if": "firmwareVersion >= 10.20",
+			"label": "LED Settings Indicator",
+			"description": "When enabled, the relay button LED indicator will flash to confirm the receipt of settings.",
```
AlCalzone, line 547:
```suggestion
```
I don't think this description is necessary to understand the parameter.

AlCalzone (changes requested):
Thanks, a few changes please.

packages/config/config/devices/0x027a/zen32.json
kaechele: Good point! It's already in the templates: https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x027a/templates/zooz_template.json#L255= (used by ZEN72). So I'll fix both this and ZEN71 not using the template.

kaechele: Do we care that ZEN32 technically doesn't have a paddle but a (relay) button instead? Because if so, I'd probably modify the template to read like ZEN71 reads currently, with no mentions of the type of switch/button.

AlCalzone: @blhoward2 what do you think?

packages/config/config/devices/0x027a/zen32.json
blhoward2: I’d make the template broader. Change it to Local Programming and then you don’t have to differentiate. I’ve tried to avoid using such specific language where it doesn’t cause confusion.

AlCalzone: wrong thread, but ok :)

packages/config/config/devices/0x027a/zen32.json
AlCalzone: @kaechele see below for the answer :)

kaechele: Thanks. Found it. I will update this PR with the proposed changes.

kaechele:
I got carried away and did a complete rehaul of the Zooz subfolder to use templates wherever this was feasible. Should I close this one and open a new PR for that?

blhoward2:
Yes, we appreciate the help.

PR #4504, opened 2022-04-25
fix(config): force scene count of Leviton VRCS2 to 4

j9brown:
The Leviton VRCS2 scene controller has 2 buttons each of which has two associated scenes for a total of 4.  Scenes 1-2 are sent when Buttons 1-2 are activated.  Scenes 3-4 are sent when Buttons 1-2 are deactivated.  This behavior is similar to the VRCS4 and VRCZ4 which have 4 buttons and 8 scenes.

I confirmed there are 4 scenes by using the Zwave PC controller to send SCENE_CONTROLLER_CONF_GET messages to the node for each scene.

### DIFF
--- a/packages/config/config/devices/0x001d/vrcs2.json
+++ b/packages/config/config/devices/0x001d/vrcs2.json
@@ -20,5 +20,9 @@
 	"firmwareVersion": {
 		"min": "0.0",
 		"max": "255.255"
+	},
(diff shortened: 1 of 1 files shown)


AlCalzone (approved):

AlCalzone:
Thanks!

PR #4522, opened 2022-04-28
fix(config): convert ADC-T2000 temperature values to partial parameters

### LINKED ISSUE #4276: Alarm.com ADC-T2000 Configuration Parameter 5
### Is your problem within Home Assistant (Core or Z-Wave JS Integration)?

NO, my problem is NOT within Home Assistant or the ZWave JS integration

### Is your problem within ZWaveJS2MQTT?

NO, my problem is NOT within ZWaveJS2MQTT

### Checklist

- [X] I have read the [changelog](https://github.com/zwave-js/node-zwave-js/blob/master/CHANGELOG.md) and my problem was **not** mentioned there.

### Describe the bug

The configuration for parameter 5 has some form of mask/offset, so that the valid range is not actually -100 to 100. For example, 704643072 in the parameter means 0 offset. As a result, zwavejs gives an error when trying to set this configuration parameter:
image

Here is how it should be set written by the developer of the device handler from smartthings:
https://community.smartthings.com/t/release-alarm-com-z-wave-thermostat-adc-t2000/74833/28

### Device information

Manufacturer: 0x0190
Model name: adc-t_2000

### How are you using `node-zwave-js`?

- [ ] `zwavejs2mqtt` Docker image (latest)
- [ ] `zwavejs2mqtt` Docker image (dev)
- [ ] `zwavejs2mqtt` Docker manually built _(please specify branches)_
- [ ] `ioBroker.zwave2` adapter _(please specify version)_
- [X] `HomeAssistant zwave_js` integration  _(please specify version)_
- [ ] `pkg`
- [ ] `node-red-contrib-zwave-js` _(please specify version, double click node to find out)_
- [ ] Manually built from GitHub _(please specify branch)_
- [ ] Other _(please describe)_

### Which branches or versions?

8.11.6

### Did you change anything?

no

### If yes, what did you change?

_No response_

### Did this work before?

No, it never worked anywhere

### If yes, where did it work?

_No response_

### Attach Driver Logfile

to be attached when required.

EiNSTeiN-:
Fix https://github.com/zwave-js/node-zwave-js/issues/4276

This adds the appropriate partial parameters for all configuration values that are temperature values. To reduce the risk of errors I copied most of this config from the ADC-T3000, it is pretty apparent that the T3000 is based on the T2000 since most config parameters are exactly the same, including min/max values and default values. Just one exception I noticed, the Swing (6) and Overshoot (7) are inverted between the two models, it's possible one of them is wrong, or they were actually inverted in the T3000 for some reason, but regardless in this PR I kept the order that currently exists.

I also ported roughly the same changes I had to make when adding the ADC-T3000 config, regarding mainly the use of `$import` and removing unnecessary descriptions.

### DIFF
--- a/packages/config/config/devices/0x0190/adc-t_2000.json
+++ b/packages/config/config/devices/0x0190/adc-t_2000.json
@@ -35,7 +35,6 @@
 		{
 			"#": "2",
 			"label": "Number of Heat Stages",
-			"description": "Heat Stages 0-3 Default is 2.",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 3,
@@ -44,7 +43,6 @@
 		{
 			"#": "3",
 			"label": "Number of Cool Stages",
-			"description": "Cool Stages 0-2 Default is 2.",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 2,
@@ -53,7 +51,6 @@
 		{
 			"#": "4",
 			"label": "Heat Fuel Type",
-			"description": "Choose type of fuel. Reality - whether unit is boiler vs forced air.",
 			"valueSize": 1,
 			"defaultValue": 1,
 			"allowManualEntry": false,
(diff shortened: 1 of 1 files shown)


AlCalzone (approved):
LGTM

PR #4565, opened 2022-05-04
feat: remember requested security classes, support inactive provisioning entries

### LINKED ISSUE #3682: Allow for "inactive" SmartStart provisioning entries, instead of deleting them
When excluding a node that has been provisioned with SmartStart, the provisioning entry will be removed automatically. This is done so that the node is not automatically re-included after the removal (devices will send out their SmartStart "inclusion" message after being excluded, which would cause a re-inclusion...).

In some scenarios it would be useful to keep the entry in the provisioning list, but mark it as "inactive". So the request here is to not remove entries on exclusion, but simply disable them. If a node advertises SmartStart inclusion, a deactivated node would not be included.

Instead of removing a provisioning entry upon node removal, mark the entry as "inactive". Users can manually remove the entry via the application if its no longer required. Applications can also allow for re-activating the entry so it can be included again.

_Use Case 1 - Seasonal Devices_
A seasonal smart plug that is only used during Christmas. Add the QR code to include it via SmartStart. When not in use, remove the node from the network. We want to keep the provisioning entry so next year we can just toggle it as "active" and it will re-add it self.

_Use Case 2 - Device Blocklists_
I want to block a certain DSK from my network, so it's never included. A persistent provisioning entry would be required to enable it. It's slightly different from the above case, but requires a persistent provisioning entry.

_Use Case 3 - S2 Bootstrapping fails_
Occasionally nodes fail S2 security bootstrapping. They must be re-included to be added with S2. If I'm using SmartStart, excluding will remove the provisioning entry, but I really want it to be included again right away. I could just re-enable the provisioning entry instead of importing the QR code again and losing any custom properties that were set (node and location in z2m). You might say, you can perform a factory reset, that removes the node but keeps the provisioning entry. That's true, but exclusion is usually easier. Also, it doesn't help with the next use case.

_Use Case 4 - Changing Configuration Parameters_
Some devices change their functionality based on configuration settings. For example, a HomeSeer switch can alternate between a Binary Switch or a Multilevel Switch (it literally changes the device type and product IDs!). I have a Zooz ZEN17 that changes endpoint behavior based on config parameters. After changing the settings, you must exclude the device and re-include. This would be easier to do if it weren't required to re-import the QR code. Factory reset is not an option.

Keep a backup of the QR code (or its text form), and re-import as needed.

**Additional context**

### LINKED ISSUE #4250: Requested Security Classes should be stored as part of the provisioning entries
... so applications can disable classes that were not requested and users can't enable them and wonder why it isn't working

AlCalzone:
This PR adds two additional (optional) fields to the `PlannedProvisioningEntry` (and by extension `IncludedProvisioningEntry`) interfaces.

The `status` field can assume the following values:
```ts
enum ProvisioningEntryStatus {
	Active,
	Inactive,
}
```
where an absent status will be assumed to be `Active` for backwards compatibility.
This can be used to disable provisioning entries. Inactive entries will be kept, but won't cause an inclusion when the corresponding node announces itself.
An entry can be set to inactive manually, or automatically during exclusion using the new possible value for the `unprovision` argument of `beginExclusion`:
```ts
await driver.controller.beginExclusion("inactive");
// this will exclude a node and set the corresponding provisioning entry to inactive afterwards
```
fixes: https://github.com/zwave-js/node-zwave-js/issues/3682

---

The optional `requestedSecurityClasses` (readonly) field keeps track of the security classes that have been requested by the device via QR code. This can be used in the provisioning entry editor to disable the checkboxes for security classes that haven't been requested, in order to avoid confusion caused by granting non-requested classes.

This is also added to the `QRProvisioningInformation` interface, so applications which copy the properties rather than pass a `QRProvisioningInformation` as a `PlannedProvisioningEntry` should take care of this new property.

fixes: https://github.com/zwave-js/node-zwave-js/issues/4250

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -169,12 +169,12 @@ Stops the inclusion process for a new node. The returned promise resolves to `tr
 ### `beginExclusion`
 
 ```ts
-async beginExclusion(unprovision?: boolean): Promise<boolean>
+async beginExclusion(unprovision?: boolean | "inactive"): Promise<boolean>
 ```
 
 Starts the exclusion process to remove a node from the network. The returned promise resolves to `true` if starting the exclusion was successful, `false` if it failed or if it was already active.
 
-The optional parameter `unprovision` specifies whether the removed node should be removed from the Smart Start provisioning list as well.
+The optional parameter `unprovision` specifies whether the removed node should be removed from the Smart Start provisioning list as well. A value of `"inactive"` will keep the provisioning entry, but disable it, preventing automatic inclusion of the corresponding nodes.
 
 ### `stopExclusion`
 
--- a/packages/core/src/security/QR.test.ts
+++ b/packages/core/src/security/QR.test.ts
@@ -43,6 +43,10 @@ describe("QR code parsing", () => {
 				SecurityClass.S2_Unauthenticated,
 				SecurityClass.S2_Authenticated,
 			],
+			requestedSecurityClasses: [
+				SecurityClass.S2_Unauthenticated,
+				SecurityClass.S2_Authenticated,
+			],
 			dsk: "51525-35455-41424-34445-31323-33435-21222-32425",
 			applicationVersion: "2.66",
 			genericDeviceClass: 0x11,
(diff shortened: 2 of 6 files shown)


AlCalzone:
cc @kpine @robertsLando

Issue #4249, opened 2022-02-18
Missing device configuration:  Neo Coolcam NAS-PD07Z


poudenes:
### Checklist

- [X] It is not in the [configuration DB](https://devices.zwave-js.io/?search=3)

- [X] It was not [merged recently](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Apr+is%3Aclosed) or has a [pending PR](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Aopen+is%3Apr)

### Which device is missing?

NAS-PD07Z

### Manufacturer ID

600

### Product Type

16

### Product ID

1824

### Firmware Version

2.6

### Do you have a manual?

No manual digital. Here configuration with all parameters:

<?xml version="1.0" encoding="utf-8"?>
<Product xmlns='https://github.com/OpenZWave/open-zwave' Revision="2">
    <MetaData>
        
        <MetaDataItem name="Name">Neo (Coolcam) NAS-PD07 Motion Sensor (PIR)</MetaDataItem>
        <MetaDataItem name="WakeupDescription">Press and hold the button for 5s until white led lights is on, then release the button before led turn off.</MetaDataItem>
        <MetaDataItem name="InclusionDescription">Set gateway in inclusion mode. Press and hold the button for 5s until white led lights is on, then release the button before led turn off. Blue led will blink with 1s interval until inclusion succesful.</MetaDataItem>
        <MetaDataItem name="ExclusionDescription">Set gateway in exclusion mode. Press and hold the button for 5s until white led lights is on, then release the button before led turn off. Blue led will blink with 0.5s interval until exclusion succesful.</MetaDataItem>
        <MetaDataItem name="ResetDescription">Press and hold button for 10s until pink led lights is on, then release the button before led turn off.</MetaDataItem>
        <ChangeLog>
            <Entry author="David J" date="20 october 2021" revision="1">Initial version, based on PD03Z file</Entry>
        </ChangeLog>
    </MetaData>
    
    
    <CommandClass id="112">
        <Value type="list" genre="config" instance="1" index="1" label="Led indicator enable" value="1" size="1">
            <Help>v2 This parameter is configured the led light on disable or enable.
            </Help>
            <Item label="Disable - Disable led blink. This configuration is not affect inclusion, exclusion and reset." value="0" />
            <Item label="Enable - Enable led blink when device detects a motion event (default)." value="1" />
        </Value>
        
[...]

PR #4576, opened 2022-05-05
fix(config): corrected labels and added new parameters for nas-pd07z config

MichaluxPL:
Updated paramInformation section. Previous version was incorrect (wrong some parameter labels and values) and didn't include all available parameters. Tested using NAS-PD07Z with firmware 2.6.

### DIFF
--- a/packages/config/config/devices/0x0258/nas-pd07z.json
+++ b/packages/config/config/devices/0x0258/nas-pd07z.json
@@ -14,238 +14,238 @@
 		"max": "255.255"
 	},
 	"paramInformation": [
-                {
-                        "#": "1",
-                        "label": "Led indicator enable",
-                        "valueSize": 1,
-                        "defaultValue": 1,
-                        "unsigned": true,
-                        "allowManualEntry": false,
-                        "options": [
-                                {
-                                        "label": "Disable",
-                                        "value": 0
-                                },
-                                {
-                                        "label": "Enable",
-                                        "value": 1
-                                }
-                        ]
-                },
-                {
-                        "#": "2",
-                        "label": "Motion Detection",
-                        "valueSize": 1,
-                        "unsigned": true,
-                        "defaultValue": 1,
-                        "allowManualEntry": false,
-                        "options": [
-                                {
-                                        "label": "Disable",
-                                        "value": 0
-                                },
-                                {
-                                        "label": "Enable",
-                                        "value": 1
-                                }
-                        ]
-                },
-                {
-                        "#": "3",
-                        "label": "Motion Alarm",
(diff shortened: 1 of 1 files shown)


MichaluxPL:
@zwave-js-bot fix lint

MichaluxPL:
fix: Updated paramInformation section. Previous version was incorrect (wrong some parameter labels and values) and didn't include all available parameters

AlCalzone:
There was a lengthy discussion about this device previously:
https://github.com/zwave-js/node-zwave-js/issues/4249

TL;DR:
There are two different variants available NAS-PD07Z and NAS-PD07ZU1, yet both seem to be sold as the same device without a clear distinction which is which. Even worse, both seem to be sold with an incorrect manual.

---

For this reason, I'm not going to blindly trust that your changes are correct. Please provide the data (device identification) and test results (logs from setting the different config parameters to their limits) we've asked from the other contributors in the linked issue, if you want to move this forward.

packages/config/config/devices/0x0258/nas-pd07z.json
```diff
@@ -10,63 +10,169 @@
...
 	],
 	"firmwareVersion": {
-		"min": "0.0",
+		"min": "2.6",
 		"max": "255.255"
 	},
 	"paramInformation": [
 		{
 			"#": "1",
-			"$import": "templates/shenzhen_neo_template.json#blink_on_motion"
```
AlCalzone, line 19: Also, please don't undo the template imports. They are done for consistency and to make some of the manufacturer-provided labels more easily understood.

packages/config/config/devices/0x0258/nas-pd07z.json
```diff
@@ -10,63 +10,169 @@
...
+			"options": [
+				{
+					"label": "Disable",
+					"value": 0
+				},
+				{
+					"label": "Enable",
+					"value": 1
+				}
+			]
```
AlCalzone, line 51: For example, this change does nothing except replace the template import with its "full" representation

packages/config/config/devices/0x0258/nas-pd07z.json
```diff
@@ -10,63 +10,169 @@
...
+			"options": [
+				{
+					"label": "Every time",
+					"value": 0
+				},
+				{
+					"label": "Only once",
+					"value": 1
+				}
+			]
```
AlCalzone, line 70: And this is also just a more verbose variant of the imported template "Motion Sensor: Prevent Re-Trigger While On"

AlCalzone (changes requested):

PR #4586, opened 2022-05-07
feat: add `sendAndReceiveData` method to `ManufacturerProprietaryCCAPI`

### LINKED ISSUE #4341: Support bidirectional Manufacturer Proprietary messages

I'm writing [some code](https://github.com/j9brown/node-zwave-pe653) to update the firmware of my Intermatic PE653 spa controller.  The protocol is quite simple but it requires sending and receiving `Manufacturer Proprietary` messages.

The `zwave-js` API already offers a [sendData function](https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/ManufacturerProprietaryCC.ts#L37) to send `Manufacturer Proprietary` messages but it does not currently support receiving replies, which I need for this project.

I'd like an API that sends a manufacturer proprietary command and returns the payload of the received manufacturer proprietary response.  Say, something like `sendAndReceiveData`.  We can safely assume that the response will have the same `manufacturerId` as the command that was sent.

I think this could be implemented by creating a subclass of `ManufacturerProprietaryCC` that is suitably decorated to expect a `ManufacturerProprietaryCC` in response.  Or perhaps `DynamicCC` could be used here.  I'm not very familiar with the zwave-js API.

I considered trying to implement this functionality on top of the available `zwave-js` API but it seems pretty tricky because the library heavily relies on reflection and I'm poking at it from the side via `zwavejs2mqtt`.  So an actual `zwave-js` API function would be best.

**Additional context**

My little afternoon project: [node-zwave-pe653](https://github.com/j9brown/node-zwave-pe653)

j9brown:
Adds an API that can be used to send ManufacturerProprietaryCC messages that expect responses from the device.

fixes: #4341

### DIFF
--- a/packages/zwave-js/src/lib/commandclass/ManufacturerProprietaryCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/ManufacturerProprietaryCC.ts
@@ -52,13 +52,12 @@ export class ManufacturerProprietaryCCAPI extends CCAPI {
 	}
 
 	@validateArgs()
-	public async sendAndReceiveData(
-		manufacturerId: number,
-		data?: Buffer,
-	): Promise<unknown> {
-		const cc = new ManufacturerProprietaryCCWithResponse(this.driver, {
+	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+	public async sendAndReceiveData(manufacturerId: number, data?: Buffer) {
+		const cc = new ManufacturerProprietaryCC(this.driver, {
 			nodeId: this.endpoint.nodeId,
 			endpoint: this.endpoint.index,
+			expectsResponse: true,
 		});
 		cc.manufacturerId = manufacturerId;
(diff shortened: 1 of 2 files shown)


AlCalzone:
@j9brown I like the approach and I tried to use the existing methods to conditionally define an expected response instead of defining a separate subclass.
Can you test if this still works with your device?

j9brown:
Looks good.  I'll give it a try!

j9brown:
LGTM.  Tested your new changes and it works fine.  Thanks!

j9brown (commented):
LGTM.  Tested these changes and it works fine.  Looks a bit cleaner too.  Thanks!

j9brown:
Hmm, do we have to add docs to ManufacturerProprietary.md for this new method or will they be automatically generated in some way?

AlCalzone:
They will be automatically generated after merging.

PR #4598, opened 2022-05-09
fix: only consider active SmartStart provisioning entries when auto-enabling listening mode

AlCalzone:


AlCalzone:
@zwave-js-bot pack this

PR #4607, opened 2022-05-10
fix(config): work around broken firmware of Intermatic PE653

j9brown:
With this configuration, the device works as expected except for the thermostat setpoint.

### DIFF
--- a/packages/config/config/devices/0x0005/pe653.json
+++ b/packages/config/config/devices/0x0005/pe653.json
@@ -55,5 +55,63 @@
 			"maxValue": 7,
 			"defaultValue": 0
 		}
+	],
+	"endpoints": {
+		"0": {
+			"associations": {
+				"1": {
+					"label": "Lifeline",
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0005/pe653.json
```diff
@@ -55,5 +55,63 @@
...
+		"0": {
+			"associations": {
+				"1": {
+					"label": "Lifeline",
+					"maxNodes": 5,
+					"isLifeline": true
+				}
+			}
+		}
+	},
```
AlCalzone: This isn't 100% clear from the documentation, but the default is endpoint 0
```suggestion
	"associations": {
		"1": {
			"label": "Lifeline",
			"maxNodes": 5,
			"isLifeline": true
		}
	},
```

AlCalzone (changes requested):
This is a seriously broken device 🤯

AlCalzone (approved):

PR #3693, opened 2021-11-10
fix(cc): update thermostat setpoint metadata when unit changes

AlCalzone:

PR #4609, opened 2022-05-11
Proof of concept: let device config files override the interview for specific CCs

### LINKED ISSUE #1625: Add compat flag to override values reported by a device
To work around this device firmware (!) issue on the zwave-js side, we need to add a compat flag that allows us to override specific values and avoid querying them from the device (because in this case the supported colors response is simply wrong).

__Originally posted by @AlCalzone in https://github.com/zwave-js/zwavejs2mqtt/issues/477#issuecomment-774475891__

### LINKED ISSUE #4588: Intermatic PE653 compatibility issues, firmware dependent
### Is your problem within Home Assistant (Core or Z-Wave JS Integration)?

NO, my problem is NOT within Home Assistant or the ZWave JS integration

### Is your problem within ZWaveJS2MQTT?

NO, my problem is NOT within ZWaveJS2MQTT

### Checklist

- [X] I have read the [changelog](https://github.com/zwave-js/node-zwave-js/blob/master/CHANGELOG.md) and my problem was **not** mentioned there.

### Describe the bug

The Intermatic PE653 doesn't seem to work correctly with zwave-js.  The nature of the problems depends on the firmware running on the device.

Firmware v3.3:
  * The thermostat isn't correctly discovered because the device uses type 7 (furnace) but zwave-js's algorithm guesses that the device has types 0 and 10 so it doesn't work.
  * The 5 binary switches are discovered correctly, can be set, and can be polled.

Firmware v3.4:
  * The same thermostat issue as v3.3.
  * The 5 binary switches are discovered correctly, but attempts to set or get their values fail because the device ignores the request for some reason. It's unclear whether this is a bug in the firmware or whether zwave-js is formatting the multichannel command encapsulation messages in a way that the device doesn't understand.

I recently upgraded the firmware of my device to v3.4 since that's what's shipped from the factory these days and what I presume most other users have, with the hope of resolving the thermostat compatibility issue. I didn't want to submit compatibility fixes for the thermostat that might break behavior for users of v3.4. That said, v3.4 seems quite broken so I may revert to v3.3 after all.

(I'd be keen to know if anyone is actually using v3.4 successfully with zwave-js.  FWIW, I originally had this device fully working with v3.3 and OpenZwave, including the thermostat.)

I'm going to keep investigating to determine a course of action to improve compatibility.

See also: [discussion](https://community.home-assistant.io/t/intermatic-pe653rc-pool-controller-and-ha/342836/12)

### Device information

Manufacturer: Intermatic
Model name: PE653
Node ID in your network: 42

### How are you using `node-zwave-js`?

- [X] `zwavejs2mqtt` Docker image (latest)
- [ ] `zwavejs2mqtt` Docker image (dev)
- [ ] `zwavejs2mqtt` Docker manually built _(please specify branches)_
- [ ] `ioBroker.zwave2` adapter _(please specify version)_
- [ ] `HomeAssistant zwave_js` integration  _(please specify version)_
- [ ] `pkg`
- [ ] `node-red-contrib-zwave-js` _(please specify version, double click node to find out)_
- [ ] Manually built from GitHub _(please specify branch)_
- [ ] Other _(please describe)_

### Which branches or versions?

zwavejs2mqtt: 6.8.1
zwave-js: 9.0.7
home id: 4146670975
home hex: 0xf7292d7f

### Did you change anything?

no

### If yes, what did you change?

_No response_

### Did this work before?

No, it never worked anywhere

### If yes, where did it work?

_No response_

### Attach Driver Logfile

[zwavejs-pe653-discovery-with-firmware-v3.4.log](https://github.com/zwave-js/node-zwave-js/files/8649189/zwavejs-pe653-discovery-with-firmware-v3.4.log)

j9brown:
This is a proof of concept of a general purpose mechanism to allow device config files to work around invalid information supplied by devices.  It introduces a new config flag "commandClasses.interview" that when present causes the node to skip the interview for a given CC and endpoint and to store the supplied values into the valueDB.

This set of patches includes a configuration file that has a rather absurd number of bugs including off-by-one errors in the thermostat setpoint bitmap and some quirks with the setpoint scale.  These changes completely resolve #4588.

Let me know what you think!

fixes: https://github.com/zwave-js/node-zwave-js/issues/1625

### DIFF
--- a/packages/config/src/devices/CompatConfig.ts
+++ b/packages/config/src/devices/CompatConfig.ts
@@ -414,6 +414,50 @@ All values in compat option commandClasses.add must be objects`,
 				this.addCCs = addCCs;
 			}
 
+			if (definition.commandClasses.interview != undefined) {
+				if (!isObject(definition.commandClasses.interview)) {
+					throwInvalidConfig(
+						"devices",
+						`config/devices/${filename}:
+	error in compat option commandClasses.interview`,
+					);
+				} else if (
+					!Object.keys(definition.commandClasses.interview).every(
+						(k) => hexKeyRegex2Digits.test(k),
+					)
+				) {
+					throwInvalidConfig(
+						"devices",
+						`config/devices/${filename}:
+	All keys in compat option commandClasses.interview must be 2-digit lowercase hex numbers!`,
+					);
+				} else if (
+					!Object.values(definition.commandClasses.interview).every(
+						(v) => isObject(v),
+					)
+				) {
+					throwInvalidConfig(
+						"devices",
+						`config/devices/${filename}:
+	All values in compat option commandClasses.interview must be objects`,
+					);
+				}
+
+				const interviewCCs = new Map<
+					CommandClasses,
+					CompatInterviewCC
+				>();
+				for (const [cc, info] of Object.entries(
+					definition.commandClasses.interview,
+				)) {
+					interviewCCs.set(
+						parseInt(cc),
+						new CompatInterviewCC(filename, info as any),
+					);
+				}
+				this.interviewCCs = interviewCCs;
+			}
+
 			if (definition.commandClasses.remove != undefined) {
 				if (!isObject(definition.commandClasses.remove)) {
 					throwInvalidConfig(
@@ -487,6 +531,10 @@ compat option alarmMapping must be an array where all items are objects!`,
 
 	public readonly alarmMapping?: readonly CompatMapAlarm[];
 	public readonly addCCs?: ReadonlyMap<CommandClasses, CompatAddCC>;
+	public readonly interviewCCs?: ReadonlyMap<
+		CommandClasses,
+		CompatInterviewCC
+	>;
 	public readonly removeCCs?: ReadonlyMap<
 		CommandClasses,
 		"*" | readonly number[]
@@ -528,6 +576,7 @@ compat option alarmMapping must be an array where all items are objects!`,
 		return pick(this, [
 			"alarmMapping",
 			"addCCs",
+			"interviewCCs",
 			"removeCCs",
 			"disableBasicMapping",
 			"disableStrictEntryControlDataValidation",
@@ -636,6 +685,69 @@ invalid endpoint index in compat option commandClasses.add`,
 	}
 
 	public readonly endpoints: ReadonlyMap<number, Partial<CommandClassInfo>>;
+}
+
+export interface CompatInterviewCCValue {
+	property: string | number;
+	propertyKey?: string | number;
+	value: unknown;
+}
+
+export class CompatInterviewCC {
+	public constructor(filename: string, definition: JSONObject) {
+		const endpoints = new Map<number, CompatInterviewCCValue[]>();
+		const parseEndpointInfo = (endpoint: number, info: JSONObject) => {
+			let parsed: CompatInterviewCCValue[] = [];
+			if (info.values != undefined) {
+				if (
+					!isArray(definition.values) ||
+					!definition.values.every(
+						(value: unknown) =>
+							isObject(value) &&
+							value.property != undefined &&
+							(typeof value.property === "string" ||
+								typeof value.property === "number") &&
+							(value.propertyKey == undefined ||
+								typeof value.propertyKey === "string" ||
+								typeof value.propertyKey === "number") &&
+							value.value != undefined,
+					)
+				) {
+					throwInvalidConfig(
+						"devices",
+						`config/devices/${filename}:
+	error in compat option commandClasses.interview, malformed value object`,
+					);
+				}
+				parsed = info.values;
+			}
+			endpoints.set(endpoint, parsed);
+		};
+		// Parse root endpoint info if given
+		if (definition.values != undefined) {
+			// We have info for the root endpoint
+			parseEndpointInfo(0, definition);
+		}
+		// Parse all other endpoints
+		if (isObject(definition.endpoints)) {
+			if (
+				!Object.keys(definition.endpoints).every((k) => /^\d+$/.test(k))
+			) {
+				throwInvalidConfig(
+					"devices",
+					`config/devices/${filename}:
+invalid endpoint index in compat option commandClasses.interview`,
+				);
+			} else {
+				for (const [ep, info] of Object.entries(definition.endpoints)) {
+					parseEndpointInfo(parseInt(ep), info as any);
+				}
+			}
+		}
+		this.endpoints = endpoints;
+	}
+
+	public readonly endpoints: ReadonlyMap<number, CompatInterviewCCValue[]>;
 }
 
 export interface CompatMapAlarmFrom {
--- a/packages/zwave-js/src/lib/commandclass/ThermostatSetpointCC.ts
+++ b/packages/zwave-js/src/lib/commandclass/ThermostatSetpointCC.ts
@@ -664,7 +664,9 @@ export class ThermostatSetpointCCReport extends ThermostatSetpointCC {
 			this.endpointIndex,
 			this._type,
 		);
-		valueDB.setValue(scaleValueId, this._scale.key);
+		if (!valueDB.hasValue(scaleValueId)) {
+			valueDB.setValue(scaleValueId, this._scale.key);
+		}
 		return true;
 	}
 
--- a/packages/zwave-js/src/lib/node/Node.ts
+++ b/packages/zwave-js/src/lib/node/Node.ts
@@ -1749,7 +1749,7 @@ protocol version:      ${this.protocolVersion}`;
 			if (instance.interviewComplete) return "continue";
 
 			try {
-				await instance.interview();
+				await this.compatInterviewCC(instance);
 			} catch (e) {
 				if (isTransmissionError(e)) {
 					// We had a CAN or timeout during the interview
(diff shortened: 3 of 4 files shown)


packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -2032,6 +2032,44 @@ protocol version:      ${this.protocolVersion}`;
...
+					this.driver.controllerLog.logNode(
+						this.id,
+						`failed to set compatibility value from device configuration for ${getCCName(
+							cc.ccId,
+						)}, endpoint ${cc.endpointIndex}: '${
+							compatValue.property
+						}' is not a registered internal value`,
+						"error",
+					);
+				}
```
AlCalzone, line 2065: I was thinking there needs to be a way to validate the overridden values, but you thought about that already. I also like limiting it to internal values. 👍️

packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -2032,6 +2032,44 @@ protocol version:      ${this.protocolVersion}`;
...
+			}
+
+			await cc.refreshValues();
```
AlCalzone, line 2068: We need to make this conditional. Probably by adding a method to the `CommandClass` class which defaults to `true` and can be overridden in subclasses. `UserCodeCC` for one does not call this during the interview.

packages/zwave-js/src/lib/commandclass/ThermostatSetpointCC.ts
```diff
@@ -664,7 +664,9 @@ export class ThermostatSetpointCCReport extends ThermostatSetpointCC {
 			this.endpointIndex,
 			this._type,
 		);
-		valueDB.setValue(scaleValueId, this._scale.key);
+		if (!valueDB.hasValue(scaleValueId)) {
+			valueDB.setValue(scaleValueId, this._scale.key);
+		}
```
AlCalzone, line 669: This was specifically added for devices that can change the setpoint unit during usage:
https://github.com/zwave-js/node-zwave-js/pull/3693

only updating this when the value doesn't exit breaks that functionality.

packages/config/src/devices/CompatConfig.ts
```diff
@@ -638,6 +687,69 @@ invalid endpoint index in compat option commandClasses.add`,
 	public readonly endpoints: ReadonlyMap<number, Partial<CommandClassInfo>>;
 }
 
+export interface CompatInterviewCCValue {
+	property: string | number;
+	propertyKey?: string | number;
+	value: unknown;
+}
+
+export class CompatInterviewCC {
```
AlCalzone, line 696: I think this (and the definition) in config files can be simplified by including the endpoint (optional, defaults to the root/0) in `CompatInterviewCCValue`, so `values` is always a flat object. You can still separate it into a map by endpoints though (or alternatively `filter()` at the usage site).

AlCalzone (changes requested):
Looks really good. A few things need to be changed though - and please add the compat flag explanation in the documentation under "File Format".

packages/zwave-js/src/lib/commandclass/ThermostatSetpointCC.ts
AlCalzone: Does the device actually send its reports with the wrong scale or is this change unnecessary maybe?

packages/zwave-js/src/lib/node/Node.ts
j9brown: Hmm, interesting. The CCs all handle this behavior in slightly different ways. What are you thoughts on this?

1. We could add a `shouldRefreshValuesDuringCompatInterview()` method that is overridden by UserCodeCC and other CCs as needed to make the refresh conditional when using compatibility flags. This alternative would result in duplicated logic between the regular interview and compat interview code paths.
2. We could add a `shouldRefreshValuesDuringInterview()` method that is overridden by UserCodeCC and other CCs to make the refresh conditional during normal or compatibility interviews and then refactor places where those CCs directly call `refreshValues()` during the interview to call this new method (or perhaps a wrapper like `refreshValuesDuringInterviewIfNeeded()`). This alternative is a bit more invasive but it keeps the code paths more convergent.
3. We could let the compat interview flag itself state what should happen by way of a `refreshValues` option. This alternative requires the least changes but it risks breaking the expected behavior of `driver.options.interview.queryAllUserCodes` in UserCodeCC.

I'm leaning towards something like option 2 because I think it's a bit tidier in the long term.

packages/zwave-js/src/lib/commandclass/ThermostatSetpointCC.ts
j9brown: This device is really broken. It sends reports with scale == 1 but it expects settings to be made with scale == 0. It just ignores settings with scale == 1.

I suppose we could work around this by adding a new internal value to control whether the scale is learned from recently received reports or kept static from the time of the interview.  What do you think?

(I'm half tempted to disassemble the firmware and patch the bugs directly but that would be a whole other project and it'd be of less benefit to other users of these devices.)

packages/config/src/devices/CompatConfig.ts
j9brown: Hmm.  I considered this at first but then I thought it might be useful to allow additional flags to be defined at the per-endpoint level.

For example, for some devices it might be useful to perform the interview as usual but then overwrite certain values after the fact. In that case, adding a `skipInterview` flag that acts on the endpoint (not a value) would make sense. Similarly, alternative 3 in one of my comments above imagines a `refreshValues` flag for an endpoint.

I also think it's better to maintain the symmetry in the syntax with `commandClasses.add` and `commandClasses.remove` since they all basically work the same way with respect to modeling each endpoint as a separate object. I don't feel strongly about it though, so let me know what you prefer.

AlCalzone: fair enough

packages/zwave-js/src/lib/commandclass/ThermostatSetpointCC.ts
AlCalzone: In that case, a compat flag disabling the scale learning would be more "correct", although this starts to go into territory I usually prefer to avoid - hyper specific flags tailored to individual devices. Lets do it though.

packages/zwave-js/src/lib/node/Node.ts
AlCalzone: Me too, I'd probably use an overload for `refreshValues(partOfInterview: boolean)` which is marked `@internal` so it doesn't get exposed to users of the library. I initially thought this feature would require something like 2. anyways, but for every single query. Maybe I'll do that at some point anyways, but that is going to require larger changes to the value definitions first.

Your solution is much less invasive with the downside that the entire interview of a CC needs to be replicated in the config file.

Speaking of, we need to be able to set metadata during the fake interview as well. There are several devices which report incorrect color channels and I'd like to use this feature for the Color Switch CC too, which sets metadata:
https://github.com/zwave-js/node-zwave-js/blob/bb924723cf234c0c153cae74143e9cf8bf5b7170/packages/zwave-js/src/lib/commandclass/ColorSwitchCC.ts#L491-L511

This will be a bit ugly since the entire metadata object will require to be repeated in the config file, but I can live with that.

packages/zwave-js/src/lib/commandclass/ThermostatSetpointCC.ts
j9brown: Sounds good to me.

packages/zwave-js/src/lib/node/Node.ts
j9brown: Gotcha.  Adding metadata should be straightforward.

In preparation for some refactoring, I took a closer look at how `refreshValues()` gets called from `interview()` and I see a few different patterns in use. Can you check my assumptions?

1. Some CCs call `refreshValues()` unconditionally right at the end of the interview.
2. Some CCs call `refreshValues()` conditionally during the interview, such as `UserCodeCC` (based on a driver option) and `ThermostatSetpointCC` (version dependent, perhaps because with version <= 2 it's querying all of the setpoint values anyhow so it's kind of refreshing values as a side effect).
3. Some CCs perform additional work during the interview after calling `refreshValues()` that's important for using the device, such as `AssociationCC` (setting the lifeline) and `BasicCC` (removing the CC if it can't obtain the value).
4. Some CCs don't call `refreshValues()` during the interview at all, such as `WakeUpCC` (doesn't have any values).

So overall, I think we can classify the behavior into 4 cases:

1. CCs that don't have any values to refresh.
2. CCs that are refreshed with discretion (UserCodeCC). Is it just because it's slow or can there be undesirable side-effects too ?
3. CCs whose interview does not have a data dependency on values and only needs to ensure that they have been populated.
4. CCs whose interview has a data dependency on values because they are needed to determine the node's capabilities, to ensure that the CC actually seems to be working, or to configure it correctly.

My little hack skips the whole interview. Unfortunately, that skips all of its potential side-effects too.

I think a better solution might be to split the interview into three (or more?) steps:

1. `interview(driver)`: Query the node's capabilities and figure out what makes it tick and set up its internal values and metadata. This step might be altered or skipped due to compat flags.
2. `configure(driver)`: Configure the node to apply side-effects (such as setting lifeline associations).
3. `refreshValues(driver, true)`: Refresh any remaining values of the node that have yet to be queried since the interview started so that they're available to the user right away (except for cases like UserCodeCC where that's considered too expensive). We could always refresh all values at this step but that might result in redundant work in cases where the interview has data dependencies on certain values and will have queried them already unless we are using compat flags to skip the interview altogether.

Thoughts?

j9brown:
Incidentally, another approach I considered instead of using compat flags to set internal values was to add a compat flag to rewrite requests or responses from the nodes.  e.g. When preparing to send command XYZ to the node, rewrite it as ABC.  Or, instead of actually sending XYZ, use this simulated response DEF.

The problem with rewriting messages is that we'd have to come up with some syntax for expressing messages and rewrite rules in device config files. Messages could be expressed as hex strings but that's not very user friendly or flexible. Rules could be expressed using a template processing language but that might be overkill. And we can't just reuse the existing CC serialization with reflection to make JSON objects to compare and rewrite because there are side-effects in the CC constructors (especially reports).

So at least for now, it seems better to solve the problem in a different way.  *shrug*

(I guess we do still have the option of adding a thermostat specific compat flag with an option to override the scale value and disable learning and an option to apply a workaround for off-by-on errors in the setpoint type bitmap encoding.)

packages/zwave-js/src/lib/node/Node.ts
AlCalzone: I like it. I'm assuming `refreshValues` checks if the condition is fulfilled when it is indicated that it is called during an interview?

AlCalzone:
> The problem with rewriting messages is that we'd have to come up with some syntax for expressing messages and rewrite rules in device config files. Messages could be expressed as hex strings but that's not very user friendly or flexible. Rules could be expressed using a template processing language but that might be overkill.

A "nicer" way would probably be hooking into the API calls via compat flags, e.g. by stating what a call to `ThermostatSetpointCCAPI.getSupported` returns. But that still has the downside that some of the information is stored as a side effect of the constructor, which would need to be refactored to be done in a different location.

Side note: I'm currently in the process of a huge refactor which will remove the hard coupling between CC / Message classes and the driver instance. Once that is done, this approach will be much simpler to implement.

I'm fine with either approach. The one followed in this PR can be more fine-grained, but is more verbose.
Replacing the API calls is currently blocked by https://github.com/zwave-js/node-zwave-js/pull/4614 (or probably another PR I'll cherry-pick out of that soon)

AlCalzone:
@j9brown I finally went forward with the approach I mentioned in my last comment here --> https://github.com/zwave-js/node-zwave-js/pull/5987

The device-specific fixes you did here are now tracked in https://github.com/zwave-js/node-zwave-js/issues/5990

PR #4622, opened 2022-05-15
fix(config): rework GE/Jasco config params to templates

disforw:
Moving all config params for the GE/Jasco devices to template for a more universal solution for configuring devices.


blhoward2:
Thanks!

I know this is still a draft, but some preliminary feedback. Use the master templates where they exist. So like in your GE template, you should be using the base_enable_disable one and then override the label and such. For some like the LED functions, there are likely master templates you can drop straight into the file. The idea is to build them up so that we can make changes in one place if need be and have it reflect everywhere. (Like enable/disable may eventually become a new bool type.)

And we try to avoid descriptions for parameters whenever possible, so if the label is clear or can be made clear, please drop the description. 

If you look in the Aeotec folder you should see what I mean.

disforw:
Can you give me an example of using the master template and overriding the title? Can I import the master template from a template? Or does that have to be done in the device?

disforw:
@zwave-js-bot fix lint

AlCalzone:
> Can you give me an example of using the master template and overriding the title?

The aforementioned Aeotec template has some examples:
https://github.com/zwave-js/node-zwave-js/blob/e4179e2b857e6567f21fba4b0af42515741be53f/packages/config/config/devices/0x0086/templates/aeotec_template.json#L373-L377

You can import recursively over multiple files, e.g. device file -> manufacturer template -> master template, and even within a file:
https://github.com/zwave-js/node-zwave-js/blob/2e6f2ead631ff26d7eb8bd97cbe25d131b83fc0c/packages/config/config/devices/0x010f/fgs214.json#L179-L180
The only thing you cannot do is a circular import.

We try to limit ourselves to the structure where a device file imports from the manufacturer template and/or the master template though.

disforw:
Amazing. I'll continue to point from device to the local template, then once that's all done, before I submit for review, I'll recursively import from the master template wherever possible.

disforw:
@zwave-js-bot fix lint

disforw:
@zwave-js-bot fix lint

disforw:
I'm really liking where this PR went. There are still a few devices where I left the parameters alone because they seemed complicated or one-off.

blhoward2:
This will take us a bit to get through as there are so many. Did you check for different ranges and such when deciding which template works?

disforw:
Yes, of course. I was extremely careful when doing this. First I moved everything to the local template. Then compared the values from the local template to the master template. Then I did a few mass replacements once I saw they were available in the master (IE 3 or 4 option LED indicator).

blhoward2:
Got it, thanks! Are you ready for us to review?

blhoward2:
If you click on Details next to the failing lint checks, you have a syntax error in one of the files.

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+			{
+				"label": "Notification",
+				"value": 1
+			},
+			{
+				"label": "Basic Set",
+				"value": 2
+			},
+			{
+				"label": "Basic report",
```
AlCalzone:
```suggestion
				"label": "Basic Report",
```

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+				"value": 3
+			}
+		]
+	},
+	"adjust_refresh": {
+		"label": "Adjust Refresh",
+		"description": "Adjusts refresh time from 1-55 seconds",
```
AlCalzone: What does this do? Is this like an automatic reporting interval? In any case, the label should be clear and the description is unnecessary.

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+		"unsigned": true,
+		"options": [
+			{
+				"label": "Default - 10 seconds",
+				"value": 0
+			},
+			{
+				"label": "Seconds",
+				"value": 55
+			}
```
AlCalzone: This option seems unnecessary

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+		"unit": "seconds",
+		"minValue": 0,
+		"maxValue": 55,
+		"defaultValue": 0,
+		"unsigned": true,
+		"options": [
+			{
+				"label": "Default - 10 seconds",
+				"value": 0
+			},
```
AlCalzone:
```suggestion
			{
				"label": "10 seconds",
				"value": 0
			},
```
We avoid mentioning "default" in the labels, since UIs add that themselves.

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+	"dimmer_steps": {
+		"label": "Dimmer Steps",
+		"description": "Number of dimming steps or levels",
+		"$import": "~/templates/master_template.json#base_1-99_nounit"
+	},
+	"dimmer_times": {
+		"label": "Dimmer Step Time",
+		"description": "Timing of dimming steps or levels",
+		"$import": "~/templates/master_template.json#dimming_timing"
+	},
```
AlCalzone: The descriptions for these and the following seem redundant.

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+	"dimmer_steps_z": {
+		"label": "Dimmer Steps (Remote Command)",
+		"description": "Number of dimming steps or levels when sent from a zwave command",
+		"$import": "~/templates/master_template.json#base_1-99_nounit"
+	},
+	"dimmer_times_z": {
+		"label": "Dimmer Step Time (Remote Command)",
+		"description": "Timing of dimming steps or levels when sent from a zwave command",
+		"$import": "~/templates/master_template.json#dimming_timing"
+	},
```
AlCalzone: Can you please make the labels consistent with this template?
https://github.com/zwave-js/node-zwave-js/blob/8e6da5d5a42b81f1542ab38b061e0102380d0bdd/packages/config/config/devices/0x0001/templates/act_template.json#L61-L76

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+	"dimmer_steps_a": {
+		"label": "Dimmer Steps (Group Command)",
```
AlCalzone: Not sure what the difference between a Z-Wave command and a Group Command is supposed to be. Maybe the first is from the controller and the latter through associations from other nodes?

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+		"description": "Set the default brightness level that the dimmer will turn on when being controlled manually",
+		"valueSize": 1,
+		"minValue": 0,
+		"maxValue": 99,
+		"defaultValue": 0,
+		"unsigned": true
+	},
+	"dim_level": {
+		"label": "Dim level",
+		"description": "Change brightness of associated light bulb(s)",
```
AlCalzone: How about this?
```suggestion
		"label": "Associated Light Brightness",
```

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+			{
+				"label": "Medium",
+				"value": 2
+			},
+			{
+				"label": "Low",
+				"value": 3
+			}
+		]
+	},
```
AlCalzone: These are the same aside from the default value, aren't they?

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+				"value": 0
+			},
+			{
+				"label": "Enabled",
+				"value": 1
+			}
+		]
+	},
+	"motion_sensor": {
+		"label": "Enable/Disable Motion Sensing",
```
AlCalzone: Please don't use Enable/Disable in parameter labels:
https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide?id=special-note-enabledisable

Also, the options should be Enable and Disable without "d"

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+			},
+			{
+				"label": "Enabled",
+				"value": 1
+			}
+		]
+	},
+	"ignore_start_level": {
+		"label": "Ignore Start-Level (Receiving)",
+		"description": "This dimmer will start dimming from its current level.",
```
AlCalzone: I think this would be clearer.
```suggestion
		"label": "Start Dimming From Current Level",
		"description": "The start level in Z-Wave commands will be ignored, if enabled",
```

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+		"unsigned": true,
+		"allowManualEntry": false,
+		"options": [
+			{
+				"label": "No",
+				"value": 0
+			},
+			{
+				"label": "Yes",
+				"value": 1
```
AlCalzone:
```suggestion
				"label": "Disable",
				"value": 0
			},
			{
				"label": "Enable",
				"value": 1
```

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+		"valueSize": 1,
+		"minValue": 0,
+		"maxValue": 255,
+		"defaultValue": 4,
+		"unsigned": true,
+		"options": [
+			{
+				"label": "1 min to 60 min",
+				"value": 60
+			},
```
AlCalzone: This seems weird. Which is it?

AlCalzone: Should this parameter rather have a "minutes" unit?

packages/config/config/devices/0x0063/templates/config_template.json
```diff
@@ -0,0 +1,468 @@
...
+			{
+				"label": "Last dim level",
+				"value": 1
+			},
+			{
+				"label": "Last state",
+				"value": 2
+			}
+		]
+	}
```
AlCalzone: We have master_templates for this, State After Power Failure.

AlCalzone (changes requested):
Here's some preliminary feedback on the template.json, since it will take a few days until I can give this a proper review.
Especially labels/descriptions could use some work to match the style guide:
https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide?id=labels
https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide?id=parameter-descriptions

Many descriptions are unnecessary or can be made so by changing the labels a bit.

packages/config/config/devices/0x0063/templates/config_template.json
disforw: I looked at this, unfortunately not. Look at the options, one counts down and one counts up.

packages/config/config/devices/0x0063/templates/config_template.json
disforw: The options in the master template do not match, I changed the language to match that from the master however.

packages/config/config/devices/0x0063/templates/config_template.json
disforw: Changed

packages/config/config/devices/0x0063/templates/config_template.json
disforw: Removed options

packages/config/config/devices/0x0063/templates/config_template.json
disforw: Done

packages/config/config/devices/0x0063/templates/config_template.json
disforw: Removed

disforw:
@zwave-js-bot fix lint

disforw:
I made all the changes to the template you requested. I even fixed a couple of grouping names to match each other. I think I’m finally done and ready for final review.

AlCalzone:
Thanks, I'll try to give you a review soon.

packages/config/config/devices/0x0063/55256_zw1002.json
```diff
@@ -33,26 +33,7 @@
...
-				{
-					"label": "LED On when light is turned On",
-					"value": 1
-				},
-				{
-					"label": "LED always Off",
-					"value": 2
-				}
-			]
+			"$import": "~/templates/master_template.json#led_indicator_three_options"
```
futurexdesign:
```suggestion
			"$import": "~/templates/master_template.json#led_indicator_four_options"
```

futurexdesign: the 55256 has a four option LED

AlCalzone:
Just a heads up: Jasco is also working on these files. I'm currently waiting on more info from them, so we can progress with both their and your changes in a way that causes the least conflicts.

AlCalzone (approved):
Thanks for your patience. Since Jasco are going to give this another pass, I'll just skip on the detailed review.
What you did here is heaps better than we had before.

AlCalzone:
@disforw just got off a call with Jasco - they are very grateful for this PR, you saved them a ton of work.
Thought I'd let you know ;)

disforw:
:) glad I could help, I have 42 Jasco devices in my house so this was big for me.

PR #4619, opened 2022-05-13
fix(config): refactor Zooz config files to use templates where possible

kaechele:
As promised in #4496 I have reworked the entire Zooz config subfolder to make use of templates wherever possible.

I recognize that this is a massive changeset, so it will take some time to review. Any help appreciated. Many changes follow simple patterns so once you get the hang of it it should become straightforward. I found having a mental model of how the `base_` templates work helps to speed up the review. E.g. `base_enable_disable` is `0 = Disable` and `1 = Enable` with a `defaultValue` of `0` (Disable).
It is also largely untested as I do not own most of the devices touched by these changes. Any guidance on how to best test this is appreciated. I'm willing to spend the effort on testing.

Happy to incorporate any feedback. E.g. you may want me to shuffle stuff around in the `master_template.json` or `zooz_template.json`. Especially the latter is getting a bit crowded by now that I was thinking maybe alphabetically sorting stuff would make sense. Zooz devices do have quite an overlap (and often not entirely consistent) it turns out.
Also Zooz documentation is of varying quality, especially the older the device is. I would have fixed it for them if their Knowledge base had been a Wiki.
I have cross referenced the documentation for each device many times to ensure that what I'm swapping out for a template is actually accurately represented by that template.
However, especially historic documentation is often unavailable and the change logs often merely allude to what parameters were changed and in which ways. So I've done my best to keep that information intact.

I have further removed `description` fields that were overly obvious.

PR #4634, opened 2022-05-19
fix(config): change param 3 label to Enable/Disable Auto Off for ZEN-21

GTerplan:
After checking the manual for the ZEN21 I found that parameter 3 was incorrectly labeled along with the primary and secondary options being flipped.  I changed the file to what it should be and included the manual for review so these changes can be committed to the master copy.
[zooz-zwave-plus-on-off-switch-zen21-ver4.0-manual.pdf](https://github.com/zwave-js/node-zwave-js/files/8721714/zooz-zwave-plus-on-off-switch-zen21-ver4.0-manual.pdf)

### DIFF
--- a/packages/config/config/devices/0x027a/zen21.json
+++ b/packages/config/config/devices/0x027a/zen21.json
@@ -86,17 +86,17 @@
 		},
 		{
 			"#": "3",
-			"label": "Enable/Disable LED Indicator",
+			"label": "Enable/Disable Auto Off",
 			"valueSize": 1,
 			"defaultValue": 0,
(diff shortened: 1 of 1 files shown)


AlCalzone:
Thanks, this is already being taken care of in this other open PR https://github.com/zwave-js/node-zwave-js/pull/4619/files#diff-d1e47a0edf5f7bc0a3dc9c205b46fcb9890bf865064cfd0a06b19084c737cb19

It seems that the meaning of the parameter was completely changed with firmware 3.0, so the fix here wouldn't be sufficient anyways.

PR #4640, opened 2022-05-19
fix(config): rework Nortek config params to templates

disforw:
Moving all config params for the Nortek devices to template for a more universal solution for configuring devices.

### DIFF
--- a/packages/config/config/devices/0x014f/lb60z-1.json
+++ b/packages/config/config/devices/0x014f/lb60z-1.json
@@ -23,39 +23,15 @@
 	"paramInformation": [
 		{
 			"#": "1",
-			"label": "Initial Bulb Brightness",
-			"valueSize": 1,
-			"defaultValue": 0,
-			"unsigned": true,
-			"allowManualEntry": false,
-			"options": [
-				{
-					"label": "Full Brightness",
-					"value": 0
-				},
-				{
-					"label": "Prior Brightness Level",
-					"value": 1
-				}
-			]
+			"$import": "templates/config_template.json#init_brightness"
 		},
 		{
 			"#": "9",
-			"label": "Dim Step Level",
-			"valueSize": 1,
-			"minValue": 1,
-			"maxValue": 99,
-			"defaultValue": 1,
-			"unsigned": true
+			"$import": "~/templates/master_template.json#dim_rate"
 		},
 		{
 			"#": "10",
-			"label": "Dim Speed",
-			"valueSize": 1,
-			"minValue": 1,
-			"maxValue": 10,
-			"defaultValue": 3,
-			"unsigned": true
+			"$import": "templates/config_template.json#dim_speed"
 		}
 	],
 	"metadata": {
--- a/packages/config/config/devices/0x014f/lb60z-1.json
+++ b/packages/config/config/devices/0x014f/lb60z-1.json
@@ -27,7 +27,7 @@
 		},
 		{
 			"#": "9",
-			"$import": "~/templates/master_template.json#dim_rate"
+			"$import": "templates/config_template.json#dim_step_zwave"
 		},
 		{
 			"#": "10",
--- a/packages/config/config/devices/0x014f/lbr30z-1.json
+++ b/packages/config/config/devices/0x014f/lbr30z-1.json
@@ -20,7 +20,7 @@
 		},
 		{
 			"#": "9",
-			"$import": "~/templates/master_template.json#dim_rate"
+			"$import": "templates/config_template.json#dim_step_zwave"
 		},
 		{
 			"#": "10",
--- a/packages/config/config/devices/0x014f/lbr30z-1.json
+++ b/packages/config/config/devices/0x014f/lbr30z-1.json
@@ -16,38 +16,15 @@
 	"paramInformation": [
 		{
 			"#": "1",
-			"label": "Dim Level Memory",
-			"valueSize": 1,
-			"defaultValue": 0,
-			"allowManualEntry": false,
-			"options": [
-				{
-					"label": "Disable dim level memory",
-					"value": 0
-				},
-				{
-					"label": "Enable dim level memory",
-					"value": 1
-				}
-			]
+			"$import": "templates/config_template.json#init_brightness"
 		},
 		{
 			"#": "9",
-			"label": "Dim / Bright Step Level",
-			"description": "How much the brightness will change with each dimming step",
-			"valueSize": 1,
-			"minValue": 1,
-			"maxValue": 99,
-			"defaultValue": 1
+			"$import": "~/templates/master_template.json#dim_rate"
 		},
 		{
 			"#": "10",
-			"label": "Dim / Bright Speed",
-			"description": "How fast the brightness will change with each dimming step",
-			"valueSize": 1,
-			"minValue": 1,
-			"maxValue": 10,
-			"defaultValue": 3
+			"$import": "templates/config_template.json#dim_speed"
 		}
 	]
 }
--- a/packages/config/config/devices/0x014f/smart_retrofit_lighting_kit.json
+++ b/packages/config/config/devices/0x014f/smart_retrofit_lighting_kit.json
@@ -14,40 +14,17 @@
 		"max": "255.255"
 	},
 	"paramInformation": [
-		{
+		{{
 			"#": "1",
-			"label": "Dim level when the light bulb is turned On",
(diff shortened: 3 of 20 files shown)


disforw:
@zwave-js-bot fix lint

disforw:
@zwave-js-bot fix lint

packages/config/config/devices/0x014f/templates/config_template.json
```diff
@@ -0,0 +1,183 @@
...
+	},
+	"power_monitor": {
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "Power Monitoring",
+		"description": "Provide a visual display of the amount of energy being consumed",
+		"defaultValue": 1
+	},
+	"energy_display_duration": {
+		"label": "LED Display Duration",
+		"description": "Sets the length of time for which the LED will reflect the current power usage",
```
AlCalzone: Description seems unnecessary. Maybe some of the information should go into the label instead. For example:
```suggestion
		"label": "LED Indicator: Power Usage Duration",
```

packages/config/config/devices/0x014f/templates/config_template.json
```diff
@@ -0,0 +1,183 @@
...
+		"label": "Dimming Steps (All-On/All-Off)"
+	},
+	"dim_rate_all": {
+		"$import": "~/templates/master_template.json#dimming_timing",
+		"label": "Dimming Rate (All-On/All-Off)"
+	},
+	"power_monitor": {
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "Power Monitoring",
+		"description": "Provide a visual display of the amount of energy being consumed",
```
AlCalzone: Description seems unnecessary

packages/config/config/devices/0x014f/wd500z5-1.json
```diff
@@ -16,90 +16,40 @@
...
 		},
 		{
 			"#": "5",
-			"label": "Ignore Start Level in Dim Command",
-			"description": "The Start Level sent in a Dim command can be ignored or not.",
-			"valueSize": 1,
-			"minValue": 0,
-			"maxValue": 1,
-			"defaultValue": 1
+			"label": "Start Level in Dim Command",
```
AlCalzone:
```suggestion
			"label": "Ignore Start Level in Dim Command",
```

AlCalzone: That made sense :)

packages/config/config/devices/0x014f/wt00z-1.json
```diff
@@ -54,231 +54,75 @@
...
-				{
-					"label": "Enable",
-					"value": 0
-				},
-				{
-					"label": "Disable",
-					"value": 1
-				}
-			]
+			"label": "Group 4 during a dim command",
```
AlCalzone:
```suggestion
			"label": "Group 4: Send Dimming Commands",
```

packages/config/config/devices/0x014f/wt00z-1.json
```diff
@@ -54,231 +54,75 @@
...
-				{
-					"label": "Enable Group 4",
-					"value": 0
-				},
-				{
-					"label": "Disable Group 4",
-					"value": 1
-				}
-			]
+			"label": "Group 4 Reporting",
```
AlCalzone:
```suggestion
			"label": "Group 4: Send Reports",
```

packages/config/config/devices/0x014f/lb60z-1.json
```diff
@@ -23,39 +23,15 @@
...
 		},
 		{
 			"#": "9",
-			"label": "Dim Step Level",
-			"valueSize": 1,
-			"minValue": 1,
-			"maxValue": 99,
-			"defaultValue": 1,
-			"unsigned": true
+			"$import": "~/templates/master_template.json#dim_rate"
```
AlCalzone: Shouldn't this be the dim step instead?
```suggestion
			"$import": "templates/config_template.json#dim_step_zwave"
```

AlCalzone: In my mind, dim_rate is **how fast** and dim_step is **how much**

packages/config/config/devices/0x014f/lbr30z-1.json
```diff
@@ -16,38 +16,15 @@
...
 		},
 		{
 			"#": "9",
-			"label": "Dim / Bright Step Level",
-			"description": "How much the brightness will change with each dimming step",
-			"valueSize": 1,
-			"minValue": 1,
-			"maxValue": 99,
-			"defaultValue": 1
+			"$import": "~/templates/master_template.json#dim_rate"
```
AlCalzone: Shouldn't this be the dim step instead?
```suggestion
			"$import": "templates/config_template.json#dim_step_zwave"
```

AlCalzone: Please double-check the other files too if they need to be adjusted

packages/config/config/devices/0x014f/smart_retrofit_lighting_kit.json
```diff
@@ -16,38 +16,15 @@
...
+			"$import": "templates/config_template.json#init_brightness"
 		},
 		{
 			"#": "9",
-			"label": "Dim / Bright Step Level",
-			"valueSize": 1,
-			"minValue": 1,
-			"maxValue": 99,
-			"defaultValue": 1
+			"$import": "~/templates/master_template.json#dim_rate"
```
AlCalzone: Shouldn't this be the dim step instead?
```suggestion
			"$import": "templates/config_template.json#dim_step_zwave"
```

AlCalzone (changes requested):
Thanks. A couple of remarks

packages/config/config/devices/0x014f/templates/config_template.json
disforw: I will change to “Usage indicator display duration”

packages/config/config/devices/0x014f/templates/config_template.json
disforw: removed

packages/config/config/devices/0x014f/smart_retrofit_lighting_kit.json
disforw: I searched for all instances of this, and only came across these 3 mentioned. The documentation does indeed call it Dim Steps.

packages/config/config/devices/0x014f/wd500z5-1.json
disforw: Actually, the "ignore" doesn't really make sense to me, and by default the param is disabled (set to 1), so I removed the word "ignore" and inverted the enable/disable options to try and simplify it. If that doesn't make sense then I will remove the invert, and change the label as requested.

disforw: see line 38 in file wt00z-1.json for comparison
see line 33 in file wd500z-1.json for comparison

disforw: once we come to a decision on this, I will change all 3 locations.

AlCalzone: So, there exists a StartLevelChange command which tells a device to start dimming. This command can include a start level indicating where the dimming should start and/or it can have a flag "Ignore Start Level", in which case the start level byte gets ignored and dimming starts at the current level.

I'm not sure if this parameter is concerned with incoming or outgoing commands, but if its for incoming ones (which I assume), "Start Dimming From Current Level" might be a less confusing label.

AlCalzone: Maybe it even makes sense for outgoing ones 🤔

disforw: hold on.. I need to re-read that..lol
does what I did with inverting the enable/disable make sense? creating a double negative for simplicity?

disforw: For Nortek devices, by default, the starting dim level is ignored and the flag is set to 1. so if we went with your suggestion of "Start Dimming From Current Level" then 1 has to be default and mean Enable and 0 will mean Disable

disforw: Dome

AlCalzone (approved):

PR #4652, opened 2022-05-25
feat: add labels to Multilevel Switch/Entry Control notifications

raman325:
We get friendly labels on the Notification CC notifications but not for these other two CCs. These labels help downstream systems understand what to present to the user.

~~Question: will `pick` work with getter functions?~~ 🎉 it does


AlCalzone (approved):
ok with my changes

PR #4661, opened 2022-05-30
fix: swap order of destroy and Driver_Failed error after restoring NVM

AlCalzone:


AlCalzone:
@zwave-js-bot pack this

marcus-j-davies:
Ok - On further review, it happened again :sweat_smile: 

Testing this patch - The resolved promise was received and reacted to, before the error - but a 2nd attempt, the other way around. 

I think this is a **Race Condition.** if an application is listening for a driver error (I.e to restart), it could do so before receiving the resolved promise.

I'm starting to think that applications (specifically those that utilise the server) may have to very very lightly throttle driver restarts, so it can continue any logic path that is the result of a pending promise being recieved.

Example:

Inside a server node app , bridging `zwave-js-server` to another language (that is monitoring the `stderr` of the node process) it could receive an error before `zwave-js-server`  has had an opportunity to respond with a successful promise

```javascript
driver.on("error", (e) => {
    if (e instanceof ZWaveError && e.code === ZWaveErrorCodes.Driver_Failed) {
        process.stderr.write("2\n");
    }
});
```

if the above is received before the promise reaches the server client - an application could prematurely restart.
unless you feel the swap really makes a difference, I'm starting to think this is more a structural consideration for client applications using `zwave-js-server`

AlCalzone:
I can try emitting the driver failed error in a setImmediate, maybe that will help.

marcus-j-davies:
Ok,

That makes sense, allowing the promise to reach the end of its context, before the error is produced.

Here if I am not mistaken.

https://github.com/zwave-js/node-zwave-js/blob/8b920a1c84bb1218e5d801355bbba134b6fc998b/packages/zwave-js/src/lib/controller/Controller.ts#L4766

AlCalzone:
@zwave-js-bot pack this

marcus-j-davies:
@AlCalzone

`nextTick` seems to have done the trick. running an NVM Restore 3-4 times, I have each time, been in receipt of the resolved promsie prior to the restore required error, therefore handling the completion before the driver/server  is taken down. 👍

PR #4685, opened 2022-06-02
feat: faster inclusion on slow networks

### LINKED ISSUE #4631: Expose some node information before bootstrapping it after inclusion
e.g. by exposing the node ID, device class and CC list as a separate event.

rossnelson:
feat: faster inclusion on slow networks

This feature adds:

- early notification of node prior to bootstrapping
- disabling the post bootstrapping interview
- allowing the user to interview nodes at will

This is generally useful in commercial installs where many devices are being
paired quickly. Many interviews being processed in a short period of
time tends to slow the inclusion process and in turn reduces install
productivity.

fixes #4631

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -858,6 +858,14 @@ A node could not be included into or excluded from the network for some reason.
 
 The process to include or exclude a node was stopped successfully. Note that these events are also emitted after a node was included or excluded.
 
+### `"node found"`
+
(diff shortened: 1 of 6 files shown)


AlCalzone (changes requested):
Thanks. I have a few nits I need to change, but I can't push here, so I'll continue this PR in https://github.com/zwave-js/node-zwave-js/pull/4692

PR #4697, opened 2022-06-08
feat(test): implement communication between mock controller and mock nodes

AlCalzone:
This PR implements the bare minimum communication between mock controller and mock nodes. At this point we can "interview" a mock node, except for the CC-specific queries.


AlCalzone:
@zwave-js-bot fix lint

PR #4706, opened 2022-06-11
fix(config): Correct Jasco 55256 configuration

futurexdesign:
The settings defined for the Jasco 55256 do not match the actual behavior of the switch.  The options as originally defined match an older version of the manual that is not correct. 

I've updated these options to reflect the actual behavior of the physical device. 

- The PDF manual from z-wavealliance.org has the incorrect settings:  https://products.z-wavealliance.org/ProductManual/File?folder=&filename=product_documents/3892/55256_QSG_v1[1].pdf 

- The updated manual from Jasco reflects the correct behavior: https://byjasco.com/mwdownloads/download/link/id/2572

The edit date on the two sets of instructions are separated by 4 days, I guess they figured out their error?

### DIFF
--- a/packages/config/config/devices/0x0063/55256_zw1002.json
+++ b/packages/config/config/devices/0x0063/55256_zw1002.json
@@ -41,11 +41,11 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "LED ON when load is Off",
+					"label": "LED On when load is On",
 					"value": 0
 				},
 				{
-					"label": "LED On when light is turned On",
+					"label": "LED On when load is Off",
 					"value": 1
 				},
 				{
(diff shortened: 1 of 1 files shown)


AlCalzone:
We need to be a bit careful here.
First there is an open PR reworking all the Jasco files: https://github.com/zwave-js/node-zwave-js/pull/4622
This might change this error, but I'm not sure. Can you check if that is the case?

And Jasco themselves are looking into cleaning their configurations up, but I haven't seen a work in progress yet.

futurexdesign:
You are correct, the templated change fixes _most_ of the issue.   I added a suggestion to switch it to the 4 option template however.   

I'll close this as unneeded.

PR #4742, opened 2022-06-24
feat: add methods for whether firmware update is in progress

raman325:
There's no reason why the driver can't return whether a firmware update is in progress. Instead of relying on the application to track it, consider cases like zwavejs2mqtt where there are two applications (z2m and zwave-js-server) that can initiate updates. IMO it needs to be centralized in the driver

### DIFF
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -534,6 +534,13 @@ export class ZWaveController extends TypedEventEmitter<ControllerEventCallbacks>
 		this.driver.cacheSet(cacheKeys.controller.supportsSoftReset, value);
 	}
 
+	public get isAnyFirmwareUpdateInProgress(): boolean {
+		for (const node of this._nodes.values()) {
+			if (node.isFirmwareUpdateInProgress) return true;
+		}
+		return false;
+	}
+
 	private _nodes: ThrowingMap<number, ZWaveNode>;
 	/** A dictionary of the nodes connected to this controller */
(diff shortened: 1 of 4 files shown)


packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -534,6 +534,13 @@ export class ZWaveController extends TypedEventEmitter<ControllerEventCallbacks>
 		this.driver.cacheSet(cacheKeys.controller.supportsSoftReset, value);
 	}
 
+	public get isAnyFirmwareUpdateInProgress(): boolean {
```
AlCalzone: Should we call this `isOTAFirmwareUpdateInProgress` in order to distinguish node updates from the OTW update for the controller (which I'll eventually add)?

Also should this be a getter/property or a method? Feels more like a method at first glance.

packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -795,6 +795,10 @@ export class ZWaveNode
 		}
 	}
 
+	public get isFirmwareUpdateInProgress(): boolean {
```
AlCalzone: Should this be a getter/property or a method? Feels more like a method at first glance.

AlCalzone (changes requested):
Thanks - two/three thoughts:

packages/zwave-js/src/lib/controller/Controller.ts
raman325: good idea, and we can make it a method

AlCalzone (approved):

PR #4758, opened 2022-06-29
feat: add support for Z-Wave JS firmware update service

AlCalzone:
This PR adds experimental support for the new firmware update service powered by https://github.com/zwave-js/firmware-updates/.
For now the support is limited to OTA upgrades and there are no available updates yet.

/cc @martinhjelmare @raman325


packages/zwave-js/src/lib/driver/FirmwareUpdateService.ts
```diff
@@ -0,0 +1,98 @@
...
+			`Unsupported hash algorithm ${hashAlgorithm} for integrity check`,
+			ZWaveErrorCodes.Argument_Invalid,
+		);
+	}
+	// TODO: Make request abort-able (requires AbortController, Node 14.17+ / Node 16)
+
+	// Download the firmware file
+	const downloadResponse = await axios.get<Buffer>(file.url, {
+		timeout: 60000,
+		maxContentLength: 10 * 1024 * 1024, // 10MB should be enough for any conceivable Z-Wave chip
```
robertsLando: I personally would place both this in a constant instead like `DOWNLOAD_TIMEOUT` and `MAX_UPDATE_SIZE`

marcus-j-davies:
Perfect!

Checking for and applying the found updates being 2 methods, implementations can schedule the update checks, and allow the front ends to use their built in notification engines, to display the available update!

image


I was going to suggest to have the ability to pass a node into the FW update check, as opposed to the numerical identifiers, but seen you already did 😅

marcus-j-davies (dismissed):

AlCalzone:
> I was going to suggest to have the ability to pass a node into the FW update check, as opposed to the numerical identifiers, but seen you already did sweat_smile

What did I do? 😅

marcus-j-davies:
Ask for the Node ID instead of all the identifiers 😄 
https://github.com/zwave-js/node-zwave-js/blob/d387473e19eae7740eac245867f512fe03109f62/packages/zwave-js/src/lib/driver/Driver.ts#L4299

docs/api/driver.md
```diff
@@ -312,6 +312,57 @@ Checks whether there is a compatible update for the currently installed config p
 
 > [!NOTE] Although the updated config gets loaded after the update, bugfixes and changes to device configuration generally require either a driver restart or re-interview of the changed devices to take effect.
 
+### `getAvailableFirmwareUpdates`
```
MartinHjelmare: What are the thoughts around placing these methods on the controller instead of on the driver?

The interface accepts a node id and previously it's been the controller that handles node related things. We also already have a method for OTA update status on the controller.

robertsLando (dismissed):

marcus-j-davies (dismissed):

docs/api/driver.md
AlCalzone: My thought was that this does HTTP requests to external services and we already have precedence for this with the config updates which also live on the driver.

For me, the Controller is more the Z-Wave chip, but I can see how in this case the boundaries aren't as clear.

Will think about it.

AlCalzone: moved to the Controller class

test/run.ts
```diff
@@ -36,6 +36,20 @@ const driver = new Driver(port, {
...
+		// try {
+		// 	console.log(`Installing update ${updates[0].version}...`);
+		// 	await wait(1000);
+		// 	await driver.controller.beginOTAFirmwareUpdate(
+		// 		2,
+		// 		updates[0].files[0],
+		// 	);
+		// } catch (e) {
+		// 	console.error(e);
+		// }
```
robertsLando, line 52: Why is this commented now?

AlCalzone: So I don't accidentally upgrade my device firmware next time I run a test.

robertsLando: I thought that was running using your fake driver 😆

AlCalzone: No that was a real-world test

robertsLando (approved):

marcus-j-davies (dismissed):

robertsLando:
@AlCalzone Could you pack this?

AlCalzone:
@zwave-js-bot pack this

robertsLando (changes requested):
Like discussed in SLACK we might want to throw after a certain amount of time if device doesn't wake up instead of wait indeterminately

Issue #4497, opened 2022-04-22
`currentValue` not updated correctly when Supervision is used


amilanov75:
### Is your problem within Home Assistant (Core or Z-Wave JS Integration)?

NO, my problem is NOT within Home Assistant or the ZWave JS integration

### Is your problem within ZWaveJS2MQTT?

NO, my problem is NOT within ZWaveJS2MQTT

### Checklist

- [X] I have checked the [troubleshooting section](https://zwave-js.github.io/node-zwave-js/#/troubleshooting/index) and my problem is **not** described there.

- [X] I have read the [changelog](https://github.com/zwave-js/node-zwave-js/blob/master/CHANGELOG.md) and my problem was **not** mentioned there.

### Describe the bug

What causes the bug?
When I increase the frequency of messages per second to more than ~10 (i.e. every 100ms) and definitely at 20 (i.e. every 50ms), syncing issues appear between the device actual state and the state reported in the Node-Red UI.

What do you observe?
When changing the brightness or turning off a group of lights (e.g. around 5x Fibaro Dimmer 2s and 5x Fibaro RGBW2s) the reported state in Node-Red UI will not update, but the device is changing it's state.

What did you expect to happen?
For example, when turning all 10x lights off, the Node-Red UI should report a Current Value = 0 against all lights. For one (or more) of the lights the value does not change from the current value e.g. 10. The light (or lights) do turn off.
Repeatedly sending a payload = 0 will not update the NR UI.

Steps to reproduce the behaviour:
By increasing the rate of messages per second to approx >10 per second, the issue occurs.
By decreasing the rate of messages per second to approx <3 per second, the issue does not occur

### Device information

Node ID = 11
Device = Fibaro RGBW2

As per images below:
Payload = 0
Reported Value in NR UI = 10

### How are you using `node-zwave-js`?

- [ ] `zwavejs2mqtt` Docker image (latest)
- [ ] `zwavejs2mqtt` Docker image (dev)
- [ ] `zwavejs2mqtt` Docker manually built _(please specify branches)_
- [ ] `ioBroker.zwave2` adapter _(please specify version)_
- [ ] `HomeAssistant zwave_js` integration  _(please specify version)_
- [ ] `pkg`
- [X] `node-red-contrib-zwave-js` _(please specify version, double click node to find out)_
- [ ] Manually built from GitHub _(please specify branch)_
- [ ] Other _(please describe)_

### Which branches or versions?

Version Information

 Module Version 
[...]

PR #4761, opened 2022-06-30
feat: improve Supervision CC usage, use automatically instead of hard-coded

AlCalzone:
The old CC specific hardcoded implementation was awkward and unflexible. The driver now automatically determines for all commands whether supervision can and should be used and does so automatically. As a result, `sendCommand` calls may now return a `SupervisionResult`.

fixes: #4716
fixes: https://github.com/zwave-js/node-zwave-js/issues/4497


packages/cc/src/cc/ConfigurationCC.ts
```diff
@@ -416,22 +417,27 @@ export class ConfigurationCCAPI extends CCAPI {
...
 			parameter: property,
 			value: targetValue,
 			valueSize: valueSize as any,
 			valueFormat,
 		});
 
-		if ((this as ConfigurationCCAPI).isSinglecast()) {
-			// Verify the current value after a delay
+		if (
+			!isUnsupervisedOrSucceeded(result) &&
```
OXINARF: Shouldn't this be `!supervisedCommandSucceeded(result)`? As it is, if it is unsupervised, it won't verify.

packages/cc/src/cc/BarrierOperatorCC.ts
```diff
@@ -232,12 +236,14 @@ export class BarrierOperatorCCAPI extends CCAPI {
 				value === BarrierState.Closed
 					? BarrierState.Closed
 					: BarrierState.Open;
-			await this.set(targetValue);
+			const result = await this.set(targetValue);
 
-			// Verify the change after a delay
-			if (this.isSinglecast()) {
+			// Verify the change after a delay, unless the command was supervised and successful
+			if (!supervisedCommandSucceeded(result)) {
```
OXINARF: Was it intended to remove the isSinglecast check?

packages/cc/src/cc/BasicCC.ts
```diff
@@ -109,34 +112,39 @@ export class BasicCCAPI extends CCAPI {
 		if (typeof value !== "number") {
 			throwWrongValueType(this.ccId, property, "number", typeof value);
 		}
-		await this.set(value);
+
+		const result = await this.set(value);
+
+		// If the command did not fail, assume that it succeeded and update the currentValue accordingly
+		// so UIs have immediate feedback
```
OXINARF: Comment is duplicated below.

packages/cc/src/cc/MultilevelSwitchCC.ts
```diff
@@ -393,81 +309,104 @@ export class MultilevelSwitchCCAPI extends CCAPI {
...
+									// eslint-disable-next-line @typescript-eslint/no-empty-function
+									void this.get().catch(() => {});
+								}
+							},
+					  }
+					: {},
+			).set(value, duration);
 
 			// If the command did not fail, assume that it succeeded and update the currentValue accordingly
 			// so UIs have immediate feedback
```
OXINARF, line 347: Comment is duplicated below.

OXINARF (commented):
Just a couple of things I noticed while looking at the PR, sorry if it is just noise, I know this is still in draft.

AlCalzone:
Thanks @OXINARF

PR #4792, opened 2022-07-11
fix(config): mark parameters of Ness Corporation 117001 as writable

kizza42:
Updated "readOnly" values to be writable as per the actual product.


kizza42:
There seems to be have been an error when we created the entry on Zwave Alliance. The config params should be writable.

marcus-j-davies:
@kizza42 

 - If the params are not read only, remove the attribute entirely.
    Its only needed, if the param is read only
    
 - You need to use  a Semantic PR title : `fix(config): correct readOnly values`

AlCalzone (approved):
Ok with my changes

PR #4809, opened 2022-07-18
feat(config): update Zooz ZEN17 to firmware 1.10

diegolrz:
Added parameters 19 and 20 per ZOOZ official documentation https://www.support.getzooz.com/kb/article/698-zen17-universal-relay-advanced-settings/

PR description here

### DIFF
--- a/packages/config/config/devices/0x027a/zen17.json
+++ b/packages/config/config/devices/0x027a/zen17.json
@@ -490,6 +490,92 @@
 			]
 		},
 		{
+			"#": "19",
+			"label": "Reverse the reported values for your selected input type on S1 C terminals",
+			"description": "Reverse the reported values for your selected input type on S1 C terminals (select the same value as you chose for parameter 2 to reverse the values for open and closed circuit)"
+			"valueSize": 1,
+			"defaultValue": 0,
+			"unsigned": true,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Values reported as normally open, not reversed (default)",
+					"value": 0
+				},
+				{
+					"label": "Leak alarm (water sensor)",
+					"value": 4
+				},
+				{
+					"label": "Heat alarm",
+					"value": 5
+				}
+				{
+					"label": "Motion alert",
+					"value": 6
+				},
+				{
+					"label": "Open/close alert (door sensor)",
+					"value": 7
+				},
+				{
+					"label": "CO alarm",
+					"value": 8
+				}
+				{
+					"label": "CO2 alarm",
+					"value": 9
+				}
+				{
+					"label": "On/off report (dry contact switch/sensor)",
+					"value": 10
+				}
+			]
+		},
+		{
+			"#": "20",
+			"label": "Reverse the reported values for your selected input type on S2 C terminals",
+			"description": "Reverse the reported values for your selected input type on S2 C terminals (select the same value as you chose for parameter 2 to reverse the values for open and closed circuit)"
+			"valueSize": 1,
+			"defaultValue": 0,
+			"unsigned": true,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Values reported as normally open, not reversed (default)",
+					"value": 0
+				},
+				{
+					"label": "Leak alarm (water sensor)",
+					"value": 4
+				},
+				{
+					"label": "Heat alarm",
+					"value": 5
+				}
+				{
+					"label": "Motion alert",
+					"value": 6
+				},
+				{
+					"label": "Open/close alert (door sensor)",
+					"value": 7
+				},
+				{
+					"label": "CO alarm",
+					"value": 8
+				}
+				{
+					"label": "CO2 alarm",
+					"value": 9
+				}
+				{
+					"label": "On/off report (dry contact switch/sensor)",
+					"value": 10
+				}
+			]
+		},
+		{
 			"#": "24",
 			"label": "DC Motor Mode",
 			"description": "Sync R1 and R2 together to prevent both being activated at the same time",
--- a/packages/config/config/devices/0x027a/zen17.json
+++ b/packages/config/config/devices/0x027a/zen17.json
@@ -66,7 +66,7 @@
 		},
 		{
 			"#": "2",
-			"$if": "firmwareVersion >= 1.4",
+			"$if": "firmwareVersion >= 1.1",
 			"label": "Input Type for S1 Terminal",
 			"description": "Note: The device must be excluded (without resetting it) and re-included after changing to/from values 4-11",
 			"valueSize": 1,
@@ -126,7 +126,7 @@
 		},
 		{
 			"#": "2",
-			"$if": "firmwareVersion < 1.4",
+			"$if": "firmwareVersion < 1.1",
 			"label": "Input Type for S1 Terminal",
 			"description": "Note: The device must be excluded (without resetting it) and re-included after changing to/from values 4-10",
 			"valueSize": 1,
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
Your file is incorrect JSON:
> Could not parse config file /home/runner/work/node-zwave-js/node-zwave-js/packages/config/config/devices/0x027a/zen17.json: JSON5: invalid character '\"' at 496:4

Also please make sure to follow the style guide, specifically https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide?id=parameter-descriptions
> unnecessary descriptions that merely restate the label must be removed
> Only include a description if it is necessary, helpful, and adds significant value.

Were these parameters added in a specific firmware version? If yes, use [conditions](https://zwave-js.github.io/node-zwave-js/#/config-files/conditional-settings?id=conditional-configuration-settings) to only expose them for these versions.

diegolrz:
> Your file is incorrect JSON:
> 
> > Could not parse config file /home/runner/work/node-zwave-js/node-zwave-js/packages/config/config/devices/0x027a/zen17.json: JSON5: invalid character '"' at 496:4
> 
> Also please make sure to follow the style guide, specifically https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide?id=parameter-descriptions
> 
> > unnecessary descriptions that merely restate the label must be removed
> > Only include a description if it is necessary, helpful, and adds significant value.
> 
> Were these parameters added in a specific firmware version? If yes, use [conditions](https://zwave-js.github.io/node-zwave-js/#/config-files/conditional-settings?id=conditional-configuration-settings) to only expose them for these versions.

I believe all the changes required are done. Let me know if there is anything else that I need to fix.

Thanks @AlCalzone

packages/config/config/devices/0x027a/zen17.json
```diff
@@ -489,6 +489,96 @@
 				}
 			]
 		},
+		{
+			"#": "19",
+			"label": "Reverse the reported values for your selected input type on S1 C terminals",
+			"description": "Select the same value as you chose for parameter 2 to reverse the values for open and closed circuit",
+			"valueSize": 1,
+        	"minValue": 0,
+        	"maxValue": 10,
```
AlCalzone: These are not necessary since they are inferred from the options.
```suggestion
```

packages/config/config/devices/0x027a/zen17.json
```diff
@@ -489,6 +489,96 @@
...
+				}
+			]
+		},
+		{
+			"#": "20",
+			"label": "Reverse the reported values for your selected input type on S2 C terminals",
+			"description": "Select the same value as you chose for parameter 2 to reverse the values for open and closed circuit",
+			"valueSize": 1,
+	       	"minValue": 0,
+        	"maxValue": 10,
```
AlCalzone:
```suggestion
```

packages/config/config/devices/0x027a/zen17.json
```diff
@@ -489,6 +489,96 @@
 				}
 			]
 		},
+		{
+			"#": "19",
+			"label": "Reverse the reported values for your selected input type on S1 C terminals",
```
AlCalzone: Label is much too long. Something like this would be better:
```suggestion
			"label": "S1 Terminal: Invert Values",
```
also:
* rename param 2 to "S1 Terminal: Input Type"
* Move this directly after param 2
* Copy the option labels from param 2, so they are consistent.

packages/config/config/devices/0x027a/zen17.json
```diff
@@ -489,6 +489,96 @@
...
+				{
+					"label": "On/off report (dry contact switch/sensor)",
+					"value": 10
+				}
+			]
+		},
+		{
+			"#": "20",
+			"label": "Reverse the reported values for your selected input type on S2 C terminals",
+			"description": "Select the same value as you chose for parameter 2 to reverse the values for open and closed circuit",
```
AlCalzone: See comment for param 19

packages/config/config/devices/0x027a/zen17.json
```diff
@@ -66,7 +66,7 @@
 		},
 		{
 			"#": "2",
-			"$if": "firmwareVersion >= 1.4",
+			"$if": "firmwareVersion >= 1.1",
```
AlCalzone: I didn't find this info, can you tell me where you did?

AlCalzone (changes requested):
Thanks - a couple more changes please.

packages/config/config/devices/0x027a/zen17.json
diegolrz: I emailed Zooz asking for the latest firmware. They said the newest file is 1.2. On their advanced parameters page it states that all of those apply from firmware 1.1. The garage mode is shown on those instructions, so anything newer than 1.1 should work.

packages/config/config/devices/0x027a/zen17.json
diegolrz: Updated as suggested on 19

packages/config/config/devices/0x027a/zen17.json
AlCalzone: Ok just to be sure: This is actually firmware 1.1 and 1.2 and not 1.10 and 1.20?
Because the website says this:
> Here is a full list of advanced settings (parameters) for the ZEN17 Universal Relay VER. **1.10** firmware:

and I know they have been numbering their firmware version weirdly (i.e. only multiples of 10) for display purposes.

diegolrz: I asked them about that… this is what they said:

“Home Assistant omits the 0 when displaying firmware files, which means firmware 1.02 displays as 1.2, as does 1.20.“

What is then the right way of setting this up in the code?

The official firmware version on the file is 1.20. It is confusing what they said on the email. I can only assume that to play nice with HA they are maintaining the firmware revisions at 1.X level.

AlCalzone: > which means firmware 1.02 displays as 1.2, as does 1.20

Yeah, thats not true. 1.02 displays as 1.2, 1.20 as 1.20.

In Z-Wave, the versions are encoded just two numbers, so what they call `1.02` is **major version 1, minor version 2**. Both can go up to 255, so if we were to pad them with leading zeroes, the "correct" way to display them would be `1.002`. This looks awkward and would also not be in line with what some manufacturers (including Zooz) use. 

In fact, we use [semver](https://semver.org/) to compare versions, except we omit the patch part.
> A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes. X is the major version, Y is the minor version, and Z is the patch version. Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.

So Z-Wave JS displays the version numbers as-is, with a `.` in the middle, which is the least ambiguous IMO, especially if you're used to software development.

AlCalzone: So to reiterate: the documented version is `1.10` and what they sent you is `1.20` as I suspected.

diegolrz: Got it. Thanks for the thorough explanation.

Do I need to change the firmware to 1.10 or is 1.1 good? Let me know if anything else is missing.

AlCalzone:
I think we're good here. I'll confirm the firmware version ranges again with Zooz, since I'm not 100% certain there.

In any case, I'll take it from here - thanks!

AlCalzone (approved):

PR #4838, opened 2022-07-23
fix(config): fix manufacturer and device labels for Heatit devices

geirra:
Cleanup of manufacturer 0x019b naming. 
Noticed this in the zwavejs device db.
heatit_zwavejs

Checked with the company and the correct name is Heatit Control AB. Thermofloor has been used shortly in the past it seems.

Have discussed the issue with @blhoward2, was suggested to make a commit

### DIFF
--- a/packages/config/config/devices/0x019b/4512688.json
+++ b/packages/config/config/devices/0x019b/4512688.json
@@ -1,5 +1,5 @@
 {
-	"manufacturer": "Heatit",
+	"manufacturer": "Heatit Controls AB",
 	"manufacturerId": "0x019b",
 	"label": "Z-Smoke Battery",
 	"description": "Wireless Smoke Detector",
--- a/packages/config/config/devices/0x019b/4512688.json
+++ b/packages/config/config/devices/0x019b/4512688.json
@@ -1,5 +1,5 @@
 {
-	"manufacturer": "Heatit Controls AB",
+	"manufacturer": "Heatit",
 	"manufacturerId": "0x019b",
 	"label": "Z-Smoke Battery",
 	"description": "Wireless Smoke Detector",
--- a/packages/config/config/devices/0x019b/zm_single_relay_16.json
+++ b/packages/config/config/devices/0x019b/zm_single_relay_16.json
@@ -2,7 +2,7 @@
 	"manufacturer": "Heatit",
 	"manufacturerId": "0x019b",
 	"label": "ZM Single Relay 16A",
-	"description": "16A Single Z-wave relay",
+	"description": "High power relay",
 	"devices": [
 		{
 			"productType": "0x0004",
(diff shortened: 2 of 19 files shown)


blhoward2 (dismissed):

packages/config/config/manufacturers.json
```diff
@@ -402,7 +402,7 @@
 	"0x0198": "Venstar Inc.",
 	"0x0199": "Wireless Maingate AB",
 	"0x019a": "Sensative AB",
-	"0x019b": "ThermoFloor",
+	"0x019b": "Heatit Controls AB",
```
AlCalzone, line 405: The most recent specification package lists "ThermoFloor" (formerly known as Heatit) for this ID.

packages/config/config/devices/0x019b/4512688.json
```diff
@@ -1,5 +1,5 @@
 {
-	"manufacturer": "Heatit",
+	"manufacturer": "Heatit Controls AB",
```
AlCalzone: The way we use this field, its more like the brand a device is being marketed as, not the actual manufacturer.
AFAIK these devices are branded "Heatit", no?

packages/config/config/manufacturers.json
AlCalzone: But if you confirmed, this is probably okay.

AlCalzone (changes requested):

packages/config/config/devices/0x019b/4512688.json
geirra: Yes thats correct, this has been updated

packages/config/config/manufacturers.json
geirra: Yes, this is confirmed a with them about a week ago.

Thermofloor owns Heatit so that´s probally why it´s listed as thermofloor in the offical docs.
And they started with their devices as thermofloor.

packages/config/config/devices/0x019b/zm_single_relay_16.json
```diff
@@ -1,8 +1,8 @@
 {
-	"manufacturer": "HeatIt",
+	"manufacturer": "Heatit",
 	"manufacturerId": "0x019b",
-	"label": "4512671",
-	"description": "Heatit ZM Single Relay 16A",
+	"label": "ZM Single Relay 16A",
+	"description": "16A Single Z-wave relay",
```
AlCalzone:
```suggestion
	"label": "ZM Single Relay 16A",
	"description": "High power relay",
```
They put this on the box, which is better than repeating the label IMO.

AlCalzone (changes requested):

AlCalzone (approved):
Ok with my one change

AlCalzone:
Thanks!

PR #4842, opened 2022-07-25
fix: more typo fixes

luzpaz:
Follow-up to 4df600f6d

Signed-off-by: luz paz <luzpaz@github.com>

### DIFF
--- a/packages/config/maintenance/importConfig.ts
+++ b/packages/config/maintenance/importConfig.ts
@@ -590,7 +590,7 @@ async function parseOZWProduct(
 		manufacturer,
 		manufacturerId: manufacturerIdHex,
 		label: productLabel,
-		description: existingDevice?.description ?? productName, // don't override the descrition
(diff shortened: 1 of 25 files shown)


packages/config/maintenance/importConfig.ts
```diff
@@ -1060,7 +1060,7 @@ function combineDeviceFiles(json: Record<string, any>[]) {
 					) == false
 				) {
 					console.log(
-						`WARNING - Detected possible firmware parameter change ${file.Identifer} -- ${file.Id} and ${test_file.Id}`,
+						`WARNING - Detected possible firmware parameter change ${file.Identifier} -- ${file.Id} and ${test_file.Id}`,
```
AlCalzone, line 1063: @blhoward2 this is a typo, right? Or is this wrong in the ZWA db files?

AlCalzone: looks like it is a typo

AlCalzone (approved):

PR #4885, opened 2022-08-05
feat: allow passing custom serial port implementation in `port` param of Driver class

sdrsdr:
Having the possibility to pass `tcp://...` as port is very handy but limiting. Allowing directly passing ZWaveSerialPortImplementation in port allows to have TLS sockets and other useful use cases. This is very minimal change with huge usability impact (at least for me)


sdrsdr:
Wow that was fast :) I was just working my way around prettier

AlCalzone (dismissed):
I like it! Fixed a few nits I found.

Any chance you could upstream the TLS socket implementation? At least one other user has asked for it so far.

sdrsdr:
My case have more that TLS in it but I guess I can show a TLS example somewhere?

sdrsdr:
An extract from my use case where a device connects to my tls server and tunnels the serial port:
 
```
import {Server, TLSSocket} from "tls";
import { Driver} from "zwave-js";

const server=new Server({
  handshakeTimeout:4000,
  cert:fs.readFileSync('/var/crt/socket.crt',{encoding:'utf8'}),
  key:fs.readFileSync('/var/crt/socket.key',{encoding:'utf8'}),
});

server.on('secureConnection',(tlsSocket) => {
  //tlsSocket.on('end',()=>{......});
  //tlsSocket.on('error',(err)=>{....});
  //tlsSocket.on('data',(data)=>{....});
  let localid=12345345; //or whatever
  const driver=new Driver({
    create(){
      console.log("driver called create for "+localid);
      return tlsSocket;
    },
    async open(port) {
      console.log("driver requested port opening for "+localid);
    },
    async close(port){
      console.log("driver requested port closing for "+localid);
      //close tlsSocket maybe
    }
  },{
    securityKeys:findKeys(localid)
  });

}

server.listen({port:5566},()=>{
  const addr=<AddressInfo>server.address();
  console.log("TLS server listening on addr "+addr.address+":"+addr.port);
})

```

But it's all the same after obtaining the connected tls socket

AlCalzone (approved):

AlCalzone:
Thanks! At some point I'll look into constructing a connection string that lets us connect to a tls socket.

PR #4887, opened 2022-08-05
fix(serial): handle `close` event while opening serial port

AlCalzone:
It was reported that the serialport open call sometimes blocks. This is an attempt at fixing it.


AlCalzone:
@zwave-js-bot pack this

PR #4892, opened 2022-08-08
test: make it possible to skip integration tests

AlCalzone:
This PR adds `.skip(...)` and `.only(...)` modifiers to the `integrationTest` method, just like `jest.describe` already has.


AlCalzone:
@zwave-js-bot approve

PR #4894, opened 2022-08-08
feat(config): add MP31ZP (rebranded MP21ZP)

AlCalzone:
Looks like there was either a mixup in the firmware and it is identical to the MP21ZP, or the manual is wrong. Essentially, this device behaves just like the MP21ZP


AlCalzone:
@zwave-js-bot approve

PR #4895, opened 2022-08-08
fix(config): param 9 for STEINEL devices, rework to templates

nechry:
PR description here
fix(config): param 9 for STEINEL devices

### DIFF
--- a/packages/config/config/devices/0x0271/motionswitch_led.json
+++ b/packages/config/config/devices/0x0271/motionswitch_led.json
@@ -77,7 +77,7 @@
 		{
 			"#": "9[0x01]",
 			"label": "Gateway Check",
-			"description": "Lamp signalizes failure to reach the gateway.",
+			"description": "",
 			"valueSize": 1,
 			"defaultValue": 1,
 			"allowManualEntry": false,
@@ -95,7 +95,7 @@
 		{
 			"#": "9[0x02]",
 			"label": "Permanently On",
-			"description": "When enabled, overrides the Gateway Check parameter",
+			"description": "",
 			"valueSize": 1,
 			"defaultValue": 0,
 			"allowManualEntry": false,
--- a/packages/config/config/devices/0x0271/motionswitch_led.json
+++ b/packages/config/config/devices/0x0271/motionswitch_led.json
@@ -35,16 +35,8 @@
 			"$import": "templates/steinel_template.json#external_light_sensor"
 		},
 		{
-			"#": "9[0x01]",
-			"$import": "templates/steinel_template.json#lamp_external_control"
-		},
-		{
-			"#": "9[0x02]",
-			"$import": "templates/steinel_template.json#lamp_gateway_check"
-		},
-		{
-			"#": "9[0x04]",
-			"$import": "templates/steinel_template.json#lamp_permanently_on"
+			"#": "9[0x07]",
+			"$import": "templates/steinel_template.json#light_behavior"
 		},
 		{
 			"#": "10",
--- a/packages/config/config/devices/0x0271/motionswitch_led.json
+++ b/packages/config/config/devices/0x0271/motionswitch_led.json
@@ -16,186 +16,150 @@
 	"paramInformation": [
 		{
 			"#": "1",
-			"label": "TIME",
-			"description": "Duration of light after motion detection [s]",
-			"unit": "s",
-			"valueSize": 2,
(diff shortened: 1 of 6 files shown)


nechry:
@zwave-js-bot fix lint

blhoward2 (changes requested):
Thanks for helping out! Your changes are great so far, but we generally try to avoid descriptions where we can. Could you please take another look and remove them if unneeded? If necessary, sometimes modifying the label (while keeping them short) makes them redundant. If something must truly be described that's fine but they clutter up the UIs.

nechry:
Hello, I simply used the labels and descriptions already present for the xled_home_2 and applied to the other models as they share the same configuration base.
I will look to simplify them or as you recommended to remove them.

I fully understand your point, I maintained for more than 8 years the configuration files for the openzwave library

AlCalzone:
Hey @nechry nice to see you at the other side of Z-Wave land.

We like to take the chance to clean up files (many of which were auto-imported from various sources at some point) when somebody touches them.

nechry:
@AlCalzone  Well noted, I think I'm getting more and more involved here.

nechry:
@blhoward2 done, always happy to contribute

packages/config/config/devices/0x0271/motionswitch_led.json
```diff
@@ -75,13 +75,40 @@
 			"defaultValue": 1
 		},
 		{
-			"#": "9",
-			"label": "SLAVE_MODE",
-			"description": "Disable local control",
+			"#": "9[0x01]",
+			"label": "Gateway Check",
```
AlCalzone: @nechry Do you have a manual which shows that this one has no "External Control" mode? I only found [this](http://manuals-backend.z-wave.info/make.php?lang=en&sku=40111665101&cert=ZC10-18076193) (near the bottom) and it looks like param 9 is the same as for the other devices.

AlCalzone (commented):
I'm going to take over and template this and the other parameters too. Just one question...

packages/config/config/devices/0x0271/motionswitch_led.json
nechry: I found this: [one](https://manuals-backend.z-wave.info/make.php?lang=en&sku=STEEMSWITCH&cert=ZC10-18076193) This is the only model I don't own.
your document seems more recent to me, so I prefer to play safe and add the bitset.

AlCalzone: Thanks, I'll go with the bitset then. Even on the one you linked it looks a bit like it is supposed to be a bitset but they entered the values weirdly.

nechry: in fact like bit 2 (LED is permanently on) override bit 0 and 1 it is true that we could simplify with a list. Again I relied on the initial config for the refactoring.

nechry: @AlCalzone do you prefer I convert to list with selection:

```json
{
  "#": "9",
  "label": "Slave mode",
  "description": "",
  "valueSize": 1,
  "minValue": 0,
  "maxValue": 4,
  "defaultValue": 2,
  "unsigned": true,
  "allowManualEntry": false,
  "options": [
  {
	  "label": "Normal mode",
	  "value": 0
  },
  {
	  "label": "Slave mode",
	  "value": 1
  },
  {
	  "label": "Normal mode with checking lifeline",
	  "value": 2
  },
  {
	  "label": "Slave mode with checking lifeline",
	  "value": 3
  },
  {
	  "label": "LED is permanently on",
	  "value": 4
  }
  ]
}
```

AlCalzone: I reworked most of the files to templates, let me just copy that. It's probably easier to understand than the bitset.

nechry: great, yes at the end a list even if behind it's bit sets it's even more user-friendly

AlCalzone (approved):

blhoward2:
> @AlCalzone  Well noted, I think I'm getting more and more involved here.

As Al said, welcome to the other side! We can always use more help with the device files. If you're going to contribute more, in the docs there is a style guide for device files and sections on templating parameters and conditional parameters. That should take out some of the mystery of where our requested changes frequently come from, to avoid less back and forth.

PR #4907, opened 2022-08-10
fix: move "notification" event args types back into `zwave-js` package

AlCalzone:
They weren't used anywhere else and this makes more sense for the event handler definitions


AlCalzone:
@zwave-js-bot approve

PR #4915, opened 2022-08-15
fix(config): add parameter 15 Invert Output to Heatit Z-Temp2

kerb101:
As described in user manual: https://media.heatit.com/1548

PR description here


AlCalzone (approved):
Found a couple of other things I did not like about this config.

Issue #4914, opened 2022-08-14
Make ZWaveNode class also emit it's events via  this.driver.emit


sdrsdr:
**Is your feature request related to a problem? Please describe.**
This library is mostly used for base of home automation systems where every node must be monitored. But to get a node monitored the code must wait for the node to become sufficiently available and then subscribe to it's events. Plus there are multiple ways a node get available: after system boot, after add  

**Describe the solution you'd like**
What I really would like is just to subscribe for some driver events where all node events get routed. This can easily be achieved with few lines of code in `Node.ts`: All value events are "captured" from the value DB and touted to   translateValueEvent method were a single line  `this.driver.emit("node "+event,this, outArg as any)` can be added to the bottom so you can now subscribe on the driver object with `on.("node value updated",...)` and get the `value updated` events  from all the nodes in the WZ network governed by the driver.
Similarly in the `onStatusChange` method of  ZWaveNode where a event is this.emit-ted  a complementary this.driver.emit can be added again prefixing the event name with `"node "` so you can get `node alive`, `node dead`, `node sleep` and `node wake up` events from the driver itself and omit subscribing for these events on every node that  goes in or out of the nework

**Describe alternatives you've considered**
This is pure convenience change not effecting any existing code. The performance impact will be minimal if any. The alternatives are somewhat complex first when the driver is ready you have to enumerate all nodes and attach your event handler then when a node is added you have to attach to that node again. There might be other steps needed that I'm not aware of just to get all node events reported to the code using the library.

**Additional context**
I can work on a PR if this feature get approved.

AlCalzone:
fixed in https://github.com/zwave-js/node-zwave-js/pull/6002

PR #4919, opened 2022-08-15
feat: make ZWaveNode class also emit it's events via this.driver.emit

sdrsdr:
As described in  #4914 it will be very hand to be able to subscribe for all network events via the driver event emitter, and this PR enables this. It's just a suggestion, I don't fully understand the typing of the event emitters but my typescript is happy with what I've done. Any suggestions are highly welcomed :)


AlCalzone:
Hey, just wanted to let you know - I haven't forgotten this, but I need to get a couple of things stabilized before QoL things like this can land.

sdrsdr:
No rush 😃 I'm using the package from my repo thanks to yarn. I'm glad that you even consider this feature

AlCalzone:
Sorry, totally lost track of this while focusing on getting the driver ready for certification.
Will take a look on monday.

sdrsdr:
Ah don't worry, I might endup maintaining a fork for our own needs so you can totally drop ths if you don't like it

packages/serial/src/ZWaveSerialPortBase.ts
```diff
@@ -76,8 +76,15 @@ export function isZWaveSerialPortImplementation(
 	);
 }
 
+type write_cb=(err:Error|null|undefined)=>void
+export interface DuplexAlike {
+	// (chunk: any, encoding?: BufferEncoding | undefined, cb?: ((error: Error | null | undefined) => void) | undefined): boolean; (chunk: any, cb?: ((error: Error | null | undefined) => void) | undefined): boolean;
+	write(chunk:Buffer,cb?:write_cb|undefined):boolean;
+} ;
```
AlCalzone, line 83: @sdrsdr can you explain why this is needed? What's your use case?

In any case this feels out of place in this PR, I'd do this separately.

AlCalzone (changes requested):
I'll rework how you forward the events as I'm not quite happy with how it is done. I'll make sure to include your initial commit, so it gets attributed to you.

packages/serial/src/ZWaveSerialPortBase.ts
sdrsdr: I'm not sure how this got here :( It's part of me experimenting with passing non full duplex complaint interface there. You see we're experimenting with WebSockets  an the ws library does not provide a duplex compatible interface so I was trying to use TS to filter down the actually used methods. Did not get me too far as it's casted multiple times to different interfaces. You can totally omit it. Sorry for the inconvenience

AlCalzone: ok, I'm happy to loosen the definitions here if it makes sense though. Just need to understand the use-case.

PR #4925, opened 2022-08-16
fix: looking up provisioning entry by node ID

AlCalzone:
Looks like this never worked unless the provisioning entry already included the node ID. This would only happen if an application edited a provisioning entry. Now we look up the DSK from the node if given a node ID that has no corresponding entry.


AlCalzone:
@zwave-js-bot pack this

PR #4926, opened 2022-08-16
feat(config): import Zooz ZEN14 from zwa templated

alvinchen1:
Don't have a ZEN14, but based the two outlets on the ZEN25 which is labeled Left and Right Outlet.  Left ZEN14 as Outlet 1 and Outlet 2

PR description here


alvinchen1:
@zwave-js-bot fix lint

AlCalzone:
@alvinchen1 your fork seems terribly out of sync. This causes the tests to fail and it is not easily fixed.
Unless you want to do some `git` magic, I suggest deleting and re-creating the fork, then repeat the steps the bot posted. Make sure to save a copy of the zen14 file first.

When you fork, don't check this box, so you get a copy of the bot's branch:
grafik

Also for the future I recommend only working in separate branches in your fork when you plan to contribute something. This way you can sync it to this repo more easily before you work on things:
grafik

PR #4945, opened 2022-08-23
feat: use Supervision for the remaining SET- and REPORT-type commands

### LINKED ISSUE #3517: Use Supervision for more SET-type commands if possible
See https://github.com/zwave-js/node-zwave-js/issues/3516#issuecomment-941617615 - we could greatly reduce the necessary traffic here.

AlCalzone:
Investigations have shown that S2-encrypted communication only works reliably if we expect some kind of feedback from the node. With this feedback we can correlate Nonce Reports and re-transmit commands that were not understood.

Depending on the verification strategy, this increases or decreases the command count for a communication attempt:
- no verification GET: +/- 0 (if device reports unsolicited) to +1 command (if no unsolicited reports)
- verification GET (now replaced with Supervision): +/- 0 (if device reports unsolicited) to -1 command (if no unsolicited reports)

also fixes: https://github.com/zwave-js/node-zwave-js/issues/3517

### DIFF
--- a/packages/cc/src/cc/MultiChannelAssociationCC.ts
+++ b/packages/cc/src/cc/MultiChannelAssociationCC.ts
@@ -1,4 +1,8 @@
-import type { IZWaveEndpoint, MessageRecord } from "@zwave-js/core/safe";
+import type {
+	IZWaveEndpoint,
+	MessageRecord,
+	SupervisionResult,
+} from "@zwave-js/core/safe";
 import {
 	CommandClasses,
 	encodeBitMask,
@@ -29,6 +33,7 @@ import {
 	commandClass,
 	expectedCCResponse,
 	implementedVersion,
+	useSupervision,
 } from "../lib/CommandClassDecorators";
 import * as ccUtils from "../lib/utils";
 import { V } from "../lib/Values";
@@ -239,7 +244,7 @@ export class MultiChannelAssociationCCAPI extends PhysicalCCAPI {
 	@validateArgs()
 	public async addDestinations(
 		options: MultiChannelAssociationCCSetOptions,
-	): Promise<void> {
+	): Promise<SupervisionResult | undefined> {
 		this.assertSupportsCommand(
 			MultiChannelAssociationCommand,
 			MultiChannelAssociationCommand.Set,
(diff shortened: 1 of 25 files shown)


packages/cc/src/cc/MultiChannelAssociationCC.ts
```diff
@@ -294,7 +300,7 @@ export class MultiChannelAssociationCCAPI extends PhysicalCCAPI {
 				endpoint: this.endpoint.index,
 				...options,
 			});
-			await this.applHost.sendCommand(cc, this.commandOptions);
+			return this.applHost.sendCommand(cc, this.commandOptions);
```
OXINARF, line 303: Shouldn't MultiChannelAssociationCCRemove have a useSupervision?

packages/cc/src/cc/UserCodeCC.ts
```diff
@@ -283,6 +285,7 @@ export class UserCodeCCAPI extends PhysicalCCAPI {
 		{ property, propertyKey },
 		value,
 	) => {
+		let result: SupervisionResult | undefined;
```
OXINARF, line 288: Aren't all below missing the useSupervision?

packages/cc/src/cc/WakeUpCC.ts
```diff
@@ -166,7 +171,7 @@ export class WakeUpCCAPI extends CCAPI {
 	public async setInterval(
 		wakeUpInterval: number,
 		controllerNodeId: number,
-	): Promise<void> {
+	): Promise<SupervisionResult | undefined> {
 		this.assertSupportsCommand(WakeUpCommand, WakeUpCommand.IntervalSet);
 
 		const cc = new WakeUpCCIntervalSet(this.applHost, {
```
OXINARF, line 177: Isn't WakeUpCCIntervalSet missing the useSupervision?

OXINARF (commented):
@AlCalzone I know this is already merged, but I had the pending comments below - which seem like bugs to me, so I didn't want to just ignore them, sorry if it is just noise.

packages/cc/src/cc/WakeUpCC.ts
AlCalzone: Thanks, those were oversights: https://github.com/zwave-js/node-zwave-js/pull/4956

PR #4952, opened 2022-08-27
fix: missing/incorrect dependencies

AlCalzone:
Without this, the wrong versions can be loaded on systems with other packages using older versions of these dependencies.


AlCalzone:
@zwave-js-bot approve

PR #4969, opened 2022-08-30
fix: randomize Supervision session ID on startup

AlCalzone:
Otherwise a driver restart may result in devices ignoring the request


AlCalzone:
@zwave-js-bot approve

AlCalzone:
@zwave-js-bot approve

PR #4971, opened 2022-08-30
fix: use highest CC version implemented by the driver when node's CC version is unknown

AlCalzone:
not sure why this ever worked, but this should help with https://github.com/home-assistant/core/issues/77466


AlCalzone:
@zwave-js-bot approve

PR #4978, opened 2022-08-31
fix: attempt endpoint communication with S0, even if S0 is not listed in the endpoint capabilities

### LINKED ISSUE #4977: If node is secure, at least attempt secure (S0) communication with endpoints
https://github.com/home-assistant/core/issues/77466#issuecomment-1232120342

> Any device that supports the Security and Multi Channel Command Classes MAY choose to support a different set of Command Classes securely for each Multi Channel End Point. An End Point with support for Security MUST report the Security Command Class as supported for that End Point.

...

> It has been found that legacy nodes do not always advertise the S0 Command Class in their Multi
Channel Capability Report and still accept all their Command Class using S0 encapsulation.
A controlling node SHOULD try to control End Points with S0 encapsulation even if S0 is not
listed in the Multi Channel Capability Report.

> The command classes supported for each endpoint securely is determined by using the Security Commands Supported Get command sent to each individual endpoint Security Encapsulated. Hence, the encapsulation order is: Security Encapsulation - Multi Channel Encapsulation - Security Commands Supported Get Command

AlCalzone:
Something in v10 changed so the endpoint security capabilities are now correctly considered when communication with endpoints. However this broke some legacy multichannel devices which do not list S0 in the endpoint capabilities and which do not respond to the secure CC query.

We now test whether an endpoint responds to secure queries and mark all endpoints CCs as secure if it does.

fixes: #4977

### DIFF
--- a/packages/zwave-js/src/lib/node/Node.ts
+++ b/packages/zwave-js/src/lib/node/Node.ts
@@ -1726,6 +1726,12 @@ protocol version:      ${this.protocolVersion}`;
 				securityClass == undefined ||
 				securityClassIsS2(securityClass)
 			) {
+				this.driver.controllerLog.logNode(
+					this.nodeId,
+					"Root device interview: Security S2",
+					"silly",
+				);
+
 				if (!this.driver.securityManager2) {
 					if (!this._hasEmittedNoS2NetworkKeyError) {
 						// Cannot interview a secure device securely without a network key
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot pack this

packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -1930,17 +1987,99 @@ protocol version:      ${this.protocolVersion}`;
...
+						}
+					} else {
+						this.driver.controllerLog.logNode(this.nodeId, {
+							endpoint: endpoint.index,
+							message: `Endpoint ${endpoint.index} is actually not using S0`,
+							level: "silly",
+						});
+						// Mark the CC as not secure again
+						endpoint.addCC(ccId, { secure: false });
+					}
```
OXINARF, line 2068: From the code above, it looks like you wanted to break here (and not continue looping for other tests).

AlCalzone: https://github.com/zwave-js/node-zwave-js/pull/4993
thanks :)

PR #5001, opened 2022-09-02
feat(config): add MyOT OpenTherm Actuator v4

czesiu20:
Configuration file for new device MyOT v4 (OpenTherm Actuator)


packages/config/config/devices/0x044c/myot4.json
```diff
@@ -0,0 +1,112 @@
...
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"paramInformation": [
+		{
+			"#": "31",
+			"$import": "~/templates/master_template.json#base_enable_disable",
+			"label": "Summer Mode",
+			"description": "Turn heating on/off"
```
AlCalzone, line 21: This feels redundant. Maybe just make the label `"Heating"` and remove the description?
Need to take care though that the values for Enable and Disable are not inverted then.

packages/config/config/devices/0x044c/myot4.json
```diff
@@ -0,0 +1,112 @@
...
+			"valueSize": 2,
+			"readOnly": true
+		},
+		{
+			"#": "95",
+			"$import": "~/templates/master_template.json#base_0-255_nounit",
+			"label": "Data report frequency",
+			"description": "Amount of minutes between sensors reports (0: continuous reporting)",
+			"defaultValue": 0,
+			"unit": "min"
```
AlCalzone:
```suggestion
			"label": "Sensor Report Interval",
			"unit": "minutes",
			"defaultValue": 0,
			"options": [
				{
					"label": "Continuous",
					"value": 0
				}
			]
```

packages/config/config/devices/0x044c/myot4.json
```diff
@@ -0,0 +1,112 @@
...
+		},
+		{
+			"#": "96",
+			"$import": "~/templates/master_template.json#base_0-255_nounit",
+			"label": "Configuration report frequency",
+			"description": "Amount of minutes between configuration parameters reports (0: continuous reporting)",
+			"defaultValue": 0,
+			"unit": "min"
```
AlCalzone:
```suggestion
			"label": "Configuration Report Interval",
			"unit": "minutes",
			"defaultValue": 0,
			"options": [
				{
					"label": "Continuous",
					"value": 0
				}
			]
```

AlCalzone (changes requested):
Thanks - needs a few changes to match the style guide though.

AlCalzone:
Sorry for taking so long - the v10 release kept me quite busy.

packages/config/config/devices/0x044c/myot4.json
czesiu20: Yes, it can be confusing, to turn off the heating you must turn it on. However, I prefer to keep the original names from the documentation. Honestly, it is really annoying when I can't find the parameters of well known device because of changed names. I have to find them by numbers. Unfortunately, renaming parameters is common practice in this project. This change makes myot.json and myot2.json inconsistent. The decision is yours.

AlCalzone: I guess consistency with the other similar devices is more important.

AlCalzone: As for the renamed labels: Some are really f-ing terrible. We prefer them to be simple and self-explanatory. Needing a manual to understand what a parameter does, isn't nice, especially when manuals get out of date due to firmware updates etc.
I understand that it can be sometimes hard to match them to the manual this way, but that is something we're willing to accept.

AlCalzone (approved):

PR #5031, opened 2022-09-09
feat: pass user agent to firmware update service

### LINKED ISSUE #5002: Allow passing user agent to retrieve available firmware updates
When fetching firmware updates, it would be nice if we can include a user agent who is requesting this data. In the case of Home Assistant I would expect it to look something like `HomeAssistant/2022.9.3 ZJS-Server/1.22.1 ZJS/10.0`

When something goes wrong with the API, it will be easier to track down where the bug is.

Pass the user agent string.

AlCalzone:
With this PR, we pass a user agent containing the driver version to the firmware update service.

For this reason, a few new options to configure the user-agent are added:
1. An optional driver option `userAgent`. This will become mandatory in future releases.
2. The method `updateUserAgent`.

Both accept a `Record<string, string>` where the key is the application/component/module name and the value is the version. In `updateUserAgent` the version can be explicitly set to `null/undefined` to remove a component.

If no user agent is configured, but the application has enabled statistics, the application name and version will be taken from there.

fixes: #5002 
/cc @balloob

### DIFF
--- a/packages/zwave-js/src/lib/controller/FirmwareUpdateService.ts
+++ b/packages/zwave-js/src/lib/controller/FirmwareUpdateService.ts
@@ -1,5 +1,6 @@
 import got, { Headers, OptionsOfTextResponseBody } from "@esm2cjs/got";
 import PQueue from "@esm2cjs/p-queue";
+import type { DeviceID } from "@zwave-js/config";
 import {
 	extractFirmware,
 	Firmware,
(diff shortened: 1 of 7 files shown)


packages/zwave-js/src/lib/controller/FirmwareUpdateService.ts
```diff
@@ -124,7 +124,9 @@ export function getAvailableFirmwareUpdates(
 	productId: number,
 	firmwareVersion: string,
 	apiKey?: string,
+	userAgent?: string,
```
balloob: @martinhjelmare @raman325

Let's add this to Home Assistant so we can at the firmware service the requests from HA

AlCalzone: I'll add a way to globally configure the user agent on the driver instance later.

AlCalzone:
PR updated. User agent can now be passed as a driver option and updated later. It also matches the usual format `app1/ver1 app2/ver2 ...` now.

PR #5036, opened 2022-09-09
fix: ensure value ID is valid and normalized before passing through

### LINKED ISSUE #5021: Normalize/clean up value IDs passed to the driver before using them

AlCalzone:
fixes: #5021

Most of the value DB already made sure of this, except for value IDs that get passed through in events. By validating and normalizing value IDs in `setValue` and `pollValue`, value IDs passed back to applications are guaranteed to be normalized.


AlCalzone:
@zwave-js-bot approve

PR #5043, opened 2022-09-10
feat(config): add Zooz ZEN04

mikebaz:
This change adds a configuration file for the ZEN04 smart plugs.  Values are based on the ZEN05 file and the info at https://www.support.getzooz.com/kb/article/1114-zen04-smart-plug-advanced-settings/.

### DIFF
--- a/packages/config/config/devices/0x027a/zen04.json
+++ b/packages/config/config/devices/0x027a/zen04.json
@@ -22,19 +22,19 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Follow Outlet",
+					"label": "Follow outlet",
 					"value": 0
 				},
 				{
-					"label": "Opposite of Outlet",
+					"label": "Opposite of outlet",
 					"value": 1
 				},
 				{
-					"label": "Always Off",
+					"label": "Always off",
 					"value": 2
 				},
 				{
-					"label": "Always On",
+					"label": "Always on",
 					"value": 3
 				}
 			]
@@ -61,11 +61,11 @@
 					"value": 0
 				},
 				{
-					"label": "Always Off",
+					"label": "Always off",
 					"value": 1
 				},
 				{
-					"label": "Always On",
+					"label": "Always on",
 					"value": 2
 				}
 			]
@@ -73,6 +73,7 @@
 		{
 			"#": "5",
 			"label": "Power Wattage Reporting Threshold",
+			"unit": "W",
 			"valueSize": 1,
 			"minValue": 5,
 			"maxValue": 50,
@@ -82,6 +83,7 @@
 		{
 			"#": "6",
 			"label": "Power Wattage Reporting Frequency",
+			"unit": "minutes",
 			"valueSize": 4,
 			"minValue": 1,
 			"maxValue": 65535,
@@ -91,6 +93,7 @@
 		{
 			"#": "7",
 			"label": "Electrical Current Reporting Threshold",
+			"unit": "0.1 A",
 			"valueSize": 1,
 			"minValue": 1,
 			"maxValue": 10,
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x027a/zen04.json
```diff
@@ -0,0 +1,114 @@
...
+				{
+					"label": "Always off",
+					"value": 1
+				},
+				{
+					"label": "Always on",
+					"value": 2
+				}
+			]
+		},
```
AlCalzone: This should use a template like the ZEN05 does here:
https://github.com/zwave-js/node-zwave-js/blob/81d6306c505b8af4d51ba8a10e645c6a8dd0f7a7/packages/config/config/devices/0x027a/zen05.json#L48-L51
except that you need the `..._prev_off_on` variant.

packages/config/config/devices/0x027a/zen04.json
```diff
@@ -0,0 +1,114 @@
...
+				},
+				{
+					"label": "Always off",
+					"value": 2
+				},
+				{
+					"label": "Always on",
+					"value": 3
+				}
+			]
```
AlCalzone: Please change to this for consistency:
```suggestion
		"label": "LED Indicator",
		"valueSize": 1,
		"defaultValue": 0,
		"allowManualEntry": false,
		"options": [
			{
				"label": "On when load is on",
				"value": 0
			},
			{
				"label": "On when load is off",
				"value": 1
			},
			{
				"label": "Always off",
				"value": 2
			},
			{
				"label": "Always on",
				"value": 3
			}
		]
```

packages/config/config/devices/0x027a/zen04.json
```diff
@@ -0,0 +1,114 @@
...
+		{
+			"#": "2",
+			"$import": "templates/zooz_template.json#auto_off_timer_duration",
+			"label": "Auto Turn-Off Timer Duration"
+		},
+		{
+			"#": "3",
+			"$import": "templates/zooz_template.json#auto_on_timer_duration",
+			"label": "Auto Turn-On Timer Duration"
+		},
```
AlCalzone: These don't need the label specified again, it comes from the template

AlCalzone (changes requested):
Thanks, just a few remarks

packages/config/config/devices/0x027a/zen04.json
AlCalzone: (looks like I messed up the indentation, but the bot will correct that)

mikebaz:
ok, I believe I correctly implemented the requested changes.  Sorry for not getting it right the first time - I know that wastes time on your end. :(

AlCalzone:
All good - this was one of the better PRs ;)
Our standards are pretty high, so it's not easy to fulfill them.

AlCalzone (approved):

gminco:
Could we be missing one extra parameter?
I see 8 on the Zooz site but after including one I'm counting 9:
image

gminco:
We're missing parameter #9, I've tried but I cannot edit the file, all it needs is this to be added:

`,
		{
			"#": "9",
			"$import": "templates/zooz_template.json#led_indicator_brightness"
		}`

AlCalzone:
@gminco I cannot find parameter 9 in [Zooz's documentation](https://www.support.getzooz.com/kb/article/1114-zen04-smart-plug-advanced-settings/). If it is actually missing, please open a new issue and share the manual or whatever states that it does exist.

gminco:
@AlCalzone it's on the documentation link you provided, but it's between parameter 1 and 2.

mikebaz:
@gminco ah you're right, I missed that.  as mentioned the update requires a new PR.  I'll get one in shortly.

PR #5054, opened 2022-09-13
fix: negative globs aren't supported by turborepo yet

AlCalzone:
except through git pathspec syntax


AlCalzone:
@zwave-js-bot approve

PR #5070, opened 2022-09-14
feat: accept more user agent components in getAvailableFirmwareUpdates

raman325:
As discussed in https://github.com/zwave-js/zwave-js-server/issues/740 , this allows the caller to add additional user agent components to calls that use the userAgent. This handles the scenario where the client of `zwave-js-server` needs to add a component to the user agent string


AlCalzone (dismissed):
Took the liberty to rewrite this a bit. I didn't like how one method did two things at once, but then only half of that was actually used sometimes.

AlCalzone (approved):

AlCalzone (dismissed):

AlCalzone (approved):

Issue #4786, opened 2022-07-08
Update Zooz device files


blhoward2:
To account for new devices, per Zooz:

Below are the new settings which have been added to the following dimmer versions so far:

ZEN72 10.10 (ver. 1.0 hardware) and 2.0 (ver. 2.0 hardware)
ZEN74 10.10 (ver. 1.0 hardware, no 2.0 hardware available at the moment)
ZEN77 10.20 (ver. 1.0 hardware), 2.10 (ver. 2.0 hardware), and 3.0 (ver. 3.0 hardware)

Please let me know if you have any questions here!

Remote Z-Wave Dimming Duration (Multilevel Commands For Group 3 And 4)

Parameter 30: Set the time it takes to get from 0% to 100% brightness on dimmers and smart bulbs directly associated with your dimmer in Groups 3 and 4 when pressing and holding the paddle (physical dimming) on your dimmer. The number entered as value corresponds to the number of seconds.
Values: 1 – 99 (seconds). Default: 5.
Size: 1 byte dec

Ramp Rate Control

Parameter 9: Adjust the ramp rate ON for your dimmer when the top paddle is pressed (physical ramp rate on). Values correspond to the number of seconds it take for the dimmer to reach full brightness when operated manually.
Values: 1 – 99 (seconds). 0 – instant on. Default: 0.
Size: 1 byte dec
 NOTE:  On all other previous firmware versions including 1.0 and 10.0, this setting was used for all ramp rate control (to adjust behavior for on/off physical and Z-Wave ramp rate). The default value for this parameter used to be 1.
 
Parameter 27: Adjust the ramp rate OFF for your dimmer when the bottom paddle is pressed (physical ramp rate off). Values correspond to the number of seconds it take for the dimmer to turn off when operated manually.
Values: 1 – 99 (seconds). 0 – instant off. Default: 2.
Size: 1 byte dec
 
Parameter 28: Adjust the ramp rate ON for your dimmer when a Z-Wave command is sent to turn the dimmer on (Z-Wave ramp rate on). Values correspond to the number of seconds it take for the dimmer to reach full brightness when operated via Z-Wave.
Values: 1 – 99 (seconds). 0 – instant on. 255 – match physical. Default: 255.
Size: 1 byte dec
 
[...]

PR #5080, opened 2022-09-18
fix: update zen72.json with more parameters

sandymac:
Correct Parameter 9 for the ZEN72 and add 27, 29, and 29 parameters Parameter 30 was not added despite being part of 
https://www.support.getzooz.com/kb/article/538-zen72-dimmer-switch-700-advanced-settings/

This should expand the device config for the Zooz ZEN72 based on the ZEN72 Dimmer Switch 700 Advanced Settings described in https://www.support.getzooz.com/kb/article/538-zen72-dimmer-switch-700-advanced-settings/

I did not add Parameter 30 "Remote Z-Wave Dimming Duration (Multilevel Commands For Group 3 And 4)" as I didn't find existing template references for that feature.

### DIFF
--- a/packages/config/config/devices/0x027a/zen72.json
+++ b/packages/config/config/devices/0x027a/zen72.json
@@ -60,7 +60,19 @@
 		},
 		{
 			"#": "9",
-			"$import": "templates/zooz_template.json#dimmer_ramp_rate"
+			"$import": "templates/zooz_template.json#dimmer_on_ramp_rate"
+		},
+		{
+			"#": "27",
+			"$import": "templates/zooz_template.json#dimmer_off_ramp_rate"
+		},
+		{
+			"#": "28",
+			"$import": "templates/zooz_template.json#zwave_on_dimmer_ramp_rate"
+		},
+		{
+			"#": "29",
+			"$import": "templates/zooz_template.json#zwave_off_dimmer_ramp_rate"
 		},
 		{
 			"#": "10",


AlCalzone:
> I didn't find existing template references for that feature.

You can just add the whole definition in that case. We'll convert to templates if that shows up more often.

kpine:
These parameters seem to be dependent on firmware version?

AlCalzone (changes requested):
What kpine and I said in the other comment

blhoward2:
@sandymac Please make these dependent on the firmware version. You can see an example in #5125.

alvinchen1:
> @sandymac Please make these dependent on the firmware version. You can see an example in #5125.

No longer think this is needed.  #5125 got merged today and does this.  This PR can be closed.

jessiewestlake:
This is still needed for hardware version 2.0 and greater as the firmware version is not reported as 10.x. For hardware version 2.0, the firmware version is reported as 2.x. Therefore, the current configuration will not allow these parameters to work for the updated hardware because the firmware version check is looking for any firmware version greater than 10.10, and therefore gets ignored for the newer firmware on the newer hardware versions.

alvinchen1:
@jessiewestlake, so I agree with you that we need version 2.0, but here's the problem.  The way that Z-wave JS determines what parameters to enable is to take the hardware version (in this case Zooz) for Manufacturer 027a and the Z-Wave Product Type and ID.  For the Zen77, that is 0x7000 0xa007.  You can see this by going to one of your Zen77 devices page in HA, go to Configure and go to Z-Wave JS Device Database.  That will take you to "https://devices.zwave-js.io/?jumpTo=[MANUFACTURER]:[PRODUCT TYPE]:[PRODUCT ID]:0.0"

From there, the code will have the hardware ID x027a:0x7000:0xa007, then it gets the firmware ID.  For this hardware device, if the firmware is greater than 10.20 it will display the new parameters.  

The problem as I mentioned in https://github.com/zwave-js/node-zwave-js/issues/4786 is we have no way of distinguishing between version 1.0 and 2.0 (or 3.0) of the hardware for Zen77 right now.  Zooz apparently decided to use the same hardware number for version 1.0 and 2.0.  The problem then becomes because we are at firmware version 10.20 with hardware version 1, we can't give version 2.0 firmware 2.0 those parameters.  Because if we did say firmware 2.0 gets these parameters, then any hardware version 1 with firmware greater than 2.0 will get the parameters as well and that may cause problems with hardware version 1.

The REAL answer here is that we need a way to distinguish between HARDWARE version 1 and HARDWARE version 2, i.e. hardware version 2 should not report as x027a:0x7000:0xa007.  This will allow us to write a separate JSON file with those parameters.  In the absence of that, there's not much more we can do unless we determine that the benefit to version 2.0 is greater than possible risk or problems of adding parameters to hardware version 1.   OR Zooz could align their firmware versioning across the hardware device id x027a:0x7000:0xa007 regardless of what hardware version it actually is.  If hardware version 2.0 and 3.0 all had the same firmware number as version1.0, then we are good.

AlCalzone:
The Z-Wave specification actually has a field for the hardware revision, we just never needed it before to distinguish devices. And I'm not sure if Zooz changed that one between HW 1 and 2. Can anyone with Z-Wave JS UI check this? The field should be under Version CC.

alvinchen1:
Sweet!  Welcome back @AlCalzone BTW.  Where would we add Version CC to the JSON.  Is there an example JSON where we've done this before for another manufacturer or device?  @AnthonyMaiorani replied to https://github.com/zwave-js/node-zwave-js/issues/4786 and has a V2.  I looked through my ZOOZ Zen15 which are supposed to have multiple hardware versions but didn't see where there was a Version CC spec.  I looked in Debug Info, Manufacturer Specfic v2 Values and Version v2 values.  I didn't see Version CC, but I did see in Version v2 [15-134-0-firmwareVersions] which I think is right...
image

Another thing I thought of we could try is instead of using firmware > x could we use firmware = x?  This would only display the parameters if the firmware matched exactly.  I can see this leading to overly bloated json files as we have to continue to change for each one.  I kinda don't like this solution, since it doesn't use the cumulative effect of firmware >x.  Just throwing it out there.

jessiewestlake:
I’m not sure where else I can look. This is my main view of the device:

C4124DB7-17E4-4DB7-9A83-1F0F704BB135

AlCalzone:
I'm not back yet, just lurking. That hardware version field looks promising. 
@jessiewestlake you'll need to use Z-Wave JS UI to find that value I'm afraid.

Whem we've confirmed that there is a difference, I can add support for this value (next year). It's currently not supported to distinguish config files.

AnthonyMaiorani:
This is what my v2 ZEN77 reports 

Screenshot 2022-12-01 at 4 07 04 PM

jessiewestlake:
> I'm not back yet, just lurking. That hardware version field looks promising. @jessiewestlake you'll need to use Z-Wave JS UI to find that value I'm afraid.
> 
> Whem we've confirmed that there is a difference, I can add support for this value (next year). It's currently not supported to distinguish config files.

I figured out how to send the commands to get the info. Unfortunately my ZEN72 ver. 2 reports this for the **Version CC hardwareVersion**:

````
DRIVER « [Node 013] [REQ] [ApplicationCommand]
         └─[Security2CCMessageEncapsulation]
           │ sequence number: 53
           └─[VersionCCReport]
               library type:      Enhanced Slave
               protocol version:  7.13
               firmware versions: 2.0
               hardware version:  1
````

jessiewestlake:
Another dead-end here on the ZEN72 VER 2.0:

````
[FirmwareUpdateMetaDataCCMetaDataReport]
manufacturer id:         634
firmware id:             40962
checksum:                0
firmware upgradable:     true
max fragment size:       28
additional firmware IDs: []
hardware version:        1
continues to function:   unknown
supports activation:     unknown
````

blhoward2:
Don’t we just need the condition to be (fw >= 1.0 && fw <2.0) || fw >10.10? Then version 2 is fw >= 2.0 && fw < 3.0? We have other devices where they renumbered like this.

jessiewestlake:
I guess that would do it for determining the hardware version through inference. Except the hardware v1 includes firmware 10.0. 

For these features it would be active if (fw >= 10.20 || (fw < 10.0 && >= 2.0))

I got the 10.20 from the documentation.

https://www.support.getzooz.com/kb/article/849-zen72-700-series-dimmer-change-log/

kpine:
> I’m not sure where else I can look. This is my main view of the device:

@jessiewestlake FYI, click the "..." overflow button that's visible in your screenshot, and then "Download diagnostics". A JSON-formatted diagnostics file for the device will be downloaded. In the file you can find all of the same values Z-Wave JS UI shows control panel shows. Probably not easy to use on a phone though.

For example, here are all the Version CC values from the diagnostics file for a ZEN72 purchased in 11/2021. It reports firmware 10.0 and hardware version 1, and based on the purchase date and change log, it must actually be a hardware version 1.0.

<details>
  <summary>Click to expand</summary>
  
  ```json
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "libraryType",
    "propertyName": "libraryType",
    "ccVersion": 3,
    "metadata": {
      "type": "number",
      "readable": true,
      "writeable": false,
      "label": "Library type",
      "states": {
        "0": "Unknown",
        "1": "Static Controller",
        "2": "Controller",
        "3": "Enhanced Slave",
        "4": "Slave",
        "5": "Installer",
        "6": "Routing Slave",
        "7": "Bridge Controller",
        "8": "Device under Test",
        "9": "N/A",
        "10": "AV Remote",
        "11": "AV Device"
      }
    },
    "value": 3
  },
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "protocolVersion",
    "propertyName": "protocolVersion",
    "ccVersion": 3,
    "metadata": {
      "type": "string",
      "readable": true,
      "writeable": false,
      "label": "Z-Wave protocol version"
    },
    "value": "7.13"
  },
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "firmwareVersions",
    "propertyName": "firmwareVersions",
    "ccVersion": 3,
    "metadata": {
      "type": "string[]",
      "readable": true,
      "writeable": false,
      "label": "Z-Wave chip firmware versions"
    },
    "value": [
      "10.0"
    ]
  },
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "hardwareVersion",
    "propertyName": "hardwareVersion",
    "ccVersion": 3,
    "metadata": {
      "type": "number",
      "readable": true,
      "writeable": false,
      "label": "Z-Wave chip hardware version"
    },
    "value": 1
  },
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "sdkVersion",
    "propertyName": "sdkVersion",
    "ccVersion": 3,
    "metadata": {
      "type": "string",
      "readable": true,
      "writeable": false,
      "label": "SDK version"
    },
    "value": "7.13.9"
  },
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "applicationFrameworkAPIVersion",
    "propertyName": "applicationFrameworkAPIVersion",
    "ccVersion": 3,
    "metadata": {
      "type": "string",
      "readable": true,
      "writeable": false,
      "label": "Z-Wave application framework API version"
    },
    "value": "10.13.9"
  },
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "applicationFrameworkBuildNumber",
    "propertyName": "applicationFrameworkBuildNumber",
    "ccVersion": 3,
    "metadata": {
      "type": "string",
      "readable": true,
      "writeable": false,
      "label": "Z-Wave application framework API build number"
    },
    "value": 405
  },
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "hostInterfaceVersion",
    "propertyName": "hostInterfaceVersion",
    "ccVersion": 3,
    "metadata": {
      "type": "string",
      "readable": true,
      "writeable": false,
      "label": "Serial API version"
    },
    "value": "unused"
  },
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "hostInterfaceBuildNumber",
    "propertyName": "hostInterfaceBuildNumber",
    "ccVersion": 3,
    "metadata": {
      "type": "string",
      "readable": true,
      "writeable": false,
      "label": "Serial API build number"
    },
    "value": 0
  },
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "zWaveProtocolVersion",
    "propertyName": "zWaveProtocolVersion",
    "ccVersion": 3,
    "metadata": {
      "type": "string",
      "readable": true,
      "writeable": false,
      "label": "Z-Wave protocol version"
    },
    "value": "7.13.9"
  },
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "zWaveProtocolBuildNumber",
    "propertyName": "zWaveProtocolBuildNumber",
    "ccVersion": 3,
    "metadata": {
      "type": "string",
      "readable": true,
      "writeable": false,
      "label": "Z-Wave protocol build number"
    },
    "value": 405
  },
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "applicationVersion",
    "propertyName": "applicationVersion",
    "ccVersion": 3,
    "metadata": {
      "type": "string",
      "readable": true,
      "writeable": false,
      "label": "Application version"
    },
    "value": "10.0.1"
  },
  {
    "endpoint": 0,
    "commandClass": 134,
    "commandClassName": "Version",
    "property": "applicationBuildNumber",
    "propertyName": "applicationBuildNumber",
    "ccVersion": 3,
    "metadata": {
      "type": "string",
      "readable": true,
      "writeable": false,
      "label": "Application build number"
    },
    "value": 43707
  }
  ```
</details>

PR #5083, opened 2022-09-19
docs: correct `json2nvm` documentation to use `--protocolVersion` flag

Tirenoth:
The json2nvm example command in the [readme](https://github.com/Tirenoth/node-zwave-js/blob/nvmeditDocumentationFix/packages/nvmedit/README.md) specifies using the "--version" flag, which only prints version information. The correct flag, according to the json2nvm help text, is "--protocolVersion"


AlCalzone (approved):
Thanks!

PR #5087, opened 2022-09-19
fix(config): add missing parameter 9 to Zooz ZEN04

mikebaz:
Correct missing parameter in #5043 reported by @gminco.


gminco (dismissed):

AlCalzone (approved):
It does make sense to group the LED-related parameters like Zooz did, so I moved the param up.

AlCalzone (approved):

PR #5102, opened 2022-09-23
docs: add link to the Z-Wave JS UI health check documentation

matejdro:
As per https://github.com/zwave-js/zwave-js-ui/discussions/2675#discussioncomment-3714547, extra link makes it easier to find the UI that network healt mentions.

### DIFF
--- a/docs/troubleshooting/network-health.md
+++ b/docs/troubleshooting/network-health.md
@@ -4,7 +4,7 @@ Only a strong and healthy Z-Wave network is a reliable Z-Wave network. Ensuring 
 
 ## Testing the connection strength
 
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
That link doesn't seem complete (it's relative to this documentation)

matejdro:
Hm, how should it be instead?

AlCalzone:
The full URL instead of the relative path.

matejdro:
Changed to full URL

AlCalzone (approved):

AlCalzone:
Thanks!

PR #5115, opened 2022-09-26
fix: check ccId for non-nullish instead of truthy in constructor

AlCalzone:


AlCalzone:
@zwave-js-bot approve

PR #5143, opened 2022-10-06
feat(config): add configuration for Honeywell Thermostat TH6320ZW2007

sormy:
Added support for TH6320ZW2007:
https://www.resideo.com/us/en/products/air/thermostats/programmable-thermostats/z-wave-t6-pro-programmable-thermostat-with-smartstart-th6320zw2007-u/

Tested by updating `devices` config for existing z-wave-server installation, restarting server, and pressing reinterview device in Home Assistant. All options have appeared in Home Assistant and are editable.

The config mostly blindly copied from th6320zw2003. I have quickly looked and it looks like at least configuration properties are at the same position as previous model th6320zw2003. New model TH6320ZW2007 has 3 more properties, I was able to trace one as humidity offset, but remaining 2 are unknown.

### DIFF
--- a/packages/config/config/devices/0x041b/th6320zw2007.json
+++ b/packages/config/config/devices/0x041b/th6320zw2007.json
@@ -1,12 +1,12 @@
 {
 	"manufacturer": "Honeywell",
-	"manufacturerId": "0x0039",
-	"label": "TH6320ZW",
-	"description": "T6 Pro Z-Wave Programmable Thermostat",
+	"manufacturerId": "0x041b",
+	"label": "TH6320ZW2007",
+	"description": "T6 Pro Z-Wave Programmable Thermostat with SmartStart",
 	"devices": [
 		{
 			"productType": "0x0011",
-			"productId": "0x0008"
+			"productId": "0x0009"
 		}
 	],
 	"firmwareVersion": {
@@ -181,6 +181,18 @@
 		{
 			"#": "42",
 			"$import": "~/templates/honeywell_template.json#temperature_offset"
+		},
+		{
+			"#": "43",
+			"$import": "~/templates/honeywell_template.json#humidity_offset"
+		},
+		{
+			"#": "44",
+			"$import": "~/templates/honeywell_template.json#temperature_resolution"
+		},
+		{
+			"#": "45",
+			"$import": "~/templates/honeywell_template.json#humidity_resolution"
 		}
 	],
 	"compat": {
--- a/packages/config/config/devices/templates/honeywell_template.json
+++ b/packages/config/config/devices/templates/honeywell_template.json
@@ -44,7 +44,7 @@
 		"valueSize": 1,
 		"minValue": 0,
 		"maxValue": 65,
-		"defaultValue": 65,
+		"defaultValue": 0,
 		"unit": "°F",
 		"options": [
 			{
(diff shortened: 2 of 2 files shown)


blhoward2:
Normally we wouldn't let the unknown parameters fly, but as I recall this is one of the weird reverse engineered ones where the manufacturer provides no information, correct? If so, I think we can go with what we can deduce.

sormy:
> Normally we wouldn't let the unknown parameters fly, but as I recall this is one of the weird reverse engineered ones where the manufacturer provides no information, correct? If so, I think we can go with what we can deduce.

I'm not sure how parameters for TH6320ZW2003 were discovered. I'm not sure if it is one of weird reversed engineered ones. I have copied parameter configuration as it is from TH6320ZW2003 to TH6320ZW2007, was able to trace humidity offset and have found there are two remaining unmapped parameters. I would like to map them but I don't know where to obtain this information.

In any case, I think 2 unmapped parameters is better than 45 unmapped parameters. If somebody will be able to trace what these remaining two are for, the configuration still can be updated and improved. The comment in the code regarding unknown parameters was left for the purpose of discoverability, so someone can see that there is something to improve here.

blhoward2:
>I would like to map them but I don't know where to obtain this information.

Before we merge this, I just want to confirm. Did you google looking for a device manual that would contain this information? Look for device files at other projects? Does this device have a Smartthings handler from which we might extract this information?

sormy:
> Did you google looking for a device manual that would contain this information? 

I have it in hands. Unfortunately, it doesn't list parameter positions, only ISU numbers so it is hard to match them. There are 53 ISU parameters in the manual and 45 in home assistant (with zwave-js integration). Manual says that it is for both TH6320ZW2003 and TH6320ZW2007 versions.

> Look for device files at other projects? 

google shows just 59 results for TH6320ZW2007, nothing helpful

> Does this device have a Smartthings handler from which we might extract this information?

I don't know.

---

I have slightly updated PR:

- Fixed humidity offset from -12 to 12 (as per manual)
- Fixed default value for base_lockout_temperature template (not a bit deal but an improvement to match with manual and default state of device)
- Found unknown 44/45 parameters - temperature/humidity resolution. These are hard to trace, adjustable only using z-wave, no control on device itself.
- Fixed temperature offset interpretation for values in C and F

It might be that TH6320ZW2003/TH6320ZW2007 have actually exactly the same parameter configuration. It might be just TH6320ZW2003 config is not fully correct. But I don't have TH6320ZW2003 in hands to check.

AlCalzone:
Isn't Honeywell part of Jasco?

Guess we could ask @jascoproducts if they have more info.

blhoward2:
I think Honeywell licenses jasco switches (or jasco licenses Honeywell's name for that purpose). I'm not sure that Jasco makes the thermostat. Honeywell makes all sorts of stuff. It's partially owned by GE, though.

blhoward2:
Have you googled and looked for a manual that may have the parameters? A device file from another project? Is there a SmartThings device file for the thermostat?Sent from my iPhoneOn Oct 8, 2022, at 8:07 PM, Artem Butusov ***@***.***> wrote:﻿

Normally we wouldn't let the unknown parameters fly, but as I recall this is one of the weird reverse engineered ones where the manufacturer provides no information, correct? If so, I think we can go with what we can deduce.

I'm not sure how parameters for TH6320ZW2003 were discovered. I'm not sure if it is one of weird reversed engineered ones. I have copied parameter configuration as it is from TH6320ZW2003 to TH6320ZW2007, was able to trace humidity offset and have found there are two remaining unmapped parameters. I would like to map them but I don't know where to obtain this information.
In any case, I think 2 unmapped parameters is better than 45 unmapped parameters. If somebody will be able to trace what these remaining two are for, the configuration still can be updated and improved.
The comment in the code regarding unknown parameters was left for the purpose of discoverability, so someone can see that there is something to improve here.

jascoproducts:
Thanks for tagging of us @AlCalzone, but @blhoward2 is correct. Jasco previously licensed the Honeywell brand specifically for connected Lighting devices only. We only produced Z-Wave In-Wall and Plug-in Lighting controls under the Honeywell brand. We no longer hold that license and do not produce Honeywell branded products any longer.

sormy:
So, is there anything I can do for this CR to be merged? There are no unknown fields anymore.

blhoward2:
Yes, I'll look at it but it'll probably be this weekend until I can do so.

runningopenloop:
From the doc: https://digitalassets.resideo.com/damroot/Original/10017/33-00588EFS.pdf

Seems the only difference between 2003 and 2007 according to document above is... (searching 2007)
```
SmartStart: (TH6320ZW2007 model only). After the initial setup and time/date
have been set, the TH6320ZW2007 thermostat will initiate SmartStart. Icon in upper
right of display will flash. Thermostat stays in SmartStart mode indefinitely until it is
included in a Z-Wave network.

Identify thermostat: (TH6320ZW2007 only). If the thermostat is included in a Z-Wave
Network. You can choose to identify the thermostat via Z-Wave app. Backlight will
flash. Touching display while flashing will stop Identify mode. This is helpful when
multiple thermostats are on the same Z-Wave Network.
```

I'm wondering if one of those codes could be to set identify thermostat mode?  I think it would be nice for a home automation alert to turn on/off this identify thermostat mode.

I'm considering this thermostat myself.

huff (approved):

masdgaf:
Was this waiting for anything other than a review? I've been running these updated files for a while without issue.

AlCalzone (approved):
LGTM

PR #5148, opened 2022-10-08
fix: add resiliency to import config routines

blhoward2:
Recently introduced corrupt entries at the Z-Wave Alliance DB cause importConfig to fail in some instances. These changes add 0x9999 placeholders to missing productId or productType entries, and if present, inserts a TODO warning in a comment at the top of the device file.

I went with a placeholder instead of putting "Unknown" or something there, as using a string would require a bunch of refactoring due to the number of different places we enforce the format and use that value to lookup manufacturer names and such. 

Screen Shot 2022-10-08 at 1 52 47 PM

### DIFF
--- a/packages/config/maintenance/importConfig.ts
+++ b/packages/config/maintenance/importConfig.ts
@@ -1560,11 +1560,9 @@ async function parseZWAProduct(
 	let output = "";
 	if (
 		newConfig.devices.filter(
-			(d) =>
-				d.productType === "0x9999" ||
-				d.productId === "0x9999" ||
-				d.manufacturerIdHex === "0x9999",
-		)
+			(d) => d.productType === "0x9999" || d.productType === "0x9999",
+		).length > 0 ||
+		newConfig.manufacturerIdHex === "0x9999"
 	) {
 		output = `// TODO: This file contains a placeholder for a productType, productID, or manufacturerId (0x9999) that must be corrected. 
 ${JSONC.stringify(normalizeConfig(newConfig), null, "\t") + "\n"}`;
--- a/packages/config/maintenance/importConfig.ts
+++ b/packages/config/maintenance/importConfig.ts
@@ -847,6 +847,25 @@ async function parseZWAFiles(): Promise<void> {
 		}
 	}
 
+	// Check for missing fields and add placeholders if missing
+	for (const device of jsonData) {
+		if (!device.ManufacturerId) {
+			device.ManufacturerId = "0x9999";
+		}
+
+		if (!device.Brand) {
+			device.ManufacturerId = "Unknown";
+		}
+
+		if (!device.ProductTypeId) {
+			device.ProductTypeId = "0x9999";
+		}
+
+		if (!device.ProductId) {
+			device.ProductId = "0x9999";
+		}
+	}
+
 	// Combine provided files within models
 	jsonData = combineDeviceFiles(jsonData);
 
@@ -865,11 +884,16 @@ async function parseZWAFiles(): Promise<void> {
 		const manufacturerName =
 			configManager.lookupManufacturer(manufacturerId);
 
-		// Add the manufacturer to our manufacturers.json if it is missing
-		if (Number.isNaN(manufacturerId)) {
-		} else if (manufacturerName === undefined && file.Brand !== undefined) {
-			console.log(`Adding missing manufacturer: ${file.Brand}`);
-			configManager.setManufacturer(manufacturerId, file.Brand);
+		if (file.ManufacturerId != "0x9999") {
+			// Add the manufacturer to our manufacturers.json if it is missing
+			if (Number.isNaN(manufacturerId)) {
+			} else if (
+				manufacturerName === undefined &&
+				file.Brand !== undefined
+			) {
+				console.log(`Adding missing manufacturer: ${file.Brand}`);
+				configManager.setManufacturer(manufacturerId, file.Brand);
+			}
 		}
 
 		/**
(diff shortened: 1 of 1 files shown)


AlCalzone:
Will try to look at this tomorrow when the kids are asleep and I have some time to kill

blhoward2:
> Will try to look at this tomorrow when the kids are asleep and I have some time to kill

Sounds good. This is fully tested now. Running the Yale locks I had a variety of conforming and non-conforming files, and after a fix or two this now works as pushed.

packages/config/maintenance/importConfig.ts
```diff
@@ -1532,9 +1556,21 @@
 	const manufacturerDir = path.join(processedDir, manufacturerIdHex);
 	await fs.ensureDir(manufacturerDir);
 
+	// Insert a TODO comment if necessary
+	let output = "";
+	if (
+		newConfig.devices.filter(
+			(d) => d.productType === "0x9999" || d.productType === "0x9999",
```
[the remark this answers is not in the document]
blhoward2: Fixed

packages/config/maintenance/importConfig.ts
```diff
@@ -847,6 +847,25 @@ async function parseZWAFiles(): Promise<void> {
...
 	}
 
+	// Check for missing fields and add placeholders if missing
+	for (const device of jsonData) {
+		if (!device.ManufacturerId) {
+			device.ManufacturerId = "0x9999";
+		}
+
+		if (!device.Brand) {
+			device.ManufacturerId = "Unknown";
```
AlCalzone:
```suggestion
			device.Manufacturer = "Unknown";
```
I guess?

packages/config/maintenance/importConfig.ts
```diff
@@ -865,11 +884,16 @@ async function parseZWAFiles(): Promise<void> {
...
 
-		// Add the manufacturer to our manufacturers.json if it is missing
-		if (Number.isNaN(manufacturerId)) {
-		} else if (manufacturerName === undefined && file.Brand !== undefined) {
-			console.log(`Adding missing manufacturer: ${file.Brand}`);
-			configManager.setManufacturer(manufacturerId, file.Brand);
+		if (file.ManufacturerId != "0x9999") {
+			// Add the manufacturer to our manufacturers.json if it is missing
+			if (Number.isNaN(manufacturerId)) {
+			} else if (
```
AlCalzone:
```suggestion
			if (
				!Number.isNaN(manufacturerId) &&
```
feels a bit more natural than an empty if

packages/config/maintenance/importConfig.ts
```diff
@@ -1532,9 +1556,21 @@ async function parseZWAProduct(
...
+			(d) => d.productType === "0x9999" || d.productId === "0x9999",
+		).length > 0 ||
+		newConfig.manufacturerIdHex === "0x9999"
+	) {
+		output = `// TODO: This file contains a placeholder for a productType, productID, or manufacturerId (0x9999) that must be corrected. 
+${JSONC.stringify(normalizeConfig(newConfig), null, "\t") + "\n"}`;
+	} else {
+		output = JSONC.stringify(normalizeConfig(newConfig), null, "\t") + "\n";
+	}
+
```
AlCalzone, line 1586: I'd do
```
let output =...
```
And then conditionally prepend the comment:
```
if (...) {
output = "...\n" + output;
}
```

(check formatting yourself, I'm on mobile 🙃)

AlCalzone (changes requested):
A few minor things

packages/config/maintenance/importConfig.ts
blhoward2: This was already there. I assumed you did that for a reason. I thought it was odd too. I'll change it.

AlCalzone: Dunno, maybe the condition evolved over time

AlCalzone:
Sorry, one more thing.
Use `!==` instead of `!=`. The only exception is checking against null/undefined.

blhoward2:
> Sorry, one more thing. Use `!==` instead of `!=`. The only exception is checking against null/undefined.

Done

blhoward2:
@AlCalzone We need to sort parameters but only where they didn't come from an original file. Ideas? Right now for a new import they end up 1, 11, 12, 2, 21, etc. I can sort them retaining the comments, but then it clobbers if we had a file manually sorted to begin with.

I could maybe add a dummy key if it is a new file (which would get stripped out in normalizeConfig, and only sort in normalizeConfig if it is there?

blhoward2:
> @AlCalzone We need to sort parameters but only where they didn't come from an original file. Ideas? Right now for a new import they end up 1, 11, 12, 2, 21, etc. I can sort them retaining the comments, but then it clobbers if we had a file manually sorted to begin with.
> 
> I could maybe add a dummy key if it is a new file (which would get stripped out in normalizeConfig, and only sort in normalizeConfig if it is there?

@AlCalzone Ok, please look at my solution. It works fine. I've tested this and it preserves the sort order for existing documents (both regular and partial params), preserves comments in existing files, and sorts new files by parameter number.

AlCalzone (approved):
Seems good. Note that the check for 0x9999 **might** introduce false positives though if a device actually uses it as an ID.
But I guess we can deal with that then.

blhoward2:
I debated that. I'm not sure how likely it is. If you're worried about it, I can also populate a key when I set 0x9999 (sort of like I did with the search) and use that to trigger the warning. Then it would only trigger if we'd set the 0x9999 versus it coming from the db.

PR #5153, opened 2022-10-08
chore: add snippets for authoring config files to VSCode

blhoward2:
Because I can never remember how to add an import, I've created a set of snippets that insert various things:
- Add device entry
- Conditional parameters
- Import from master device file (with and without using the clipboard contents as the template name
- Import from manufacturer device file (with and without using the clipboard contents as the template name
- New Parameters (with manual entry)
- New Parameters (without manual entry)
- Insert allowManualEntry scaffold with two options
- Insert a Disable option (for when manual entry is allowed)
- Insert compat section, with all of the true/false compat flags in a drop-down

Hitting tab jumps you from piece to piece where you'd need to change things.

Someday maybe I'll get around to writing an extension, but for now this is helpful (to me at least).

### DIFF
--- a/.vscode/deviceSnippets.code-snippets
+++ b/.vscode/deviceSnippets.code-snippets
@@ -134,10 +134,55 @@
 		"prefix": ["Insert Compat Flag"],
 		"body": [
 			"\"compat\": {",
-			"$LINE_COMMENT This device incorrectly ${1:[problem]}",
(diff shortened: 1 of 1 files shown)


AlCalzone (dismissed):
Haven't checked them in depth but I like it.

What about some common compat flags that are hard to remember, like disabling Supervision CC?

blhoward2:
Ok, I added disabling supervision, removing a CC, and preserving endpoints. All of the true/false compat flags were already supports.

@robertsLando let's go ahead and merge and we can always add more later.

AlCalzone (approved):

PR #5156, opened 2022-10-11
feat(config): zooz zen72 zen74 zen77 param 27 - 30

alvinchen1:
Finishing up Issue [4786](https://github.com/zwave-js/node-zwave-js/issues/4786).  
Added parameter 27, 28, 29 to zen77.  
Added parameter 30 to zen72, zen74 and zen77.  

Could not obtain hardware identifier for zen77 2.0 or 3.0 to get firmware versions listed there.  
Recommend closing 4786 and when hardware identifiers for zen77 2.0 and 3.0 show up, can write configuration files for them.  Also this should take care of parameter 30 which was in question on PR [5080](https://github.com/zwave-js/node-zwave-js/pull/5080).  
Recommend closing PR 5080.


blhoward2:
We can just ask Zooz for the information that you need. Please list our specifically what you need and I'll get that info.

alvinchen1:
The original request included this:

ZEN77 10.20 (ver. 1.0 hardware), 2.10 (ver. 2.0 hardware), and 3.0 (ver. 3.0 hardware)

See https://github.com/zwave-js/node-zwave-js/issues/4786#issuecomment-1264221997
User reported having Zooz Zen77 v2.0.  But when he went to Z-Wave JS Device Database link, it took him to 

Product Manufacturer: 0x027a
Product Identifiers: 0x7000:0xa007

This is the same Product Identifier as version 1.0. If Zooz can provide us a different Product Identifier for vers 2.0 and vers 3.0 of Zooz77 then my thought was we could write the config files for each.  But since version 2.0 appears to be reporting the same identifiers as version 1.0, I don't see a way we can write the config file where:

IF version 1.0 and firmware > 10.20 then Parameter 27 - 30
IF version 2.0 and firmware > 2.10 then Parameter 27 - 30
IF version 3.0 and firmware > 3.0 then Parameter 27 - 30

If all three configs are the in the same file, then won't it present Parameter 27 - 30 if the firmware is > 3.0 since it can't tell between version 1.0, 2.0 and 3.0?  Or am I missing something other means of writing a firmware and version check?

blhoward2:
Zooz changes the fw version with the different hardware versions. So fw >2.10 && fw <3.0, it is hardware version 2. Etc.

alvinchen1:
> Zooz changes the fw version with the different hardware versions. So fw >2.10 && fw <3.0, it is hardware version 2. Etc.

So, maybe I don't understand but, if that's the case, what else needs to be done to merge?

AlCalzone:
@blhoward2 ?

packages/config/config/devices/0x027a/zen77.json
```diff
@@ -155,6 +155,26 @@
 			"#": "26",
 			"$if": "firmwareVersion >= 10.0",
 			"$import": "templates/zooz_template.json#local_programming"
+		},
+		{
+			"#": "27",
+			"$if": "firmwareVersion >= 10.20",
```
TaiSHiNet, line 161: This (and further conditionals) should follow the following logic https://github.com/zwave-js/node-zwave-js/pull/5156#issuecomment-1280222835

```
IF version 1.0 and firmware > 10.20 then Parameter 27 - 30
IF version 2.0 and firmware > 2.10 then Parameter 27 - 30
IF version 3.0 and firmware > 3.0 then Parameter 27 - 30
```

Reason for that is Zooz changes the firmware version based on the hardware version

TaiSHiNet:
> > Zooz changes the fw version with the different hardware versions. So fw >2.10 && fw <3.0, it is hardware version 2. Etc.
> 
> So, maybe I don't understand but, if that's the case, what else needs to be done to merge?

I made a comment on the code on what needs addressing

alvinchen1:
See https://github.com/zwave-js/node-zwave-js/pull/5244.  There is NO current way of distinguishing between HARDWARE version 1.0, 2.0 and 3.0. Zooz presents all Zen77 with the same hardware identifier regardless of version.  We looked to see if Version CC would work, but it also does not distinguish them.  Therefore there may not currently be anything that can be done to distinguish between hardware versions. Unfortunately that means, that we can merge firmware > 10.20, but there is nothing we can currently do about using the additional checks for Hardware Version AND Firmware version.

My 2 cents... merge so that hardware version 1 can get the settings.  Once we have a way to distinguishing hardware version 2 and 3, we can write the settings files for them.  Otherwise, reject and close.

blhoward2:
I don’t think that’s right. They usually bump the fw version for a hardware revision and we can tell via the fw version. Are you positive that hasn’t happened here?

TaiSHiNet:
> I don’t think that’s right. They usually bump the fw version for a hardware revision and we can tell via the fw version. Are you positive that hasn’t happened here?

Indeed, based on the [changelog](https://www.support.getzooz.com/kb/article/850-zen77-700-series-s2-dimmer-change-log/) versions 1.x and 10.x (why 10? I don't know) refer to HW 1.0, versions 2.x for hardware 2.0 and versions 3.x for hardware 3.0

blhoward2:
10.10 was due to a miscommunication with their programmers. They wanted to use 1.10 and 1.20 to ensure consistency among platforms. They told us that at the time.

blhoward2:
Zooz confirmed this, so the files just need to be conditioned on the firmware version number. The version 1 split can be done with an || between the two tests. 

 Firmware starting with 1.x or 10.x applies to VER. 1.0, firmware starting with 2.x applies to VER. 2.0 and for the ZEN77, firmware starting with 3.x applies to VER. 3.0.”

PR #5181, opened 2022-10-26
fix(config): allow higher minimum dim level for Inovelli LZW31-SN, FW 1.57+

kreene1987:
Per: https://community.inovelli.com/t/maximum-mininum-level-of-45/2727/13

Only applies to FW v 1.57+, please confirm applied properly.


wrender:
oh. Thanks for doing this. Is it possible for me to override these files locally on my own home assistant, that has zwave js?  So I can test?

AlCalzone (approved):

PR #5195, opened 2022-11-07
fix: indicator identify command is v3

kpine:
CC API IndicatorCC.identify fails even for CC v3, this fixes the version check.


AlCalzone:
Whoops

AlCalzone (approved):

PR #5198, opened 2022-11-07
fix(config): swap parameter numbers of myot4

czesiu20:
Parameter numbers 30 and 31 must be swapped.


AlCalzone:
You might want to **reset** your master branch to ours after this is merged and then work in branches only. Makes the git history clearer.

AlCalzone (approved):

PR #5206, opened 2022-11-13
fix(config): correct Aeotec Multisensor 7, param 2

papadave1984:
parameter 2 should have a range of 0 to 30 seconds not 30 to 3600 seconds

PR description here


AlCalzone:
Where are you getting the upper bound of 30 from?
grafik

AlCalzone (approved):
ok with my change

PR #5220, opened 2022-11-24
feat: prevent parallel FW updates and soft/hard reset during ongoing FW updates

raman325:
Per https://github.com/zwave-js/zwave-js-server/pull/772#discussion_r1028144666 , we won't let hard resets or soft resets occur while a firmware update is in progress. Al has also mentioned that two firmware updates should not happen in parallel (at least for now), so this PR protects against those as well

### DIFF
--- a/packages/core/src/error/ZWaveError.ts
+++ b/packages/core/src/error/ZWaveError.ts
@@ -158,7 +158,7 @@ export enum ZWaveErrorCodes {
 	/** Gets thrown when parsing an invalid QR code */
 	Security2CC_InvalidQRCode,
 
-	/** The firmware update process is already active */
+	/** The firmware update process is already active on this node */
 	FirmwareUpdateCC_Busy = 1500,
 	/** The selected firmware target is not upgradable */
 	FirmwareUpdateCC_NotUpgradable,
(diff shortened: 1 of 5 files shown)


packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -4650,6 +4659,16 @@ ${associatedNodes.join(", ")}`,
 			);
 		}
 
+		// Don't let two firmware updates happen in parallel
+		if (this.isAnyOTAFirmwareUpdateInProgress()) {
```
AlCalzone, line 4663: I'm sure this change will get complaints from users who queue a bunch of firmware updates for sleeping nodes. However I think it makes sense, at least until I can implement some kind of task scheduling in the driver.

packages/core/src/error/ZWaveError.ts
```diff
@@ -176,6 +176,9 @@ export enum ZWaveErrorCodes {
 	/** An firmware file with an unsupported format was provided */
 	Unsupported_Firmware_Format,
 
+	/** A firmware update is already in progress preventing this action from proceeding */
+	Firmware_Update_In_Progress_Error,
```
AlCalzone: I'm guessing you want to distinguish this from `FirmwareUpdateCC_Busy`?

AlCalzone: In any case, I renamed it to match the other flag.

AlCalzone (approved):
Ok with my changes

PR #5223, opened 2022-11-26
fix(config): add metadata for Sensative strips

LordMike:
Add metadata to Sensative strips

### DIFF
--- a/packages/config/config/devices/0x019a/11_02_011.json
+++ b/packages/config/config/devices/0x019a/11_02_011.json
@@ -335,5 +335,12 @@
 			"maxValue": 120,
 			"defaultValue": 0
 		}
-	]
+	],
+	"metadata": {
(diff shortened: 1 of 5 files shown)


packages/config/config/devices/0x019a/11_02_011.json
```diff
@@ -335,5 +335,12 @@
 			"maxValue": 120,
 			"defaultValue": 0
 		}
-	]
+	],
+	"metadata": {
+		"inclusion": "Set your controller in add or remove mode (see your controller’s manual). Follow the instruction below for Wake up. A long LED blink indicates that the add/remove was successful.\n\nTo wake up Strips manually for communication with the Z-Wave controller, move the magnet towards and away from the rounded edge. Repeat 3 times.",
```
AlCalzone: Can you make this and the following line consistent across these files? I think I prefer "inclusion mode" and "exclusion mode"

AlCalzone (changes requested):

packages/config/config/devices/0x019a/11_02_011.json
LordMike: There's a mix of different texts all over the place, here, other config  files, and various manuals from manufacturers.. -.-'

AlCalzone (approved):

AlCalzone:
Took a bit to review, but thanks :)

PR #5244, opened 2022-12-07
fix(config): support hardware version 2.0 for Zooz ZEN71, add parameter 17

namomitk:
The hardware version 2.0 of the switch uses different versioning. Can this be changed to support the different firmware versioning?

"$if": "firmwareVersion >= 10.0",
to

Hardware Version 1.0 is currently Firmware Version 10.10 Hardware Version 2.0 is currently Firmware Version 2.10 (>= 2.0 has the parameter 17)

This would capture both hardware versions without showing the parameter for those on FW version 1.0 or 1.01

https://www.support.getzooz.com/kb/article/1102-zen71-700-series-on-off-switch-change-log/

### DIFF
--- a/packages/config/config/devices/0x027a/zen71.json
+++ b/packages/config/config/devices/0x027a/zen71.json
@@ -76,7 +76,7 @@
 		},
 		{
 			"#": "17",
-			"$if": "firmwareVersion >= 10.0",
+			"$if": "firmwareVersion >= 2.0",
 			"$import": "templates/zooz_template.json#local_programming"
 		}
 	],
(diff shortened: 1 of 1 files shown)


AlCalzone:
I guess if we find a solution for https://github.com/zwave-js/node-zwave-js/pull/5156, we can do something similar here.

Simply changing it to >= 2.0 won't work because it doesn't capture params that get added in versions between 2 and 10, but are not present in the 10.x firmware.

namomitk:
Fair enough. I worked around the issue for myself. I'll keep an eye on the other issue. Thanks. =)

AlCalzone:
Hmm actually it does make sense here, we just can't use `>= 2.x` for parameters that get added after 2.0

AlCalzone (approved):

PR #5267, opened 2022-12-20
Update nas-wr01ze.json

mulbuck:
PR description here
In my current usage of this device, those are the information coming back from the device.
I'd like it to be integrated in zwave-js before migrating


AlCalzone:
The same fingerprint already exists in wr01z.json, so it cannot be just added here.

Can you clarify the difference and why you think it needs to be in this file?

PR #5269, opened 2022-12-20
fix(config): allow higher pulse factor for NQ-9021

hannut:
Pulse factor needs to be 100,000 for most electricity meters in Finland. According to below post the maximum is 1,000,000 for this device.

* https://forum.fibaro.com/topic/24261-northq-nq-9021-powermeter/


AlCalzone (approved):
According to the forum post you linked, this parameter was changed in firmware 3.24. I modeled the parameter with a condition accordingly.

PR #5273, opened 2022-12-22
fix(config): add new product ID for 46201/ZW4008

BrandtWoolf:
Adding support for the GE/Enbrighten In-Wall Paddle Switch, ZW4008 /42601-2, with the product ID of `0x3036`. 

Logs to show productId when interviewing the device:

```
received response for manufacturer information:
                                    manufacturer: GE/Jasco (0x63)
                                    product type: 0x4952
                                    product id:   0x3036
```

### DIFF
--- a/packages/config/config/devices/0x0063/46201_zw4008.json
+++ b/packages/config/config/devices/0x0063/46201_zw4008.json
@@ -8,6 +8,11 @@
 			"productType": "0x4952",
 			"productId": "0x3135",
 			"zwaveAllianceId": 4143
+		},
+    {
+			"productType": "0x4952",
+			"productId": "0x3036",
+			"zwaveAllianceId": 4143
 		}
 	],
 	"firmwareVersion": {


BrandtWoolf:
I see that productId `0x3036` is duplicate in [14291_zw4005.json](https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x0063/14291_zw4005.json) however the device I purchased is labeled as ZW4008. 

image

**Reasoning for this PR is home assistant does not recognize this new switch for manufacturer and model.**  

Based on what I have found here in the [device listing](https://devices.zwave-js.io/?jumpTo=0x0063:0x4952:0x3036:0.0), this switch should be properly identified but it is not. 

How do I proceed to get this device supported?

AlCalzone:
@jascoproducts can you clear up the confusion?

BrandtWoolf:
@AlCalzone would you like me to open an issue on this? I went straight for the PR thinking this would be a shoo-in but clearly ran into some problems

BrandtWoolf:
Dug the other box out of the trash, I apparently got 2 different model switches is my order which then make 14291_zw4005 correct for the device in question. 

Still doesn't answer why home assistant doesn't assign the model and manufacture but will take that issue elsewhere.

AlCalzone:
An issue would be ideal 👍🏻

PR #5275, opened 2022-12-26
fix(config): add wakeup instructions for Vision Security ZP3111-5

honikos:
this is helpful to configure the device immediately


AlCalzone (approved):
ok with my change

PR #5280, opened 2022-12-30
fix(config): add fingerprint for Fibaro FGWP102

dacwe:
Got a late 2022 batch of these wall plugs with this product id.

image


dacwe:
Shout out if there is anything more I need to do to get this PR merged.

AlCalzone (approved):

PR #5282, opened 2023-01-01
feat(config): add Namron 16 A switch and 4-channel remote control

Wuerger:
Add Namron devices:
- 16 A switch (4512746)
- 4-channel remote control (4512711)


AlCalzone (approved):
Cleaned this up a bit, ok with my changes

PR #5286, opened 2023-01-02
fix(config): correct low temp threshold for Aeotec aerQ

AlCalzone:


AlCalzone:
@zwave-js-bot approve

PR #5288, opened 2023-01-02
feat: add readonly `rfRegion` property to `Controller` class

AlCalzone:
This PR adds the `rfRegion` property to the `Controller` class. This automatically gets queried during the controller identification phase now and updated when using `setRFRegion` or `getRFRegion`.

Outside of these, the RF region of the controller isn't going to change, so applications can now display/use the RF region without having to query it and handle errors.

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -971,6 +971,15 @@ enum InclusionState {
 }
 ```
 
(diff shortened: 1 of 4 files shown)


AlCalzone:
@robertsLando :)

PR #5309, opened 2023-01-05
feat: support pre-filling the DSK and scanning DSK-only QRs

### LINKED ISSUE #5205: Smart Start: QR Code inclusion
A controller providing the SmartStart functionality SHOULD provide a QR Code scanning capability. If the controller offers a QR Code scanning capability:

- [x] It MUST support the addition of nodes using QR Code format defined in [10] in its Provisioning List when scanning the QR Code.
- [ ] It SHOULD support scanning of S2 only QR codes representing the DSK String prefixed with “zws2dsk:”.(example: “zws2dsk:34028-23669-20938-46346-33746-07431-56821-14553”) in order to simplify the S2 bootstrapping process.

_Note: We should also support un-prefixed DSK QR codes_

AlCalzone:
@raman325 @robertsLando 

This is an attempt at supporting QR codes that only include the DSK (prefixed or unprefixed). I tried to add this functionality to the API without breaking changes - doing this as part of `parseQRCodeString` would require changes to the return type which in turn would require application-level checks to decide what to do with the info.

The downside is that after scanning, applications need to first check if the QR code string is a valid DSK (`tryParseDSKFromQRCodeString`) and if yes use the S2 inclusion strategy with pre-filled DSK.
If not, attempt to parse the encoded info from the QR code string as they do now.

Thoughts?

fixes: #5205

### DIFF
--- a/packages/core/api.md
+++ b/packages/core/api.md
@@ -2312,6 +2312,11 @@ export enum TransmitStatus {
     OK = 0
 }
 
+// Warning: (ae-missing-release-tag) "tryParseDSKFromQRCodeString" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
+//
+// @public
+export function tryParseDSKFromQRCodeString(qr: string): string | undefined;
+
 // Warning: (ae-missing-release-tag) "TXReport" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
 //
 // @public (undocumented)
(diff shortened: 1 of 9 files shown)


robertsLando (dismissed):
LGTM for me 👍🏼

raman325:
so if I understand this correctly, the existing logic would not work for this type of QR code, both before and after the change. After this change, we'd have to add explicit support by adding a logic branch that checks the QR code against this new function and branches off to inclusion strategy with pre-filled DSK if it's this type of QR code?

If I got that right, this looks like a great way to solve this problem without breaking anything.

AlCalzone:
Correct

robertsLando:
@AlCalzone Could that function be 'safe' so I can use it directly on frontend?

AlCalzone:
done

robertsLando:
Perfect! That would make it even easier to integrate :)

AlCalzone:
@zwave-js-bot pack this

PR #5323, opened 2023-01-12
fix: distinguish between protocol and SDK version on 500 series

### LINKED ISSUE #5160: sdkVersion reported for some 500-series controllers looks like the protocol version
Submitting as discussed in Discord.

I've seen some reports of sdkVersions of 500-series controllers looking incorrect. One instance was a Zooz ZST10 reporting SDK version "6.4.0". Another is a firmware-upgraded Nortek reporting "6.10.0". Neither of these are valid Z-Wave SDK versions for 500-series (only for 300/400). The versions actually appear to be the protocol version (or controller library version), as protocol version 6.10 corresponds to SDK version 6.84.0, and protocol version 6.04 corresponds to SDK version 6.81.3. The latter is the certified SDK version of the Zooz ZST10, so that matches.

My guess is that `protocolVersionToSDKVersion` needs to be called somewhere for these controllers to convert the protocol version to the proper SDK version. Or, it's fine as-is and really is meant to be the protocol version.

Attached is a log of the Nortek controller interview showing this.

[Nortek.txt](https://github.com/zwave-js/node-zwave-js/files/9768044/Nortek.txt)

AlCalzone:
fixes: #5160

### DIFF
--- a/packages/zwave-js/api.md
+++ b/packages/zwave-js/api.md
@@ -1013,6 +1013,8 @@ export class ZWaveController extends TypedEventEmitter<ControllerEventCallbacks>
     get productId(): number | undefined;
     // (undocumented)
     get productType(): number | undefined;
+    // (undocumented)
+    get protocolVersion(): string | undefined;
     provisionSmartStartNode(entry: PlannedProvisioningEntry): void;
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot approve

PR #5361, opened 2023-01-25
fix(config): additional cleanup for Inteset motion sensor

AlCalzone:


AlCalzone:
@zwave-js-bot approve

PR #5362, opened 2023-01-25
feat: firmware flasher utility

AlCalzone:
This PR adds a CLI utility that uses the `zwave-js` driver to flash firmware updates on a controller.


AlCalzone:
@zwave-js-bot approve

Issue #5360, opened 2023-01-25
Discard all pending transactions when starting OTW update, do additional checks not to communicate with the controller


alexruffell:
### Checklist

- [ ]  I am not using Home Assistant. Or: a developer has told me to come here.
- [X] I have checked the troubleshooting section and my problem is not described there.
- [X] I have read the changelog and my problem is not mentioned there.

### Deploy method

Docker

### Z-Wave JS UI version

8.8.0

### ZwaveJS version

10.4.0

### Describe the bug

I initiated an upgrade of my Aeotec S7 using the new OTW functionality and soon after starting the update I got these errors:

<img width="584" alt="image" src="https://user-images.githubusercontent.com/4006944/214606103-3aa3353d-acc7-4056-b69d-a7b4eaa93f23.png">

Driver: Failed to recover from bootloader. Please flash a new firmware...

Problem is that neither HA or Zwave PC Controller see the stick as connected. It sees the COM port but not the Zwave stick so I cannot flash a new firmware or even restore the NVM I backed up just before testing this.

<img width="591" alt="image" src="https://user-images.githubusercontent.com/4006944/214607220-b44c2d53-d8dc-4ebd-93a0-960b9d9ea777.png">

### To Reproduce

Update to 7.19.0 using Zwave JS UI OTW. The file I used was BRD4206A as it is the same board type as the one on Aeotec's site.

### Expected behavior

n/a

Daniel-dev22:
What firmware did you use? There's one for zooz and one for aeotec.

alexruffell:
I used: zwave_ncp_serial_api_controller_BRD4206A.gbl version 7.19.0

alexruffell:
@Daniel-dev22 Is there any tool that I can use to flash the stick that doesn't expect to find a zwave controller? Zwave PC Controller won't let me even try and the same goes with ZWave JS UI. I searched Simplicity Studio but did not see anything... but I am also not sure where to look.

AlCalzone:
I waited for this to happen 🙈
Will make a small utility later today or tomorrow.

robertsLando:
Try to enable boot loader only mode under zwave settings and then trigger the fwupdate 

![2023-01-25_18-11](https://user-images.githubusercontent.com/11502495/214632869-584ff5de-556f-4fc0-882d-eefa1a0fcbd9.png)

alexruffell:
[...]

PR #5365, opened 2023-01-26
fix: use 700 series OTW method when already in bootloader mode

AlCalzone:
next fix for https://github.com/zwave-js/node-zwave-js/issues/5360


AlCalzone:
@zwave-js-bot approve

PR #5368, opened 2023-01-27
fix: prevent ongoing transactions from messing with the bootloader

### LINKED ISSUE #5360: Discard all pending transactions when starting OTW update, do additional checks not to communicate with the controller

AlCalzone:
fixes: #5360 

The main issue was that soft-resetting restarts the auto-provisioning mode. Since this request doesn't get an answer, it used to get re-transmitted while in bootloader mode - and worse, while transferring a firmware update.
We now reject all pending transactions twice (once before and once after pausing the send queue) to be extra sure. And we prevent the auto-provisioning when we're about to enter the bootloader.

There's probably a more elegant fix for this, but this would require reworking the state machines again.

### DIFF
--- a/packages/zwave-js/src/lib/driver/Driver.ts
+++ b/packages/zwave-js/src/lib/driver/Driver.ts
@@ -2350,11 +2350,14 @@ export class Driver
 
 		this.isSoftResetting = false;
 
-		// Resume sending
-		this.unpauseSendThread();
-		// Soft-resetting disables any ongoing inclusion, so we need to reset
-		// the state that is tracked in the controller
-		this._controller?.setInclusionState(InclusionState.Idle);
+		// This is a bit hacky, but what the heck...
+		if (!this._enteringBootloader) {
+			// Resume sending
+			this.unpauseSendThread();
+			// Soft-resetting disables any ongoing inclusion, so we need to reset
+			// the state that is tracked in the controller
+			this._controller?.setInclusionState(InclusionState.Idle);
+		}
 	}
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot approve

alexruffell:
@AlCalzone I am afraid the issue happened again (got stuck in bootloader mode when trying to do OTW within Zwave JS UI) with a different USB controller (Zooz ZST10-700). I am trying to restore it using the method you provided but was unable to both within Zwave JS and using PowerShell.

Edit: Without @latest it was using version 10.5.0 and was getting stuck at "flashing". Using the latest flag it used 10.8.0 and it flashed the stick.

PR #5409, opened 2023-02-03
fix: rename `label` property on `Endpoint` class to `endpointLabel`

AlCalzone:
We noticed that the `Node.label` property overrides the endpoint label for the root endpoint with the device label, so we avoid this in a patch before this feature gets widespread use.


AlCalzone:
@zwave-js-bot approve

PR #5416, opened 2023-02-06
feat(compat): add flag to override report timeout for single devices

AlCalzone:
We already have a driver option for this, which is necessary for devices that respond slowly before being identified.
This compat option is the same, but for individual (identified) devices. This should make communication with slow devices more reliable without introducing unnecessary delays globally.


AlCalzone:
@zwave-js-bot approve

PR #5417, opened 2023-02-06
fix(config): disable Supervision and increase report timeout for Yale Smart Door Lock

AlCalzone:


AlCalzone:
@zwave-js-bot approve

PR #5420, opened 2023-02-06
feat: create values for compat-mapped notifications

### LINKED ISSUE #2890: Create metadata for mapped notification values during the interview
It looks like these mappings don't cause values to be created until the V1 alarm is actually reported by the node. It's a little bit of a challenge to generate real smoke and CO alarms. Is it possible to create the values from the mappings themselves ahead of time?

_Originally posted by @kpine in https://github.com/zwave-js/node-zwave-js/issues/2888#issuecomment-864665821_

AlCalzone:
Notification values mapped through compat flags were previously "invisible" until the corresponding notification actually came in. With this change, those values are now created during the interview like for actual v2+ notification values:

grafik

In this example, the Lock state is mapped.

fixes: #2890

### DIFF
--- a/packages/cc/src/cc/NotificationCC.ts
+++ b/packages/cc/src/cc/NotificationCC.ts
@@ -612,6 +612,54 @@ export class NotificationCC extends CommandClass {
 			this.ensureMetadata(applHost, NotificationCCValues.alarmLevel);
 		}
 
+		// Also create metadata for values mapped through compat config
+		const mappings = applHost.getDeviceConfig?.(this.nodeId as number)
+			?.compat?.alarmMapping;
+		if (mappings) {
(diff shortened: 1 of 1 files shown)


webmogul1:
Do you know by any chance what those alarm type and alarm level mean?
Ring Retrofit has a pair for each zone (endpoint) but values never change.  I can see there are updates when sensor status changes but values are always the same.

image

AlCalzone:
The device is likely using the standardized notifications and does not use the alarm values, thus they are always 0. We don't hide them because there's no way of knowing automatically if they are used.

You can just ignore them.

webmogul1:
Ok. Thanks!!!!!!

PR #5437, opened 2023-02-08
feat: add new SerialAPI Setup commands

AlCalzone:
These were found in the Gecko SDK sources but are not documented yet. The controller methods now switch between the 8 and 16 bit variants of the get/set powerlevel methods depending on support.


AlCalzone:
@zwave-js-bot approve

PR #5455, opened 2023-02-11
fix: move serialport mock to `/mock` subpath export

AlCalzone:
Somehow this slipped through the tests. The serialport mocks should not end up in the "main" exports, because they use dev-dependencies.


AlCalzone:
@zwave-js-bot approve

PR #5488, opened 2023-02-22
fix(config): update Dome Wireless Siren

rmcginty:
Fixed some case inconsistencies and added options for strobe per device documentation.  Tested on the unit I have and it does indeed change whether the LED ring flashes or not.

### DIFF
--- a/packages/config/config/devices/0x021f/dms01.json
+++ b/packages/config/config/devices/0x021f/dms01.json
@@ -223,48 +223,24 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Primary Notification",
+					"label": "Primary notification",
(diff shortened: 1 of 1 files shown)


AlCalzone (approved):
OK with my changes

rmcginty:
> OK with my changes

Nice! I didn't know about templates. I'll remember that for the future. Thanks!

AlCalzone:
If you want to read a bit: https://zwave-js.github.io/node-zwave-js/#/config-files/using-templates

rmcginty:
> If you want to read a bit: https://zwave-js.github.io/node-zwave-js/#/config-files/using-templates

Awesome! Definitely will read up on it.

PR #5552, opened 2023-03-09
fix: assign return routes after adding associations

### LINKED ISSUE #5158: [feat] Assign return route during device association
I have struggled to create associations between my permanently powered devices (e.g., wall-mounted scene controllers) and my battery-powered devices (e.g., Springs Window Fashions battery powered cellular shades).  The association only ever seemed to work when the battery-powered device was already "awake" (e.g., in motion and constantly sending updated position reports, or in inclusion mode).  The secondary controllers that came with the shades functioned based on group association, but must have some secret sauce for "waking up" the shades when they send a command.

I discovered this post: https://drzwave.blog/2021/02/01/controlling-flirs-via-associations/comment-page-1/?unapproved=24900&moderation-hash=f64b22b07106e08d9a6d3af673985532#comment-24900 on the Dr ZWave bog about creating return routes for Frequently Listening Routing Slaves (FLiRS), which it appears the battery-powered shades are.  When I create an association using the Windows-based ZWave PC Controller, I can specify to assign a "return route" (apparently the equivalent of the SerialAPI command ZW_AssignReturnRoute( SourceNodeID, DestNodeID, callback) which is SerialAPI command 0x46 (see section 4.4.4 of [INS13954](https://www.silabs.com/documents/public/user-guides/INS13954-Instruction-Z-Wave-500-Series-Appl-Programmers-Guide-v6_81_0x.pdf)).)  Now my wall-mounted switch acts as a secondary controller for the shades using direct group association, and no problem with the device sleeping through the commands.

It would be great if this were a feature that could be implemented in to the ZWaveJS UI, so I can avoid the use of a Windows laptop and Zwave PC Controller to create this kind of association to a battery-powered FLiRS

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/issues/5158

### DIFF
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -4335,7 +4335,7 @@ ${associatedNodes.join(", ")}`,
 	/**
 	 * Adds associations to a node or endpoint
 	 */
-	public addAssociations(
+	public async addAssociations(
 		source: AssociationAddress,
 		group: number,
 		destinations: AssociationAddress[],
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #5555, opened 2023-03-10
fix: don't create values for unsupported Door Lock CC features

### LINKED ISSUE #5554: Door Lock ...Status values should only be included in `getDefinedValueIDs` if they are supported

AlCalzone:
fixes: #5554


AlCalzone:
@zwave-js-bot automerge

PR #5560, opened 2023-03-10
feat: automatically refresh some CC values regularly/on wakeup

### LINKED ISSUE #2589: Automatic Battery CC refresh for non-listening devices

### LINKED ISSUE #5558: Schedule automatic CC value refreshes
**Meter CC:**
Unless unsolicited Meter Report Commands are received, a controlling node MUST:
• Probe the current meter readings for each supported scale (and rate type if v4 nodes) at least
every 6 hours for listening nodes.
• Probe the current meter readings for each supported scale (and rate type if v4 nodes) when the
supporting node issues a Wake Up Notification Command for sleeping nodes.

**Multilevel Sensor CC:**
Unless unsolicited Multilevel Sensor Report Commands are received, a controlling node MUST:
• Probe the current reading for each supported sensor type at least every 6 hours for listening
nodes.
• Probe the current reading for each supported sensor type when the supporting node issues a
Wake Up Notification Command for sleeping nodes.

**Notification CC:**
A controlling node MUST probe state/event notifications from pull nodes
• at least every 6 hours for listenting nodes.
• when the supportig node issues a Wake up Notification Command for sleeping nodes.

**Battery CC:**
Unless unsolicited Battery Report Commands are received, a controlling node SHOULD Probe the
current battery level at least every month

AlCalzone:
With this PR, the following CCs have their values queried regularly if some of them haven't been updated recently:
* Battery CC (7 days)
* Multilevel Sensor CC (6 hours)
* Meter CC (6 hours)
* Notification CC (pull-mode nodes only, 6 hours)

To distribute the potential queries a bit, we first check if a refresh is necessary is checked every hour ± 10 minutes for listening and FLiRS nodes, randomized when the node is ready. For sleeping nodes we check on every wake up. Only if the values haven't been updated, e.g. through an unsolicited report, they will be refreshed.

fixes: #5558
fixes: #2589

### DIFF
--- a/packages/cc/api.md
+++ b/packages/cc/api.md
@@ -1725,6 +1725,8 @@ export class BatteryCC extends CommandClass {
     interview(applHost: ZWaveApplicationHost_2): Promise<void>;
     // (undocumented)
     refreshValues(applHost: ZWaveApplicationHost_2): Promise<void>;
+    // (undocumented)
+    shouldRefreshValues(this: SinglecastCC_2<this>, applHost: ZWaveApplicationHost_2): boolean;
 }
 
 // Warning: (ae-missing-release-tag) "BatteryCCGet" is exported by the package, but it is missing a release tag (@alpha, @beta, @public, or @internal)
@@ -1800,7 +1802,7 @@ export const BatteryCCValues: Readonly<{
             readonly endpoint: number;
             readonly property: "lowTemperatureStatus";
         };
-        readonly is: (valueId: ValueID_2) => boolean;
+        readonly is: (valueId: ValueID) => boolean;
         readonly meta: {
             readonly label: "Battery temperature is low";
             readonly writeable: false;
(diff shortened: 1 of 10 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #5567, opened 2023-03-14
feat(config): add Vesternet branded Z-Wave devices

martyn-vesternet:
Hope this PR request meets the guidelines, happy to make any changes if required.

Adds support for 10 Vesternet branded Z-Wave devices (9 based on Sunricher hardware, 1 based on Everspring hardware).  

Should cover all applicable Configuration Parameters, Inclusion / Exclusion / Reset instructions as well as having links to PDF manuals. Z-Wave Alliance ID's are included where available.

All have been tested internally and we will react to any user feedback as needed.


packages/config/config/devices/0x0330/VES-ZW-WAL-009.json
AlCalzone, line 1: This device has the same identification as [this one](https://github.com/zwave-js/node-zwave-js/blob/2e6f2ead631ff26d7eb8bd97cbe25d131b83fc0c/packages/config/config/devices/0x0330/heatit_z-push_button_8.json).
Since the other one is a bit messy (Sunricher Heatit ...), I'd prefer to keep the new one here, but we should probably include the factory reset parameter (if it actually exists).

packages/config/config/devices/0x0330/VES-ZW-REM-013.json
AlCalzone, line 1: This seems to be a copy of [this file](https://github.com/zwave-js/node-zwave-js/blob/543e8608554c223d088315279aceaaa22bf7899a/packages/config/config/devices/0x0330/zv9001k12-dim-z4.json).

I suppose we could/should remove the device ID from the other file? Not sure who's the OEM for whom.

packages/config/config/devices/0x0330/VES-ZW-WAL-008.json
AlCalzone, line 1: And this file has the same identification as [this one](https://github.com/zwave-js/node-zwave-js/blob/e086382f7e567154cfb7eb901b32a11bee83005a/packages/config/config/devices/0x0330/zv9001k4-dim-g2.json).

Again not sure who's the manufacturer/OEM here. Your new file seems the most complete though.

packages/config/config/devices/0x0330/VES-ZW-HLD-016.json
```diff
@@ -0,0 +1,127 @@
...
+			"#": "3",
+			"label": "Send Basic Report on State Change",
+			"$import": "~/templates/master_template.json#base_enable_disable",
+			"defaultValue": 1
+		},
+		{
+			"#": "4",
+			"label": "Over Current Protection",
+			"description": "Triggered at over 16A.",
+			"$import": "~/templates/master_template.json#base_enable_disable",
```
AlCalzone: `$import` should come immediately after `#`. Please apply everywhere.

packages/config/config/devices/0x0330/VES-ZW-HLD-016.json
```diff
@@ -0,0 +1,127 @@
...
+				},
+				{
+					"label": "Toggle",
+					"value": 1
+				}
+			]
+		},
+		{
+			"#": "10",
+			"label": "Power Report Absolute Threshold",
```
AlCalzone: We typically call this `Power Change Report Threshold`. Please apply to the other parameters similarly.

packages/config/config/devices/0x0330/VES-ZW-HLD-016.json
```diff
@@ -0,0 +1,127 @@
...
+			"description": "Report when power changes by W value, unit is 1W.",
```
AlCalzone: This description seems unnecessary, also we don't want to repeat the unit which is already available via the `unit` property. Please apply throughout.

packages/config/config/devices/0x0330/VES-ZW-HLD-016.json
```diff
@@ -0,0 +1,127 @@
...
+			"options": [
+				{
+					"label": "Disable",
+					"value": 0
+				}
+			]
+		},
+		{
+			"#": "13",
+			"label": "Report Interval",
```
AlCalzone:
```suggestion
			"label": "Energy/Power Report Interval",
```
and remove the description please.

packages/config/config/devices/0x0330/VES-ZW-MOT-018.json
```diff
@@ -0,0 +1,248 @@
...
+				{
+					"label": "Disable",
+					"value": 0
+				}
+			]
+		},
+		{
+			"#": "3",
+			"label": "Working Mode",
+			"description": "Change mode of device",
```
AlCalzone:
```suggestion
```
unnecessary description

packages/config/config/devices/0x0330/VES-ZW-MOT-018.json
```diff
@@ -0,0 +1,248 @@
...
+			"#": "5",
+			"label": "Over Current Protection",
+			"description": "Triggered when module goes over current limit.",
+			"$import": "~/templates/master_template.json#base_enable_disable",
+			"defaultValue": 1
+		},
+		{
+			"#": "6",
+			"label": "Start Positioning Calibration",
+			"description": "ONLY VALID IN SHUTTER WITH POSITIONING MODE. Enable to begin, will revert to Disabled once completed.",
```
AlCalzone:
```suggestion
			"description": "Only valid in Shutter Mode with positioning. Set to 'Calibrate' to begin, will revert to 'Idle' once completed.",
```

packages/config/config/devices/0x0330/VES-ZW-MOT-018.json
```diff
@@ -0,0 +1,248 @@
...
+			"options": [
+				{
+					"label": "Not Calibrated",
+					"value": 0
+				},
+				{
+					"label": "Calibrated",
+					"value": 1
+				}
+			]
```
AlCalzone: I suppose this would make more sense given the above description
```suggestion
			"options": [
				{
					"label": "Idle / Not calibrated",
					"value": 0
				},
				{
					"label": "Calibrate",
					"value": 1
				}
			]
```

packages/config/config/devices/0x0330/VES-ZW-MOT-018.json
```diff
@@ -0,0 +1,248 @@
...
+		},
+		{
+			"#": "7",
+			"label": "Configuration Of Momentary Switch Inputs",
+			"description": "ONLY VALID IN LIGHT MODE. Switch inputs accept a variety of push combinations (press, hold & released), configure how the outputs react to them here.",
```
AlCalzone:
```suggestion
			"label": "Momentary Switch Input Behavior",
			"description": "Only valid in Light Mode.",
```

packages/config/config/devices/0x0330/VES-ZW-MOT-018.json
```diff
@@ -0,0 +1,248 @@
...
+					"value": 2
+				},
+				{
+					"label": "Short Press Only, Toggle Both Outputs",
+					"value": 3
+				},
+				{
+					"label": "Hold & Release Only, Toggle Both Outputs",
+					"value": 4
+				}
```
AlCalzone: Option labels should be in Sentence case, not Title Case.

packages/config/config/devices/0x0330/VES-ZW-MOT-018.json
```diff
@@ -0,0 +1,248 @@
...
+		{
+			"#": "8",
+			"label": "Enable Inclusion / Exclusion Using External Switch On Input S1",
+			"$import": "~/templates/master_template.json#base_enable_disable",
+			"defaultValue": 1
+		},
+		{
+			"#": "9",
+			"label": "Enable / Disable Sending Of Central Scene Commands In Response To Inputs",
+			"description": "Switch inputs accept a variety of push combinations (press, hold & released), configure if Central Scene Commands are sent in response to them.",
```
AlCalzone:
```suggestion
```
unnecessary description

packages/config/config/devices/0x0330/VES-ZW-MOT-018.json
```diff
@@ -0,0 +1,248 @@
...
+				{
+					"label": "Enable For S2 Input Only",
+					"value": 3
+				}
+			]
+		},
+		{
+			"#": "10",
+			"label": "Power Report Absolute Threshold",
+			"description": "Report when power changes by W value, unit is 1W.",
```
AlCalzone: See comments above regarding these parameters

packages/config/config/devices/0x0330/VES-ZW-SOC-28.json
```diff
@@ -0,0 +1,78 @@
...
+			"options": [
+				{
+					"label": "Disabled (socket stays off)",
+					"value": 0
+				},
+				{
+					"label": "Enabled (socket resumes previous state)",
+					"value": 1
+				}
+			]
```
AlCalzone: We have a template for this
```suggestion
			"$import": "~/templates/master_template.json#state_after_power_failure_off_prev"
```

packages/config/config/devices/0x0330/VES-ZW-SOC-28.json
```diff
@@ -0,0 +1,78 @@
...
+		},
+		{
+			"#": "2",
+			"label": "LED Function",
+			"description": "LED turns ON when socket is ON",
```
AlCalzone:
```suggestion
			"label": "LED Indicator",
```

packages/config/config/devices/0x0330/VES-ZW-SWI-014.json
```diff
@@ -0,0 +1,173 @@
...
+			"#": "5",
+			"label": "Over Current Protection",
+			"description": "Triggered when module goes over current limit.",
+			"$import": "~/templates/master_template.json#base_enable_disable",
+			"defaultValue": 1
+		},
+		{
+			"#": "7",
+			"label": "Configuration Of Momentary Switch Inputs",
+			"description": "Switch inputs accept a variety of push combinations (press, hold & released), configure how the outputs react to them here.",
```
AlCalzone: See similar comments in VES-ZW-MOT-18 file

AlCalzone (changes requested):
Thanks a lot for this. I'd like to see a few changes for consistency, and we need to figure out what to do with the 3 files that have duplicated identifiers.

packages/config/config/devices/0x0330/VES-ZW-WAL-009.json
martyn-vesternet: As far as I'm aware it doesn't support a factory reset parameter ..... it doesn't really make sense for a battery powered device as you'd have to wake it up for it to accept the parameter!

packages/config/config/devices/0x0330/VES-ZW-REM-013.json
martyn-vesternet: Sunricher are the manufacturer, Vesternet (and others) rebrand the devices.  I'm not sure if Z-Wave JS has the ability to specifiy something like "white label" along the lines of how Zigbee2MQTT does it?

packages/config/config/devices/0x0330/VES-ZW-WAL-008.json
martyn-vesternet: I think with some of the existing files they're a bit sparse on the details, or have taken the literal naming off the default Sunricher specs which aren't really clear.  "2 group single color wall mounted remote" doesn't really describe the device accurately IMO, we call it a "4 Button Wall Controller" as it seems to better reflect its Central Scene functionality.

packages/config/config/devices/0x0330/VES-ZW-WAL-009.json
AlCalzone: Got it.

packages/config/config/devices/0x0330/VES-ZW-REM-013.json
AlCalzone: The field in the config files is called `manufacturer` for legacy reasons, but we actually use it as the brand. To distinguish multiple re-branded devices, we need some way to automatically identify them, which can work if everyone uses a different combination of manufacturer ID, product ID and product type. If that's not the case, I suppose the "safe" way is to keep the original manufacturer.

packages/config/config/devices/0x0330/VES-ZW-WAL-008.json
AlCalzone: I like that. We need to decide on the brand/manufacturer name though, if this device is sold under multiple names without changing IDs.

packages/config/config/devices/0x0330/VES-ZW-REM-013.json
martyn-vesternet: Yeah it's a tough call, those that rebrand typically aren't Z-Wave Alliance registered (so they won't have their own manufacturer ID).  From our point of view we've used our name, label and description because that's what's physically on our products as well as in the manuals.  

If we keep the original (for example Sunricher ZV9001K12-DIM-Z 4 Channel Dimmer Remote Control) then we know from experience on other platforms that customers get confused because the details don't match when they install our devices.

It's also compounded a little because some of the devices haven't yet made their way into the Z-Wave Alliance database, so official manufacturer label / description and manual aren't readily available.

We're also conscious that some of these rebranded devices aren't actually the same even though they purport to be based on the manufacturerID, productType and productID. 

For example the Vesternet dimmer module is Sunricher product SR-ZV9100A-S but there are other variants SR-ZV9100A-A, SR-ZV9100A-B, etc.  They all report the same manufacturerID, productType and productID though, but we can only vouch for the features and functionality in our version.  It's possible that the others might be different - that HeatIt battery switch above comes to mind, does it actually have that reset Parameter?  Our version definitely doesn't .....

With Zigbee2MQTT it's a little easier to deal with re-branded devices.  They have a "whiteLabel" field that helps:

fingerprint: [{modelID: 'HK-SL-DIM-A', softwareBuildID: '2.5.3_r52'}], <------ What's actually in the device
model: 'VES-ZB-DIM-004', <------ Vesternet product code
vendor: 'Vesternet', <------ Vesternet 
description: 'Zigbee dimmer', <------ Vesternet product description
whiteLabel: [{vendor: 'Sunricher', model: 'SR-ZG9040A'}], <------ manufacturer and manufacturer product code

There's still some element of "first come first served" though, because it will still match any device with that modelID (although it's further constrained to the software build which helps). 

I think perhaps a wider discussion on how to handle this situation might be needed for the future?  On Hubitat they've started using a Wizard to choose what device you're actually installing and on Homey they use Apps so the user chooses the App to install the device under.  Perhaps in the future for Z-Wave JS it might be possible to offer a selection of matching device configurations to choose from in the event that there are multiple matches.

I guess the alternative is to leave the existing config files as-is (the other re-branders were "first", so that's fair), we'll be adding all our config files to our Github repo anyway, so customers can always override the existing ones if they find that they don't quite work fully / correctly.

packages/config/config/devices/0x0330/VES-ZW-HLD-016.json
martyn-vesternet: Should this be "Energy/Power/Voltage/Current Report Interval" since it covers all of them?  Just conscious that removing the description makes it less clear what Parameter covers?

packages/config/config/devices/0x0330/VES-ZW-HLD-016.json
martyn-vesternet: Should the unit just contain W / V / C or should (can) it contain the increment?

I'd added the description to make it clear what the increment is, so for V and W it's "1" but for Current it's "0.1"

packages/config/config/devices/0x0330/VES-ZW-MOT-018.json
martyn-vesternet: This one's a bit tricky, you can only write value 1 to the Parameter, to start the calibration.  

But it reads back as either 0 (not yet calibrated, or failed to calibrate) or 1 (calibrated).

Not sure how best to handle that since 1 can mean two different things, start or completed?

Maybe:

```
"description":  "Only valid in Shutter Mode with positioning. Set to 'Calibrate / Calibrated' to begin, will show 'Not Calibrated' or 'Calibrate / Calibrated' once complete."

		"options": [
			{
				"label": "Not Calibrated",
				"value": 0
			},
			{
				"label": "Calibrate / Calibrated",
				"value": 1
			}
		]
```

packages/config/config/devices/0x0330/VES-ZW-HLD-016.json
martyn-vesternet: Is "Power Change Report Absolute Threshold" OK to differentiate from "Power Change Report Percentage Threshold"?  

Or should it just be "Power Change Report Threshold" and the difference inferred from the unit "W" vs "%" ?

packages/config/config/devices/0x0330/VES-ZW-REM-013.json
AlCalzone: Are you running a custom firmware on your rebranded devices? We can use the firmware version to identify specific devices. Or if there is any other field, it might be worth adding as another way of identification.

packages/config/config/devices/0x0330/VES-ZW-HLD-016.json
AlCalzone: If there are both settings in a single file, I'd say "Power Report: Change Threshold" for "W" and "Power Report: Relative Change Threshold" for "%".
If there's only one, I think the variant from above is enough.

packages/config/config/devices/0x0330/VES-ZW-HLD-016.json
AlCalzone: You can contain the multiplier in the unit.

packages/config/config/devices/0x0330/VES-ZW-HLD-016.json
AlCalzone: I'd be fine with that. We just try to keep labels short and typically devices report all they can on an interval.

packages/config/config/devices/0x0330/VES-ZW-REM-013.json
martyn-vesternet: No, it's whatever they come with from Sunricher unfortunately, they do sometimes change over time with each new shipment which makes it hard to keep up with version changes :-( 

For the duplicates, I can add the exact firmware version for each device that I've been testing with, but I've had mine for a while now so they might not match what's currently shipping. I can add more in the future if customers report that their devices don't match correctly.

Is that done in the "firmwareVersion" field?, I see there's a min / max, how would I explicitly match multiple versions like:

1.6
2.2.5
2.2.6

packages/config/config/devices/0x0330/VES-ZW-HLD-016.json
martyn-vesternet: Cool thanks, the HLD-016 only has W but the DIM-001 has both so I'll apply that there.

packages/config/config/devices/0x0330/VES-ZW-HLD-016.json
martyn-vesternet: OK, will adjust as needed :-)

packages/config/config/devices/0x0330/VES-ZW-REM-013.json
AlCalzone: > would I explicitly match multiple versions

Currently not possible. A decent approximation would be a range from 1.6 to 2.2.6 and hope that it works.

This doesn't solve the issue of detecting that a device is a Vesternet brand though, since others could have gotten the same firmware version, right?

martyn-vesternet: Yes, I don't believe that the firmware version is unique so it would only be a way to match explicitly with the devices that we know are ours.

packages/config/config/devices/0x0330/VES-ZW-MOT-018.json
martyn-vesternet: I noticed the availability of conditional parameters - https://github.com/zwave-js/node-zwave-js/blob/master/docs/config-files/conditional-settings.md

I don't suppose it's possible to use the value of another parameter as a condition to change / enable / hide other parameters?

So for example with this calibration parameter, if the current value is 0 then it could show the label for the 2nd option as "start calibration" and if the current value is 1 then it could show it as "calibrated".

It would also be useful to be able to hide parameters that are only applicable when the device is configured in a specific mode, e.g. those that say "only valid in light mode" to only be shown when that's true.

packages/config/config/devices/0x0330/VES-ZW-REM-013.json
AlCalzone: What a mess... Let's try that though - better than the status quo.

packages/config/config/devices/0x0330/VES-ZW-MOT-018.json
AlCalzone: Sadly we can't correlate config parameters in that way yet. I've opened an issue for that, we have some other use case too: https://github.com/zwave-js/node-zwave-js/issues/5583

About the option values: let's take your suggestion. I might have misunderstood the original description.

packages/config/config/devices/0x0330/VES-ZW-REM-013.json
martyn-vesternet: OK will do, I just committed most of the other changes earlier on.  I'll tweak the firmware field later on and resubmit.

packages/config/config/devices/0x0330/VES-ZW-MOT-018.json
martyn-vesternet: Ah yes, I was trying to rack my brain to remember what other devices had similar dynamic parameters.  I think some Qubino and Secure (Horstmann) do.

packages/config/config/devices/0x0330/VES-ZW-REM-013.json
martyn-vesternet: Latest commit should hopefully cover everything discussed!

AlCalzone:
Apologies for the delay - I had some surprise vacation days I had to take before end of march. Will look at this now.

AlCalzone (approved):

PR #5570, opened 2023-03-14
feat: include reason for low security in `"node added"` parameter

### LINKED ISSUE #3233: Include reason for failed bootstrapping in `"node added"` parameter

AlCalzone:
For users it is currently pretty opaque why a node was not added with Security S0/S2. With this PR, we surface a multitude of reasons as part of the `"node added"` event, so applications can give some more details on why the secure inclusion failed.

The `InclusionResult` contained in the 2nd event parameter
```ts
// "node added"
(node: ZWaveNode, result: InclusionResult) => void
```
now has a property `lowSecurityReason` when `lowSecurity` is true. This is an enum value with the following possibilities:
```ts
enum SecurityBootstrapFailure {
	/** Security bootstrapping was canceled by the user */
	UserCanceled,
	/** The required security keys were not configured in the driver */
	NoKeysConfigured,
	/** No Security S2 user callbacks (or provisioning info) were provided to grant security classes and/or validate the DSK. */
	S2NoUserCallbacks,
	/** An expected message was not received within the corresponding timeout */
	Timeout,
	/** There was no possible match in encryption parameters between the controller and the node */
	ParameterMismatch,
	/** Security bootstrapping was canceled by the included node */
	NodeCanceled,
	/** The PIN was incorrect, so the included node could not decode the key exchange commands */
	S2IncorrectPIN,
	/** There was a mismatch in security keys between the controller and the node */
	S2WrongSecurityLevel,
	/** Some other unspecified error happened */
	Unknown,
}
```

fixes: #3233

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -1183,9 +1183,42 @@ The second argument gives additional info about the inclusion result.
 <!-- #import InclusionResult from "zwave-js" -->
 
 ```ts
-interface InclusionResult {
-	/** This flag warns that a node was included with a lower than intended security, meaning unencrypted when it should have been included with Security S0/S2 */
-	lowSecurity?: boolean;
+type InclusionResult =
+	| {
+			/** This flag warns that a node was included with a lower than intended security, meaning unencrypted when it should have been included with Security S0/S2 */
+			lowSecurity?: false;
+	  }
+	| {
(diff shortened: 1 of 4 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #5574, opened 2023-03-14
fix: use stricter checks when receiving a `DeviceResetLocallyNotification`

### LINKED ISSUE #5509: Device Reset Locally Notification: Don't accept longer payloads than expected
ref: https://github.com/home-assistant/core/issues/77966

If a Central Scene CC gets bit-flipped to a Device Reset Locally Notification, that's not good.

AlCalzone:
We now discard this CC when it has any payload at all, because this means it was "created" by a bit flip.

fixes: #5509


AlCalzone:
@zwave-js-bot automerge

PR #5581, opened 2023-03-15
fix: value timestamps are no longer updated for optimistic value updates

AlCalzone:
This PR fixes the edge case of optimistic value updates mentioned in https://github.com/zwave-js/node-zwave-js/pull/5556

We now only update the timestamp of a value when
- the node sends an unsolicited update
- the node responds to GET-type requests
- a supervised SET-type call was successful


AlCalzone:
@zwave-js-bot automerge

PR #5585, opened 2023-03-16
fix: during interview of `Schedule Entry Lock CC`, set timezone if necessary and supported

### LINKED ISSUE #5505: Schedule Entry Lock CC: Timezone offset queried but not saved or exposed as value

### LINKED ISSUE #5562: Schedule Entry Lock CC: Set timezone during interview if Time CC V2 is not supported

AlCalzone:
... instead of simply querying and forgetting it

fixes: #5562

This also...
closes: https://github.com/zwave-js/node-zwave-js/issues/5505
... because it makes more sense to just configure it IMO instead of exposing a value that's hard to interact with.

### DIFF
--- a/packages/cc/src/cc/ScheduleEntryLockCC.ts
+++ b/packages/cc/src/cc/ScheduleEntryLockCC.ts
@@ -1,6 +1,7 @@
 import {
 	CommandClasses,
 	encodeBitMask,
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #5590, opened 2023-03-18
feat(config): new configuration for ad370 device

kimjohnsson:


kimjohnsson:
The z-wave alliance doesn't specify a Product Type ID for this device.. How do I handle that? I put the `productType` same as `productId` but not sure this is right...

AlCalzone:
If you use Z-Wave JS UI, it should show you the full ID when you open the node "dropdown".

packages/config/config/devices/0x0060/ad370.json
```diff
@@ -0,0 +1,209 @@
+{
+	"manufacturer": "Everspring",
+	"manufacturerId": "0x0060",
+	"label": "AD370",
+	"description": "Z-Wave 700 MOS Dimmer",
+	"devices": [
+		{
+			"productType": "0x0033",
+			"productId": "0x0033",
+			"zwaveAllianceId": [4514, 4648]
```
AlCalzone, line 10: According to the ZWA website, these are two different manufacturers (Everspring and NEXA). Do you know if they use the same IDs? Often those re-branded devices end up with a different manufacturer ID or so.

packages/config/config/devices/0x0060/ad370.json
```diff
@@ -0,0 +1,209 @@
...
+	],
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"paramInformation": [
+		{
+			"#": "1",
+			"label": "Switch Type",
+			"description": "Def 0",
```
AlCalzone, line 21: Unnnecessary description.

packages/config/config/devices/0x0060/ad370.json
```diff
@@ -0,0 +1,209 @@
...
+			"valueSize": 2,
+			"minValue": 0,
+			"maxValue": 1,
+			"defaultValue": 0,
+			"unsigned": true
+		},
+		{
+			"#": "2",
+			"label": "Auto-on time",
+			"description": "Def 0, second",
```
AlCalzone, line 31: Unnnecessary description. Units should be provided via the `unit` property.

packages/config/config/devices/0x0060/ad370.json
```diff
@@ -0,0 +1,209 @@
...
+			"valueSize": 2,
+			"minValue": 1,
+			"maxValue": 100,
+			"defaultValue": 10,
+			"unsigned": true
+		},
+		{
+			"#": "9",
+			"label": "MOS operation mode",
+			"description": "Def 2",
```
AlCalzone, line 101: Is there an explanation for the values? Those should probably be `options`.

packages/config/config/devices/0x0060/ad370.json
```diff
@@ -0,0 +1,209 @@
...
+			"valueSize": 2,
+			"minValue": 0,
+			"maxValue": 2,
+			"defaultValue": 2,
+			"unsigned": true
+		},
+		{
+			"#": "16",
+			"label": "Wire connection type",
+			"description": "0: unknow, 1:2 wire, 2:3 wire",
```
AlCalzone, line 171: This should be converted to options.

AlCalzone (changes requested):
This does not match our style guide, especially regarding labels and descriptions (which we avoid in favor of more useful properties like unit, options, ...).

Please give https://zwave-js.github.io/node-zwave-js/#/config-files/style-guide?id=labels (and the following sections) a good read and apply it throughout.

PR #5475, opened 2023-02-16
feat: support Security S2 Multicast and sending large commands with Transport Service CC

AlCalzone:
This PR implements support for `Security S2` Multicast and the entire machinery that's necessary for it, including the sending half of `Transport Service CC`.

It also implements the wait-for-nonce transmit verification strategy I mentioned in https://github.com/zwave-js/node-zwave-js/issues/5471#issue-1586352002, for cases where we send an S2-encrypted, unsupervised, singlecast SET-type command. This will introduce short delays for nodes that support S2, but have flawed Supervision support that had to be disabled.

Furthermore, `getBroadcastNode` and `getMulticastGroup` have been deprecated. Instead, two new methods were added, which automatically group the nodes by security class:
```ts
getMulticastGroups(nodeIDs: number[]): VirtualNode[]
getBroadcastNodes(): VirtualNode[]
```
This makes it possible to send multicasts to nodes with mixed security classes (1 multicast per security class). Broadcast may desugar to several multicasts, unless only insecure nodes are in the network.
Nodes using S0 are ignored, because they don't support either.

TODO:

- [x] bunch of FIXME comments
- [x] actually sending
- [x] automatically choosing Supervision
- [x] waiting after S2 singlecast without Supervision
- [x] SPAN+MPAN extension requires Transport Service :)

fixes: https://github.com/zwave-js/node-zwave-js/issues/3270
fixes: https://github.com/zwave-js/node-zwave-js/issues/4835
fixes: https://github.com/zwave-js/node-zwave-js/issues/5490
fixes: https://github.com/zwave-js/node-zwave-js/issues/1119

PR #5593, opened 2023-03-20
feat: simplify working with mixed-security multicast

AlCalzone:
It has been found that working with the new methods introduced in https://github.com/zwave-js/node-zwave-js/pull/5475 is really awkward. So this PR deprecates them and reverts the public multicast/broadcast API back to the old one that just returns a single `VirtualNode` for both.

Instead, grouping nodes by security classes and targeting multiple groups at once now happens internally.


AlCalzone:
This is almost ready. Mixed security classes now actually result in multiple multicasts, one per class. All that's left to do is make sure the `SupervisionResult`s get passed back to the caller.

AlCalzone:
@zwave-js-bot automerge

PR #5595, opened 2023-03-23
fix(config): add 0x1351 product ID for MCOHome A8-9

EiNSTeiN-:
I purchased this air quality sensor and it has with a different product ID than any of the ones in the existing configuration file. I added the product Id to the list, that should solve the issue and let me set the configuration.

image

### DIFF
--- a/packages/config/config/devices/0x015f/a8-9.json
+++ b/packages/config/config/devices/0x015f/a8-9.json
@@ -6,8 +6,7 @@
 	"devices": [
 		{
 			"productType": "0xa803",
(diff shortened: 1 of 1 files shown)


EiNSTeiN-:
I have the US version of the A8-9, that might explain the different product ID?

AlCalzone:
Can you sort the entries in that array so the ones with lower product ID come first? So opposite of how it is now.
Other than that lgtm

AlCalzone (approved):

AlCalzone:
Thanks!

PR #5614, opened 2023-04-03
fix: add legacy/simple door sensor value to restore old behavior

AlCalzone:
https://github.com/zwave-js/node-zwave-js/pull/5394 introduced a change for the `Door state` notification variable that was more breaking than anticipated, due to applications automatically generating binary sensors for two-state notification values.

This PR adds a new value `Door state (simple)` which only has two states and behaves just like the now 4-state `Door state` variable did before:

| Value sent from device   | Door state                      | Door state (simple) |
|--------------------------|---------------------------------|---------------------|
| open                     | open (22)                       | open (22)           |
| open in regular position | open in regular position (5632) | open (22)           |
| open in tilt position    | open in tilt position (5633)    | open (22)           |
| closed                   | closed (23)                     | closed (23)         |

### DIFF
--- a/packages/cc/api.md
+++ b/packages/cc/api.md
@@ -11992,6 +11992,43 @@ export const NotificationCCValues: Readonly<{
             readonly supportsEndpoints: false;
         };
     };
+    doorStateSimple: {
+        readonly id: {
+            commandClass: CommandClasses.Notification;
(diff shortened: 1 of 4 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #5617, opened 2023-04-03
fix(lint): ensure that firmware version ranges of files do not overlap

AlCalzone:
Turns out our checks for duplicate device files only accounted for identical firmware ranges, not overlapping ones. Now we do


AlCalzone:
@zwave-js-bot automerge

PR #5619, opened 2023-04-03
feat: add support for preferred configs with overlapping version ranges

AlCalzone:
This is mainly meant to support re-branded devices with narrow version ranges more easily, without having to split the whitelabel configs into multiple parts.
Required for https://github.com/zwave-js/node-zwave-js/pull/5567/


AlCalzone:
@zwave-js-bot automerge

PR #5627, opened 2023-04-04
fix: assume Notification CC is push mode, unless proven otherwise

### LINKED ISSUE #5626: Issue with Zipato mini keypad rfdi (wintop)
Hi,

After upgrading from zwavejs2mqtt 1.7.2 to 1.8.0 / 1.8.1 and 1.9.0

my 2 wintop keypads are acting up.

in the old 1.7.2, when entering og leaving, press home or away and hold up the rfdi tag.

in 1.8.x it's the same, BUT under 1.8.X it's like the wintop keypads wakes up, and resends the last action.

so when leaving the home, we are the alarm, but if the second keypad was last used to disarm, the the alarm are turned off and on durring the day

then event in HA (listen to events)  

there is no difference the notification in HA of a real event or the "resend"

The "resend" happens with the "wakeup", if that helps.

This was NOT an issue before 1.8.x and I had it running for at least 2.5 years

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/issues/5626

### DIFF
--- a/packages/cc/src/cc/NotificationCC.ts
+++ b/packages/cc/src/cc/NotificationCC.ts
@@ -489,8 +489,11 @@ export class NotificationCC extends CommandClass {
 				/* ignore */
 			}
 		}
-		// If everything failed, fall back to "pull"
-		return "pull";
+		// If everything failed, e.g. because the node is V1/V2, assume this is a "push" node.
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #5632, opened 2023-04-05
fix: correct low battery alarm mapping for Yale and Kwikset locks

AlCalzone:
As a followup to https://github.com/zwave-js/node-zwave-js/pull/5630 and as described in https://github.com/zwave-js/node-zwave-js/issues/1543#issuecomment-1497218026,
the alarm-to-battery mapping in these locks wasn't correct.

While the notification variables are called "Replace battery soon/now", they belong to the "Battery maintenance status", which is not meant for indicating a low battery but rather a rechargeable battery that is near its end of life.

This PR corrects this error, which further allows using Battery reports to idle these values. (followup PR)

### DIFF
--- a/packages/config/config/devices/0x0090/910.json
+++ b/packages/config/config/devices/0x0090/910.json
@@ -228,6 +228,9 @@
 			},
 			{
 				"$import": "templates/kwikset_template.json#alarm_map_low_battery"
+			},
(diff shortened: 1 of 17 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #5633, opened 2023-04-05
feat: idle Battery level status notifications using Battery CC Reports

AlCalzone:
As a followup to https://github.com/zwave-js/node-zwave-js/pull/5632 and as mentioned in https://github.com/zwave-js/node-zwave-js/issues/1543#issuecomment-1497218026, this PR adds a heuristic to the Battery CC that will reset the corresponding Notification CC values for "low battery" to idle if the battery is near full charge.


AlCalzone:
@zwave-js-bot automerge

PR #5634, opened 2023-04-05
feat: expose method on Node class to manually idle notifications, do not auto-idle V1-V7

AlCalzone:
This PR adds a new method `manuallyIdleNotificationValue` on the `ZWaveNode` class which allows manually idling "stuck" notifications.

To better align with the specifications, we now limit the timed auto-idling of notification values to V1-V7 of the Notification CC.


AlCalzone:
@zwave-js-bot automerge

packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -3460,7 +3513,9 @@ protocol version:      ${this.protocolVersion}`;
 				// We don't know what this notification refers to, so we don't force a reset
 				allowIdleReset = false;
 			} else if (valueConfig.type === "state") {
-				allowIdleReset = valueConfig.idle;
+				// CL:0071.01.52.07.2: A controlling node SHOULD NOT use timeouts to consider a state variable
+				// to be idle [...] for v8 or newer supporting nodes.
+				allowIdleReset = command.version >= 8 && valueConfig.idle;
```
OXINARF, line 3518: According to the comment, shouldn't it be `command.version < 8`, only allowing timeout reset when it is v1-v7? You might also want to change the comment where this variable is used:
`// Nodes before V8 (and some misbehaving V8 ones) don't necessarily reset the notification to idle.`

OXINARF (commented):
I'm confused on what was intended here:

- the title of PR says "_do not auto-idle V1-V7_"
- but the description says "_To better align with the specifications, we now limit the timed auto-idling of notification values to V1-V7 of the Notification CC._"

And then in the code, as pointed out by my comment below, it also seems that it doesn't reflect the comment.

AlCalzone:
Title should have been "only auto idle v1-v7"

OXINARF:
Then it seems to me that code is wrong.

AlCalzone:
🙈 agreed

AlCalzone:
I've reverted that part of the PR. We already only did it on an opt-in basis, and this needs to be working for v8+ devices too.

PR #5639, opened 2023-04-07
fix: revert change to notification auto-idle

AlCalzone:
PR #5634 introduced changes to auto-idling which were the opposite of the intended change. On top of that, the intended change would clash with the fact that we need to allow auto-idling notification values for misbehaving v8 devices.

So this PR reverts those changes. Auto-idling is enabled for all devices with the corresponding compat flag set.


AlCalzone:
@zwave-js-bot automerge

PR #5647, opened 2023-04-11
fix: idle "Keypad state" notifications on keypad events

AlCalzone:
as discussed in https://github.com/zwave-js/node-zwave-js/issues/1543#issuecomment-1498247929
When an event is received that indicates the keypad was just used, we should assume the keypad state has returned to idle (not disabled, not busy).


AlCalzone:
@zwave-js-bot automerge

PR #5655, opened 2023-04-12
test: move mock responses for Notification CC into testing lib

AlCalzone:
I found those to be repeated a bit too often in test files. Also there's now a framework for defining CC-specific mock capabilities, which can then be used by the mock implementations.


AlCalzone:
@zwave-js-bot automerge

PR #5657, opened 2023-04-13
fix(config): correct link to manual for Aeotec ZWA011

fclinton:
Manual was a dead link, so updated to correct link


AlCalzone (approved):

AlCalzone:
thanks!

Issue #5660, opened 2023-04-13
 Z-WAVE-SERVER: Cannot send this message because it would exceed the maximum payload length! (ZW0215) 


thecobra666:
### Is your problem within Home Assistant (Core or Z-Wave JS Integration)?

YES, BUT a Home Assistant developer has told me to come here

### Is your problem within Z-Wave JS UI (formerly ZwaveJS2MQTT)?

NO, my problem is NOT within Z-Wave JS UI

### Checklist

- [X] I have checked the [troubleshooting section](https://zwave-js.github.io/node-zwave-js/#/troubleshooting/index) and my problem is **not** described there.

- [X] I have read the [changelog](https://github.com/zwave-js/node-zwave-js/blob/master/CHANGELOG.md) and my problem was **not** mentioned there.

### Describe the bug

What causes the bug?
Call a scene to set a light to a color and brightness level, when the light is already on.

What do you observe?
Says error ZW0215. 

What did you expect to happen?
The light to change color.

Set a scene in HA, call it = ok
Call another scene = nok error.
Turn off the light
turn it immediately back on
Call the scene = ok

### Device information

Manufacturer: Qubino
Model name:  ZMNKAD1
Node ID in your network: 60
Device ID 345-82-8 (0x0159-0x0008-0x0052) 

### How are you using `node-zwave-js`?

- [X] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker image (latest)
- [ ] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker image (dev)
- [ ] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker manually built _(please specify branches)_
- [ ] `ioBroker.zwave2` adapter _(please specify version)_
- [ ] `HomeAssistant zwave_js` integration  _(please specify version)_
- [ ] `pkg`
- [ ] `node-red-contrib-zwave-js` _(please specify version, double click node to find out)_
- [ ] Manually built from GitHub _(please specify branch)_
- [ ] Other _(please describe)_

### Which branches or versions?

zwave-js-ui: 8.13.1.33a6f57
zwave-js: 10.14.0
home id: 3439378454
home hex: 0xcd00c016

### Did you change anything?

no

### Did this work before?

Yes (please describe)

### If yes, where did it work?

50/50 of the times this works, so rather doesn't work unless turning off the light.

### Attach Driver Logfile

```
[...]

PR #5662, opened 2023-04-14
fix: try automatically disabling explorer frames for too large payloads

AlCalzone:
This should fix issues like in #5660 for cases where transport service isn't supported.


AlCalzone:
@zwave-js-bot automerge

PR #5663, opened 2023-04-14
fix(config): disable Supervision CC for Homeseer HS-FLS100

devWaves:
Update hs-fls100 - Remove supervision. Fix to set parameters

the Homeseer HS-FLS100+ was unable to have parameters set from zwave-js


AlCalzone (approved):

AlCalzone:
thanks!

PR #5675, opened 2023-04-18
fix(cc): always write metadata for meter reports and add `ccSpecific` info

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #5689, opened 2023-04-20
feat(config): add new parameters 10-13 to Zooz ZEN04

cybergrimes:
Zooz released firmware 1.3 for ZEN04 device. Adding parameters 10 through 13 for enable/disable energy reporting and additional reporting frequency control.

https://www.support.getzooz.com/kb/article/1387-zen04-smart-plug-change-log/

> Firmware: 1.30 (released 4/2023 as OTA)
> 
>     Fixed bug: reduced the delay after the plug reports a Wattage change when on and the device connected to the smart plug is suddenly unplugged. 
>     Added new parameters to disable energy reporting values (parameters 10 and 11)
>     Added new parameters to set energy reporting frequency (parameters 12 and 13)

### DIFF
--- a/packages/config/config/devices/0x027a/zen04.json
+++ b/packages/config/config/devices/0x027a/zen04.json
@@ -56,8 +56,15 @@
 			"$import": "~/templates/master_template.json#state_after_power_failure_prev_off_on"
 		},
 		{
+			"#": "10",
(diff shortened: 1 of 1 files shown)


AlCalzone (approved):
Did a few changes for consistency. Also FW version is 1.30, not 1.3.

PR #5690, opened 2023-04-21
feat(config): add missing parameter 9 to Logic Group ZDB5100

LordMike:
In firmware `1.8` of the Logic Group ZDB5100, parameter 9 was added to control the brightness on the next turning on of the switch.

The setting has two ranges besides 0 for "No effect": 1-99 to affect only the next turning on, and 129-228 to affect _every_ turning on. I picked an option type with manual input allowed, but maybe something else is better suited? I tried to communicate the specifics in description


AlCalzone (approved):
ok with my changes

PR #5692, opened 2023-04-23
Update siren.json

yycTommy:
modify the secondary notification Length to -1 (line 75) as described in the Wink documentation for the siren.  255 is not within the assigned range. http://manuals-backend.z-wave.info/make.php?lang=en&sku=Wink%20Siren&cert=ZC10-17065664

PR description here


AlCalzone:
-1 (signed) is `0xff` and 255 (unsigned) is also `0xff`. Since the parameter is marked as `unsigned`, the current definition is correct.

PR #5696, opened 2023-04-24
fix: allow logging to console when log to file is active

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #5698, opened 2023-04-25
fix: handle non-ready controller when testing if CCs should be discarded

AlCalzone:
reported on Discord


AlCalzone:
@zwave-js-bot automerge

PR #5699, opened 2023-04-25
fix: limit decrypt attempts to 1 after initializing the SPAN

AlCalzone:
otherwise we'd try decrypting with the same IV 5 times.


AlCalzone:
@zwave-js-bot automerge

PR #5701, opened 2023-04-25
fix: report commands supported via S0

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #5704, opened 2023-04-25
fix: respond to CC version queries

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #5706, opened 2023-04-25
fix: add Supervision CC to NIF, don't advertise controlled CCs

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #5714, opened 2023-04-27
feat: implement mock-server to expose simulated controller via tcp

AlCalzone:
We've had this functionality for a while in integration tests, but there are use cases in development where talking to a simulated controller and nodes makes sense. This PR adds a simple mock-server which exposes a fake serial port via TCP that zwave-js can connect to. Behind that port is fake controller with fake nodes whose capabilities can be configured.

Example config for a network with the controller and a node that supports Version CC and Window Covering CC with all parameters:
```js
// filename: mock_config.js

// @ts-check
const { CommandClasses } = require("@zwave-js/core");
const { ccCaps } = require("@zwave-js/testing");

module.exports.default = {
	nodes: [
		{
			id: 2,
			capabilities: {
				commandClasses: [
					CommandClasses.Version,
					ccCaps({
						ccId: CommandClasses["Window Covering"],
						isSupported: true,
						version: 1,
						supportedParameters: new Array(24)
							.fill(0)
							.map((_, i) => i),
					}),
				],
			},
		},
	],
};
```

This can be executed using
```
yarn mock-server -c mock_config.js
```
in the `zwave-js` repo. It should also be possible to run this in any project zwave-js is installed in, but this is untested.

The config file may be `.js` or `.json`, but `.js` is recommended because it allows the use of exports from `zwave-js`.

### DIFF
--- a/packages/zwave-js/bin/mock-server.js
+++ b/packages/zwave-js/bin/mock-server.js
@@ -1,0 +1,58 @@
+// @ts-check
+const { MockServer } = require("../build/Testing");
+const { readFileSync } = require("fs");
+const path = require("path");
+
+const args = process.argv.slice(2);
+
+if (args.includes("--help") || args.includes("-h")) {
+	// Print a help text explaining the usage of this script and mention the options it supports
+	console.log(`
+Usage: node ${path.basename(__filename)} [options]
+
+Options:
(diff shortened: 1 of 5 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #5730, opened 2023-05-02
fix(cc): fall back to query Basic CC even if not advertised

AlCalzone:
When no actuator CC is supported, the controller MUST check if Basic CC is. We now do that, even if Basic CC is not advertised as supported anywhere.


AlCalzone:
@zwave-js-bot automerge

PR #5731, opened 2023-05-02
fix: handle Transport Service CC with unequal fragment sizes, do not encapsulate CRC16 or Transport Service in Security S0/S2

AlCalzone:
This PR fixes issues with incorrect encapsulation of the "outer" encapsulation CCs CRC16 and Transport Service. It also fixes how incoming Transport Service CC sessions with unequal fragment sizes are handled.


AlCalzone:
@zwave-js-bot automerge

PR #5732, opened 2023-05-02
fix(cc): only parse daily entry lock schedule if any weekday is selected

### LINKED ISSUE #5729: `Schedule Entry Lock` CC `getDailyRepeatingSchedule` always returns
Whatever is the `userId/slotId` requested it always returns:

```
{
weekdays: [ [length]: 0 ],
startHour: 0,
startMinute: 0,
durationHour: 0,
durationMinute: 0
}
```

Quoting @AlCalzone 

> Ok that seems to be because the node responds with a zeroed response instead of an empty one

AlCalzone:
fixes: #5729


AlCalzone:
@zwave-js-bot automerge

PR #5735, opened 2023-05-02
fix(cc): several fixes for Window Covering CC

### LINKED ISSUE #5734: Several bugs in `Window Covering CC`
- WindowCoveringCCSet should not expect a response
- Open sends the wrong value
- Close up/Close down expect `false` instead of `true`.

AlCalzone:
fixes: #5734


AlCalzone:
@zwave-js-bot automerge

PR #5738, opened 2023-05-03
fix: add compat flag to disable periodic value refreshes

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #5741, opened 2023-05-03
fix: more fixes for `Window Covering CC`

AlCalzone:
This PR fixes a couple of issues found while working with the Window Covering CC.


AlCalzone:
@zwave-js-bot automerge

PR #5743, opened 2023-05-03
fix(config): add 700 series variant of SimonTech Roller Blind

davidcalleja:
I have new Roller Blind in my home and his signature is `x0267-0x0102-0x0000`

### DIFF
--- a/packages/config/config/devices/0x0267/10002080-13x.json
+++ b/packages/config/config/devices/0x0267/10002080-13x.json
@@ -39,6 +39,10 @@
 		{
 			"productType": "0x0004",
 			"productId": "0x0177"
+		},
+		{
+			"productType": "0x0102",
+			"productId": "0x0000"
 		}
 	],
 	"firmwareVersion": {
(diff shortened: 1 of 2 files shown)


AlCalzone:
I assume you have confirmed that the config parameters are the same for your device?
This ID looks so different from the others, it may be an entirely new hardware revision.

davidcalleja:
Yes, is a bit weird but... is same device with same instructions... but this one have `z-wave 2` I have tried to test it in my zwave-js-ui (I have it in docker) but is not working, do you know exactly how to test if it works?

Z-Wave JS UI display diferences

Screenshot 2023-05-04 at 18 21 10

Instructions of new device are this

AlCalzone:
Ok looks like a new 700 series version. I'll double check later and merge. Thanks!

AlCalzone:
The changes were significant enough (including differences in config parameters) that I decided to move this to a new file.

AlCalzone (approved):

PR #5747, opened 2023-05-04
fix: improve "unnecessary endpoint" heuristic

### LINKED ISSUE #4718: Consider not hiding "unnecessary" endpoints if they implement different switch-type classes
Need to think about the implications. For example ZEN30 has Multilevel and Binary Switch for a Dimmer and Relay.

At the very least, it should be possible to use the endpoints in associations

AlCalzone:
fixes: #4718


AlCalzone:
@zwave-js-bot automerge

PR #5760, opened 2023-05-05
fix(cc): only discard unknown sensors on V4 and lower

### LINKED ISSUE #5748: Multilevel Sensor reports with unknown (but supported) scales must not be dropped

AlCalzone:
On V5 and higher, we can test if the device actually supports the sensors, and preserve them.

fixes: #5748


AlCalzone:
@zwave-js-bot automerge

PR #5783, opened 2023-05-10
fix: incoming command check for lower-than-expected security level

### LINKED ISSUE #5752: S0-encapsulated commands are not discarded if the highest security class is S2

AlCalzone:
fixes: #5752


AlCalzone:
@zwave-js-bot automerge

PR #5792, opened 2023-05-12
feat(cc): add `states` property to `boolean` value metadata

### LINKED ISSUE #5750: Modify metadata of writeonly booleans which accept `true` and `false`
for https://github.com/zwave-js/zwave-js-ui/issues/3075

AlCalzone:
Until now, some writeonly `boolean` values would only accept `true` (e.g. reset meters), whereas others (start/stop level change) would accept `true` and `false`. To make this more apparent, this PR adds a `states` property to `boolean` metadata.

```ts
	/** Possible values and their meaning */
	states?: {
		true?: string;
		false?: string;
	};
```

This field is optional.
If it is missing, both `true` and `false` are possible values, but without special meaning
If it exists, it defines which values are accepted, and assigns them a label/meaning. For example
```ts
states: {
  true: "Reset"
}
```
means the value will **only** accept `true` as a value. The corresponding button should have the label "Reset".

```ts
states: {
  true: "Start",
  false: "Stop"
}
```
means both `true` and `false` are accepted, where setting `true` will start a process and `false` will stop it again. The corresponding buttons should have the labels "Start" and "Stop", respectively.

fixes: #5750

### DIFF
--- a/packages/cc/api.md
+++ b/packages/cc/api.md
@@ -1609,6 +1609,9 @@ export const BasicCCValues: Readonly<{
         readonly is: (valueId: ValueID_2) => boolean;
         readonly meta: {
             readonly label: "Restore previous value";
+            readonly states: {
+                true: string;
+            };
             readonly readable: false;
             readonly type: "boolean";
             readonly writeable: true;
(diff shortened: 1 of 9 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #5796, opened 2023-05-15
fix: make `Driver` and `ZWave[Application]Host` interfaces compatible for external projects

### LINKED ISSUE #5602: Driver is not assignable to ZWaveHost (external projects)

AlCalzone:
This PR exposes some previously `@internal` properties on the driver. Most of those are NOT meant to be public nor are they meant to be used by applications. However they need to be visible so TypeScript does not complain when a driver instance is passed to a method that accepts a host interface.

fixes: #5602


AlCalzone:
@zwave-js-bot automerge

PR #5803, opened 2023-05-15
fix(cc): correct return type for `Configuration CC` `setBulk` command

AlCalzone:
Previously the return type was incorrect, so a `ConfigurationCCBulkReport` instance was returned instead of the `SupervisionResult`.


AlCalzone:
@zwave-js-bot automerge

PR #5806, opened 2023-05-15
feat: add `allowManualEntry` property to numeric metadata to indicate whether `states` are only informational

AlCalzone:
Some numeric metadata can benefit from having named `states`, but if they are supposed to be writable, we need a way to indicate whether the `states` are simply informational or limit what the user can enter.
We copy this attribute from configuration metadata which shares the same logic.


AlCalzone:
@zwave-js-bot automerge

PR #5807, opened 2023-05-15
fix(cc): clean up `Window Covering CC` values

AlCalzone:
Those were a bit of a mess, with functionality duplicated and no values to perform a level change. This PR clears that up, and also marks `targetValue` with the new `allowManualEntry` property.


AlCalzone:
@zwave-js-bot automerge

PR #5809, opened 2023-05-16
fix(cc): always include duration field in serialized buffer for CC versions that support it

### LINKED ISSUE #5749: Binary/Multilevel Switch CC must include the duration for supporting nodes

AlCalzone:
fixes: #5749


AlCalzone:
@zwave-js-bot automerge

Issue #5798, opened 2023-05-15
Window Covering CC: Hide Multilevel Switch CC


AlCalzone:
> A controlling node MUST NOT interview and provide controlling functionalities for the Multilevel
Switch Command Class for a node (or endpoint) supporting this Command Class, as it is a fully
redundant and less precise application functionality.

PR #5812, opened 2023-05-17
fix: hide Multilevel Switch CC in favor of Window Covering CC

AlCalzone:
This PR fulfills a certification requirement. Unfortunately this is very likely breaking for applications, especially if they don't support Window Covering CC yet.

fixes: #5798


AlCalzone:
@zwave-js-bot automerge

PR #5820, opened 2023-05-19
fix: consider endpoints when force-adding CCs mandatory for Z-Wave+ v2 devices, add logs

AlCalzone:
The added logs should remove some of the guessing why e.g. Supervision is used for a device that doesn't seem to support it.


AlCalzone:
@zwave-js-bot automerge

PR #5822, opened 2023-05-19
feat(config): add/fix params for Intermatic PE653

philh30:
Fix params 1, 2 & 22 and add params 3-21, 23, 32-47, 49-50 for Intermatic PE653.

### DIFF
--- a/packages/config/config/devices/0x0005/pe653.json
+++ b/packages/config/config/devices/0x0005/pe653.json
@@ -15,45 +15,466 @@
 	},
 	"paramInformation": [
 		{
-			"#": "2",
+			"#": "1[0x02]",
+			"$import": "~/templates/master_template.json#base_options_nounit",
+			"label": "Installed Pump Type",
+			"valueSize": 2,
+			"options": [
+				{
+					"label": "One Speed",
+					"value": 0
+				},
+				{
+					"label": "Two Speed",
+					"value": 1
+				}
+			]
+		},
+		{
+			"#": "1[0x01]",
+			"$import": "~/templates/master_template.json#base_enable_disable",
+			"label": "Booster (Cleaner) Pump Installed",
+			"valueSize": 2
+		},
+		{
+			"#": "1[0xff00]",
+			"label": "Booster (Cleaner) Pump Operation Mode",
+			"description": "Set the filter pump mode to use when the booster (cleaner) pump is running.",
+			"valueSize": 2,
+			"defaultValue": 1,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Disable",
+					"value": 1
+				},
+				{
+					"label": "Circuit 1",
+					"value": 2
+				},
+				{
+					"label": "VSP Speed 1",
+					"value": 3
+				},
+				{
+					"label": "VSP Speed 2",
+					"value": 4
+				},
+				{
+					"label": "VSP Speed 3",
+					"value": 5
+				},
+				{
+					"label": "VSP Speed 4",
+					"value": 6
+				}
+			]
+		},
+		{
+			"#": "2[0xff00]",
 			"label": "Fireman Timeout",
-			"description": "Minutes that pump will continue to run after heater shutoff",
-			"valueSize": 1,
+			"description": "Minimum time pump will continue to run after heater shutoff for cooldown. Circuit 5 will be assigned to the heater if >=0. Set to -1 to disable.",
+			"valueSize": 2,
+			"minValue": -1,
+			"maxValue": 15,
+			"defaultValue": -1,
+			"unit": "minutes",
+			"options": [
+				{
+					"label": "No heater installed",
+					"value": -1
+				},
+				{
+					"label": "Heater installed with no cooldown",
+					"value": 0
+				}
+			]
+		},
+		{
+			"#": "2[0x01]",
+			"$import": "~/templates/master_template.json#base_enable_disable",
+			"label": "Heater Safety Setting",
+			"description": "Prevent the heater from turning on while the pump is off.",
+			"valueSize": 2
+		},
+		{
+			"#": "3[0xff000000]",
+			"label": "Water Temperature Offset",
+			"valueSize": 4,
+			"minValue": -5,
+			"maxValue": 5,
+			"unit": "°F",
+			"defaultValue": 0
+		},
+		{
+			"#": "3[0xff0000]",
+			"$import": "#paramInformation/3[0xff000000]",
+			"label": "Air/Freeze Temperature Offset"
+		},
+		{
+			"#": "3[0xff00]",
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0005/pe653.json
```diff
@@ -15,45 +15,466 @@
...
+				{
+					"label": "Two Speed",
+					"value": 1
+				}
+			]
+		},
+		{
+			"#": "1[0x01]",
+			"$import": "~/templates/master_template.json#base_enable_disable",
+			"label": "Booster (Cleaner) Pump Installed",
```
AlCalzone, line 36: This sounds like a readonly param. Is it?

packages/config/config/devices/0x0005/pe653.json
```diff
@@ -15,45 +15,466 @@
...
+					"value": 6
+				}
+			]
+		},
+		{
+			"#": "2[0xff00]",
 			"label": "Fireman Timeout",
-			"description": "Minutes that pump will continue to run after heater shutoff",
-			"valueSize": 1,
+			"description": "Minimum time pump will continue to run after heater shutoff for cooldown. Circuit 5 will be assigned to the heater if >=0. Set to -1 to disable.",
```
AlCalzone: Not sure if "Fireman Timeout" is a commonly used expression. Maybe there's a better label which also makes the description unnecessary?

"Set to -1 to disable" should be removed and instead merged with the label for that option.

packages/config/config/devices/0x0005/pe653.json
```diff
@@ -15,45 +15,466 @@
...
+			"valueSize": 2,
+			"minValue": -1,
+			"maxValue": 15,
+			"defaultValue": -1,
 			"unit": "minutes",
+			"options": [
+				{
+					"label": "No heater installed",
```
AlCalzone, line 83:
```suggestion
					"label": "Disable / No heater installed",
```
maybe?

packages/config/config/devices/0x0005/pe653.json
```diff
@@ -15,45 +15,466 @@
...
+			"#": "49",
+			"$import": "#paramInformation/32",
+			"label": "Variable Speed Pump - Max Speed",
+			"defaultValue": 3450
+		},
+		{
+			"#": "50[0xff000000]",
+			"$import": "~/templates/master_template.json#base_options_nounit",
+			"label": "Freeze Protection: Temperature",
+			"description": "Temperature at air/freeze sensor when freeze protection starts. Set to 0 to disable freeze protection.",
```
AlCalzone: The description seems unnecessary. Also is the user meant to enter any other values than 0, 40-44? If not, add `"allowManualEntry": false`

packages/config/config/devices/0x0005/pe653.json
```diff
@@ -15,45 +15,466 @@
...
+					"label": "44 °F",
+					"value": 44
+				}
+			]
+		},
+		{
+			"#": "50[0x010000]",
+			"$import": "~/templates/master_template.json#base_enable_disable",
+			"label": "Freeze Protection: Circuit 1",
+			"description": "Turn on the circuit when system is in freeze protection mode. Requires connected air/freeze sensor.",
```
AlCalzone: How about this?
```suggestion
			"label": "Freeze Protection: Turn On Circuit 1",
			"description": "Requires connected air/freeze sensor.",
```
with 5 identical parameters, the alternative would end up cluttering UIs a lot.

packages/config/config/devices/0x0005/pe653.json
```diff
@@ -15,45 +15,466 @@
...
+			"valueSize": 4
+		},
+		{
+			"#": "50[0x020000]",
+			"$import": "#paramInformation/50[0x010000]",
+			"label": "Freeze Protection: Circuit 2"
```
AlCalzone: If we do this, give all partial params for 50 the same treatment.

packages/config/config/devices/0x0005/pe653.json
```diff
@@ -15,45 +15,466 @@
...
 		},
 		{
-			"#": "22",
-			"label": "Pool - Spa Mode",
-			"description": "Controller Mode",
-			"valueSize": 1,
-			"unit": "Decimal",
+			"#": "4",
+			"label": "Circuit 1 Schedule 1",
+			"description": "Start time (first two bytes, little endian) and stop time (last two bytes, little endian) of schedule in minutes past midnight, e.g. 12:05am (0x0500) to 3:00pm (0x8403) is entered as 83919875. Set to 4294967295 (0xffffffff) to disable.",
```
AlCalzone, line 318: These will be a lot of repeated and long descriptions. Maybe the other descriptions should just refer to this one?

AlCalzone (changes requested):
Thanks. A couple of remarks

packages/config/config/devices/0x0005/pe653.json
philh30: It should be, but it isn't. Its value needs to be linked to 1[0xff00] to work properly - disabled when 1[0xff00] is disabled and enabled when 1[0xff00] has a circuit/pump speed selected. The PE653 will accept a mismatch though, and having 1[0x01] disabled will block the pump speed in 1[0xff00] from operating. I'm not sure if there are any ill effects from having 1[0x01] enabled while 1[0xff00] is disabled.

The handheld PE953 controller that ships with the PE653 doesn't expose this as a separate config option to the user, and just handles the logic to tie this to 1[0xff00].

AlCalzone: got it

packages/config/config/devices/0x0005/pe653.json
philh30: Good point. "Fireman Timeout" makes some sense in that it's controlling the fireman switch on the heater, but "Heater Cooldown Period" would be more descriptive. I think "Set to -1 to disable" can be skipped on the option label as "No heater installed" should be descriptive enough.

packages/config/config/devices/0x0005/pe653.json
philh30: Agreed on the description - will remove.

The allowable values are just 0, 40-44. The `base_options_nounit` template provides `"allowManualEntry": false` so it's not necessary here.

packages/config/config/devices/0x0005/pe653.json
philh30: As noted above, I think we can skip this as "No heater installed" should be sufficient.

philh30:
@AlCalzone I think I've addressed everything, but let me know if there are additional changes you'd like to see. Thanks for the review!

AlCalzone (approved):

AlCalzone:
closing and re-opening to trigger tests

PR #5827, opened 2023-05-22
fix: correct the serial API commands to set/get/delete priority routes

### LINKED ISSUE #5711: Investigate Priority Route differences between 500 and 700 series specs
GetPriorityRoute has an additional byte after the node ID in the 500 series specs, which is supposed to indicate what kind of route it is. The gecko_sdk firmware code seems to contain that too.

Also, SetPriorityRoute seems to accept a command without repeater info (shorter command?), which can clear the priority route.

AlCalzone:
fixes: #5711


AlCalzone:
@zwave-js-bot automerge

PR #5828, opened 2023-05-22
fix: ignore inclusion notification for existing nodes

### LINKED ISSUE #5781: Healed nodes not finishing re-interview
### Checklist

- [ ]  I am not using Home Assistant. Or: a developer has told me to come here.

### Deploy method

Docker

### Z-Wave JS UI version

8.15.0.f48c52b

### ZwaveJS version

10.17.1

### Describe the bug

When I attempt to heal any (or all) nodes on my network, the healing completes successfully, but then the node's interview status changes to "none" and has a forever spining wheel and I loose all ability to control the device as all of its functions are no longer accessible. Rebooting zwave brings all this back. I looked in the logs and just after any individual node heals, I see the following (error at the end):

This is what I see before a repair:

image

and after (until a reboot occurs)

image

The error occurs twice in the logs once during and once just after the heal process has finished. It doesn't occur outside of this process and is referencing provisioning which I believe is related to the interview process this is stuck on.

2023-05-09 08:35:00.965 ERROR APP: Unhandled Rejection, reason: TypeError: Cannot use 'in' operator to search for 'provisioning' in undefined

### To Reproduce

Heal any node and wait for the heal to complete

### Expected behavior

I'd expect to be able to use the node after healing and for the interview to complete as expected.

### Additional context

Not sure if this is the UI or the driver, but we'll start here since the error is "APP". Below is an extract from my logs of the start of the heal and the error after it is successful.

[z-ui_current.log](https://github.com/zwave-js/zwave-js-ui/files/11433031/z-ui_current.log)

AlCalzone:
fixes: #5781

### DIFF
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -2031,6 +2031,18 @@ export class ZWaveController extends TypedEventEmitter<ControllerEventCallbacks>
 			const nodeId = msg.nodeId;
 			const nodeInfo = msg.nodeInformation;
 
+			// It can happen that this is received for a node that is already part of the network:
+			// https://github.com/zwave-js/node-zwave-js/issues/5781
+			// In this case, ignore this message to prevent chaos.
+
+			if (this._nodes.has(nodeId)) {
+				this.driver.controllerLog.print(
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #5829, opened 2023-05-22
fix: avoid crash when value DB contains non-JSON key

AlCalzone:
reported on Discord


AlCalzone:
@zwave-js-bot automerge

Issue #5597, opened 2023-03-26
Simplify working with Indicator values that must/may be set together


AlCalzone:
- [x] API to set timeout (#5830)
- [ ] remove values to set timeout
- [ ] toggle

PR #5830, opened 2023-05-22
feat(cc): add value and API to control indicator timeout

AlCalzone:
part 1 of #5597


AlCalzone:
@zwave-js-bot automerge

PR #5136, opened 2022-10-01
feat(config): add Dawon PM-S240-ZW, disable Supervision

AlCalzone:
fixes: #5132

Issue #5132, opened 2022-09-29
[question] My switch does not work in v7.0.0 and later versions.


jsj0318:
Hello.
First of all, I apologize for my poor English.

SWITCH : Dawon DNS Zwave Switch

I am inquiring because the switch that worked well before zwavejs2mqtt v6.14.1 has not been working since the update.

v6.14.1 and earlier debug log (work good)
---------------------------------------------------------------------------------------------------------------------
2022-09-29 08:40:50.575 INFO ZWAVE: Node 39: value updated: 37-1-currentValue false => true
2022-09-29 08:40:50.581 INFO ZWAVE: Node 39: value updated: 37-1-targetValue false => true
2022-09-29 08:40:50.615 INFO ZWAVE: Node 39: value updated: 113-1-alarmType 0 => 0
2022-09-29 08:40:50.619 INFO ZWAVE: Node 39: value updated: 113-1-alarmLevel 0 => 0
2022-09-29 08:40:50.639 INFO ZWAVE: Node 39: value updated: 113-1-Power Management-Mains status 2 => 3
2022-09-29 08:40:52.986 INFO ZWAVE: Node 39: value updated: 37-1-currentValue true => true

v.7.0.0 and later versions debug log (not work)
---------------------------------------------------------------------------------------------------------------------
2022-09-28 16:54:41.269 INFO Z-WAVE: Calling api writeValue with args: [
  {
    nodeId: 39,
    commandClass: 37,
    endpoint: 1,
    property: 'targetValue'
  },
  true,
  {},
  [length]: 3
]

The sensor values for that switch are displayed well.
Is there anything I need to check? Is there any more information I need to tell you?

Thank you.

robertsLando:
Please try to re-interview your device

jsj0318:
> Please try to re-interview your device

I tried it several times, but it didn't work.
exclusion and Inclusion same.

robertsLando:
@jsj0318 Try uploading driver logs so please

AlCalzone:
Please make a [driver log](https://zwave-js.github.io/zwave-js-ui/#/troubleshooting/generating-logs?id=driver-logs), loglevel `debug` and attach it here as a file (drag & drop into the text field).

jsj0318:
I am attaching the log file you mentioned.

The node number of the non-operating switch is
35, 38, 39, 51. The on-off test was done with node 51.

v6.14.1 (work good)
[...]

PR #5833, opened 2023-05-23
feat(config): add Dawon PM-S140-ZW, PM-S340-ZW and KR frequencies

techinpark:
The 2-button device is registered,  but the 3-button device is missing, so we created a PR to add it. 

#5136 Related PR is here


packages/config/config/devices/0x018c/pm-240-zw-3button.json
```diff
@@ -0,0 +1,25 @@
+{
+  "manufacturer": "Dawon",
+  "manufacturerId": "0x018c",
+  "label": "PM-S240-ZW",
```
AlCalzone: I think this is supposed to be
```suggestion
  "label": "PM-S340-ZW",
```
at least according to [this comment](https://github.com/zwave-js/node-zwave-js/issues/5132#issuecomment-1264405046) I apparently never followed up.

Can this be?

AlCalzone: If that's the case, I suppose we should also add the other product types mentioned in https://github.com/zwave-js/node-zwave-js/issues/5132#issuecomment-1264369553

AlCalzone (changes requested):

packages/config/config/devices/0x018c/pm-240-zw-3button.json
techinpark: @AlCalzone  you right! 3 button type model is `PM-S340-ZW` 👍 [6b2846c](https://github.com/zwave-js/node-zwave-js/pull/5833/commits/6b2846c888fc1670d525f4b9ebc76d18fe59cf3b) fixed here

AlCalzone:
Thanks! I'll hijack this PR to also add the 1-button variant and the other device IDs

AlCalzone (dismissed):

AlCalzone (approved):

PR #5836, opened 2023-05-24
chore: upgrade api-extractor

AlCalzone:
this should solve issues with the `satisfies` syntax


AlCalzone:
@zwave-js-bot automerge

Issue #5236, opened 2022-12-02
Value changes to undefined do not get emitted 


AlCalzone:
It looks like the `undefined` value change doesn't get emitted. This is something I will have to investigate.

@robertsLando can the UI handle if a number gets set to "unknown"? In that case, they could enable the driver option as a workaround (can you explain how?).

_Originally posted by @AlCalzone in https://github.com/zwave-js/node-zwave-js/discussions/4804#discussioncomment-4293390_

robertsLando:
Yep it can, answered in the discussion

AlCalzone:
Whoops didn't mean to tag you here

PR #5843, opened 2023-05-25
refactor: distinguish between not (yet) known (`undefined`) and known to be unknown (`null`)

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/issues/5864
fixes: #5236

This change needs some explanation, as it has stirred up some controversy in https://github.com/zwave-js/node-zwave-js/discussions/4804 (although some of that is probably due to a misunderstanding).

Several CCs use the value 254 to indicate that a state is unknown. However this value often lies outside of the normal range of values, e.g. 0-99 for multilevel switches. Other CCs translate the raw numeric values to other type, e.g. Binary Switch CC. So simply preserving the 254 is no option, because it would throw off applications which expect a value in the defined range or of the correct type.
Previously, we used `"unknown"` or `undefined` (based on the driver option `preserveUnknownValues`), which both have downsides aswell:
- The string `"unknown"` is obviously not a number or boolean, which could cause errors
- `undefined` gets "filtered" out by the value DB, because it also means **no value**. As a result changes **to** `unknown` would cause no events
- `undefined` could also mean the data hasn't been queried yet, but it is impossible to distinguish from the `unknown` case

---

To solve this dilemma, this PR changes how we handle unknown and not known values:
It introduces two signal types and values with the same name:
- `NOT_KNOWN` is an alias for `undefined` and indicates that a value is **not known (yet)**, either for a value that has not been queried yet, or where no response was received
- `UNKNOWN_STATE` is an alias for `null` and indicates that some information has been queried, but the corresponding state is **(known to be) unknown**. E.g. a multilevel device without intermediate position support - it knows that it is neither fully open, nor fully closed, but not the exact position.

Two related utility types have been introduced in an attempt to preserve this distinction through the type system:
- `MaybeNotKnown<T>` is `T | NOT_KNOWN` and indicates that something is either of the type T, or `NOT_KNOWN = undefined`
- `MaybeUnknown<T>` is `T | UNKNOWN_STATE` and indicates that something is either of the type T, or `UNKNOWN_STATE = null`

These types replace the former `Maybe<T>` type.

Many method signatures and properties which previously used `undefined` to indicate an uncertainty have been updated to use either of the previous types to make it clear what an absence of a value means.

Likewise, some CC values can now be `null / UNKNOWN_STATE`. The `preserveUnknownValues` driver option has been removed.

Application developers should double-check for common sources of errors, e.g.:
- The comparisons `!= null` and `!= undefined` fail to distinguish between `NOT_KNOWN` and `UNKNOWN_STATE`
- Nullish coalescing (`?.`, `??`, `??=`)
- Expecting a number or boolean when the value is not `undefined`


AlCalzone:
@zwave-js-bot automerge

PR #5857, opened 2023-05-30
fix: do not mark a node as awake when it sends a `NonceGet`

### LINKED ISSUE #5854: Event "CAN" was sent to stopped service "serialAPICommand"
### Checklist

### Deploy method

PKG executable

### Z-Wave JS UI version

8.17.1

### ZwaveJS version

10.22.2

### Describe the bug

Hello

I've upgraded my Home Assistant today and had to move from `zwavejs2mqtt` to `zwave-js-ui`.  
Everything works as expected and I could migrate everything, but I have some issues with the new Z-Wave JS server.

I've 7 similar Z-Wave switches (Aeotec WallMote Quad) and some of them have issues.

When I press a button, it works. However, when I press the next button (within a couple of seconds), it starts to blink and the event doesn't work.

At the Z-Wave JS UI server I suddenly see a lot of `Timeout ACK`, `Dropped TX` and `Dropped RX`. When I open the debug log I can see this:

```
<<< Me pressing the first button >>>
2023-05-18 19:31:26.480 INFO Z-WAVE: [Node 007] Is now awake
2023-05-18 19:31:26.536 INFO Z-WAVE: [Node 007] Value notification: 91-0-scene-003 0
<<< Me pressing the second button 1s later; no output in the log but 4s later the node goes to sleep >>>
2023-05-18 19:31:30.835 INFO Z-WAVE: [Node 007] Is now asleep
Warning: Event "CAN" was sent to stopped service "serialAPICommand". This service has already reached its final state, and will not transition.
Event: {"type":"CAN"}
Warning: Event "ACK" was sent to stopped service "serialAPICommand". This service has already reached its final state, and will not transition.
Event: {"type":"ACK"}
Warning: Event "ACK" was sent to stopped service "serialAPICommand". This service has already reached its final state, and will not transition.
Event: {"type":"ACK"}
```

I didn't had these issues before the upgrade and I didn't change anything on the configuration / setup. It is more than one WallMote which is now experiencing these issues, so it's unlikely a hardware issue (worked before upgrade, multiple devices fail at the same time after the upgrade).

Do you've any idea how to fix that or where I can start debugging?

### To Reproduce

Press different WallMote buttons within a couple of seconds.

### Expected behavior

Event will be sent to HA, but it's not sent. 
Instead there's an error (see log above).

### Additional context

Please note I already [opened an issue in the HA repo](https://github.com/home-assistant/core/issues/93245), but a developer sent me here :)

Here's my additional log: [zwave.log](https://github.com/zwave-js/zwave-js-ui/files/11583005/zwave.log)

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/issues/5854

### DIFF
--- a/packages/zwave-js/src/lib/node/Node.ts
+++ b/packages/zwave-js/src/lib/node/Node.ts
@@ -2703,7 +2703,12 @@ protocol version:      ${this.protocolVersion}`;
 		}
 
 		// If we're being queried by another node, treat this as a sign that the other node is awake
-		if (command.constructor.name.endsWith("Get")) {
+		if (
+			command.constructor.name.endsWith("Get") &&
+			// Nonces can be sent while asleep though
+			!(command instanceof SecurityCCNonceGet) &&
+			!(command instanceof Security2CCNonceGet)
+		) {
 			this.markAsAwake();
 		}
 


AlCalzone:
@zwave-js-bot automerge

Issue #5869, opened 2023-06-01
Rework support for auto-discovered Bitfield config parameters


AlCalzone:
**Bitfield**
```
└─[ConfigurationCCPropertiesReport]
		parameter #:         48
		next param #:        49
		value size:          1
		value format:        BitField
		max value:           1, 2, 3, 4, 5, 6, 7, 8
		default value:
		alters capabilities: false
		readonly:            false
		advanced:            false
		bulk support:        false


	└─[ConfigurationCCInfoReport]
			parameter #:       48
			info:              This value is bit mask. Used to enable/disable measur
ement reports for various sensors that is more than the upper limit value or l
ess than the lower limit value.
			0 - Disabled
			1 - Temperature upper level
			2 - Humidity upper level
			4 - Lux upper level
			8 - UV upper level
			16 - Temperature lower level
			32 - Humidity lower level
			64 - Lux lower level
			128 - UV lower level
			reports to follow: 0
```
I think we should adopt the way the bitfield is described here, e.g. as an object with keys 1,2,4,8,... and boolean values instead of the JS Set which doesn't serialize well.

PR #5870, opened 2023-06-02
fix: translate `BitField` config params to partial params

AlCalzone:
fixes: #5869


AlCalzone:
@zwave-js-bot automerge

PR #5887, opened 2023-06-05
fix: remove support/control mark from NIF, omit non-mandatory CCs

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #5897, opened 2023-06-07
feat: change return type of `setValue` to include context on the execution

AlCalzone:
This PR changes the return type of `setValue` from `boolean` to `SetValueResult`. While not strictly necessary, this change enables applications to give better feedback to users why a command failed and make better decisions what to do with successful commands.


AlCalzone:
@zwave-js-bot automerge

PR #5899, opened 2023-06-07
fix(config): remove duplicated battery info on endpoint 0 for Fibaro FGT001

kvaster:
Fix regression. On fibaro thermostat we have 3 endpoints. 0 and 1 are duplicated - main, 2 - is an endpoint for external sensor. With this fix we remove reporting battery on endpoint 0, cause we have same reports on endpoint 1.

### DIFF
--- a/packages/config/config/devices/0x010f/fgt001.json
+++ b/packages/config/config/devices/0x010f/fgt001.json
@@ -140,6 +140,13 @@
 		// The device has two endpoints with different device classes, but both need to exist
 		"preserveEndpoints": [1, 2],
 		// Not sure if necessary, but this can prevent missing updates on endpoint 1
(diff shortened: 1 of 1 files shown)


AlCalzone:
Are there any values on endpoint 0 that are not duplicated? Seeing a driver log of a re-interview would be useful.

kvaster:
Attaching zwave logs. Logs were done with my patch applied. I can redo this in case logs will be bit informative cause of this.

[zwavejs_2023-06-09.log](https://github.com/zwave-js/node-zwave-js/files/11705215/zwavejs_2023-06-09.log)

AlCalzone (approved):
Ok with my clarifying comment

PR #5905, opened 2023-06-12
fix: do not check CC support when responding to Get requests

AlCalzone:
When responding to `Get` requests we may be responding to a node that only controls a CC but does not support it. Thus we don't require support on the target node.


AlCalzone:
@zwave-js-bot automerge

PR #5906, opened 2023-06-12
fix: remove force-adding of CC support for Z-Wave+ v2

AlCalzone:
This PR removes a hack we added to force-add support for certain mandatory CCs after a node was determined to be Z-Wave+ v2. It is not worth the headaches and https://github.com/zwave-js/node-zwave-js/pull/5905 is a better solution.


AlCalzone:
@zwave-js-bot automerge

PR #5907, opened 2023-06-12
fix: do not use Supervision for sending `TimeCC` Reports

AlCalzone:
... as they are usually sent in response to a Get request


AlCalzone:
@zwave-js-bot automerge

PR #5913, opened 2023-06-13
fix: support decoding certain commands with other security classes than the highest one

### LINKED ISSUE #5700: Respond to S2 Commands Supported Get received with a lower security class
Also, omit the S2 CC itself:
> A supporting node MUST NOT advertise the Security 2 Command Class in its S0/S2 Commands Supported Report list

Also, omit CCs that were already reported as non-secure (in the NIF)

AlCalzone:
With this PR, we attempt decoding S2-encapsulated commands with lower granted keys if using the highest one fails. This allows us to answer the secure command queries which are used by other controllers to learn our security classes.

fixes: #5700


AlCalzone:
@zwave-js-bot automerge

PR #5914, opened 2023-06-13
fix: discard GET-type commands received via multicast or broadcast

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #5916, opened 2023-06-14
fix: fall back to "General: Lifeline" profile when assigning lifeline

AlCalzone:
Despite all the heuristics, there was still an edge case where we couldn't determine any lifeline and the issues commands were of no help either.
In this case we now check if group 1 has the "General: Lifeline" profile, and if yes, use it.


AlCalzone:
@zwave-js-bot automerge

PR #5917, opened 2023-06-14
fix: during proxy bootstrapping, `inclusionOptions` can be `undefined`

AlCalzone:
This was reported before, not sure where. This happens during proxy bootstrapping S2 for an inclusion controller.


AlCalzone:
@zwave-js-bot automerge

PR #5924, opened 2023-06-16
fix: store MPAN when receiving MPAN extension, slice 13 bytes only

AlCalzone:
TODO: rebase onto master, remove debug logging


AlCalzone:
@zwave-js-bot automerge

PR #5925, opened 2023-06-16
fix: make guessing the firmware file format case-insensitive

AlCalzone:
Apparently some manufacturers distribute files with uppercase extensions.


AlCalzone:
@zwave-js-bot automerge

PR #5929, opened 2023-06-19
feat: accept `manufacturerData` as hex string in `ThermostatModeCC.set` API

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #5931, opened 2023-06-19
fix: allow other devices to configure a single lifeline, send it reset notifications

AlCalzone:
Certification expects us to send a `Device Reset Locally Notification` via the lifeline on a factory reset. Although this doesn't seem to make a lot of sense when we're the primary controller, we now implement the bare minimum for supporting this.


AlCalzone:
@zwave-js-bot automerge

PR #5934, opened 2023-06-20
feat: support the "identify" function

### LINKED ISSUE #5716: Indicator CC: Must support Identify
The Root Device of a node MUST support the Indicator Command Class, version 3 and support the Indicator ID 0x50 (Identify) and Properties ID 0x03, 0x04 and 0x05.

AlCalzone:
fixes: #5716

With this PR, we now **support** the `Indicator CC`, which was previously only **controlled**. Although the specs are a bit confusing regarding controllers, and this seems totally unnecessary for a gateway with UI, this appears to be a certification requirement.

We now advertise support for indicator ID `0x50`, properties `0x03, 0x04, 0x05`, a.k.a. identify. When such a command is received, the new `"identify"` controller event gets emitted. Applications must listen for this and show some form of identification. The specs recommend this:
> The node is RECOMMENDED to use a visible LED for an identify function if it has an LED. If the node is itself a light source, e.g. a light bulb, this MAY be used in place of a dedicated LED.

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -1413,3 +1413,16 @@ enum ControllerFirmwareUpdateStatus {
 	OK = 0xff,
 }
 ```
+
+### `"identify"`
+
+This is emitted when another node instructs Z-Wave JS to identify itself using the `Indicator CC`, indicator ID `0x50`. The callback has no arguments:
+
+```ts
+() => void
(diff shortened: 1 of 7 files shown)


AlCalzone:
@zwave-js-bot automerge

Issue #5772, opened 2023-05-08
Wakeup CC: Certification requires IntervalSet, even if the interval and destination is already correct


AlCalzone:
This happens because we first check the current settings and do not re-set them if they are already correct. We might have to discuss this with the ZWA.

AlCalzone:
Fixed in CTT 3.7.0

PR #5936, opened 2023-06-20
fix: move weird pre-interview into the normal interview process

AlCalzone:
This mainly concerns setting up the wakeup destination (in reference to #5772), but also some hacky lifeline and assigning a return route which is already done before.


AlCalzone:
@zwave-js-bot automerge

PR #5938, opened 2023-06-23
feat: extract neighbor discovery into separate method, cleanup healing process, assign routes to all association destinations

AlCalzone:
With this PR, neighbor discovery is now available outside of a healing process. This also gave a chance to clarify the assign/delete return route methods.

If a node has more than 4 association targets, healing now assigns routes to all of them. Previously, only 4 were assigned due to a misunderstanding.


AlCalzone:
@zwave-js-bot automerge

PR #5939, opened 2023-06-23
ci: check updates to API report as part of documentation generation

AlCalzone:
Requiring this as part of every PR is slowing me down too much, especially since the API generation seems to require two builds locally.


AlCalzone:
@zwave-js-bot automerge

PR #5941, opened 2023-06-23
feat: assign custom return routes, cache priority return routes

AlCalzone:
This PR adds a missing piece of functionality for managing return routes. Previously, there were two ways to do this:
- Let the controller assign them, which usually works fine, but can produce subpar results
- Assign priority routes, which are always used first and can introduce permanent latency if they fail

With this PR, applications are now able to assign custom return routes without going through the corresponding serial API. This uses the internal `Z-Wave Protocol CC`, which is normally only used by the controller as part of serial API calls, so it should at least be considered an inofficial way.

Routes set this way are cached, so they can be read back from cache, since there's no Z-Wave API to query routes from a node. In addition, priority return routes are now also cached, so they can also be read back.
It should be noted that all this caching is done on a best-effort basis. There might be some edge cases where the controller assigns new routes without Z-Wave JS knowing.

---

Also, this PR introduces a consistent naming scheme for the route assignment methods. Methods that deal with multiple routes have been renamed to plural:
- `assignReturnRoute` -> `assignReturnRoutes` 
- `assignSUCReturnRoute` -> `assignSUCReturnRoutes`
- `deleteReturnRoute` -> `deleteReturnRoutes` 
- `deleteSUCReturnRoute` -> `deleteSUCReturnRoutes`

The original methods still exist, but are now deprecated.

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -602,14 +602,16 @@ assignPrioritySUCReturnRoute(
 -   `assignPriorityReturnRoute` sets the priority route from node `nodeId` to the destination node.
 -   `assignPrioritySUCReturnRoute` does the same, but with the SUC (controller) as the destination node.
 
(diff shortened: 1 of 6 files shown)


AlCalzone:
TODO: Discovering node neighbors seems to do an AssignSUCReturnRoute behind the scenes 🙄

AlCalzone:
@zwave-js-bot automerge

PR #5946, opened 2023-06-26
fix(config): clarify "Hand Button" action for `ZVIDAR Z-CM-V01 Smart Curtain Motor`

brianegge:
PR description here


AlCalzone (changes requested):
Cool, can you work this information into the labels, preferably short and preferably without needing the `"description"` field?
Maybe something like "Pushbutton Behavior", options "Open blinds" / "Close blinds".

AlCalzone (approved):

PR #5956, opened 2023-06-28
feat(config): update Swidget devices to match their June 8 2023 spec

michaelkkehoe:
Swidget have updated their ZWave specification: https://files.swidget.com/Integrations/UserManuals/Swidget_Z-Wave_Insert_Configuration.pdf

This PR updates their device config to match the specification.

* [fix] a number of wrong parameter sizes
* [update] A number of changed default values
* [remove] Parameters that are no longer exposed


AlCalzone:
Is there any specific firmware version where these changes start? Ideally we wouldn't want to break for users that already use these.

michaelkkehoe:
@AlCalzone these devices are currently pre-sale (I have some beta versions). So won't be breaking for the public.

AlCalzone (approved):

PR #5958, opened 2023-06-28
refactor: simplify state machine architecture

AlCalzone:
The state machines for communicating with the controller and nodes are quite involved and very hard to debug, although I rewrote them like 3 times already.
This PR has the goal to massively reduce their architecture. Managing the queue is now done in the driver, instead of the state machine. The driver now also executes the Serial API command machines itself, making everything more obvious.

As a result, 3 out of 4 of the state machines involved with sending commands are gone, and the logic for edge cases and error handling now lives in the driver.

TODO:

- [x] Test that SendDataAbort works
- [x] Test if a transaction may be started before doing it
- [x] Test that S0/S2 nonce stuff still works
- [x] Test that communication with sleeping devices still works
- [ ] Improve error message when a ping does not get moved to the wakeup queue


AlCalzone:
@zwave-js-bot automerge

PR #5966, opened 2023-07-03
fix: do not discard `Transport Service` and `CRC16 CC` for low security

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #5967, opened 2023-07-03
fix: reduce S2 duplicate detection window to 1

### LINKED ISSUE #5965: Reduce S2 duplicate detection window to 1
...or investigate if we can work around the CTT reusing the same sequence numbers for each test case.

AlCalzone:
fixes: #5965


AlCalzone:
@zwave-js-bot automerge

PR #5970, opened 2023-07-05
fix: clear custom SUC return route cache after assigning priority SUC return route

### LINKED ISSUE #5968: [bug] Custom SUC routes cache not updated when setting new SUC priority return route
Dunno if this happens also with non suc return routes too

AlCalzone:
fixes: #5968


AlCalzone:
@zwave-js-bot automerge

PR #5972, opened 2023-07-05
fix: automatically use `...SUC...` variant when assigning custom/priority return routes with the controller as the destination

### LINKED ISSUE #5957: Prevent assigning return routes to the controller without using the ...SUC... methods

AlCalzone:
With this PR, `assignCustomReturnRoutes` and `assignPriorityReturnRoute` now internally call their `...SUC...` variants when called with the controller ID as the destination. This is meant as a transition before preventing it entirely in the next major version --> #5971

fixes: #5957


AlCalzone:
@zwave-js-bot automerge

PR #5987, opened 2023-07-08
feat: add compat flag to override (almost) arbitrary CC API queries

### LINKED ISSUE #1625: Add compat flag to override values reported by a device

AlCalzone:
A lot of device issues are caused by incorrect responses to certain queries. This PR attempts to solve this issue by declaratively overriding what a CC API call returns and which values need to be stored in the value DB. It is set up this way, because device interviews internally use the CC API methods.
By hooking into this process, we simply skip the original API call and simulate the result.

This allows us to avoid interview-specific implementations and provides a framework to solve this issue for most problematic devices.

closes: https://github.com/zwave-js/node-zwave-js/pull/4609
fixes: #1625

### DIFF
--- a/docs/config-files/file-format.md
+++ b/docs/config-files/file-format.md
@@ -423,13 +423,14 @@ A frequent reason for device not "working" correctly is that they respond to que
 				"numDailyRepeatingSlots": 1
 			},
 			// Which values should be stored in the value DB when the API method is called (optional).
-			// The keys are the names of statically defined values for the given CC,
-			// see the CC documentation for available values. Dynamic values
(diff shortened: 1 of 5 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #5991, opened 2023-07-10
fix(config): override schedule slot count for `P-KFCON-MOD-YALE`

### LINKED ISSUE #5778: Add a compat flag to override the schedule slot count
Yale Smart Door Lock  P-KFCON-MOD-YALE (0x0129-0x0007-0x0001) has 20 users, reports 20 schedule slots, but only seems to have a single slot per user (slot ID is ignored).

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/issues/5778


AlCalzone:
@zwave-js-bot automerge

PR #5998, opened 2023-07-10
feat(config): add Heatit Z-HAN2

geirra:
Adds config file for AMSreader from Heatit with Z-wave 800 chip.
Config file is verified and tested

Device is currently not listed in ZWA database.


AlCalzone (approved):
Thanks, did a few minor modifications

PR #5999, opened 2023-07-11
feat(config): add Heatit ZM Dimmer

geirra:
adds support for heatit zm dimmer.
The data on ZWA is wrong here, so have corrected it in the config file.

Parameters and are defined pr manual, tried to use template where i could.


AlCalzone (dismissed):
Did a few changes for consistency and to disambiguate params 10 and 15

AlCalzone (approved):

PR #6027, opened 2023-07-18
fix(config): add new product id to Fakro ZWS12

Sennevds:
Adds extra productId 0x0116 to ZWS12n. Tested this locally and works


Sennevds:
@zwave-js-bot rename commit

AlCalzone (approved):

PR #6024, opened 2023-07-17
fix: answer S0 nonce request during Get-Report command flow

AlCalzone:
This PR fixes a regression that was introduced in https://github.com/zwave-js/node-zwave-js/pull/5958 which caused the driver not to respond to `S0 Nonce Get` requests while it was waiting for a response to its own ongoing query.

Apparently we had no test in our test suite that properly tests an S0 encapsulated Get-Report message exchange.

fixes: https://github.com/zwave-js/node-zwave-js/issues/6015
fixes: https://github.com/zwave-js/zwave-js-ui/issues/3169 (as soon as Z-Wave JS UI bumps)

PR #6028, opened 2023-07-18
fix: handle some commands in a separate "immediate" queue

AlCalzone:
With this change, some commands - specifically nonce reports, supervision reports and Transport Service - are handled in a separate immediate queue. Transactions in that queue may always be started, so they can be interlaced with other commands, for example responding to `NonceGet` or `SupervisionGet` from one node while waiting for another node to send a reply.
Messages generated by each queue are going into another queue - the serial API queue, which simply gets drained as fast as possible.

This should fully restore the functionality of the nested state machines we got rid of in #5958 without their complexity.

As an added bonus, looping through the transaction queue(s) is now much easier on the eyes since it is now a simple for-await-of loop.

TODO:

- [x] restore idle detection
- [x] revert #6024 which is now handled by this change too

### DIFF
--- a/packages/cc/src/cc/Security2CC.ts
+++ b/packages/cc/src/cc/Security2CC.ts
@@ -198,7 +198,7 @@ export class Security2CCAPI extends CCAPI {
 				// Only try sending a nonce once
 				maxSendAttempts: 1,
 				// Nonce requests must be handled immediately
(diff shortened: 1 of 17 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6058, opened 2023-07-24
fix: preserve auto-discovered params that are missing from config file

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6059, opened 2023-07-24
fix: only create auto-discovered partials that exist

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6060, opened 2023-07-24
fix: do not override metadata for existing config params when querying param properties

AlCalzone:
We mainly scan the param properties to auto-discover undocumented parameters. For params where we have a config file, we do not want to override that information with what the device reports.

The way this was previously done resulted in the param information from the config file being partially preserved, but marked with `isFromConfig: false`, which is obviously wrong.


AlCalzone:
@zwave-js-bot automerge

PR #6061, opened 2023-07-24
docs: explain how to use custom log transports

### LINKED ISSUE #2142: Document how to use the default formats for custom log transports
**Checklist:**

- [ ] I am **not** using HomeAssistant. **Or:** a developer has told me to come here.
- [ ] I am **not** using ZWaveJS2MQTT. **Or:** a developer has told me to come here.
- [x] I have read the changelog and my problem was **not** mentioned there.

**Describe the bug**

Since the log files can get pretty big I'm using winston file transport with maxsize/maxFiles configured and this worked great in pre v7 but when I upgraded to version 7.0.0 today, I'm only seeing "undefined" in the log file. I have attached the log file.

- [ ] **No**, it never worked anywhere
- [x] **Yes**, in: pre v7
    
- [x] Pkg

**To Reproduce**

Use this as a custom transport in logConfig:

```
const rotationTransportation = new winston.transports.File({
  filename: './zwave/logs/zwave.log',
  maxsize: 5*1000000,
  maxFiles: 5
})
```

**Logfile:**

[zwave.log](https://github.com/zwave-js/node-zwave-js/files/6195910/zwave.log)

### LINKED ISSUE #3019: Options for integrating zwave-js logging with the parent application's logs
I am building out a project that makes use of zwave-js as one of many home automation focused plugins.  However, when running, zwave-js uses its internal logging to dump preformatted logs straight to the terminal.

I want to implement my own logging solution for my project, and would like a way to capture zwave-js log messages and integrate it with my solution.  This would allow my repl to work while zwave-js is running, and permit me to capture zwave log messages and display them on the web ui.

Ideally, I would be able to receive log packets from a zwave-js event, and re-emit them into my logging framework.  Raw text would be acceptable, but if zwave-js uses key/value ancillary data along with the messages that would be even better.

Exploring around, there is some sort of mechanism for passing custom transports, as hinted at by the existence of https://github.com/zwave-js/log-transports.  However, this does not seem to be documented yet, and I have yet to find an example of how to attach these into zwave-js.

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/issues/3019
fixes: #2142

### DIFF
--- a/docs/_sidebar.md
+++ b/docs/_sidebar.md
@@ -33,6 +33,7 @@
     -   [Send custom messages](usage/custom.md)
     -   [External config DB location](usage/external-config.md)
     -   [Remote serial port over TCP](usage/tcp-connection.md)
+    -   [Custom log transports](usage/log-transports.md)
 
 -   Troubleshooting
 
--- a/docs/api/driver.md
+++ b/docs/api/driver.md
@@ -445,17 +445,26 @@ interface LogConfig {
 }
 ```
 
--   `enable`: If `false`, logging will be disabled. Default: `true`.
+By default, Z-Wave JS has two internal transports, a file transport and a console transport. Both share the following options:
+
+-   `enable`: If `false`, the internal transports will be disabled. Default: `true`.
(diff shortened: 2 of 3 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6029, opened 2023-07-18
fix: infinite loop persisting 4-byte bitfield parameters with the high bit set

AlCalzone:
Looks like we fell into a JavaScript trap here :(

fixes: #6012

PR #6063, opened 2023-07-24
fix: accidental revert of #6029

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6067, opened 2023-07-25
feat: fall back to embedded config directory when looking up `$import` target

### LINKED ISSUE #3471: Load templates from embedded dir if missing from user config dir
I was customizing a device config file, but it failed to load because a template file was not available in the user config directory. I don't have any intention of editing the template file, so I would just like to store the single device file. One of the easy things about the user config directory is I can just throw in a device file w/o having to worry about naming and directory structure. For a file that requires templates that becomes slightly more complicated. It also means if the config file is being used long-term, I don't need to maintain the template, only the single device file.

When loading a user config file and the template file is not found, try to fallback to the embedded template file.

Save the template(s) into the user config dir with an identical directory structure compared to the embedded dir. This reduces the relative ease of customizing a device config though.

**Additional context**
```text
2021-10-05T05:52:32.925Z CONFIG   Error parsing config file zw062.json: Could not parse config file /usr/src/app
                                  /store/config/templates/master_template.json: ENOENT: no such file or director
                                  y, open '/usr/src/app/store/config/templates/master_template.json'
                                  Import stack:
                                    in #base_0-100_nounit
                                    in /usr/src/app/store/config/zw062.json
```

```text
$ pwd
/srv/apps/z2m/data/config
$ ls
14287.json  zw062.json
```

AlCalzone:
This change allows user-defined templates to reference the master template using `~/` without having to copy them into the priority directory. If the referenced template is not found in the priority config directory, it will be looked up in the internal config dir instead.
This also works for manufacturer templates, but they must be referenced with a `~/` path too, e.g. `~/0x1234/templates/manufacturer_template.json` instead of `templates/manufacturer_template.json`.

fixes: #3471


AlCalzone:
@zwave-js-bot automerge

PR #6070, opened 2023-07-25
feat: enable 16 bit node IDs in Serial API communication

AlCalzone:
This PR switches the encoding of node IDs in serial API commands to 16 bits which lays the foundation for Z-Wave Long Range support.

Coincidentally it uncovered a bug in the controller firmware:
https://community.silabs.com/s/question/0D58Y0000AcPqWrSQK/7x-firmware-setpriorityroute-behaves-erratically-when-used-with-16bit-node-ids-host-api-specification-missing-node-id


AlCalzone:
@zwave-js-bot automerge

PR #6071, opened 2023-07-25
docs: fix link in V11 Migration guide

marcus-j-davies:
I noticed an invalid link as I was migrating the NR plug-in to V11.
I _think_  this is correct - as currently its a 404


AlCalzone (approved):
I checked the other links in the docs, leading / is also wrong. I changed it.

PR #6073, opened 2023-07-25
feat: add `getDateAndTime` method to `ZWaveNode`

kdknigga:
This adds the getDateAndTime method to ZWaveNode to query and return a summary of the state of the node's clock.


AlCalzone:
> How can I tell if the node is using DST or not?

We can't. Maybe it makes more sense to just return the dstOffset and standardOffset configured on the device, so we can use that to decide if the device is on the correct offsets.

Alternatively Z-Wave JS includes a helper method `getDSTInfo` to return when the current or next DST is. You can use that to figure out which of the two offsets the device **should** be using. I prefer the other option though.

packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -5994,6 +5994,131 @@ ${formatRouteHealthCheckSummary(this.id, otherNode.id, summary)}`,
 		return true;
 	}
 
+	/**
+	 * Returns the current date, time and timezone (or a subset of those) on the node using one or more of the respective CCs.
+	 */
+	public async getDateAndTime(): Promise<object> {
```
AlCalzone: I suggest the following type as a return type instead of `object`. Put it in `_Types.ts` in the same directory:
```ts
export type DateAndTime = AllOrNone<{
	hour: number;
	minute: number;
	second: number;
}> &
	AllOrNone<{
		year: number;
		month: number;
		day: number;
	}> &
	AllOrNone<{
		dstOffset: number;
		standardOffset: number;
	}>;
```
This makes it clear that you'll get groups of properties, all time properties and/or all date properties and/or all timezone properties.

packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -5994,6 +5994,131 @@ ${formatRouteHealthCheckSummary(this.id, otherNode.id, summary)}`,
...
+		const response: {
+			hour?: number;
+			minute?: number;
+			second?: number;
+			timezone?: number;
+			weekday?: number;
+			day?: number;
+			month?: number;
+			year?: number;
+		} = {};
```
AlCalzone: Then this would be
```suggestion
		const response: DateAndTime = {};
```

packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -5994,6 +5994,131 @@ ${formatRouteHealthCheckSummary(this.id, otherNode.id, summary)}`,
...
+				if (result) {
+					// Time Parameters is all UTC per the spec
+					response.hour = result.getUTCHours();
+					response.minute = result.getUTCMinutes();
+					response.second = result.getUTCSeconds();
+					response.timezone = 0;
+					response.weekday = result.getUTCDay();
+					response.day = result.getUTCDate();
+					response.month = result.getUTCMonth();
+					response.year = result.getUTCFullYear();
```
AlCalzone: But you'll probably have to use `Object.assign` for these assignments
```ts
Object.assign(response, {
  hour: result.getUTCHours(),
  // ...
});
```

packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -5994,6 +5994,131 @@ ${formatRouteHealthCheckSummary(this.id, otherNode.id, summary)}`,
...
+				const result = await timeParametersAPI.get();
+				if (result) {
+					// Time Parameters is all UTC per the spec
+					response.hour = result.getUTCHours();
+					response.minute = result.getUTCMinutes();
+					response.second = result.getUTCSeconds();
+					response.timezone = 0;
+					response.weekday = result.getUTCDay();
+					response.day = result.getUTCDate();
+					response.month = result.getUTCMonth();
```
AlCalzone: Careful, JS months are 0 based for some strange reason.
```suggestion
					response.month = result.getUTCMonth() + 1;
```

AlCalzone (changes requested):

packages/zwave-js/src/lib/node/Node.ts
kdknigga: > ```ts
> export type DateAndTime = AllOrNone<{
> 	hour: number;
> 	minute: number;
> 	second: number;
> }> &
> ```

Clock returns weekday instead of second.

AlCalzone: Then probably this:
```ts
export type DateAndTime = AllOrNone<{
  hour: number;
  minute: number;
}> &
  (
    | { weekday?: Weekday; second?: undefined }
    | { weekday?: undefined; second?: number }
  ) &
  AllOrNone<{
    year: number;
    month: number;
    day: number;
  }> &
  AllOrNone<{
    dstOffset: number;
    standardOffset: number;
  }>;
```

kdknigga:
From a node that supports Time v1 and Schedule Entry Lock v3:

```
{
hour: 13,
minute: 11,
second: 18,
day: 1,
month: 8,
year: 2023,
standardOffset: -360,
dstOffset: -300
}
```

From a node that supports only Clock:

```
{ hour: 13, minute: 53, weekday: 'Tuesday' }
```

AlCalzone (approved):
Thanks! Just had some small additions to the docs I did myself.
Also added the `satisfies DateAndTime` check, which noticed that you were incorrectly converting the weekday to a string. We use the `Weekday` enum elsewhere, so we should stick with it here.

kdknigga:
Thank you!

AlCalzone:
re-opening to trigger build again

Issue #5335, opened 2023-01-16
Second Thermostat Setpoint not being found for Intermatic PE653 z-wave pool controller


bluemantwo:
Is your problem within Home Assistant (Core or Z-Wave JS Integration)?
NO, my problem is NOT within Home Assistant or the ZWave JS integration

Is your problem within Z-Wave JS UI (formerly ZwaveJS2MQTT)?
NO, my problem is NOT within Z-Wave JS UI

Checklist
 I have checked the [troubleshooting section](https://zwave-js.github.io/node-zwave-js/#/troubleshooting/index) and my problem is not described there.

 I have read the [changelog](https://github.com/zwave-js/node-zwave-js/blob/master/CHANGELOG.md) and my problem was not mentioned there.

Describe the bug

I have an Intermatic PE653 pool controller.  Product Manufacturer: 0x0005, Product Identifiers: 0x5045:0x0653 

I had it included on my ISY944i and was able to access both Thermostat setpoints: one for the Pool which was Type Heating (Type 1) and one for the spa which was Type Furnace (Type 7).  When I included the device in my Home Assistant zwave-js, it only finds the 1 setpoint for Heating.  Attached is the log created by doing a re-interview of the device (Device ID 19).  Inclusion starts at 1:45:36 Zulu in the log.  It seem to not make any reference to the other setpoint.  Any ideas?  My PE653 is running firmware v39.

From Manufacturer: 

`The THERMOSTAT_SETPOINT command Class is fully supported as documented in the Z-Wave Command
Class Specification for support setpoint types HEATING # 1 (setpoint type 1) and FURNACE (setpoint type
#7). The remaining setpoint types are unsupported. The HEATING #1 setpoint is used for setting the POOL
target temperature and the FURNACE setpoint is used for setting the SPA target temperature. `

[zwavejs_2023-01-15.log](https://github.com/zwave-js/node-zwave-js/files/10422003/zwavejs_2023-01-15.log)

Also, if it helps, here is a SmartThings device handler that works with both setpoints:

[intermatic-pe653-pool-control-system.groovy.txt](https://github.com/zwave-js/node-zwave-js/files/10468559/intermatic-pe653-pool-control-system.groovy.txt)

AlCalzone:
[...]

PR #6076, opened 2023-07-26
fix: ensure endpoint capability report matches requested endpoint

AlCalzone:
found while debugging https://github.com/zwave-js/node-zwave-js/issues/5335


AlCalzone:
@zwave-js-bot automerge

PR #6077, opened 2023-07-26
fix(config): force use of Multi Channel CC v1 for all versions of PE653

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6086, opened 2023-07-27
feat: include intermediate health check results in progress callback

AlCalzone:
This PR adds a parameter with the most recent health check result to the `onProgress` callback. This way applications can show some results before the health check is done.

for https://github.com/zwave-js/node-zwave-js/issues/4075


AlCalzone:
@zwave-js-bot automerge

PR #6089, opened 2023-07-28
fix: actually set 1 as the CC version when assuming V1 is supported

### LINKED ISSUE #6048: Unable to control old Leviton switches (VRS05/VRS15)
I have several old Leviton switches (and dimmers) around the house - those are of the first generation of Z-Wave from about 10-15 years ago and I've been using them w/o problems with several different controllers and hubs throughout the years.

I'm setting up everything now in Z-Wave JS UI and noticed that I cannot control them either directly from the UI or from HomeAssistant via WebSocket. Moreover, in the UI under "Binary Switch v0" section there's "Current value" field, but there's NO "Target value" at all, which I normally get for other devices with e.g. "Binary Switch v1":

binary-switch-v0
binary-switch-v1

I'm using Zooz 800-series ZST39 LR stick. The problematic devices are Leviton VRS05 and VRS15 switches, but I haven't tried dimmers yet.

They report Z-Wave protocol version 2.9 and Z-Wave chip firmware version "0.4" or "0.5"

I wonder if this is due to "v0" in the "Binary Switch" class:

leviton-vrs15-values

Or if it's a driver issue...

I would appreciate any help and gladly provide logs or try different options. Thanks in advance!

AlCalzone:
fixes: #6048

### DIFF
--- a/packages/cc/src/cc/VersionCC.ts
+++ b/packages/cc/src/cc/VersionCC.ts
@@ -462,6 +462,7 @@ export class VersionCC extends CommandClass {
 					} supports version 1...`,
 					level: "warn",
 				});
+				endpoint.addCC(cc, { version: 1 });
 			}
 		};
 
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6101, opened 2023-07-31
fix: message sequencing between controller and node mocks

AlCalzone:
This PR fixes an issue in integration tests and the mock-server that becomes apparent when the `txDelay` of nodes is low. In these cases, due to how the mock communication was previously set up, the node response would often be handled before the ACK.

We now defer the transmission simulation to an async queue (the "air") to better decouple the mock controller from mock nodes.


AlCalzone:
@zwave-js-bot automerge

PR #6108, opened 2023-08-01
fix(nvmedit): correct NVM during conversion if it incorrectly contains flag for "on other network" 

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6111, opened 2023-08-02
fix: update of `currentValue` after successful supervised turn-on `setValue` call

### LINKED ISSUE #6110: Supervised setValue for Multilevel Switch to 255 with duration changes currentValue to 255

AlCalzone:
This fixes an edge case for supervised `setValue` calls where the `targetValue` is 255 and a duration is specified. In this case we'd always set `currentValue` equal to `targetValue` after the command execution was successful, regardless what `targetValue` actually is.
We now use the same check as for immediately successful supervised commands, and query the current value if targetValue is 255.

fixes: https://github.com/zwave-js/node-zwave-js/issues/6110
fixes: https://github.com/home-assistant/core/issues/97595


AlCalzone:
@zwave-js-bot automerge

Issue #6105, opened 2023-08-01
Binary Switch vundefined


MickaelC26:
Hello everyone, 
I added this json (customized)
[mh-s411-7102.zip](https://github.com/zwave-js/zwave-js-ui/files/12226090/mh-s411-7102.zip)

Then, I included a MCOHome MH-S411 panel switch to Zwave JS UI.
Everything seems fine, but when I turn On the switch (using targetValue or currentValue), the currentVallue change to True then to False one sec later, but the light stay On.
If I turn Off the switch (using targetValue or currentValue), the light switch Off.

I also remarked that, just after the inclusion, I only have the Binary Switch v2 command class, but as soon as I change the switch state once, a new command class appear (Binary Switch vundefined). Those currentValue doesn't have the same problem as Binary Switch v2, but I can't trigger anything using them. (37-1-currentValue or target value won't trigger my switch, they are just following 37-0- state)
 
I also tried witthout the MH-S411's json loaded, but I get the same result.
Could someone enlight me on this issue ?

Thanks in advance !
![image](https://github.com/zwave-js/zwave-js-ui/assets/123379128/2a34e81a-dd49-4852-b800-48f07cae439d)

![image](https://github.com/zwave-js/zwave-js-ui/assets/123379128/7e50db45-f177-4cb0-aede-e81df37c17a6)

[mh-s411 debug info.txt](https://github.com/zwave-js/zwave-js-ui/files/12226184/mh-s411.debug.info.txt)

robertsLando:
@MickaelColombieSnC Did you tried to re-interview your device?

MickaelC26:
Many times, I even tried to exclude/include with three differents controllers, always getting the same issue

AlCalzone:
Seems like the device is reporting from endpoints 1 and 2 after you control endpoint 0. The info above seems to indicate that the device only has one binary switch on endpoint 0 though.

Please make a [driver log](https://zwave-js.github.io/zwave-js-ui/#/troubleshooting/generating-logs?id=driver-logs) of a re-interview, loglevel `debug` and attach it here as a file (drag & drop into the text field).

MickaelC26:
[z-ui_2023-08-01.log](https://github.com/zwave-js/node-zwave-js/files/12228767/z-ui_2023-08-01.log)

Hi, here are the log, I made the reinterview, and then triggered the endpoint 0 that made the 1 and 2 appear.

AlCalzone:
[...]

PR #6120, opened 2023-08-03
feat(config): add new MCOHome MH-S411/S412 models

MickaelC26:
- removed unused command class for both
- added "preserveEndpoints": [1] for MH-S411 (https://github.com/zwave-js/node-zwave-js/issues/6105)

PR description here


packages/config/config/devices/0x015f/mh-s411-7102.json
```diff
@@ -0,0 +1,231 @@
...
+					"label": "Reset",
+					"value": 85
+				}
+			]
+		}
+	],
+	"compat": {
+		"commandClasses": {
+			"remove": {
+				"Scene Actuator Configuration": {
```
AlCalzone, line 220: Why is this being removed? Is it not supported or did you do this because it creates so many objects?

packages/config/config/devices/0x015f/mh-s411-7102.json
```diff
@@ -0,0 +1,231 @@
...
+					"endpoints": "*"
+				},
+				"Battery": {
```
AlCalzone, line 224: Same question. According to your logfile, this is a FLiRS device with a battery. Or is that incorrect?
```
is always listening:   false
is frequent listening: 250ms
```

packages/config/config/devices/0x015f/mh-s411-7102.json
MickaelC26: Hello, it's a non dimming panel switch, I removed them because it creates 950 useless endpoints.

packages/config/config/devices/0x015f/mh-s411-7102.json
MickaelC26: This isn't a battery powered devices, I don't know why it is reconized as this. I removed those endpoint because they are useless aswell. If you know a way to make this product type reconnized correctly, I would happily test it

AlCalzone: > I don't know why it is reconized as this

Because the device says so.

> If you know a way to make this product type reconnized correctly

The manufacturer would have to fix their firmware for this.

We can remove the battery support here, but as far as the mesh is concerned, the device will be treated as a frequently-listening battery-powered device. Meaning you can send it commands as usual without having to wake it, but it won't act as a repeater for other devices.

packages/config/config/devices/0x015f/mh-s411-7102.json
AlCalzone: They aren't useless. They configure which Scene Actuator commands are sent to associated devices.

I found this manual where the functionality is specifically mentioned (page 3)
[62015601.pdf](https://github.com/zwave-js/node-zwave-js/files/12258761/62015601.pdf)
Seems wrong to remove that, although I agree creating hundreds of values isn't an ideal way to expose them.

AlCalzone: Hm on the other hand, there are configuration parameters that expose the same functionality.
I'm fine with removing then.

packages/config/config/devices/0x015f/mh-s411-7102.json
MickaelC26: Well, I'll tell it to the manufacturer. I keep you updated

AlCalzone (approved):
The config parameters could use a bit of cleanup, but since they are consistent with the other variants of this device, I'll accept it.

PR #6121, opened 2023-08-03
fix: improve heuristic to refresh legacy nodes when receiving NIF

AlCalzone:
As reported in https://github.com/zwave-js/node-zwave-js/discussions/6117, marking an association as a lifeline association in the config file seems to prevent the automatic refresh for legacy devices after them sending a NIF.

With this PR, we now also refresh the values if we don't know for certain that any lifeline group sends actuator or sensor commands.


AlCalzone:
@zwave-js-bot automerge

PR #6140, opened 2023-08-08
feat: add node to controller `identify` event callback

raman325:
Adds the node that triggered the identify event to the event callback


AlCalzone (approved):
Thanks

PR #6141, opened 2023-08-10
build: migrate codegen to TS transformers

AlCalzone:
When done, this will allow us to go back to simply invoking TypeScript to build the entire project.

Further reading: https://github.com/nonara/ts-patch/discussions/29


AlCalzone:
We've moved in the opposite direction now that TS is going to be a native binary

PR #6142, opened 2023-08-10
fix: consider redirects when determining firmware update filename

AlCalzone:
This came up with Aeotec where the official URL redirects to a "less pretty" one that contains the filename.


AlCalzone:
@zwave-js-bot automerge

PR #6145, opened 2023-08-10
feat(mock-server): support loading a directory of mocks

AlCalzone:
With this PR, the `mock-server` accepts a directory containing multiple mock files, as an alternative to loading single files.
Each file can define one or more mocks, with the following rules:
- Only a single file may define the controller mock
- Each node ID may only be defined once


AlCalzone:
@zwave-js-bot automerge

PR #6152, opened 2023-08-14
fix: preserve user codes during re-interview when not querying them

### LINKED ISSUE #4979: Keep user codes through re-interview if they are not queried automatically

AlCalzone:
fixes: #4979


AlCalzone:
@zwave-js-bot automerge

PR #6154, opened 2023-08-14
fix: always enable soft-reset on 700+ series controllers

AlCalzone:
Experience has shown that the problems caused by soft reset are limited to 500 series and older controllers (shutting down, reconnecting with a different port, etc...)
On 700+ series, soft-resetting is however required in lots of cases. Here, having soft-reset disabled by the users rather causes problems.

This PR changes how we handle the `enableSoftReset` driver option. On 700+ series, this is now ignored and soft-reset is always performed, regardless of user preference. On 500 series and older, we still respect the setting, unless a stick is known not to support soft reset.


AlCalzone:
/cc @robertsLando
When this lands, please update the config option description to mention that disabling soft reset only affects 500 series and older.

AlCalzone:
@zwave-js-bot automerge

robertsLando:
@AlCalzone done

PR #6155, opened 2023-08-14
fix(config): disable Supervision for Kwikset HC620, add to naughtly list

AlCalzone:
This should fix the issue explained in https://github.com/zwave-js/node-zwave-js/discussions/5566, seeing that Kwikset are not interested in fixing this.


AlCalzone:
@zwave-js-bot automerge

PR #6159, opened 2023-08-14
feat: auto-configure controller region and TX power on startup

### LINKED ISSUE #6149: Set controller region and tx power settings on startup, based on configuration

There have been several occasions where a user upgrades their 700-series controller, and the region and power level settings are reset. For example, the region was US, but the f/w upgrade changes it to EU. Now suddenly everything in the network will be dead, and uninformed users are left wondering how to troubleshoot. Worse, if you're using Home Assistant to upgrade, it does not even allow you to change the region, so your only choice is to change the software.

Allow region and tx settings to be specified in the driver settings. The driver can apply these settings on startup to ensure they are always correct. If the settings are not configured, use the current behavior and don't set it on startup.

Something like:

```
Get RF Region
If RF Region result != region configuration and region configuration is set:
    Set RF Region to configuration value if supported
    If Set RF Region is not supported:
        log an error to indicate region not supported and doesn't match configuration
```

Keep the current behavior.

**Additional context**
This is how all both of the SiLabs controller gateways have been designed (zipgateway and now zpc).

AlCalzone:
fixes: #6149

### DIFF
--- a/docs/api/driver.md
+++ b/docs/api/driver.md
@@ -844,6 +844,23 @@ interface ZWaveOptions extends ZWaveHostOptions {
 		scales: Partial<Record<string | number, string | number>>;
 	};
 
+	/**
+	 * RF-related settings that should automatically be configured on startup. If Z-Wave JS detects
+	 * a discrepancy between these settings and the actual configuration, it will automatically try to
+	 * re-configure the controller to match.
+	 */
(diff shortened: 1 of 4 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6162, opened 2023-08-14
feat: configure default transition duration and volume for each node

### LINKED ISSUE #6107: Store default setValue options on nodes

AlCalzone:
This PR adds two properties to the node class that allows configuring default transition durations (and volume) on a per-node basis. These will be used when a `setValue` call has no corresponding `valueChangeOptions`.

fixes: https://github.com/zwave-js/node-zwave-js/issues/6107


AlCalzone:
@zwave-js-bot automerge

PR #6165, opened 2023-08-15
fix: during startup, query serial API capabilities first

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6166, opened 2023-08-15
fix: include patch version in request to firmware update service

AlCalzone:
I noticed that the firmware versions passed to the update service never included the patch version (x.y.**z**). Now they do.


AlCalzone:
@zwave-js-bot automerge

PR #6169, opened 2023-08-15
fix: disallow firmware version ranges where min > max

### LINKED ISSUE #3775: Missing checks
During the fw split deprecation works, I observed some checks we're missing:

1. Device file fw min was higher than its max
2. A duplicate definition existed because the device was included in one file with a range 0.0-255.255, and another where the range was less than full (e.g. 3.0-5.0).

AlCalzone:
fixes: #3775


AlCalzone:
@zwave-js-bot automerge

Issue #2909, opened 2021-06-25
Maintenance task / notification / alert list


blhoward2:
Add metadata for devices that are unknown in the cache allowing the UI to alert the user that a config file is now available and that the device should be reinterviewed.

Consider doing the same for new custom device files (which would require comparing the custom device file to the device cache each time).

AlCalzone:
Thinking about this in a broader sense, we could add a maintenance task list for nodes. Points could include:
- [ ] New config file -> re-interview necessary --> Already done differently, but could be tied in here
- [ ] Bad node health (building on the statistics, e.g. when a node has a high error count)
- [x] Firmware update available (based on the firmware service we've been discussing)
- [ ] Device spamming
- [ ] Bad signal strength
- [ ] [Slow communication with IP-based controllers](https://github.com/zwave-js/node-zwave-js/issues/7085#issuecomment-2273380734)
- [ ] Timeouts during the interview
- [ ] ... other ideas?

robertsLando:
Hehehe implementing all that points would be awesome! Point 3 require an entire new project to handle firmwares uploads by manufacturers, I can do that but I need to know what specific features we need there

blhoward2:
What about flagging that the device spams us with excessive traffic?

robertsLando:
Is there a way to get the 'signal' status of a device, I mean how good is the communication with it

AlCalzone:
There is - zwave-js doesn't use that info yet.

blhoward2:
How about a known critical flaw that disables the device if we push a flag in a config file? Just as an insurance policy. I’m thinking like on ozw there was a bug that caused all Leviton devices to crash the driver. You disable the device and flag it so the driver at least comes up, then the user knows to downgrade. This would be a last resort for things that absolutely don’t work and will bring down the driver if operated.

Perhaps a lesser Known Issues flag for the rare occasion that we know a device doesn’t work and we can’t work around it, but the driver isn’t endangered.

PR #6170, opened 2023-08-15
feat: store a hash of the device config used during the interview

### LINKED ISSUE #4079: Track which device file was the basis of an interview and flag devices with outdated interviews
Hash the device file used during a device interview and compare the hash value of current device files at startup. Also export a function so that UIs can call this function after device files are pulled (zjs2mqtt). Flag devices with outdated interviews so that UIs can display a reinterview flag or prompt on the device. 

Per prior conversations, we don't want to automatically reinterview because a device could be a battery device or there could be an issue, and all information about a device is lost after triggering a reinterview.

AlCalzone:
This PR implements a hash method that condenses the configuration file used during an interview into a buffer that can easily be compared. This is stored and can be used to keep track of changes to config files which likely require a re-interview to be picked up.

fixes: https://github.com/zwave-js/node-zwave-js/issues/4079
for: https://github.com/zwave-js/node-zwave-js/issues/2909

### DIFF
--- a/docs/api/node.md
+++ b/docs/api/node.md
@@ -736,6 +736,16 @@ Many devices using `Notification CC` do not idle their notification values, sinc
 
 > [!NOTE] This method will only do something if the node supports `Notification CC`, and the selected notification variable has an idle state.
 
+### `hasDeviceConfigChanged`
+
+```ts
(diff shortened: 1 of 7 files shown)


AlCalzone:
@zwave-js-bot automerge

Issue #2723, opened 2021-05-28
Detect an unresponsive stick and reset it


cinadr:
### Version

**Checklist:**

- [x] I am **not** using HomeAssistant.
- [x] I have checked the troubleshooting section and my problem is **not** described there.
- [x] I have read the changelog problem was **not** mentioned there.

Build/Run method

- [ ] Docker
- [ ] PKG
- [ ] Snap package
- [x] Manually built (git clone - yarn install - yarn run build ) Windows x64 Server

zwavejs2mqtt version: 4.4.0.5238242
zwave-js version: 7.7.0

### Describe the bug
After a while the zwavejs.log shows 
```
CNTRLR   Failed to execute controller command after 2/3 attempts. Scheduling next try in 1100 ms.
```
and nothing works or responds. After restarting zwavejs2mqtt all comes back to live.
In Zwavejs2mqtt.log:
```
ERROR ZWAVE: Node x error while updating Neighbors: Failed to send the message after 3 attempts
```
or 
```
ERROR ZWAVE: Error while writing true on 5-37-1-targetValue: Failed to send the message after 3 attempts
```
But the undelying zwavejs error always as above.

### To Reproduce

Start with yarn start and wait for several hours before this appears. Tested for 3-4 days and it is consequently dies every time. Approx since 4.2 version. (Not sure on what version this occured first).

### Expected behavior

Stabile controller operation.

### Additional context

Controller is AEOTEC ZWAVE USB STICK.

### Attached logs
[logs.zip](https://github.com/zwave-js/zwavejs2mqtt/files/6560646/logs.zip)

AlCalzone:
The log seems clean up until the point where it suddenly dies. Did anything else change? USB Port or something?

cinadr:
No nothing to mention or relevant.

cinadr:
Same with version 4.5.
[logs.zip](https://github.com/zwave-js/node-zwave-js/files/6565229/logs.zip)
What to do to get more info on this? Any option to switch to trace?
A simple restart solve the issue, no need to unplug the key or restart the host. 
Might Windows be a problem?

AlCalzone:
I don't think there's anything to be logged when nothing happens. I guess I need to add a detection when the stick no longer responds and soft-reset it.

> Might Windows be a problem?

I've never had any problems with Windows.

cinadr:
One more thing to add here:
[...]

PR #6174, opened 2023-08-16
feat: track controller status, do not mark nodes as dead when controller is jammed

### LINKED ISSUE #4191: Failure to send/execute network commands (status Fail) shouldn't mark the node as dead
Maybe there should be a status OK/Jammed for the controller?

AlCalzone:
In a lot of cases where nodes get temporarily/randomly marked dead, the issue is actually the controller being unable to transmit. However, until now this was not visible without looking at driver logs.

This PR adds a new `status` property and `"status changed"` event to the controller, which are used to expose the controller's status to applications and highlight potential issues. For now, only the `Ready` and `Jammed` statuses are used, `Unresponsive` is reserved for a future PR.

When sending a command fails with `TransmitStatus.Fail`, the corresponding node is no longer marked dead. Instead, the controller gets marked `Jammed`, and sending the command is retried after a second.

fixes: https://github.com/zwave-js/node-zwave-js/issues/4191
prepares for: https://github.com/zwave-js/node-zwave-js/issues/2723

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -1189,6 +1189,29 @@ Returns the ID of the controller in the current network.
 * readonly supportsTimers: boolean
 -->
 
+### `status`
+
+```ts
+readonly status: ControllerStatus
+```
+
+This property tracks the status of the controller. Valid values are:
+
+<!-- #import ControllerStatus from "zwave-js" -->
+
+```ts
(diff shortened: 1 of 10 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6176, opened 2023-08-16
fix: check `maySoftReset` in `softReset()` instead of driver options

### LINKED ISSUE #5190: Investigate if soft reset is done after NVM backup
There are reports that it isn't. However this is necessary for 700 series <7.19

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/issues/5190


AlCalzone:
@zwave-js-bot automerge

PR #6178, opened 2023-08-17
fix: discard `Meter CC` and `Multilevel Sensor CC` reports when the node does not support them

### LINKED ISSUE #5510: Meter CC / Multilevel Sensor CC / ...: Ignore reports on non-supporting nodes
ref: https://github.com/home-assistant/core/issues/77966

This will probably cause breaks for a bunch of devices which don't report all CCs they support, but it should prevent reports being incorrectly attributed to other nodes in case of flipped bits in the incoming frames.

AlCalzone:
This PR helps with situations where reports get corrupted on air and get attributed to a different node ID. We may consider expanding this check to all CCs, but this will likely result in breaks for some users where devices don't report all their capabilities.

fixes: #5510

### DIFF
--- a/packages/cc/src/cc/BinarySensorCC.ts
+++ b/packages/cc/src/cc/BinarySensorCC.ts
@@ -2,8 +2,6 @@ import {
 	CommandClasses,
 	MessagePriority,
 	ValueMetadata,
-	ZWaveError,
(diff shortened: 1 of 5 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6180, opened 2023-08-17
fix: abort inclusion when node is already part of the network

### LINKED ISSUE #4805: [Bug]: Including a device twice makes it inaccessible until excluded and included again
### Checklist

### Deploy method

Manually built (git clone - yarn build - yarn install

### Zwavejs2Mqtt version

6.13.0

### ZwaveJS version

9.5.0

### Describe the bug

Including a device that's already included will make it become inaccessible even if the given name and/or location matches the already included one.

### To Reproduce

Include a zwave device, then, include it once again.

### Expected behavior

Either the inclusion gets declined or the device gets a new name/location, and it still accessible, even if under a new path/id/mqtt topic.

### Additional context

Excluding then Including the device fixes this

AlCalzone:
fixes: #4805


AlCalzone:
@zwave-js-bot automerge

PR #6187, opened 2023-08-18
fix: unknown security class when node does not support S0 and/or S2

### LINKED ISSUE #6098: Security info not showing correctly
Hi,

I wanted to move Aetoec multisenzor 6 from S0 security to no security, but wanted to keep the same node ID. So I manually reset the device and proceed with "Replace failed node" process. Device was successfully added, it works fine, but in GUI I see question mark in security column. I tried also re-interview node, but nothing helped.

How can I fix this? Version is 8.21.2
[Capture](https://github.com/zwave-js/zwave-js-ui/assets/7306524/428291f8-ee46-482f-a9b5-34aedbfba20e)

AlCalzone:
fixes: #6098

### DIFF
--- a/packages/zwave-js/src/lib/node/Node.ts
+++ b/packages/zwave-js/src/lib/node/Node.ts
@@ -137,7 +137,6 @@ import {
 	RssiError,
 	SecurityClass,
 	SupervisionStatus,
-	UNKNOWN_STATE,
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6192, opened 2023-08-21
refactor: validate device IDs before installing firmware update instead of when checking for updates

### LINKED ISSUE #6177: Firmware updates: Check with cached information, verify before installing
Otherwise, battery powered devices might never get updates offered, unless a user actively goes searching.

AlCalzone:
With this PR, checking the availability of firmware updates now uses cached information, instead of requesting it fresh from the nodes. One one hand, this reduces the likelyhood of [this issue](https://github.com/home-assistant/core/issues/80398) happening, on the other hand, it prevents users from having to wake up sleeping devices in strange locations when manually checking for updates.

Since the check for changed information now needs to be done before applying the update, we now store the device information used to request updates in the returned objects. To accomodate for this, the signature of `firmwareUpdateOTA` needed to be changed to accept the full `FirmwareUpdateInfo` object.

fixes: #6177

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -1003,7 +1003,11 @@ interface FirmwareUpdateInfo {
 }
 ```
 
-where each entry in `files` looks like this:
+where the `device` field stores which device the update is for
+
+<!-- #import FirmwareUpdateDeviceID from "zwave-js" -->
+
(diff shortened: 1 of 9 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6207, opened 2023-08-23
fix: transform temporary source file when formatting docs

### LINKED ISSUE #3617: Documentation generation: import not found when used twice
Can be reproduced in controller.md when adding 2nd import of `PlannedProvisioningEntry`

AlCalzone:
Instead of transforming the original program in memory, which messes up subsequent transform attempts, we now copy the relevant source code to a temporary project and transform that.
Using emit transformers would be the cleaner option, but this approach needs less change to the code and works.

fixes: #3617


AlCalzone:
@zwave-js-bot automerge

PR #6209, opened 2023-08-24
fix(docs): move `Driver.interviewNode` to `Node.interview`

marcus-j-davies:
As titled.

I'm not sure what format wizardry happens with the docs, so I may be missing a trick or 2?
This method was moved in V10.


AlCalzone (approved):
Thanks, I found one more mention

PR #6216, opened 2023-08-25
fix(config): extend version range for Vesternet VES-ZW-DIM-001

kris-g:
Update firmwareVersion max to account for newer version of the Vesternet 2-Wire Capable Dimmer module.


AlCalzone:
@martyn-vesternet Can you confirm this?

martyn-vesternet:
@AlCalzone yes no problem with that.

(we currently have a whole new batch of devices in testing, so we'll push some of our own updates to take into account version changes as soon as possible)

@kris-g thanks for the assist :-)

AlCalzone:
Thanks for confirming!

AlCalzone (approved):

PR #6219, opened 2023-08-28
fix: don't abort transaction while still retrying

AlCalzone:
This PR fixes an issue found while trying to reproduce https://github.com/home-assistant/core/issues/98491

Calling `sendCommand` (either directly or indirectly) with `maxSendAttempts > 1` for an unreachable node would result in the Promise being resolved after attempt 1, although the entire transaction would be rejected later.


AlCalzone:
@zwave-js-bot automerge

PR #6224, opened 2023-08-29
test: add regression test for setValue with temporary communication failure

AlCalzone:
I used this to reproduce https://github.com/home-assistant/core/issues/98491
Turns out, https://github.com/zwave-js/node-zwave-js/pull/6219 already fixes that, but having a regression test cannot hurt.


AlCalzone:
@zwave-js-bot automerge

PR #6225, opened 2023-08-29
refactor: remove Sentry error reporting

AlCalzone:
As of this PR, we no longer use Sentry for error reporting. While it occasionally did catch crashes, the vast majority of reported errors were from normal usage, to be handled by applications.

This means we need to find a new way to report missing device configurations.


AlCalzone:
@zwave-js-bot automerge

PR #6230, opened 2023-08-30
fix(config): delay value refresh for ZW500D

AlCalzone:
for https://github.com/zwave-js/node-zwave-js/discussions/6117


AlCalzone:
@zwave-js-bot automerge

PR #6231, opened 2023-08-30
fix: validate start/stop time and date in `Schedule Entry Lock CC` commands

### LINKED ISSUE #5727: Add validation on `Schedule Entry Lock` CC set methods
Seems that `setWeekDaySchedule` and `setYearDaySchedule` has no validation on start/stop values. This means that if I set a start that is after the stop the commands executes correctly

AlCalzone:
fixes: #5727


AlCalzone:
@zwave-js-bot automerge

PR #6233, opened 2023-08-30
feat: optimize order of node communication during startup

AlCalzone:
Prior to this PR, communication with nodes on startup was sequenced purely by node ID, which leads to unresponsive devices delaying the `"ready"` event for responsive devices.

We now use the following heuristic to prioritize devices during this phase:
- Fully-interviewed devices come first (often just need a ping or nothing at all)
- Always listening devices come first, then FLiRS, then sleeping devices
- Among those, the order is determined by the `lastSeen` flag, more recently active devices come first
- Lastly, we sort ascending by node ID

This should ensure that fully-interviewed responsive devices are ready almost immediately on startup, even if they have a high node ID.


AlCalzone:
@zwave-js-bot automerge

PR #6240, opened 2023-09-04
fix: device config has changed when hashes are NOT equal

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6243, opened 2023-09-05
test: implement test runner that can handle the full root-relative paths

AlCalzone:
Until now, executing test files was always a bit awkward, since `ava` expects the paths to be relative to the package, not to the monorepo root.
This PR adds a test runner that "translates" the paths to 1 or more `ava` calls and allows testing the current file with <kbd>F5</kbd>


AlCalzone:
@zwave-js-bot automerge

PR #6254, opened 2023-09-06
fix: argument type for `updateLogConfig` and `updateOptions`

AlCalzone:
Turns out `DeepPartial` drilled a bit too deep into `winston`'s `Transport`s. Found in https://github.com/zwave-js/zwave-js-server/pull/1034


AlCalzone:
@zwave-js-bot automerge

PR #6257, opened 2023-09-06
fix(cc): default to RF unprotected if not given for `Protection CC` V2+

### LINKED ISSUE #6246: Double-check Protection CC encoding
There is a new Inovelli firmware 1.22 that is prompting to update, but is still getting the same error.  
This firmware also appears to address an issue specific to Home Assistant;
_"Fixed the bug that the V1 version of Protection Set Command would cause RF state of Protection to become enabled. This mostly affected Home Assistant users as it is still using V1 of this command class."_

_Originally posted by @Platup in https://github.com/zwave-js/node-zwave-js/issues/5081#issuecomment-1707032509_

AlCalzone:
fixes: #6246

Apparently, the RF protection flag was not set in some situation. We now default to `Unprotected` in this case.

### DIFF
--- a/packages/cc/src/cc/ProtectionCC.ts
+++ b/packages/cc/src/cc/ProtectionCC.ts
@@ -508,7 +508,7 @@ export class ProtectionCCSet extends ProtectionCC {
 			);
 		} else {
 			this.local = options.local;
-			this.rf = options.rf;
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6261, opened 2023-09-07
feat: allow putting mock configs outside the repo

AlCalzone:
With this PR, configs for the `mock-server` can be put in locations where they don't have direct access to the `zwave-js` node-modules. When this is desired, the mock configs need to be refactored so their default export is a function that accepts an object with a scoped `require` function:

**Before:**
```js
const { CommandClasses } = require("@zwave-js/core");
const { ccCaps } = require("@zwave-js/testing");
const { ThermostatMode, ThermostatSetpointType } = require("zwave-js");

module.exports.default = {
	// config
}
```

**After:**
```js
module.exports.default = ({ require }) => {
	const { CommandClasses } = require("@zwave-js/core");
	const { ccCaps } = require("@zwave-js/testing");
	const { ThermostatMode, ThermostatSetpointType } = require("zwave-js");
		
	return {
		// config
	}
}
```

### DIFF
--- a/packages/zwave-js/bin/mock-server.js
+++ b/packages/zwave-js/bin/mock-server.js
@@ -82,7 +82,12 @@ Each node ID may only be used once in mock configs. Node ID ${nodeConfig.id} is 
  */
 function getConfig(filename) {
 	if (filename.endsWith(".js")) {
-		return require(filename).default(childRequire);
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6263, opened 2023-09-08
feat(config): add Heatit Z-TRM6

geirra:
This adds config file for Heatit floor thermostat Z-TRM6


geirra:
There seems to a issue with setpoints, should be resolved before merge.

geirra:
as for the setpoints, it seems like we have to define a decimal point.
So i can set 24,5 but not 24,0 as that is translated to 24

Tried looking for a compat flag, but didn´t find any

geirra:
Added compat override float, seems to work fine.

will do some more testing

packages/config/config/devices/0x019b/z-trm6.json
```diff
@@ -0,0 +1,469 @@
...
+			"minValue": 30,
+			"maxValue": 65535,
+			"defaultValue": 840,
+			"unit": "seconds",
+			"unsigned": true
+		},
+		{
+			"#": "21",
+			"label": "Action After Error",
+			"description": "Device will attempt to turn on again after timeout period",
```
AlCalzone:
```suggestion
			"label": "Turn On Delay After Error",
```

packages/config/config/devices/0x019b/z-trm6.json
```diff
@@ -0,0 +1,469 @@
...
+			"valueSize": 2,
+			"minValue": 0,
+			"maxValue": 65535,
+			"defaultValue": 0,
+			"unit": "seconds",
+			"unsigned": true,
+			"options": [
+				{
+					"label": "Turn off (Display error)",
```
AlCalzone:
```suggestion
					"label": "Stay off (Display error)",
```

AlCalzone: together with the label change above this seems to make more sense?

packages/config/config/devices/0x019b/z-trm6.json
```diff
@@ -0,0 +1,469 @@
...
+			"valueSize": 1,
+			"minValue": 1,
+			"maxValue": 10,
+			"defaultValue": 2,
+			"unit": "10%",
+			"unsigned": true
+		},
+		{
+			"#": "26",
+			"label": "Thermostat State Update Interval",
```
AlCalzone:
```suggestion
			"label": "Thermostat State Report Interval",
```

packages/config/config/devices/0x019b/z-trm6.json
```diff
@@ -0,0 +1,469 @@
...
+			"valueSize": 2,
+			"minValue": 0,
+			"maxValue": 65535,
+			"defaultValue": 43200,
+			"unsigned": true,
+			"unit": "seconds",
+			"options": [
+				{
+					"label": "On change",
```
AlCalzone:
```suggestion
					"label": "Changes only",
```

packages/config/config/devices/0x019b/z-trm6.json
```diff
@@ -0,0 +1,469 @@
...
+				},
+				{
+					"label": "Enabled",
+					"value": 1
+				}
+			]
+		},
+		{
+			"#": "29",
+			"label": "Size of Load",
```
AlCalzone:
```suggestion
			"label": "Load Power",
```
maybe?

packages/config/config/devices/0x019b/z-trm6.json
```diff
@@ -0,0 +1,469 @@
...
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 99,
+			"defaultValue": 0,
+			"unsigned": true,
+			"unit": "100W",
+			"options": [
+				{
+					"label": "Uses power metering value",
```
AlCalzone:
```suggestion
					"label": "Use measured value",
```

packages/config/config/devices/0x019b/z-trm6.json
```diff
@@ -0,0 +1,469 @@
...
+			"options": [
+				{
+					"label": "Enable",
+					"value": 0
+				},
+				{
+					"label": "Disable",
+					"value": 1
+				}
+			]
```
AlCalzone: We often call this Local Control. Would this make sense here too? In that case, the meaning of Enable and Disable would have to be switched.

packages/config/config/devices/0x019b/z-trm6.json
```diff
@@ -0,0 +1,469 @@
...
+		},
+		{
+			"#": "3",
+			"label": "Sensor Value(sen)",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 7,
+			"defaultValue": 0,
+			"unsigned": true,
+			"unit": "KΩ",
```
AlCalzone:
```suggestion
			"unit": "kΩ",
```

packages/config/config/devices/0x019b/z-trm6.json
```diff
@@ -0,0 +1,469 @@
...
+				{
+					"label": "33",
+					"value": 4
+				},
+				{
+					"label": "47",
+					"value": 5
+				},
+				{
+					"label": "6,8",
```
AlCalzone: Is this supposed to be a comma between 6 and 8? Seems like a strange sequence of numbers in that case.

packages/config/config/devices/0x019b/z-trm6.json
```diff
@@ -0,0 +1,469 @@
...
+			"options": [
+				{
+					"label": "Uses power metering value",
+					"value": 0
+				}
+			]
+		}
+	],
+	"compat": {
+		"overrideFloatEncoding": {
```
AlCalzone, line 447:
```suggestion
		// The device doesn't accept setpoints with automatic precision
		"overrideFloatEncoding": {
```

packages/config/config/devices/0x019b/z-trm6.json
```diff
@@ -0,0 +1,469 @@
+{
+	"manufacturer": "Heatit",
+	"manufacturerId": "0x019b",
+	"label": "Z-TRM6",
+	"description": "Floor thermostat",
```
AlCalzone:
```suggestion
	"description": "Floor Thermostat",
```

packages/config/config/devices/0x019b/z-trm6.json
```diff
@@ -0,0 +1,469 @@
...
+			]
+		},
+		{
+			"#": "16",
+			"label": "Active Display Brightness",
+			"valueSize": 1,
+			"minValue": 1,
+			"maxValue": 10,
+			"defaultValue": 10,
+			"unit": "10%",
```
AlCalzone: All units with a number in them should have a space between the number and the actual unit
```suggestion
			"unit": "10 %",
```
please do the same for the other params too.

AlCalzone (changes requested):
Thanks - a few remarks.

packages/config/config/devices/0x019b/z-trm6.json
AlCalzone: Ok seems to be intended. Weird... We should turn it into a decimal point for consistency though.
```suggestion
					"label": "6.8",
```

packages/config/config/devices/0x019b/z-trm6.json
geirra: agreed

packages/config/config/devices/0x019b/z-trm6.json
geirra: should be fixed for all params this now

packages/config/config/devices/0x019b/z-trm6.json
AlCalzone: @geirra I think you missed this one

packages/config/config/devices/0x019b/z-trm6.json
AlCalzone: @geirra I think you missed this one

geirra: sorry, they where hidden in the filter, should be fixed now.

AlCalzone (approved):
Thanks. I did the last tweaks myself.

PR #6266, opened 2023-09-11
fix: restart transaction when moving it to wakeup queue

AlCalzone:
Fixes the issue reported in https://github.com/home-assistant/core/issues/99736

Not sure which specific change caused this, but the message generator for transactions moved to the wakeup queue after failing with `NoACK` was no longer reset. This resulted in them immediately resolving to `undefined` when continuing. For queries, this also resulted in the command time to include the time the node was asleep, causing an extremely long timeout waiting for the node's response, which would never come since no command was actually sent.


AlCalzone:
@zwave-js-bot automerge

PR #6296, opened 2023-09-21
fix: attempt soft-resetting controller when callback times out

### LINKED ISSUE #6260: All send attempts fail after aborting SendData with timed out callback
https://gist.github.com/robarnold/c94d61f91fa4c78002e87e4805c4ca95 has a log that shows this. We should consider soft-resetting the controller when this happens for a longer stretch of time.

AlCalzone:
This PR implements a workaround for controllers that get stuck while transmitting and never respond with the SendData callback. According to the Serial API specs, this
> is an exception that should never happen

but it has recently started happening frequently on the 7.19.x firmware line. Turns out that just aborting the transmission isn't enough to recover. As a result, we now use the existing recovery mechanism for missing ACKs from an unresponsive controller and destroy the driver when soft-resetting isn't enough.

fixes: https://github.com/zwave-js/node-zwave-js/issues/6260 (hopefully)


AlCalzone:
@zwave-js-bot automerge

PR #6297, opened 2023-09-21
chore: implement ESLint rules for correct CC implementations

AlCalzone:
We used to have this as a custom script, but it wasn't active by accident and doing it via ESLint is better anyways.


AlCalzone:
@zwave-js-bot automerge

PR #6310, opened 2023-09-22
feat(config): add Sunricher SR-ZV2833PAC

### LINKED ISSUE #5850: Missing device configuration: sunricher SR-ZV2833PAC
### Checklist

### Which device is missing?

sunricher SR-ZV2833PAC

### Manufacturer ID

0x0330

### Product Type

0x0400

### Product ID

0xd105

### Firmware Version

1.0.0

no

### Do you have a manual?

https://www.sunricher.com/media/resources/manual/SR-ZV2833PAC-Instruction.pdf
[debug-info.json.txt](https://github.com/zwave-js/node-zwave-js/files/11581839/debug-info.json.txt)

apella12:
Found this in OpenSmartHouse DB Should close #5850 as help wanted, so giving it a go.  Do not have the device, so can't test

PR description here


apella12:
I see an issue with the allowManualEntry from the wizard extracted file.  Not sure exactly what that means.  It seems to be picking up the "Limit to Options" flag in the OpenSmartHouse DB.  What that means in OpenHAB terms is only the options shown are allowed, whereas I literally read allowManualEntry as meaning as the opposite, so need to understand better to fix without messing this up for a device owner.

packages/config/config/devices/0x0330/zv2833pac.json
```diff
@@ -0,0 +1,211 @@
...
+		"max": "255.255"
+	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 5,
+			"isLifeline": true
+		},
+		"2": {
+			"label": "Launch 1",
```
AlCalzone, line 23: Do we know what those additional associations do? Maybe there's a better name? Or are these just 4 control groups?

packages/config/config/devices/0x0330/zv2833pac.json
```diff
@@ -0,0 +1,211 @@
...
+		"5": {
+			"label": "Launch 4",
+			"maxNodes": 5
+		}
+	},
+	"paramInformation": [
+		{
+			"#": "2",
+			"label": "External Switch Type",
+			"description": "External Switch Type",
```
AlCalzone:
```suggestion
```
We avoid descriptions that duplicate or merely restate the labels.
Same for several others.

packages/config/config/devices/0x0330/zv2833pac.json
```diff
@@ -0,0 +1,211 @@
...
+					"value": 2
+				}
+			]
+		},
+		{
+			"#": "3",
+			"label": "Duration of the Multilevel Start Level Change Up/Down",
+			"description": "Set the duration of the Multilevel Start Level Change Up/Down",
+			"valueSize": 1,
+			"minValue": 0,
```
AlCalzone, line 67: This needs a unit

packages/config/config/devices/0x0330/zv2833pac.json
```diff
@@ -0,0 +1,211 @@
...
+			"defaultValue": 8
+		},
+		{
+			"#": "4",
+			"label": "Associations in Z-Wave network Security Mode",
+			"description": "Associations in Z-Wave network Security Mode",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 15,
+			"defaultValue": 15
```
AlCalzone: I guess this is meant to toggle the use of Security for the 4 non-lifeline groups? In this case, this should be split into 4 [partial parameters](https://zwave-js.github.io/node-zwave-js/#/config-files/file-format?id=partial-parameters), one per association group.

packages/config/config/devices/0x0330/zv2833pac.json
```diff
@@ -0,0 +1,211 @@
...
+				},
+				{
+					"label": "button pressed and held",
+					"value": 4
+				},
+				{
+					"label": "button released",
+					"value": 8
+				}
+			]
```
AlCalzone: This looks like they are meant to be combined, so this needs one partial parameter per scene (Pressed 1x, Pressed 2x, Press and hold, Release).
I'm sure there is a similar device already that this definition can more or less be copied from.

Same goes for params 6-8, but they can most likely be `$import`-ed from param 5 -> https://zwave-js.github.io/node-zwave-js/#/config-files/using-templates?id=example-3

packages/config/config/devices/0x0330/zv2833pac.json
```diff
@@ -0,0 +1,211 @@
...
+			"label": "Function of the connected switches",
+			"description": "Function of the connected switches",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 2,
+			"defaultValue": 0,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "function 1",
```
AlCalzone: is there an explanation what functions 1-3 mean?

AlCalzone (changes requested):

apella12:
Just a comment before I get into this; This disabuses me of the idea I could rip through 4-5 of these by simply extracting from the Open Smart House DB.  I was starting to suspect that was too easy to be true.  Now I'll have to read the docs ;-)

AlCalzone:
Yeah, we have probably the highest standards of all projects when it comes to config params.

packages/config/config/devices/0x0330/zv2833pac.json
apella12: From the manual posted, these are four control groups that do different things based on the based on the switch type - param 2 and function param 9.  This is a complicated device and the owner is going to have to read the manual several times (IMO)

packages/config/config/devices/0x0330/zv2833pac.json
apella12: Done

packages/config/config/devices/0x0330/zv2833pac.json
apella12: Done

packages/config/config/devices/0x0330/zv2833pac.json
apella12: I did my best (at my current level of skill) in the next commit.  Let me know if okay.

packages/config/config/devices/0x0330/zv2833pac.json
apella12: ditto from comment above

packages/config/config/devices/0x0330/zv2833pac.json
apella12: Like comment above the functions mean something different based on the association groups and run several paragraphs

AlCalzone (dismissed):
Thanks, I had a few more changes for consistency, but I did them myself.

AlCalzone (approved):

PR #6312, opened 2023-09-23
feat: make timeout to send devices back to sleep configurable, lower default to 250 ms

apella12:
On change from 1000 to 500: I do not think shorter sleep is worth a user parameter that then has to be checked. Active battery time is 12.5mA; sleep 1uA, so 1/2 second savings is 26 extra daya a year (700 device). More for 500 devices.

By the time this is called all pending messages are cleared, so all the work is done.

### DIFF
--- a/packages/zwave-js/src/lib/driver/Driver.ts
+++ b/packages/zwave-js/src/lib/driver/Driver.ts
@@ -253,6 +253,7 @@ const defaultOptions: ZWaveOptions = {
 		report: 1000, // ReportTime timeout SHOULD be set to CommandTime + 1 second
 		nonce: 5000,
 		sendDataCallback: 65000, // as defined in INS13954
+		sendToSleep: 250, // The default should be enough time for applications to react to devices waking up
 		refreshValue: 5000, // Default should handle most slow devices until we have a better solution
 		refreshValueAfterTransition: 1000, // To account for delays in the device
 		serialAPIStarted: 5000,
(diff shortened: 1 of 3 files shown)


AlCalzone (approved):
I lowered the default timeout to 250ms, since there should really not be a reason to keep the device awake for longer, other than giving the application more time to do something.

Also, I decided to make this configurable after all, in case someone has a weird edge case that needs a higher timeout.

AlCalzone:
@zwave-js-bot automerge

apella12:
First thanks. Second with the OpenHab binding I have been using 250 ms.  I wasn't sure you would buy that so hedged originally at 500 ms. ;-)

Lastly, at least with my understanding, the description `The default should be enough time for applications to react to devices waking up` could tempt users to change this is for unrelated problems. What I assumed the flow was 1) wakeup notice by device 2) after acking wakeup notice, ZJS checks for queued messages, sends them and waits until completion (with Acks) 3) Queue empty, wait 250 ms and send back to sleep. Maybe `"Sleep delay after all queued messages processed"` I'm also assuming that it the queued commands don't get processed, ZJS marks the node as asleep again and the sleep message is never sent? Anyway I don't understand the "for applications to react" means. I didn't want to open a can of worms.

AlCalzone:
Maybe what I wrote is worded badly. `Enough time for applications to react` refers to the time applications get after wakeup or each processed command to queue additional commands before the device gets sent back to sleep.

E.g. they might want to refresh a value, or do things based on the result of a command. Considering that the chain to the actual application might be quite long (e.g. `node-zwave-js` -> `Z-Wave JS UI` -> `Z-Wave JS Server` -> TCP connection -> HA Python Library -> Home Assistant Addon -> Home Assistant`) , and Z-Wave JS may be running on slower hardware, I didn't dare to go much lower

apella12:
Thanks.  That makes sense.  I have observed short periods with an empty queue (at least in the OpenHab zwave binding) while the next command gets created, (generally with the initial interview) where the data from the prior command was needed to formulate the next message.  I'm assuming that besides the empty queue check, zwave-js also checks if the device is fully configured before sending the sleep?
I was not advocating for lower. In the battery PR that was merged in OH zwave, I used 500 ms, but was testing 250 ms locally. I have some Sensative sticks that make me paranoid about battery life. ;-)

AlCalzone:
> I'm assuming that besides the empty queue check, zwave-js also checks if the device is fully configured before sending the sleep?

Not explicitly, but as long as the interview isn't completed, there should always be pending messages on the queue.

PR #6317, opened 2023-09-25
fix(config): update some descriptions for Shelly Wave 1PM

QubinoHelp:
fix: minor changes in parameters

### DIFF
--- a/packages/config/config/devices/0x0460/qnsw-001P16.json
+++ b/packages/config/config/devices/0x0460/qnsw-001P16.json
@@ -56,7 +56,7 @@
 		{
 			"#": "17",
 			"$import": "~/templates/master_template.json#state_after_power_failure_prev_off",
-			"label": "O1: State After Power Failure",
+			"label": "O1: State After Power Failure"
 		},
 		{
 			"#": "19",
--- a/packages/config/config/devices/0x0460/qnsw-001P16.json
+++ b/packages/config/config/devices/0x0460/qnsw-001P16.json
@@ -57,9 +57,6 @@
 			"#": "17",
 			"$import": "~/templates/master_template.json#state_after_power_failure_prev_off",
 			"label": "O1: State After Power Failure",
-			"minValue": 0,
-			"maxValue": 1,
-			"unsigned": true
 		},
 		{
 			"#": "19",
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0460/qnsw-001P16.json
```diff
@@ -70,15 +74,16 @@
...
+			"description": "The timer resets to zero each time the Device receives an OFF command",
 			"valueSize": 2,
 			"minValue": 0,
 			"maxValue": 32535,
 			"defaultValue": 0,
 			"unsigned": true
 		},
 		{
-			"#": "25",
-			"label": "O1: Auto-On/Off Timer Unit",
```
AlCalzone, line 81: I switched 23 and 25 on purpose because it IMO makes sense to have the unit right next to the timers.

packages/config/config/devices/0x0460/qnsw-001P16.json
```diff
@@ -56,11 +56,15 @@
 		{
 			"#": "17",
 			"$import": "~/templates/master_template.json#state_after_power_failure_prev_off",
-			"label": "O1: State After Power Failure"
+			"label": "O1: State After Power Failure",
+			"minValue": 0,
+			"maxValue": 1,
+			"unsigned": true
```
AlCalzone:
```suggestion
```
These are unnecessary. The imported template already contains everything that's needed.

packages/config/config/devices/0x0460/qnsw-001P16.json
```diff
@@ -250,6 +255,7 @@
 		{
 			"#": "201",
 			"label": "Serial Number: Part 1",
+			"description": "This parameter contains a part of devices serial number.",
```
AlCalzone, line 255: IMO these descriptions are redundant with the label.

AlCalzone (changes requested):

packages/config/config/devices/0x0460/qnsw-001P16.json
QubinoHelp: understood

packages/config/config/devices/0x0460/qnsw-001P16.json
QubinoHelp: Correct, will remove it in the next update, I just wanted to get trough with the first one, so that I can prepare config files for the rest devices.

AlCalzone: Works for me. Let me know if you're done with this one. The test failure is due to a formatting error I can fix.

QubinoHelp: That's ok for me, will add other two devices later this week, you may fix the error.

AlCalzone (approved):

PR #6322, opened 2023-09-25
chore: implement ESLint rule to avoid unnecessary min/maxValue in config parameters

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6336, opened 2023-09-28
feat(config): add Eurotronic Comet Z (700 series)

### LINKED ISSUE #5464: Missing device configuration: Eurotronic Comet Z-Wave
### Checklist

### Which device is missing?

Eurotronic Comet Z-wave

### Manufacturer ID

0x0148

### Product Type

0x0004

### Product ID

0x0003

### Firmware Version

11.26.4

https://products.z-wavealliance.org/products/1601

### Do you have a manual?

https://manual.zwave.eu/backend/make.php?lang=en&sku=EURECOMET&cert=

*The device is detected but manufacturer does not show in Home Assistant, looks like a product ID increment as the config DB has product 1,2 and not 3*
The parameters are enumerated in discovery, as below, but are not editable or named to edit once interview completes (attached)

2023-02-14T18:03:52.164Z CNTRLR « [Node 113] received information for parameter #1:
                                  parameter name:      LCD_invert
                                  value format:        UnsignedInteger
                                  value size:          1 bytes
                                  min value:           0
                                  max value:           1
                                  default value:       0
                                  is read-only:        false
                                  is advanced (UI):    false
                                  has bulk support:    true
                                  alters capabilities: false
comet z no parameter edit
comet z

apella12:
This is intended to close #5464

PR description here
https://[manual.zwave.eu/backend/make.php?lang=en&sku=EURECOMET&cert=](https://manual.zwave.eu/backend/make.php?lang=en&sku=EURECOMET&cert=)


AlCalzone (approved):
Thanks - I made a few changes:
- It seems that this version is the 700 series revision, while the old one was 500. I clarified that in case it becomes relevant in the future
- As we don't have a config file for versions <= 11 and the fingerprint is different from other variants of the device, I extended the version range. I prefer not having gaps in the detection when we don't know that it is necessary.
- Added units, removed unnecessary descriptions, etc.
- Importing the `Enable/Disable` template and then changing the options seems wrong. I inlined the template.

cadwizzard:
Thanks!!
This is the 700 version, and a great reason to swap out the devices from the predecessor, the (Eurotronic) Spirit, (Also re-branded Aeotec)

jakubbaginski:
Hi. 
It seems that the basic commands are not compatible and require similar compatibility section as for sprint - off/heat/energy/....

PR #6338, opened 2023-09-28
chore: extend auto-unsigned rule to exclusive predefined options

AlCalzone:
As a follow-up to #6325, the `auto-unsigned` rule now also considers predefined options with `allowManualEntry: false`.


AlCalzone:
@zwave-js-bot automerge

PR #6345, opened 2023-09-29
chore: add ESLint rule to enforce consistency in device config files and prevent some common errors

AlCalzone:
This PR adds the following checks as ESLint rules:

- Title Case and Sentence case is used consistently where required by the style guide
- Config parameter units are consistent with the style guide
- CC names are spelled consistently, e.g. `Binary Switch` instead of `Switch Binary`
- Strings in config files are not surrounded by whitespace
- "empty" parameter descriptions like `""`, `"0"` are not specified
- option labels don't re-specify their value, e.g. `"0: Disable"`
- Option labels and descriptions don't mention that they are the default -> use `defaultValue` instead


AlCalzone:
@zwave-js-bot automerge

PR #6359, opened 2023-10-02
fix: normalize result of `Controller.getAvailableFirmwareUpdates` to always include `channel` field

raman325:
The V1 API for the firmware update service doesn't include `channel` in the payload so we have to insert it


AlCalzone (changes requested):
I think we should rather normalize the response to always include the channel here:
https://github.com/zwave-js/node-zwave-js/blob/f43abb0fc92524ffee83df4e0cbcc39cc8ec97e4/packages/zwave-js/src/lib/controller/FirmwareUpdateService.ts#L226-L229

raman325:
good idea, I will adjust this PR accordingly

AlCalzone (approved):

PR #6361, opened 2023-10-04
fix: limit number of retries when sending a command results in the transmit status `Fail`

AlCalzone:
The controller immediately responding with a transmit status of `Fail` usually means that it is unable to transmit due to being RF jammed. Apparently there are other, non-temporary reasons for this status, so our logic of permanently retrying would result in an infinite loop.

With this PR, commands in this situation are retried up to 5 times before giving up. This should resolve some of the issues users are experiencing in the v12 release.


AlCalzone:
@zwave-js-bot automerge

PR #6362, opened 2023-10-04
fix: avoid calling `assignReturnRoutes` with controller as target

### LINKED ISSUE #6357: Asking for a "rebuild routes" on a particular node triggers error message
### Checklist

### Deploy method

Snap

### Z-Wave JS UI version

9.0.3

### ZwaveJS version

12.0.2

### Describe the bug

When trying to ask for a rebuild route on a particular node (I mean by going via the node advanced options), there is an error stating something about SUC ... that ... literally SUCks :D 

image

I guess it's something related to the recent revamp of the backend API.

### To Reproduce

Go into the dashboard of devices
Select one victim amongst the available nodes
Expand the tab of that node
On the right upper part of that node tab, click on Advanced
Then Advanced Options
Then "Rebuild routes"

### Expected behavior

Routes are getting rebuilt without error message.

### Additional context

As a side node, rebuilding ALL the routes, so via the option on the right lower corner of screen (blue circle with lines) , so I mean rebuilding all the routes in the network seems to work.
BUT 

The animation of the turning circle in the status column doesn't show as it used to.
No turning blue circle to indicate which nodes are still rebuilding their routes.

image

AlCalzone:
fixes: #6357

### DIFF
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -4175,7 +4175,7 @@ supported CCs: ${
 			}
 
 			// 2. re-create the SUC return route, just in case
-			node.hasSUCReturnRoute ||= await this.assignSUCReturnRoutes(nodeId);
+			node.hasSUCReturnRoute = await this.assignSUCReturnRoutes(nodeId);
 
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6364, opened 2023-10-04
fix(config): clarify parameters and units for Everspring AN158

EmielWitteman:
PR description here


AlCalzone (changes requested):
Thanks, but please use the `"unit"` property for this, not the description.
https://zwave-js.github.io/node-zwave-js/#/config-files/file-format?id=paraminformation

We don't have a separate multiplier, so `"100ms"` would be the correct unit for param 1 here.

AlCalzone:
@zwave-js-bot automerge

AlCalzone (approved):

PR #6372, opened 2023-10-05
fix: move workaround for 500 series controller bug to master template, enable for Zooz ZST10

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6373, opened 2023-10-05
fix: limit recovery after missing callback to `SendData` commands

### LINKED ISSUE #6371: Exclude some commands from automatic recovery from missing callback
Some can repeatedly fail and end up in a restart loop:
- RequestNodeInfo: https://github.com/zwave-js/node-zwave-js/pull/6244#issuecomment-1748676807

AlCalzone:
Doing it for others may depend on correct end node behavior a bit too much.

Fixes: #6371


AlCalzone:
@zwave-js-bot automerge

PR #6375, opened 2023-10-06
fix(config): add params 7, 18, 19 to Zooz ZEN71 FW 10.20

cybergrimes:
Zooz has added parameters 18 and 19 to the ZEN71 device beginning with firmware 10.20 for 700 series and firmware 3.10 for 800 series. I believe I've isolated the options for both versions using conditional parameter settings "firmwareVersion >= 3.10 && firmwareVersion < 10.00 && firmwareVersion >= 10.20"

ZEN71 changelog: https://www.support.getzooz.com/kb/article/1102-zen71-700-series-on-off-switch-change-log/

ZEN71 parameter settings: https://www.support.getzooz.com/kb/article/535-zen71-on-off-switch-700-advanced-settings/

thank you!


AlCalzone (approved):
Thanks!
I had to fix the condition as you mixed up an AND (&&) with an OR (||), resulting in an impossible condition. Also Zooz firmware versions are weird, because they went from 10 to 2 to 3, so your original condition also missed the 2.x release line.

Added param 7 on top.

PR #6378, opened 2023-10-09
fix: increase `response` timeout to 30 seconds and maximum to 60 seconds

### LINKED ISSUE #6374: Consider upping the range and default for the `response` timeout
Some 500 series controller seem to take a long time before sending this response (although the correct behavior would be responding immediately and waiting before the callback).

AlCalzone:
Not sure why, but some 500 series controllers take upwards of 10 seconds sometimes to respond after trying to communicate with a node, even if successfully. Having a lower timeout has no benefit, but will cause the transmission to abort earlier.

fixes: https://github.com/zwave-js/node-zwave-js/issues/6374


AlCalzone:
@zwave-js-bot automerge

PR #6388, opened 2023-10-11
fix: add compat flag to set time as UTC, enable for Namron thermostats

### LINKED ISSUE #6386: setting clock: timezone applied twice
### Checklist

### Deploy method

Docker

### Z-Wave JS UI version

9.0.3.c37f5a4

### ZwaveJS version

12.0.2

### Describe the bug

Z-Wave JS sends clock to nodes, applying its own timezone, also to devices which already have a timezone setting. Therefore, nodes will show time = UTC + timezone_controler + timezone_node (one timezone too much).

### To Reproduce

How to reproduce: in Z-WaveJS UI -> [node] -> advanced -> set clock, I push clock from server to nodes. Operation completes successfully, and nodes get updated. However, device shows wrong clock, off by yet-another timezone setting. In my case: Z-WaveJS UI is deployed on docker, setting timezone (UTC+1). Nodes are thermostats, with a configured setting of the same timezone (UTC+1), among their node configuration.

### Expected behavior

Server could optionally send clock, without timezone correction (i.e. in UTC). Having nodes in UTC would not be the best workaround, because of automation (programmed on node).

### Additional context

Controller is an usb, Silicon Labs, Aeotec 7-Stick, f/w. 7.19.1

AlCalzone:
fixes: #6386

### DIFF
--- a/docs/config-files/file-format.md
+++ b/docs/config-files/file-format.md
@@ -506,3 +506,8 @@ By default, `Multilevel Switch CC::Set` commands are ignored, because they are m
 ### `treatDestinationEndpointAsSource`
 
 Some devices incorrectly use the multi channel **destination** endpoint in reports to indicate the **source** endpoint the report originated from. When this flag is `true`, the destination endpoint is instead interpreted to be the source and the original source endpoint gets ignored.
+
+### `useUTCInTimeParametersCC`
+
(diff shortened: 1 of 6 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6400, opened 2023-10-15
fix: export `ConfigurationCCAPISetOptions`

raman325:
Need this type available for the server to fix the `setRawConfigParameterValue` command so we can address https://github.com/zwave-js/certification-backlog/issues/12


AlCalzone (approved):

AlCalzone:
Ooops, that one wasn't correct. The file is auto-generated.

Issue #6399, opened 2023-10-15
ZST39 is probably corrupting the ACK after a soft reset


jtbraun:
### Is your problem within Home Assistant (Core or Z-Wave JS Integration)?

NO, my problem is NOT within Home Assistant or the ZWave JS integration

### Is your problem within Z-Wave JS UI (formerly ZwaveJS2MQTT)?

NO, my problem is NOT within Z-Wave JS UI

### Checklist

- [X] I have checked the [troubleshooting section](https://zwave-js.github.io/node-zwave-js/#/troubleshooting/index) and my problem is **not** described there.

- [X] I have read the [changelog](https://github.com/zwave-js/node-zwave-js/blob/master/CHANGELOG.md) and my problem was **not** mentioned there.

### Describe the bug

What causes the bug?
The issue of a soft reset to a ZST39 during the startup phase.

What do you observe?
Soft reset is issued, and an 0x86 byte is immediately received and discarded. Then the SerialAPIStarted message comes in, and later the driver complains that an ACK was never received.

What did you expect to happen?
ACK would be received.

  1. Get a Zooz ZST39 800-series stick
  2. Update to latest (7.19 I think?)
  3. Run the `test/run.ts` aka "Debug locally" test in vscode
  4. See screenshotted below.

![image](https://github.com/zwave-js/node-zwave-js/assets/1192545/14bf6b6d-3032-40cc-9e3b-bbaf1976ee83)

Modifying `packages/serial/src/parsers/SerialAPIParser.ts` to also accept 0x86 as "ACK" gets past the error.

NOTE: I've reported the issue to Zooz, if I hear back from them I'll let you know. In the meantime, there probably needs to be a quirk flag or something that is set in the serialAPIParser specifically when softreset is waiting on an ACK, that treats a discarded byte as the missing ACK. :-/

### Device information

Manufacturer: ZOOZ
Model name: ZST39
Node ID in your network: 1

### How are you using `node-zwave-js`?

- [ ] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker image (latest)
- [ ] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker image (dev)
- [ ] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker manually built _(please specify branches)_
- [ ] `ioBroker.zwave2` adapter _(please specify version)_
- [ ] `HomeAssistant zwave_js` integration  _(please specify version)_
- [ ] `pkg`
[...]

PR #6401, opened 2023-10-16
feat: support Z-Wave Long Range

### LINKED ISSUE #5253: ZWave Long Range Support
With the 700 and 800 chipsets Silicon Labs added a new parallel channel/protocol to ZWave named Long Range. Devices that utilize ZWave Long Range are starting to be released.
https://community.inovelli.com/t/z-wave-800-series-2-1-switch-on-off-dimmer-project-phoenix/11428
https://atrim.co/zwave/stick-hub/
https://www.thesmartesthouse.com/products/zooz-800-series-z-wave-long-range-usb-stick-zst39

Support for ZWave Long Range frequency and protocol on the 700 and 800 chipsets.

The UI allows you to select the US Long Range region and some grepping implies that initial work towards ZWave Long Range was done. Though it looks like it is not complete.

**Additional context**
https://sdomembers.z-wavealliance.org/document/dl/897 sections 5 and 6

### LINKED ISSUE #6158: switching to 16-bit node IDs is currently done twice during startup
... once before and once after soft-reset.

jtbraun:
This PR adds support for Z-Wave Long Range

fixes: https://github.com/zwave-js/node-zwave-js/issues/5253
fixes: https://github.com/zwave-js/node-zwave-js/issues/6158


packages/cc/src/cc/Security2CC.ts
```diff
@@ -94,13 +94,26 @@ function getAuthenticationData(
...
+		offset += 2;
+	}
+
+	ret.writeUInt32BE(homeId, offset);
+	offset += 4;
+	ret.writeUInt16BE(commandLength, offset);
+	offset += 2;
 	// This includes the sequence number and all unencrypted extensions
-	unencryptedPayload.copy(ret, 8, 0);
+	unencryptedPayload.copy(ret, offset, 0);
```
jtbraun: If either nodeid is 256+, you have to use uint16BE to encode both.

packages/cc/src/cc/ZWaveLRProtocolCC.ts
```diff
@@ -0,0 +1,238 @@
...
+	extends ZWaveLRProtocolCCNodeInformationFrame
+{}
+
+interface ZWaveLRProtocolCCSmartStartIncludedNodeInformationOptions
+	extends CCCommandOptions
+{
+	nwiHomeId: Buffer;
+}
+
+// BUGBUG: this is exactly equal to the ZWaveProtocolCommand.SmartStartIncludedNodeInformation, can we reuse/inherit that somehow?
```
jtbraun: As comment, how much sharing do we want to do between these classes (or not)?

packages/cc/src/cc/ZWaveLRProtocolCC.ts
```diff
@@ -0,0 +1,238 @@
+import {
```
jtbraun: While this command class is defined in the specs, It's not exposed that I saw in any command class list, so I'm not sure where/how this would be used. Maybe if you had a secondary LR controller in the network? Unknown. We may not need to include this at all.

packages/cc/src/cc/ZWaveLRProtocolCC.ts
```diff
@@ -0,0 +1,238 @@
...
+
+@CCCommand(ZWaveLRProtocolCommand.NOP)
+export class ZWaveLRProtocolCCNOP extends ZWaveLRProtocolCC {}
+
+interface ZWaveLRProtocolCCNodeInformationFrameOptions
+	extends CCCommandOptions, NodeInformationFrame
+{}
+
+// BUGBUG: how much of this can we share with existing stuff? Can we use a ZWaveProtocolCCNodeInformationFrameOptions field to do the `isLongRange` stuff?
+// BUGBUG: how much can we share also with the Smart Start things below that are VERY close to this stuff?
```
jtbraun: Same, can we leverage base classes and combine a lot of this with the existing `ZWaveProtocolCC`?

packages/cc/src/lib/_Types.ts
```diff
@@ -1670,6 +1670,22 @@ export enum ZWaveProtocolCommand {
 	SmartStartInclusionRequest = 0x28,
 }
 
+export enum ZWaveLRProtocolCommand {
+	NOP = 0x00,
+
+	// BUGBUG: all defined above, can they be shared?
```
jtbraun: These have the same # assignments, but are in a separate namespace. Add the missing ones to the list above, or leave these separate (or maybe just reference `Foo = ZWaveProtocolCommand.Foo.value` or something?

packages/core/src/capabilities/CommandClasses.ts
```diff
@@ -130,6 +130,7 @@ export enum CommandClasses {
 	"Z-Wave Plus Info" = 0x5e,
 	// Internal CC which is not used directly by applications
 	"Z-Wave Protocol" = 0x01,
+	"Z-Wave Long Range Protocol" = 0x04,
```
jtbraun: Any preference on this string?

packages/core/src/capabilities/NodeInfo.ts
```diff
@@ -124,32 +124,47 @@ export function parseCCList(payload: Buffer): {
...
 		// CCs after the support/control mark are controlled
+		// BUGBUG: does the "mark" and support/control convention apply to isLongRange?
 		if (cc === CommandClasses["Support/Control Mark"]) {
 			isAfterMark = true;
 			continue;
 		}
 		(isAfterMark ? ret.controlledCCs : ret.supportedCCs).push(cc);
 	}
+	// BUGBUG: isLongRange prohibits CC from 0x00..0x20 from being advertised here, as does 4.3.2.1.1.17
+	// BUGBUG: how do >0xFF CC get advertised? I don't immediately see a mechanism for indicating a multi-byte CC
```
jtbraun: This seems like a validation miss?

packages/core/src/capabilities/NodeInfo.ts
```diff
@@ -124,32 +124,47 @@ export function parseCCList(payload: Buffer): {
...
+		listEnd = offset + listLength;
+		validatePayload(payload.length >= listEnd);
+	}
+	while (offset < listEnd) {
 		// Read either the normal or extended ccId
 		const { ccId: cc, bytesRead } = parseCCId(payload, offset);
 		offset += bytesRead;
 		// CCs before the support/control mark are supported
 		// CCs after the support/control mark are controlled
+		// BUGBUG: does the "mark" and support/control convention apply to isLongRange?
```
jtbraun: I think the answer to this was "yes, this code already supports multi-byte values"

packages/core/src/capabilities/NodeInfo.ts
```diff
@@ -124,32 +124,47 @@ export function parseCCList(payload: Buffer): {
...
-		sum(supportedCCs.map((cc) => (isExtendedCCId(cc) ? 2 : 1)))
+	const bufferLength = (isLongRange ? 1 : 0)
+		+ sum(supportedCCs.map((cc) => (isExtendedCCId(cc) ? 2 : 1)))
 		+ (controlledCCs.length > 0 ? 1 : 0) // support/control mark
 		+ sum(controlledCCs.map((cc) => (isExtendedCCId(cc) ? 2 : 1)));
 
 	const ret = Buffer.allocUnsafe(bufferLength);
 	let offset = 0;
+	if (isLongRange) {
+		// BUGBUG: validate bufferLength - 1 is <= 0xFF
```
jtbraun: If this length can't be encoded in a byte, what's an appropriate error action here? Throw an error? Other?

packages/core/src/capabilities/NodeInfo.ts
```diff
@@ -215,34 +230,49 @@ export type NodeInformationFrame =
...
+		if (speedExtension & 0b001) {
+			supportedDataRates.push(100000);
+		}
+		if (supportedDataRates.length === 0) {
+			supportedDataRates.push(9600);
+		}
 	}
 
-	const protocolVersion = buffer[offset] & 0b111;
+	// BUGBUG: what's the correct protocol version here for long range?
```
jtbraun: I couldn't find this in the specs.

packages/core/src/capabilities/NodeInfo.ts
```diff
@@ -252,21 +282,30 @@ export function parseNodeProtocolInfo(
...
+			// BUGBUG: is Controller correct for default && isLongRange?
 			nodeType = NodeType.Controller;
 			break;
 	}
 
-	const hasSpecificDeviceClass = !!(capability & 0b100);
-	const supportsSecurity = !!(capability & 0b1);
+	const hasSpecificDeviceClass = isLongRange || !!(capability & 0b100);
+	// BUGBUG: can we assume security is true?
+	const supportsSecurity = isLongRange || !!(capability & 0b1);
```
jtbraun: Given that I couldn't include a node except with S2 security, and they don't define the bit for LR versions of this struct, I suspect the answer is "security is always available if LR", but I couldn't find that in the spec.

packages/core/src/consts/index.ts
```diff
@@ -10,6 +10,12 @@ export const NODE_ID_MAX = MAX_NODES;
 /** The number of bytes in a node bit mask */
 export const NUM_NODEMASK_BYTES = MAX_NODES / 8;
 
+/** The number of node ids in a long range "segment" (GetLongRangeNodes response) */
+export const NUM_LR_NODES_PER_SEGMENT = 128;
+
+/** The number of bytes in a long range node bit mask segment */
+export const NUM_LR_NODEMASK_SEGMENT_BYTES = NUM_LR_NODES_PER_SEGMENT / 8;
```
jtbraun, line 17: Anyone like better names?

packages/core/src/values/Primitive.ts
```diff
@@ -231,15 +235,17 @@ export function encodeFloatWithScale(
...
+export function parseBitMask(
+	mask: Buffer,
+	startValue: number = 1,
+	numBits: number = mask.length * 8,
+): number[] {
 	const ret: number[] = [];
-	for (let index = 1; index <= numBits; index++) {
-		const byteNum = (index - 1) >>> 3; // id / 8
-		const bitNum = (index - 1) % 8;
+	for (let index = 0; index < numBits; index++) {
```
jtbraun, line 244: I personally found this confusing... we do 1 based indexing in the `for` statement, only to subtract off `1` everywhere below. I just removed the `1` everywhere.

packages/serial/src/message/Constants.ts
```diff
@@ -52,8 +52,8 @@ export enum FunctionType {
 	UNKNOWN_FUNC_MEMORY_PUT_BUFFER = 0x24,
 
 	EnterBootloader = 0x27, // Leave Serial API and enter bootloader (700+ series only). Enter Auto-Programming mode (500 series only).
-	UNKNOWN_FUNC_UNKNOWN_0x28 = 0x28, // ??
-
+	UNKNOWN_FUNC_UNKNOWN_0x28 = 0x28, // ZW_NVRGetValue(offset, length) => NVRdata[], see INS13954-13
+
```
jtbraun: Unrelated to long range, but I found a reference doc from Silicon Labs 500 series chips that gave the `FunctionType` values for a lot of unknown ones.

packages/serial/src/parsers/SerialAPIParser.ts
```diff
@@ -44,6 +44,12 @@ export class SerialAPIParser extends Transform {
 						this.push(MessageHeaders.ACK);
 						break;
 					}
+					case 0x86: {
+						// This is _maybe_ a corrupted ACK byte from a ZST39 after a soft reset, transform it and keep it
+						this.logger?.ACK("inbound");
+						this.push(MessageHeaders.ACK);
+						break;
+					}
```
jtbraun: GROSS. See [issue](https://github.com/zwave-js/node-zwave-js/issues/6399) for details. This probably shouldn't be committed here, I'm happy to back it out.

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -121,6 +121,7 @@ import {
 	ApplicationUpdateRequestNodeAdded,
 	ApplicationUpdateRequestNodeInfoReceived,
 	ApplicationUpdateRequestNodeRemoved,
+	ApplicationUpdateRequestSmartStartHomeIDLRReceived,
```
jtbraun: Name? It's easy to miss the `LR` in there. Maybe `LongRange` instead?

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -897,7 +905,7 @@ export class ZWaveController
 		const apiCaps = await this.driver.sendMessage<
 			GetSerialApiCapabilitiesResponse
 		>(
-			new GetSerialApiCapabilitiesRequest(this.driver),
+			new GetSerialApiCapabilitiesRequest(this.driver, {mysteryValue: 3}),
```
jtbraun: This is a holdover from when I couldn't get inclusion working. SimplicityStudio adds a mystery single byte field to this command with value 3. it's not documented, and I think the inclusion problems were security related. Probably should back this out.

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1022,11 +1030,13 @@ export class ZWaveController
 						)
 					}`,
 				);
+				this._supportsLongRange = resp == RFRegion["USA (Long Range)"];
 			} else {
 				this.driver.controllerLog.print(
 					`Querying the RF region failed!`,
 					"warn",
 				);
+				this._supportsLongRange = false;
```
jtbraun: *sigh* ... I can't tell what's supposed to be the definitive reference for this. I'd REALLY _want_ to use the long range bit in the `SerialAPIStarted` message, but that's sent after a soft reset, and I think we have many cases/sticks (including ser2net) that require we DON'T do a soft reset. This was the next best option I could find.

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1283,6 +1293,46 @@ export class ZWaveController
...
+
+		// BUGBUG: Dummy check? SimplicityStudio adds this extra 3 argument to this command. I suspect that's not the secret sauce for LR inclusion, but S2 encryption is.
+		await this.driver.sendMessage<
+			GetSerialApiCapabilitiesResponse
+		>(
+			new GetSerialApiCapabilitiesRequest(this.driver, {mysteryValue: 3}),
+			{
+				supportCheck: false,
+			},
+		);
```
jtbraun: See above, this can probably get removed.

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1283,6 +1293,46 @@ export class ZWaveController
...
+
+		// fetch the list of long range nodes until the controller reports no more
+		const lrNodeIds: Array<number> = [];
+		let lrChannel : LongRangeChannel | undefined;
+		const maxPayloadSize = await this.getMaxPayloadSize();
+		let maxPayloadSizeLR : number | undefined;
+		if (this.isLongRange()) {
+			const nodePage = 0;
+			while (true) {
+				const nodesResponse = await this.driver.sendMessage<
```
jtbraun: This loop should probably go into a helper function called `getLongRangeNodes` or something?

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1283,6 +1293,46 @@ export class ZWaveController
...
+			{
+				supportCheck: false,
+			},
+		);
+
+		// fetch the list of long range nodes until the controller reports no more
+		const lrNodeIds: Array<number> = [];
+		let lrChannel : LongRangeChannel | undefined;
+		const maxPayloadSize = await this.getMaxPayloadSize();
+		let maxPayloadSizeLR : number | undefined;
```
jtbraun: Do we want to cache these off into member variables, or just report them for now?

I did NOT include APIs for setting some of these things, as I don't know what we intend to expose in JS UI, etc.

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1283,6 +1293,46 @@ export class ZWaveController
...
+						listStartOffset128: nodePage,
+					}),
+				);
+				lrNodeIds.push(...nodesResponse.nodeIds);
+				if (!nodesResponse.moreNodes) {
+					break;
+				}
+			}
+
+			// TODO: restore/set the channel
```
jtbraun: We probably need a driver constructor argument to choose a long range channel to set, yeah?

And maybe TXPower, etc.

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1283,6 +1293,46 @@ export class ZWaveController
...
+			const lrChannelResp = await this.driver.sendMessage<GetLongRangeChannelResponse>(new GetLongRangeChannelRequest(this.driver));
+			lrChannel = lrChannelResp.longRangeChannel;
+
+			// TODO: fetch the long range max payload size and cache it
```
jtbraun: I didn't see where we use the existing zwave maxpayloadsize to verify/check/limit packets elsewhere, do we?

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1323,6 +1377,9 @@ export class ZWaveController
 			nodeIds.unshift(this._ownNodeId!);
 		}
 
+		// BUGBUG: do nodes need to implicitly know that they were a long range node? Or are long range nodes determined 100% by their nodeID values being >= 256?
+		// The controller is the odd-man out, as it's both. Let's assume that apart from explicit exclusion we don't need to know for now.
+		nodeIds.push(...lrNodeIds);
```
jtbraun: I couldn't find a spot in the serial API that really cared "this is a long range node" or not.

We _probably_ don't want people assuming that from the nodeID though, so there's probably a boolean parameter `isLongRange` or something that we need to instantiate on the `Node` objects?

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1573,11 +1636,26 @@ export class ZWaveController
...
 
+		// BUGBUG: fix me
+		// if (options.isLongRange && !this.isLongRange()) {
+		// 	throw new ZWaveError(
+		// 		`Invalid long range inclusion on a non-LR host`,
+		// 		ZWaveErrorCodes.Argument_Invalid,
+		// 	)
+		// }
+		const isLongRange = this.isLongRange();
+
```
jtbraun: I couldn't actually get a note to include as long range here.

Morever, we probably want whether we're including as long range or not to be an option/parameter passed in, yeah?

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1573,11 +1636,26 @@ export class ZWaveController
...
 		await this.pauseSmartStart();
 
 		this.setInclusionState(InclusionState.Including);
 		this._inclusionOptions = options;
+		// BUGBUG: todo: cache the options used to start inclusion so they can be re-used for stopping, etc
+		this._inclusionFlags = {
+			highPower: true,
+			networkWide: true,
+			protocolLongRange: isLongRange,
+		};
```
jtbraun: It seems like you'd want to re-use these same values when issuing the stop commands later? Otherwise you risk starting a long range and stopping a non--long range inclusion?  the spec was unclear on when these flags matter and don't.

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1649,6 +1726,11 @@ export class ZWaveController
 			strategy: InclusionStrategy.SmartStart,
 			provisioning: provisioningEntry,
 		};
+		this._inclusionFlags = {
+			highPower: true,
+			networkWide: true,
+			protocolLongRange: provisioningEntry.isLongRange,
```
jtbraun: I didn't understand all the inclusion pathways well enough to know what the right thing to do here is. There's a separate smartstart message for LR nodes (Update Smart Start HomeID (LR) = 0x87). Again, it wasn't clear if that's a statement about "I'm long range capable" or "I must be included long range". If the former, what is here is correct. If the latter, we'll want to allow the user to decide _somwhere_ whether this node is being added as long range or normal. At the time they add/verify the DSK, there should probably be a "as long range if possible" checkbox?

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1928,19 +2004,32 @@ export class ZWaveController
...
 
+		// BUGBUG: fix me, same logic as beginInclusion, probably an incoming option...
+		// if (options.isLongRange && !this.isLongRange()) {
+		// 	throw new ZWaveError(
+		// 		`Invalid long range inclusion on a non-LR host`,
+		// 		ZWaveErrorCodes.Argument_Invalid,
+		// 	)
+		// }
+		const isLongRange = this.isLongRange();
+
```
jtbraun: Again, all this "include as long range or not" logic needs overhauled once we have a plan for how the UI/clients will interact with the driver about it.

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -2065,6 +2152,7 @@ export class ZWaveController
 			}
 		} else if (
 			msg instanceof ApplicationUpdateRequestSmartStartHomeIDReceived
+			|| msg instanceof ApplicationUpdateRequestSmartStartHomeIDLRReceived
```
jtbraun: Here's the LR-specific message mentioned above...

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -4309,6 +4397,14 @@ ${associatedNodes.join(", ")}`,
...
 	 */
 	public async assignSUCReturnRoutes(nodeId: number): Promise<boolean> {
+		if (nodeId >= 0x100) {
+			this.driver.controllerLog.logNode(nodeId, {
+				message: `Skipping SUC return route because isLR...`,
+				direction: "outbound",
+			});	
+			return true;
+		}
+
```
jtbraun: This should _probably_ be conditioned on a per-node flag/property like "supports routing" or something. I think that's actually specified/fetched during the interview process or something for nodes, yes?

Long range nodes don't mesh, so don't need routes, and have no SUC.

packages/zwave-js/src/lib/controller/Inclusion.ts
```diff
@@ -132,6 +133,11 @@ export type InclusionOptions =
 		 * This is not recommended due to the overhead caused by S0.
 		 */
 		forceSecurity?: boolean;
+
+		/**
+		 * Force long range. If not provided, will default to long range iff the controller supports it, and not otherwise.
+		 */
+		isLongRange?: boolean;
```
jtbraun: This goes back to "how will the UI specify it want's to do long range inclusion or not". Whatever options there are should probably go here, but I'd like to hear from the maintainers how they'd like it to work.

packages/zwave-js/src/lib/controller/Inclusion.ts
```diff
@@ -158,6 +164,11 @@ export type InclusionOptions =
 		strategy:
 			| InclusionStrategy.Insecure
 			| InclusionStrategy.Security_S0;
+
+		/**
+		 * Force long range. If not provided, will default to long range iff the controller supports it, and not otherwise.
+		 */
+		isLongRange?: boolean;
```
jtbraun: Same API questions.

packages/zwave-js/src/lib/controller/Inclusion.ts
```diff
@@ -171,6 +182,13 @@ export type InclusionOptionsInternal =
 		provisioning: PlannedProvisioningEntry;
 	};
 
+// BUGBUG: better way to do this?
+export type InclusionFlagsInternal = {
```
jtbraun: This is the "preserver the flags used at START inclusion for STOP inclusion" struct. I'm new to typescript, I wasn't sure if there was a better way or something. Does this need to be a type? Or just an Interface?

packages/zwave-js/src/lib/controller/Inclusion.ts
```diff
@@ -219,6 +237,10 @@ export interface PlannedProvisioningEntry {
 	/** The device specific key (DSK) in the form aaaaa-bbbbb-ccccc-ddddd-eeeee-fffff-11111-22222 */
 	dsk: string;
 
+	/** The device must be included via ZWave Long Range */
+	// BUGBUG: by adding this here we probably broke the API used by JSUI, etc...
```
jtbraun: Yeah, I probably did. Maybe this should be optional/undefined so it's backwards compatible?

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -5345,6 +5345,7 @@ ${handlers.length} left`,
 	}
 
 	/** @internal */
+	// BUGBUG: do we need to replicate this for Long Range?
```
jtbraun: Here's where the new command class could be used, I just didn't understand how it's supposed to be exercised if it never is advertised in any of the CC lists?

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiCapabilitiesMessages.ts
```diff
@@ -15,10 +15,48 @@ import {
 const NUM_FUNCTIONS = 256;
 const NUM_FUNCTION_BYTES = NUM_FUNCTIONS / 8;
 
+export interface GetSerialApiCapabilitiesRequestOptions
+	extends MessageBaseOptions
+{
+	mysteryValue: number | undefined,
+}
+
```
jtbraun: See above, probably should remove this.

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
```diff
@@ -215,3 +218,115 @@ export class GetSerialApiInitDataResponse extends Message {
...
+@expectedResponse(FunctionType.GetLongRangeNodes)
+@priority(MessagePriority.Controller)
+export class GetLongRangeNodesRequest extends Message {
+	public constructor(
+		host: ZWaveHost,
+		options:
+			| MessageDeserializationOptions
+			| GetLongRangeNodesRequestOptions,
+	) {
+		super(host, options);
```
jtbraun: Getting the list of nodes is part of setting up the serial API, but I could equally see this going in with the other LR messages in the new file below...

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
```diff
@@ -215,3 +218,115 @@ export class GetSerialApiInitDataResponse extends Message {
...
+
+		if (gotDeserializationOptions(options)) {
+			// BUGBUG: validate length at least
```
jtbraun: Some existing code validates buffer lengths, some does not. I'm assuming I should be calling `ValidateBuffer` or whatever here...?

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
```diff
@@ -215,3 +218,115 @@ export class GetSerialApiInitDataResponse extends Message {
...
+			this.listStartOffset128 = this.payload[0];
+		} else {
+			this.listStartOffset128 = options.listStartOffset128;
+		}
+	}
+
+	public listStartOffset128: number;
```
jtbraun: I hate this name. Got a better one?

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
```diff
@@ -215,3 +218,115 @@ export class GetSerialApiInitDataResponse extends Message {
...
+		);
+
+		this.payload[0] = this.listStartOffset128;
+		return super.serialize();
+	}
+}
+
+export interface GetLongRangeNodesResponseOptions extends MessageBaseOptions {
+	moreNodes: boolean;
+	listStartOffset128: number;
```
jtbraun: Same name, same hate. Got a better one?

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
```diff
@@ -215,3 +218,115 @@ export class GetSerialApiInitDataResponse extends Message {
...
+	public constructor(
+		host: ZWaveHost,
+		options:
+			| MessageDeserializationOptions
+			| GetLongRangeNodesResponseOptions,
+	) {
+		super(host, options);
+
+		if (gotDeserializationOptions(options)) {
+			// BUGBUG: validate length at least
```
jtbraun: Same thing... payload length validation here, and a few lines down.

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
```diff
@@ -215,3 +218,115 @@ export class GetSerialApiInitDataResponse extends Message {
...
+		);
+		nodeBitMask.copy(this.payload, 3);
+
+		return super.serialize();
+	}
+
+	private listStartNode(): number {
+		return 256 + NUM_LR_NODES_PER_SEGMENT * 8 * this.listStartOffset128;
+	}
+
```
jtbraun: The equations given in the spec for this say `255 +`, but the given examples result in this being `256 +`. What's here is correct experimentally. Maybe the spec was using 1-based math and I missed it?

packages/zwave-js/src/lib/serialapi/capability/LongRangeSetupMessages.ts
```diff
@@ -0,0 +1,169 @@
...
+	type MessageBaseOptions,
+	type MessageDeserializationOptions,
+	MessageType,
+	expectedResponse,
+	gotDeserializationOptions,
+	messageTypes,
+	priority,
+} from "@zwave-js/serial";
+
+// TODO: move to header?
```
jtbraun: which other header should this sort of thing live in, if any?

packages/zwave-js/src/lib/serialapi/capability/LongRangeSetupMessages.ts
```diff
@@ -0,0 +1,169 @@
...
+export interface SetLongRangeChannelResponseOptions extends MessageBaseOptions {
+	responseStatus: number;
+}
+
+// BUGBUG:  move someplace common, see Spec 4.2.5
+//
+// Although, that section refers to a "callback message", and SetZWaveLongRangeChannel doesn't define a callback message...
+export enum ResponseStatus {
+	FAILED = 0x00,
+}
```
jtbraun: The `ReportStatus` here is supposed to be a common thing, but in looking through other things that use it, I didn't see one that was implemented. It's _supposed_ to be used with messages that then have a callback message, so that a failure can stop the callback from occuring, but this message doesn't have a callback message...

Seems like a case where this should have just been pass/fail on it's own, and someone in the spec committee decided to re-use a status field that wasn't necessarily applicable?

packages/zwave-js/src/lib/serialapi/capability/LongRangeSetupMessages.ts
```diff
@@ -0,0 +1,169 @@
...
+@messageTypes(MessageType.Response, FunctionType.SetLongRangeChannel)
+export class SetLongRangeChannelResponse extends ResponseStatusMessageBase {}
+
+export interface LongRangeShadowNodeIDsRequestOptions
+	extends MessageBaseOptions
+{
+	shadowNodeIds: number[];
+}
+const LONG_RANGE_SHADOW_NODE_IDS_START = 2002;
+const NUM_LONG_RANGE_SHADOW_NODE_IDS = 4;
```
jtbraun, line 112: Probably should be in a header somewhere.

Also, I didn't provide an API to set this stuff. Presumably these are for when you want your LR-controller to show up with a LR nodeId? I don't know.

packages/zwave-js/src/lib/serialapi/network-mgmt/AddNodeToNetworkRequest.ts
```diff
@@ -142,11 +146,14 @@ export class AddNodeToNetworkRequest extends AddNodeToNetworkRequestBase {
 	public highPower: boolean = false;
 	/** Whether to include network wide */
 	public networkWide: boolean = false;
+	/** Whether to include as long-range or not */
+	public protocolLongRange: boolean = false;
```
jtbraun: The ZST39 stick I have refused to honor this bit during normal, insecure inclusion. No clue why. Bug, feature? The spec seems to say it should do it, but I don't know.

jtbraun (commented):
Here's a bunch of places where I could use some feedback, suggestions, etc.

jtbraun:
I'm not sure how to request a review to get some feedback? I'm going to assume the maintainers watch for PRs occasionally. @blhoward2 ?

blhoward2:
You can summon @AlCalzone, but I'm sure he'd see it in the morning anyway. He keeps a close watch.

jtbraun:
I've got a support ticket out to zooz on the "'normal' inclusion doesn't seem to respect the protocol=1 flag" issue. I suspect it's a bug in the ZST39, or undocumented behavior in SILabs' reference design.

The big issues for @AlCalzone right this second are:
* Opinions on API changes for how to indicate include/exclude longrange/normal.
* Opinions on names. There's a mix right now of `lr*`, `*LR*` and `*LongRange*`.
* I haven't taken any stabs at test coverage yet... if you have preference for what and how much, let me know.

There's a ton of `BUGBUG/TODO` comments right now, I've tried to explain in comments what the questions/issues are.
There's also pre-existing tests that appear to be broken by this. I'll need to look into those.

With the ZST39/ZEN32 I was testing, I have to say "long range" currently looks like an over-promise/under-deliver situation. I got about twice the range of the normal AEOTEC stick in a similar location. While that's good, the star network topology means that the overall network reach is going to be smaller. It's a big shame that they didn't allow for routes, or expand the spec to allow for more than the ~240 nodes in the mesh topology. My signaling environment hasn't been optimized at all, and I am indoors, but I'd hoped for better. 👎

AlCalzone:
@jtbraun thanks for this! I'll make sure to take a good look at it soon and give you some feedback, I'm currently dealing with all the weird controller firmware issues that cause problems with the new "unresponsive controller" recovery feature.

Just FYI if it wasn't clear - the `ZWaveProtocolCC` stuff is mostly used internally and for testing, as these classes are normally used by the controller under the hood. We use it in a few locations to hack our way around missing features in the Serial API.

packages/serial/src/parsers/SerialAPIParser.ts
jtbraun: This is fixed in https://github.com/zwave-js/node-zwave-js/pull/6409. I'll need to remove this and rebase here.

packages/cc/src/cc/ZWaveLRProtocolCC.ts
AlCalzone: This may be used in our testing framework, but unless you're sure about these things, I'd leave it out for now.

packages/cc/src/cc/Security2CC.ts
```diff
@@ -94,13 +94,26 @@ function getAuthenticationData(
...
+	if (nodeIdSize == 1) {
+		ret[offset++] = sendingNodeId;
+		ret[offset++] = destination;	
+
+	} else {
+		ret.writeUint16BE(sendingNodeId, offset);
+		offset += 2;
+		ret.writeUint16BE(destination, offset);
+		offset += 2;
+	}
```
AlCalzone:
```suggestion
	ret.writeUIntBE(sendingNodeId, offset, nodeIdSize);
	offset += nodeIdSize;
	ret.writeUIntBE(destination, offset, nodeIdSize);
	offset += nodeIdSize;
```

packages/cc/src/lib/_Types.ts
AlCalzone: Leave separate IMO, or leave it out completely together with the CC

packages/core/src/capabilities/CommandClasses.ts
AlCalzone: The specs call it `"Z-Wave Long Range"` Command Class, let's stick to that.

packages/core/src/capabilities/NodeInfo.ts
AlCalzone: parseCCId should take care of that.

packages/core/src/capabilities/NodeInfo.ts
```diff
@@ -124,32 +124,47 @@ export function parseCCList(payload: Buffer): {
...
 	let isAfterMark = false;
-	while (offset < payload.length) {
+	let listEnd = payload.length;
+	if (isLongRange) {
+		validatePayload(payload.length >= offset + 1);
+		const listLength = payload[offset++];
+		listEnd = offset + listLength;
+		validatePayload(payload.length >= listEnd);
+	}
+	while (offset < listEnd) {
```
AlCalzone: Why is there a change necessary to this method? It's intended to take a list of command classes encoded as bytes and split it into two arrays. The caller should make sure the passed buffer starts at a list of CCs.

packages/core/src/capabilities/NodeInfo.ts
AlCalzone: This is parsing code, no need to validate IMO.

packages/core/src/capabilities/NodeInfo.ts
AlCalzone: Why is there a change necessary to this method? It's intended to do the opposite of `parseCCList`. The caller should make sure to put the resulting buffer in the correct place.

packages/core/src/capabilities/NodeInfo.ts
```diff
@@ -215,34 +230,49 @@ export type NodeInformationFrame =
 export function parseNodeProtocolInfo(
 	buffer: Buffer,
 	offset: number,
+	isLongRange: boolean = false,
```
AlCalzone, line 218: Where did you find the information you used to update this method? I'm not sure a change is necessary, as the controller should forward a properly formatted NIF, no matter if the protocol-level frame it received is using the Long Range CC or not.

packages/core/src/capabilities/NodeInfo.ts
AlCalzone: The specs for the Z-Wave CC say this:
> Nodes supporting the S0 or S2 Command Classes shall set this field to 1

but in practice I don't think this flag is used by anything. We don't at least and the controller isn't involved with S0/S2.

Given this
> Nodes join a Z-Wave Long Range network using SmartStart Inclusion

I'd agree that Long Range = Security S2.

packages/core/src/capabilities/NodeInfo.ts
AlCalzone: Given that the NIF for Long Range lists these bits as reserved, I'd guess it's either 1 or not interesting. What do the frames received from your controller contain here?

packages/core/src/values/Primitive.ts
AlCalzone: Not sure why I did it this way. Maybe this was more in line with some texts in the specs.

packages/serial/src/parsers/SerialAPIParser.ts
AlCalzone: This can be removed after merging the master branch.

packages/core/src/consts/index.ts
AlCalzone: Fine with me

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone:
```suggestion
	ApplicationUpdateRequestSmartStartLongRangeHomeIDReceived,
```
maybe?

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone: 700/800 series controllers generally have no issue with soft reset. In fact, we ignore the driver option for those controllers and always soft-reset them, because it's required to fix some issues.

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone: Would be nice to have. AFAIK not all controllers support this, so I couldn't rely on it and used some values found in the 500 series specs.
If we can assume that all 700+ series controllers support the long range equivalent, then please store it somewhere.

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone: yeah

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone: I guess - not sure if we're supposed to configure this. Being able to do it doesn't hurt though.
Both a Driver option and API method.

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone: We don't, for reasons explained above:
https://github.com/jtbraun/node-zwave-js/blob/d1e23fc6a2e58c15eff86d31cf78951d8b62476b/packages/zwave-js/src/lib/serialapi/transport/SendDataMessages.ts#L195-L200

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1305,7 +1355,11 @@ export class ZWaveController
   controller role:            ${this._isPrimary ? "primary" : "secondary"}
   controller is the SIS:      ${this._isSIS}
   controller supports timers: ${this._supportsTimers}
-  nodes in the network:       ${initData.nodeIds.join(", ")}`,
+  zwave nodes in the network: ${initData.nodeIds.join(", ")}
+  max payload size:           ${maxPayloadSize}
+  LR nodes in the network:    ${lrNodeIds.join(", ")}
```
AlCalzone: Will be interesting to see if this list becomes extraordinarily long.

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone: Might be interesting to know, even if only for troubleshooting. If a user has connectivity issues, generating new routes won't help if the node is LR.

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone: If I'm not mistaken, Long Range only supports Smart Start inclusion. When doing that, there should probably be a way to decide which protocol to use though - if that is possible?

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone: Agreed. I think we always use high power and network wide though, as the other options are deprecated. So I guess the only thing we should remember is Long Range or not.

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone: I need to do some more reading on this. This sounds like the user should be able to choose this by setting a flag in the provisioning list:
grafik

I'm not yet sure though how the controller decides this? Maybe the node sends both a Z-Wave and a Long Range Smart Start request with different Auth/NWI IDs, and the application can then decide which one to start the inclusion for?

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1593,8 +1670,7 @@ export class ZWaveController
 			await this.driver.sendMessage(
 				new AddNodeToNetworkRequest(this.driver, {
 					addNodeType: AddNodeType.Any,
-					highPower: true,
-					networkWide: true,
+					...this._inclusionFlags,
```
AlCalzone: I don't think we can include a LR node this way, given that it must use SmartStart?

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone: I guess that's where we decide how to include the node. Can you check if you receive both these commands when powering up a LR node? And what's in them?

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone: > Long range nodes don't mesh, so don't need routes, and have no SUC.

...yet

But I agree. Guess if we remember that a node is LR at startup or after inclusion, we can use that to skip.

packages/zwave-js/src/lib/controller/Inclusion.ts
AlCalzone: For these "constant" shapes, an interface is preferred. But I think we don't need this, just the flag LR or not.

packages/zwave-js/src/lib/driver/Driver.ts
AlCalzone: As said before, ignore it.

packages/zwave-js/src/lib/serialapi/application/ApplicationUpdateRequest.ts
```diff
@@ -223,3 +223,13 @@ export class ApplicationUpdateRequestSmartStartHomeIDReceived
...
+
+@applicationUpdateType(ApplicationUpdateTypes.SmartStart_HomeId_Received)
+export class ApplicationUpdateRequestSmartStartHomeIDReceived
+	extends ApplicationUpdateRequestSmartStartHomeIDReceivedBase
+{}
+
+@applicationUpdateType(ApplicationUpdateTypes.SmartStart_HomeId_LR_Received)
+export class ApplicationUpdateRequestSmartStartHomeIDLRReceived
+	extends ApplicationUpdateRequestSmartStartHomeIDReceivedBase
+{}
```
AlCalzone: We should test if parsing the node ID may need to be done differently for the LR variant. `parseNodeID` decides on the configured nodeIdType of the controller, which may be 8 bit. Not sure what happens then when a LR node requests inclusion.

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
AlCalzone: Yep this should be in a separate file. In general I try to group them by `FunctionType`.

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
AlCalzone: We generally validate buffer lengths for the command classes, as there are tons of nodes that send broken frames. The controller firmware is generally much more predictable.

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
AlCalzone: pageNumber? segmentNumber?

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
```diff
@@ -215,3 +218,115 @@ export class GetSerialApiInitDataResponse extends Message {
...
+	public listStartOffset128: number;
+
+	public serialize(): Buffer {
+		this.payload = Buffer.allocUnsafe(
+			1,
+		);
+
+		this.payload[0] = this.listStartOffset128;
```
AlCalzone:
```suggestion
		this.payload = Buffer.from([this.listStartOffset128]);
```

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
AlCalzone: I think 256 makes more sense. Not the first time that the specs are wrong.

AlCalzone: The `* 8` seems wrong though?

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
```diff
@@ -215,3 +218,115 @@ export class GetSerialApiInitDataResponse extends Message {
...
+		return super.serialize();
+	}
+
+	private listStartNode(): number {
+		return 256 + NUM_LR_NODES_PER_SEGMENT * 8 * this.listStartOffset128;
+	}
+
+	private listEndNode(): number {
+		return 256
+			+ NUM_LR_NODES_PER_SEGMENT * 8 * (1 + this.listStartOffset128);
```
AlCalzone: Also there should be a -1 here, since 128 with offset 0 would be from 256 to 383, since the next segment starts at 384.

packages/zwave-js/src/lib/serialapi/capability/LongRangeSetupMessages.ts
AlCalzone: Maybe in `packages/core/src/capabilities/Protocols.ts`, or a new LR-related file in that folder

packages/zwave-js/src/lib/serialapi/capability/LongRangeSetupMessages.ts
AlCalzone: > It's _supposed_ to be used with messages that then have a callback message, so that a failure can stop the callback from occuring, but this message doesn't have a callback message...

In that case the response indicates the final result of the command flow, not the callback.

I don't think we need the enum value or a separate field. Just do this like we do here for example:
https://github.com/jtbraun/node-zwave-js/blob/d1e23fc6a2e58c15eff86d31cf78951d8b62476b/packages/zwave-js/src/lib/serialapi/capability/SerialAPISetupMessages.ts#L899-L907

packages/zwave-js/src/lib/serialapi/capability/LongRangeSetupMessages.ts
AlCalzone: same as the others I suppose -> `packages/core/src/consts/index.ts`

AlCalzone: IMO if we have these messages, we should expose an API on the controller for them.

AlCalzone (dismissed):
Thanks a lot for this. I've tried to answer most of the questions so you can finally continue. I'll probably have to play with this a bit to see how it all works out before we can merge it.

packages/core/src/capabilities/NodeInfo.ts
jtbraun: You're right, that the length check could be moved to the caller.

packages/core/src/capabilities/NodeInfo.ts
jtbraun: See `specs\Z-Wave Stack Specifications\Z-Wave and Z-Wave Long Range Network Layer Specification.pdf`. Long Range added their own Node Information Frame Command in section 6.3.1.2. The non-long-range frame (section 4.3.2.1) has 3 bytes for Basic/Generic/Specific device classes, and then a list of command classes for the remainder of the packet.

The long range frame has Generic and Specific device class bytes (no basic), and then a byte that is the "command list length" (in bytes, as I recall), and then the list of command classes.

Why they felt the need to replace these fields and add the seemingly redundant length, I'm not sure.

If we don't need/want the long range command class in this change (it wasn't needed to get pairing and comms to work). I think this stuff can be backed out.

packages/zwave-js/src/lib/controller/Controller.ts
jtbraun: Oh really? When I was playing around separately with ser2net, soft resetting the controller would cause it to de/re-enumerate over USB. That would cause the file handle ser2net had open to error out, which caused ser2net to close the TCP connection, and then zwave-js/ui would get upset that the ser2net socket disappeared out from under it.

packages/zwave-js/src/lib/controller/Controller.ts
jtbraun: Ah, so that method probably needs updated to take a `isLongRange` boolean, and return this value if set?

packages/zwave-js/src/lib/controller/Controller.ts
jtbraun: > If I'm not mistaken, Long Range only supports Smart Start inclusion.

That was certainly my experience, though I didn't see that in the specs anywhere. 
The LR test spec only talks about smart start inclusion, though. Let's just assume that's the case, I'll rework this to only support long range inclusion via smart start.

packages/zwave-js/src/lib/controller/Controller.ts
jtbraun: My assumption was that during smart start inclusion, the controller is told whether or not the being-included-node is long-range capable by whether it receives an 0x85 or an 0x87 tagged Update SmartStart HomeID command.

In response, the controller issues a SmartStartInclude command, and can choose whether it wants to add the node as long-range or not by setting the Protocol bit = 1 (LR) or not.

However the User/UI starts this process off (scanning the QR code, or confirming the DSK) would need a LR/not checkbox that would eventually make it's way into the protocol bit.

I haven't verified this behavior though, it's just what it seems like should happen given the definition of that Protocol bit.

image

packages/zwave-js/src/lib/controller/Controller.ts
jtbraun: Sure.

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
jtbraun: > The * 8 seems wrong though?

Yeah, I think that was a holdover from when I had "NUM BYTES PER SEGMENT" there instead.

packages/zwave-js/src/lib/serialapi/network-mgmt/AddNodeToNetworkRequest.ts
jtbraun: From @AlCalzone 's comments above, it's probably because LR nodes only support inclusion via SmartStart.

packages/zwave-js/src/lib/controller/Controller.ts
kpine: Take a look at [How to set up Z-Wave Long Range](https://community.silabs.com/s/article/how-to-set-up-z-wave-long-range?language=en_US), it confirms both of your assumptions.

packages/core/src/capabilities/NodeInfo.ts
jtbraun: Good question, I'm not sure. I'll have to check.

packages/zwave-js/src/lib/serialapi/application/ApplicationUpdateRequest.ts
jtbraun: You've got `ZWaveHost.nodeIdType` documented as 
> 	/** How many bytes a node ID occupies in serial API commands */

As I recall, the init code always sets up the serial API to use 2-byte IDs. I'll have to go back and look at the specs for this message to see if it's specified that 2-byte nodeIDs are always present here, or only if the API has been configured for them, or what.

packages/zwave-js/src/lib/controller/Controller.ts
jtbraun: Ah, yep. LR inclusion only via smart-start. It also says:
>  In other words, during learn mode, a end node that support LR will send SmartStart Prime on both classic Z-Wave and Z-Wave LR PHY, both request are send up to the host on the controller and it is the host's responsibility to determine which PHY is used for inclusion.

So the user will need to choose how they want to include a node, which we'll need to remember, and when we get both of the PRIME packets, we'll ignore the one that doesn't match the user's inclusion preference, I guess.

packages/zwave-js/src/lib/serialapi/capability/GetSerialApiInitDataMessages.ts
jtbraun: I was using it as `node < end`, so not -1, but this isn't used right now, I'll just remove it.

packages/zwave-js/src/lib/controller/Controller.ts
jtbraun: `getLongRangeNodes` now

packages/zwave-js/src/lib/controller/Controller.ts
jtbraun: Filling in from elsewhere:
* yes, LR inclusion MUST be smart start only
* LR nodes will send both normal and LR prime messages, and its up to the user/controller to respond to one or the other. So the DSK-providing API will need to include a "use LR" boolean that we honor. If the user sets it to a value that's not supported, we should probably just not include the node.

packages/zwave-js/src/lib/controller/Controller.ts
jtbraun: Does it hurt to just leave all these here, rather than have many places we set highPower/networkWide to true?

packages/zwave-js/src/lib/controller/Controller.ts
jtbraun: yes, I'll clean this up.

jtbraun:
Ugh... some of the tests fail on windows for windows reasons... things like CRLF line endings messing with test vector fetching code, or things like `__filename` not matching the contents of a stack trace because the driver letter capitalization or forward slash/backslash-as-dir-separator varying.

Maybe I can get these to run in WSL, otherwise I'll have to finish the test fixes when I later get access again to a linux box.

blhoward2:
That's weird since I'm pretty sure @AlCalzone develops it in windows. Al?

AlCalzone:
I can take a look if you like.
I used to develop on Windows, so this did work at some point.

amluto:
> With the ZST39/ZEN32 I was testing, I have to say "long range" currently looks like an over-promise/under-deliver situation. I got about twice the range of the normal AEOTEC stick in a similar location.

I haven't look at the API at all, and I don't entirely know what I'm talking about, but I did find this:

https://www.silabs.com/documents/login/presentations/design-with-z-wave-to-extend-your-wireless-range-1-mile-tw.pdf

In the US regulatory region at least, there should be a 17dB link budget improvement with LR, and that should be good for a 4-8x range improvement.  (And a *bigger* improvement if both ends of the link use 30dBm, and supposedly Zooz could set up their hardware like that.)

So I'm wondering if the TX power is configured correctly.

amluto:
> So I'm wondering if the TX power is configured correctly.

I tried to educate myself more, and I'm rather confused.

The most recent serial API documentation I could find is here:

https://www.silabs.com/documents/public/user-guides/INS12350-Serial-API-Host-Appl.-Prg.-Guide.pdf

and it has basically nothing to say.  `SERIAL_API_SETUP_CMD_TX_POWERLEVEL_SET` allows setting a field called `NormalTxPower`.  I think, but I'm not entirely sure, that "normal" means "not LR".

But then I found this "Knowledge" article that has a screenshot from the PC controller application:

https://community.silabs.com/s/article/z-wave-700-how-to-adjust-tx-power-on-controller-devices?language=en_US

And there's a setting called Max LR Tx Power, and it has at least one choice "Mode14dBm", which is presumably the correct choice unless you know that the hardware supports something higher.  Google knows nothing about "Mode14dBm", and Bing finds some not-obviously-relevant hits on some datasheets on LCSC in Chinese.

I'm wondering if there is some missing documentation here.

AlCalzone:
DrZWave has a blog post on LR range: https://drzwave.blog/2022/11/23/how-far-does-z-wave-long-range-reach/

I think it also depends a lot on the hardware. The official [devkits](https://www.silabs.com/development-tools/wireless/z-wave/z-wave-800-dev-kit?tab=overview) come with external antennas, but the Zooz stick only has a PCB antenna.

AlCalzone:
I've started taking a look at this again. @jtbraun if you don't mind, I'll take this PR over. Some notes for future me:

After setting the RF region to US (Long Range), the `SerialAPIStarted` command indicates that LR is supported:
```
13:34:46.202 DRIVER » [REQ] [SerialAPISetup]
                        command: SetRFRegion
                        region:  USA (Long Range)
...
13:34:47.462 DRIVER « [REQ] [SerialAPIStarted]
                        wake up reason:        WatchdogReset
                        watchdog enabled:      false
                        generic device class:  0x02
                        specific device class: 0x07
                        always listening:      false
                        supports Long Range:   true
```
so we should be able to use this flag for feature detection.


Node IDs above 255 are definitely LR. Node IDs below are definitely NOT LR.

AlCalzone:
I've updated the logic to pass the protocol selection using the provisioning entry. LR Inclusion is done exclusively via SmartStart, so this is how to include a LR node:
```ts
		driver.controller.provisionSmartStartNode({
			dsk: "...",
			protocol: Protocols.ZWaveLongRange,
			securityClasses: [SecurityClass.S2_Authenticated],
		});
```

Exclusion works as usual, no changes needed there.

IMO all that's missing now is a bit of cleanup and consistency - I'll do this tonight probably.

Thanks for the amazing head start @jtbraun

jtbraun:
> I've started taking a look at this again. @jtbraun if you don't mind, I'll take this PR over.
> Thanks for the amazing head start @jtbraun

I don't mind at all, and you're very welcome, I'm happy to have helped.

AlCalzone (changes requested):
TODO:
- [x] split the methods in packages/core/src/capabilities/NodeInfo.ts into LR / classic
- [x] move long-range related methods out of GetSerialApiInitData.ts
- [ ] fix parsing of GetNodeInfoResponse - currently it doesn't know whether the node it is for is ZWLR or classic
- [ ] documentation

packages/zwave-js/src/lib/serialapi/application/ApplicationUpdateRequest.ts
```diff
@@ -28,7 +28,7 @@ import { buffer2hex, getEnumMemberName } from "@zwave-js/shared";
 export enum ApplicationUpdateTypes {
 	SmartStart_NodeInfo_Received = 0x86, // An included smart start node has been powered up
 	SmartStart_HomeId_Received = 0x85, // A smart start node requests inclusion
-	SmartStart_HomeId_LongRange_Received = 0x87, // A start start long range note requests inclusion
+	SmartStart_LongRange_HomeId_Received = 0x87, // A start start long range note requests inclusion
```
jtbraun: Looks like I left a typo in this comment.

AlCalzone:
@jtbraun not sure if you just reviewed or Github is acting up - the review comments you supposedly added 6 minutes ago are weeks/months old.
Anyways, just to avoid any conflicts - I'm working on this and will deal with the remaining issues.

jtbraun:
I think what happened is:
* I had started an old review with a LOT of comments detailing questions for other reviewers (you).
* I never hit the "complete review" button, so you likely never saw any of those comments :-(
* I saw a typo in some of the code you edited today/yesterday that originated from me, so I commented about it so you'd see it, and closed my review.
* Presumably because of all the above, you got spammed with a ton of comments from me, instead of the one thing about typos.

Re: typo: Search for "// A start start" in [packages/zwave-js/src/lib/serialapi/application/ApplicationUpdateRequest.ts](https://github.com/zwave-js/node-zwave-js/pull/6401/files/8715fced2d75906a9926244b90663accf8f1d2e5..f9585f09bcf08b56fbf1de74054a9c442c726a06#diff-4d59b88d71766a354847dad2d9e438593b20fe0151c9323010cad4a6a3fe9a2e)

AlCalzone (changes requested):
The most recent PR fixes/refactors the node protocol info parsing. The specs are a bit confusing, but all fields are always there. It's just the actual NIF that's transmitted that doesn't have everything.

However, parsing a GetNodeProtocolInfoResponse currently doesn't know whether it should interpret the contents as ZWLR or ZW classic.

TODO:
- [x] fix parsing of GetNodeInfoResponse
- [x] documentation

AlCalzone (approved):

Issue #6402, opened 2023-10-16
Unresponsive controller recovery: Improvements


AlCalzone:
Re: https://github.com/home-assistant/addons/issues/3234#issuecomment-1764115803

- [x] just increasing the response timeout is not enough. We should probably go back to 10s, issue a Send Data Abort when this is triggered and then wait for the callback as usual.
- [x] Try reopening the serial port if soft-reset failed. If that doesn't help, move on.
- [x] Introduce a soft timeout for missing callback. Once elapsed, issue SendDataAbort, then wait until the callback timeout elapses or a callback is received.
- [x] Add setting to disable this.

jschollenberger:
Thanks for your work on resolving these issues.
I just updated to 12.x and encountering issues with my HUSBZB-1 controller. I tested this latest commit and while everything loads and works at the start, I then get the "Driver: Recovering unresponsive controller failed. Restarting the driver... (ZW0100)" error even with soft reset disabled. I rolled back to 11.14.3 where everything works, but unfortunately Home Assistant 2023.10 requires the newer version. Should I start a new issue with the logs or would you prefer we use this one?

AlCalzone:
You can share them here

jschollenberger:
I misspoke when I said I was running this latest commit. I realized I was running the master branch of zwave-js-ui, not the node application. I'm not actually sure how to do that.
Anyway, here are the relevant portions of my log file where it times out after sending a command and loops. Perhaps this is already what you addressed here, but I'm not sure how to build and test it.

```
...
2023-10-16 17:16:58.120 INFO Z-WAVE: [Node 017] Value added 17-114-0-manufacturerId => 99
2023-10-16 17:16:58.120 INFO Z-WAVE: [Node 017] Value added 17-114-0-productType => 18756
2023-10-16 17:16:58.120 INFO Z-WAVE: [Node 017] Value added 17-114-0-productId => 12337
2023-10-16 17:16:58.137 INFO Z-WAVE: [Node 017] Value added 17-119-0-name => Bedroom Dimmer
2023-10-16 17:16:58.137 INFO Z-WAVE: [Node 017] Value added 17-119-0-location => Bedroom
2023-10-16 17:16:58.148 INFO Z-WAVE: [Node 017] Value added 17-134-0-libraryType => 6
2023-10-16 17:16:58.148 INFO Z-WAVE: [Node 017] Value added 17-134-0-protocolVersion => 3.67
[...]

PR #6403, opened 2023-10-16
fix: when SendData callback times out continuously, mark node as dead and do not restart driver 

AlCalzone:
Restarting the driver as a response to timed out SendData callbacks is problematic. It seems that on older controllers, unresponsive nodes can also cause this, so restarting will just end up in a loop due to the nodes being pinged on startup.

Instead, we now treat a missing SendData callback different from a missing controller ACK. In the former case, we now soft-reset once and if this doesn't help, mark the node as dead like we used to before v12.

for #6402


AlCalzone:
@zwave-js-bot automerge

PR #6404, opened 2023-10-16
fix(config): shenzhen parameter corrections

cdnninja:
Adjustments to align some of the elements of motion to: https://www.manualslib.com/manual/1347623/Shenzhen-Nas-Pd02z.html?page=11#manual

Started looking at the sensitivity since I can't seem to change that locally and found a few more.

### DIFF
--- a/packages/config/config/devices/0x0258/templates/shenzhen_neo_template.json
+++ b/packages/config/config/devices/0x0258/templates/shenzhen_neo_template.json
@@ -1,11 +1,11 @@
 {
 	"motion_sensitivity": {
 		"label": "Motion Sensor: Sensitivity",
-		"description": "8 = highest; 255 = lowest",
+		"description": "1 = highest; 4 = lowest",
 		"valueSize": 1,
-		"minValue": 8,
-		"maxValue": 255,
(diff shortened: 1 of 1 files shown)


cdnninja:
I have found conflicting info on this device.    More research needed.

AlCalzone:
Have you confirmed that these changes apply to all devices using the templates?
AFAIK the Shenzhen Neo devices are quite inconsistent in this regard.

cdnninja:
Different manuals for what appears to be the same device conflict.  I am planning to test to see if my device ID takes the values or not.  This will help me narrow this down.      Shenzhen doesn't make this easy.

cdnninja:
Just for reference for future.  Two manuals exist: 

https://www.manualslib.com/manual/1644574/Neo-Coolcam-Motion-Sensor-2.html?page=7#manual
https://www.manualslib.com/manual/1347623/Shenzhen-Nas-Pd02z.html?page=9#manual

They appear to be same product but small differences in the manual.   
I am doing hands on testing on values to see what takes and what doesn't.    This is to solve an issue where my device is always reporting non existent motion.

PR #6408, opened 2023-10-17
fix: abort transmission after response timeout, do not retry, do not restart controller unless callback is missing

AlCalzone:

This PR changes how timed out Send Data responses are handled. Previously, we'd retry the command up to 3 times without awaiting the end of the command cycle, which can likely put the controller in a weird state, possibly triggering the "unresponsive controller" recovery unnecessarily.

Now we abort the transmission and wait for the callback to be received (expected to be with status NoACK), then treat it like a NoACK (node marked dead/asleep).

If the callback doesn't come, the recovery will kick in, possibly restarting the controller.

Since we now actively abort, the default timeout was reduced back to 10s, as not responding to incoming commands for an extended amount of time can cause further problems. Users regularly experiencing these timeouts should fix their mesh or increase the timeout.

### DIFF
--- a/docs/api/driver.md
+++ b/docs/api/driver.md
@@ -684,10 +684,10 @@ interface ZWaveOptions extends ZWaveHostOptions {
 		byte: number; // >=1, default: 150 ms
 
 		/**
-		 * How long to wait for a controller response. Usually this timeout should never elapse,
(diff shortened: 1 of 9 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6411, opened 2023-10-17
feat: abort `Send Data` commands early after timeout, reduce callback timeout

AlCalzone:

This PR introduces an additional timeout for `Send Data` commands. If this elapses while waiting for a callback, the transmission gets aborted. The driver will still wait for the callback, but unless the controller is actually unresponsive, this should now come relatively quickly.

The default for this is 20 seconds and the callback timeout has been lowered from 65 seconds to 30 seconds. While the specs explicitly mention the 65 seconds, waiting this long has shown to cause problems when nodes are reporting and expect a response.
Some commercial hubs abort around 8-10 seconds, but that seems a tad too low. I may revisit this timeout though.

This change may alleviate some of the issues with unresponsive controllers, assuming they'll honor the abort request and eventually send the callback. If not, the recovery now kicks in faster.

### DIFF
--- a/docs/api/driver.md
+++ b/docs/api/driver.md
@@ -689,8 +689,17 @@ interface ZWaveOptions extends ZWaveHostOptions {
 		 */
 		response: number; // [500...60000], default: 10000 ms
 
-		/** How long to wait for a callback from the host for a SendData[Multicast]Request */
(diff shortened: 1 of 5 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6412, opened 2023-10-17
feat: driver option presets

AlCalzone:
This PR adds support for driver option presets, allowing us to curate a set of useful option presets, starting with:
| Preset | Description |
| --- | --- |
| `SAFE_MODE` | Increases several timeouts to be able to deal with controllers and/or nodes that have severe trouble communicating. This should not be enabled permanently, as it can decrease the performance of the network significantly |
| `BATTERY_SAVE` | Sends battery powered nodes to sleep more quickly in order to save battery. |
| `AWAKE_LONGER` | Sends battery powered nodes to sleep less quickly to give applications more time between interactions. |

### DIFF
--- a/docs/api/driver.md
+++ b/docs/api/driver.md
@@ -5,12 +5,29 @@ The driver is the core of this library. It controls the serial interface, handle
 ## Constructor
 
 ```ts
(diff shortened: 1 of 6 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6413, opened 2023-10-17
fix: export all option types for `Configuration CC`

AlCalzone:
Followup to https://github.com/zwave-js/node-zwave-js/pull/6400, this is the correct fix, the other won't stick due to the file being auto-generated.


AlCalzone:
@zwave-js-bot automerge

Issue #5937, opened 2023-06-22
Add compat flag to treat any SET command as a report


AlCalzone:
re https://github.com/AlCalzone/ioBroker.zwave2/issues/1011
or https://github.com/zwave-js/node-zwave-js/issues/6290, using a controlling device as input

XLDom:
Hello, I'm not a big user of git.... (i don't understand the errors)
@AlCalzone if you can see what is blocking please: https://github.com/zwave-js/node-zwave-js/pull/6415

thanks for all

PR #6415, opened 2023-10-17
feat: add compat flag to treat BinarySwitchSet and ThermostatModeSet command as a report

XLDom:
Hello, i've try to do the job for BinarySwitchSet and ThermostatModeSet command as a report.
(exemple for SRT321 HRT4-ZW)

for this : https://github.com/zwave-js/node-zwave-js/issues/5937

I've tested : it's ok (value updated)

just one thing I don't understand: The commands arrive during their first update (I don't know why)
image
if you have un idea ?

and if you can push this update please.

Thx you.


AlCalzone (changes requested):
I'd prefer to have one compat flat, probably `treatSetAsReport`, which is an array of command names (e.g. ["BinarySwitchCCSet", "ThermostatModeCCSet"]), instead of potentially having to add one or more compat flags per command class to map.

You don't have to implement the mapping for all of them, but laying the foundation for this once would be good.

AlCalzone:
> just one thing I don't understand: The commands arrive during their first update (I don't know why)

This is because the devices don't **support** the CCs (allow being controlled), but they only **control** them in other devices. Unless specifically handled by a compat flag, Z-Wave JS only exposes values for supported CCs.

XLDom:
> I'd prefer to have one compat flat, probably `treatSetAsReport`, which is an array of command names (e.g. ["BinarySwitchCCSet", "ThermostatModeCCSet"]), instead of potentially having to add one or more compat flags per command class to map.
> 
> You don't have to implement the mapping for all of them, but laying the foundation for this once would be good.

Yes, I also thought about this (use array or end with "Set"), but the problem is that BinarySwitchCCSet returns "currentValue", and ThermostatModeCCSet retun "mode". How can I know (in code) the command name to use (currentValue or mode) ?

And how to use on this code for example ? `this._valueDB.setValue(BinarySwitchCCValues.currentValue.endpoint(command.endpointIndex), command.targetValue);`

AlCalzone:
> but the problem is that BinarySwitchCCSet returns "currentValue", and ThermostatModeCCSet retun "mode"

Yeah we still need CC-specific code to handle this like you did in `Node.ts`. That shouldn't prevent us from unifying the compat flag into one though. We'll stick with the two you already added to `Node.ts` and add others on demand.

Issue #6405, opened 2023-10-16
Thermostat Mode getSupportedModes override not working correctly


poelstra:
### Is your problem within Home Assistant (Core or Z-Wave JS Integration)?

NO, my problem is NOT within Home Assistant or the ZWave JS integration

### Is your problem within Z-Wave JS UI (formerly ZwaveJS2MQTT)?

NO, my problem is NOT within Z-Wave JS UI

### Checklist

- [X] I have checked the [troubleshooting section](https://zwave-js.github.io/node-zwave-js/#/troubleshooting/index) and my problem is **not** described there.

- [X] I have read the [changelog](https://github.com/zwave-js/node-zwave-js/blob/master/CHANGELOG.md) and my problem was **not** mentioned there.

### Describe the bug

**What causes the bug?**

My Eurotronic Spirit thermostatic valve (well, a clone brand) supports a "Manufacturer Specific" mode as one of the Thermostat Modes. There's already a warning added to it (https://github.com/zwave-js/node-zwave-js/pull/4144) but I'd thought to use the now-available compat mechanism to just fix it.

I copied the existing spirit.json to a new custom config (in store/config/spirit.json in ZWave-JS UI), removed the existing warning and added the following:
```json
	"compat": {
		"overrideQueries": {
			// The device does not report "Manufacturer specific" as a supported thermostat mode
			"Thermostat Mode": [
				{
					"method": "getSupportedModes",
					"result": [0, 1, 11, 15, 31] // Off / Heat / Energy heat / Full power / Manufacturer specific
				}
			]
		}
	}
```

I then pressed Save in UI's settings to reload, clicked the re-interview button for my device.

**What do you observe?**

In the logs, I get (as expected):

```
2023-10-15T16:23:39.259Z CNTRLR   [Node 007] Interviewing Thermostat Mode...
2023-10-15T16:23:39.259Z CNTRLR » [Node 007] querying supported thermostat modes...
2023-10-15T16:23:39.259Z CNTRLR   [Node 007] API call getSupportedModes for Thermostat Mode CC overridden by a c
                                  ompat flag.
2023-10-15T16:23:39.259Z CNTRLR « [Node 007] received supported thermostat modes:
                                  · Off
                                  · Heat
                                  · Energy heat
                                  · Full power
                                  · Manufacturer specific
```

But in the UI I then get all modes listed (including Manufacturer specific):

[...]

PR #6420, opened 2023-10-18
fix(config): correct overridden `thermostatMode` metadata for ZME_FT

AlCalzone:
reported in https://github.com/zwave-js/node-zwave-js/issues/6405


AlCalzone:
@zwave-js-bot automerge

PR #6435, opened 2023-10-19
fix: throttle `firmware update progress` events for OTA

### LINKED ISSUE #6394: Throttle firmware update progress events
Currently they are being emitted faster than Z-UI can handle them, leading to the progress indicator lagging behind while the window is open.

AlCalzone:
fixes: #6394


AlCalzone:
@zwave-js-bot automerge

PR #6448, opened 2023-10-23
fix: check for existing node associations on root, refresh after remove

### LINKED ISSUE #5222: Associations are duplicated for root endpoint
Noticed after assigning the right associations with my devices it seems to duplicate the associations to endpoint 0 ?

Looks to be duplicating something:
image

[zwave-js-ui-store.zip](https://github.com/zwave-js/zwave-js-ui/files/10067833/zwave-js-ui-store.zip)

I dont know if the logs is showing anything here related to endpoint 0 duplicate but this is the latests logs i have.

AlCalzone:
Turns out we didn't check for existing associations correctly when migrating root lifeline from node association to endpoint association.

fixes: https://github.com/zwave-js/node-zwave-js/issues/5222


AlCalzone:
@zwave-js-bot automerge

PR #6450, opened 2023-10-23
fix: allow setting `Sound Switch CC` volume

### LINKED ISSUE #6320: Sound Switch CC `volume` properly marked as writeable, but setting it is unsupported
Just something I noticed via code inspection. The [`volume`](https://zwave-js.github.io/node-zwave-js/#/api/CCs/SoundSwitch?id=volume) value is flagged as writable (type `UInt8` not `ReadOnlyUInt8`), but if you attempt to set the value, you will get an unsupported exception:

https://github.com/zwave-js/node-zwave-js/blob/34fa8a56e1b1f97c5c624c32fd8e24e678b48b55/packages/cc/src/cc/SoundSwitchCC.ts#L323-L325

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/issues/6320

Setting this property doesn't do anything immediately, it just pre-sets the volume for following play commands, e.g. by setting the `toneId`

### DIFF
--- a/packages/cc/src/cc/SoundSwitchCC.ts
+++ b/packages/cc/src/cc/SoundSwitchCC.ts
@@ -286,6 +286,23 @@ export class SoundSwitchCCAPI extends CCAPI {
 					0x00, /* keep current tone */
 					value,
 				);
+			} else if (property === "volume") {
+				if (typeof value !== "number") {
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6451, opened 2023-10-23
fix: allow setting `Sound Switch CC` volume to other values than 0

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6452, opened 2023-10-23
fix: populate tone names in metadata for `Sound Switch CC` `defaultToneId`

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6457, opened 2023-10-24
fix: preserve `stateful` and `secret` flags for dynamic CC values

AlCalzone:
https://github.com/zwave-js/node-zwave-js/pull/5467 added support for these flags, and made it so they couldn't be overridden by dynamic metadata. However the logic didn't consider that `.options` is always static, even on dynamic values, while `.meta` isn't - leading to the `stateful` and `secret` flags being ignored for dynamic CC values.


AlCalzone:
@zwave-js-bot automerge

PR #6460, opened 2023-10-24
feat(config): update fingerprints for Vesternet devices

martyn-vesternet:
updated firmware version constraints to match newer variants of Vesternet devices.

### DIFF
--- a/packages/config/config/devices/0x0330/ves-zw-hld-016.json
+++ b/packages/config/config/devices/0x0330/ves-zw-hld-016.json
@@ -11,7 +11,7 @@
 	],
 	"firmwareVersion": {
 		"min": "2.2.5",
-		"max": "2.4"
+		"max": "2.4.6"
 	},
 	"paramInformation": [
 		{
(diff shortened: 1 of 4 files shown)


AlCalzone (approved):

martyn-vesternet:
@AlCalzone I believe these should be in the latest release 12.2.3 but it doesn't look like they've been picked up.

Running HA Supervised docker Z-Wave JS UI version 9.3.0 which also reports Z-Wave JS version 12.2.3.  Checking in the file system of the container and the old config files are still present.

DB reports 12.2.3 also but still showing old config files, for example - https://devices.zwave-js.io/?jumpTo=0x0330:0x0200:0xd00f:1.4

That one should now be showing 1.4 - 1.8 for firmware field instead of 1.4 - 1.7

Just wondered if I did something wrong with the commit, or something is amiss somewhere?

AlCalzone:
This will be in the next release (probably today). The PR was merged just after the release was cut.

martyn-vesternet:
Apologies, I should have checked the commit order and not just the date!

PR #6464, opened 2023-10-27
fix(config): tweak z-trm6 options

geirra:
This is mainly a cosmetic fix for HA.
These predefined options, but since we allow manual entry Home-Assistant doesn´t list the options.

### DIFF
--- a/packages/config/config/devices/0x019b/z-trm6.json
+++ b/packages/config/config/devices/0x019b/z-trm6.json
@@ -47,6 +47,7 @@
 			"minValue": 0,
 			"maxValue": 5,
 			"defaultValue": 1,
+			"allowManualEntry": false,
 			"options": [
 				{
 					"label": "Floor",
@@ -82,6 +83,7 @@
 			"minValue": 0,
 			"maxValue": 7,
 			"defaultValue": 0,
+			"allowManualEntry": false,
 			"options": [
 				{
 					"label": "10",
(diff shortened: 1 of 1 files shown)


geirra:
We should also consider if we should use the the more technical sensor descriptions. 
Home-Assistant has some override code that switches the temp sensor depending on operation mode. 

Example: https://github.com/home-assistant/core/blob/dev/homeassistant/components/zwave_js/discovery.py#L548

Did a quick discussion on this topic in zwave_dev but, but there was no clear recommendation. As the current layout is more in line with the new simple stile of zwave-js, it might be better to keep todays code.

The only benefit in HA would be that we could reuse some code, but it´s not much.

Basically  if param2 should be changed to something like this.

``` 
		{
			"#": "2",
			"label": "Sensor Mode",
			"valueSize": 1,
			"minValue": 0,
			"maxValue": 5,
			"defaultValue": 1,
			"allowManualEntry": false,
			"options": [
				{
					"label": "F-Mode, Floor",
					"value": 0
				},
				{
					"label": "A-Mode, Internal",
					"value": 1
				},
.........
				{
					"label": "PWR-Mode, Power regulator",
					"value": 5
				}
			]
		},

AlCalzone (approved):
Thanks. I also removed the min/max value entries which are now unnecessary.

As for the label question, I don't think application code should rely too much on labels, aside from `Disable/Enable` maybe.
I'm not able to keep track what would be broken by label changes otherwise.

PR #6474, opened 2023-10-30
fix(config): add wakeup instructions for Nexia ZSENS930

The00Dustin:
In addition to wakeup instructions, I also made quoting of the button name consistent.


The00Dustin:
I've botched this enough that I'm going to redo it.

PR #6477, opened 2023-10-31
fix: reopen serial port if controller is missing ACKs after soft-reset

AlCalzone:
With this PR, we no longer restart the driver when the controller does not ACK our commands.
Instead, we now reopen the serial port, hoping that this restarts the controller (at least it should for some 500 series ones). If it doesn't help, we just carry on and hope that the controller recovers itself at some point.


AlCalzone:
@zwave-js-bot automerge

PR #6483, opened 2023-10-31
fix: auto-remove failed SmartStart nodes when bootstrapping times out

### LINKED ISSUE #6482: Smart Start "Phantom Nodes" Not Removed
### Discussed in https://github.com/zwave-js/node-zwave-js/discussions/6479

<sup>Originally posted by **@The00Dustin** October 31, 2023</sup>
### Checklist

### Describe the issue

What is happening?
At approximately 22:48 UTC, I removed battery tabs from a Smart Start device and reassembled it, I noticed two new unidentified nodes in Z-Wave JS UI (45 & 46).  After waiting a bit expecting the second one to work, I triggered inclusion on the device (and not in Z-Wave JS UI) at approximately 22:50 UTC.  This lead to a new node that successfully interviewed.

I expected the first node ID to be re-used, but now I am under the impression that each node ID that wasn't successful should have been removed automatically.  I did use remove dead node to remove these, so it is possible that they could have been removed automatically if I had waited.

Anything else we should know?
Based on past experience with a mains-powered device after some cause of this behavior was resolved, I suspected that the key to triggering my experience was to remove power early in the interview of the Smart Start device.  As such, I suspect that the battery connection was not stable in this case until after I reassembled the device.  After all of this, I realized I needed to include with no security for association to another device, so the final node ID after disabling Smart Start is actually 48.

### Software versions

Driver (node-zwave-js): 12.2.1

Z-Wave JS UI: 9.2.3

Home Assistant Z-Wave Integration: 1.33.0

Home Assistant Z-Wave JS Addon: UI 2.2.3

### Device information

This issue doesn't seem to be device-specific.  Node IDs were mentioned in issue description.

### Checklist

### Upload Logfile

[zwave-js-ui-store.zip](https://github.com/zwave-js/node-zwave-js/files/13214693/zwave-js-ui-store.zip)
</div>

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/issues/6482

Most of the code as already in place, but
1. unexpected errors during the bootstrapping wouldn't return a failure reason
2. timeouts while waiting for a command would not be exposed as a timeout, but an unspecified error (see 1.)

### DIFF
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -2769,6 +2769,17 @@ supported CCs: ${
 				return SecurityBootstrapFailure.UserCanceled;
 			};
 
+			const abortTimeout = async () => {
+				this.driver.controllerLog.logNode(node.id, {
+					message:
+						`Security S2 bootstrapping failed: a secure inclusion timer has elapsed`,
+					level: "warn",
+				});
+
+				await abort();
+				return SecurityBootstrapFailure.Timeout;
+			};
+
 			// Ask the node for its desired security classes and key exchange params
 			const kexParams = await api
 				.withOptions({ reportTimeoutMs: inclusionTimeouts.TA1 })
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6517, opened 2023-11-28
feat(config): add a new productId and add parameters to 14297/ZW1002 outlet

allenbenz:
Label is the same but I got one with a newer product Id.

Also added the parameters/metadata from the other zw1002 entry as the configuration parameters are the same. (as written on the manual in the box)

### DIFF
--- a/packages/config/config/devices/0x0063/14297_zw1002.json
+++ b/packages/config/config/devices/0x0063/14297_zw1002.json
@@ -7,10 +7,26 @@
 		{
 			"productType": "0x4952",
 			"productId": "0x3233"
+		},
+		{
+			"productType": "0x4952",
+			"productId": "0x3236"
 		}
 	],
 	"firmwareVersion": {
 		"min": "0.0",
 		"max": "255.255"
+	},
+	"paramInformation": [
+		{
+			"#": "3",
+			"$import": "~/templates/master_template.json#led_indicator_three_options"
+		}
+	],
+	"metadata": {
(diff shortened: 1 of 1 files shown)


allenbenz:
@AlCalzone anything that needs to change to make this PR acceptable?

packages/config/config/devices/0x0063/14297_zw1002.json
```diff
@@ -7,10 +7,26 @@
 		{
 			"productType": "0x4952",
 			"productId": "0x3233"
+		},
+		{
+			"productType": "0x4952",
+			"productId": "0x3236"
 		}
```
AlCalzone, line 14: @JascoProducts can you confirm if these two are actually the same product from a Z-Wave perspective? Do both have the LED indicator config param?

jascoproducts: > @jascoproducts can you confirm if these two are actually the same product from a Z-Wave perspective? Do both have the LED indicator config param?

These are essentially the same products and have identical functionality. Both are the 14297 In-Wall Outlet, TR, Lt Almd but 0x3236 has S2 and Smart Start while 0x3233 does not.

They both have the LED configuration parameter:

image

AlCalzone: Are both 500 series?

jascoproducts: > Are both 500 series?

Yes.

AlCalzone: Thanks, then we can treat them as identical

AlCalzone (approved):

PR #6525, opened 2023-11-30
feat: expose rebuild routes progress as a controller property

AlCalzone:
For some reason we were storing the progress but not exposing it before.


AlCalzone:
@zwave-js-bot automerge

PR #6526, opened 2023-11-30
fix: only expose currentValue on devices that use Basic CC for reporting

### LINKED ISSUE #6516: Ensure Basic CC is NOT supported by sensors
So far it looks like the issue is limited to sensors. I've checked the specification and for all reporting type devices (this includes multisensors), the spec is pretty clear:
> Basic Command Class MUST NOT be supported

At least my Multisensor 6 though does report support for the CC and responds when queried, which is where the entity seems to come from (at least after reporting motion once).

I think the solution would be forcibly marking Basic CC for all sensors as **unsupported**, and if they issue Basic commands or report support, treat this as **controlling** Basic CC.

_Originally posted by @AlCalzone in https://github.com/zwave-js/node-zwave-js/discussions/6494#discussioncomment-7680041_

AlCalzone:
Since HA now creates a light entity for devices supporting Basic CC, it was noticed that some sensors that cannot be controlled "support" Basic CC. This is now forbidden per the Z-Wave specifications, so we assume that the devices actually **control** Basic CC and do not **support** it. For these devices, Z-Wave JS now only exposes the `currentValue`:

grafik

fixes: https://github.com/zwave-js/node-zwave-js/issues/6516

### DIFF
--- a/maintenance/schemas/deviceClasses.json
+++ b/maintenance/schemas/deviceClasses.json
@@ -28,6 +28,9 @@
 						"controlledCCs": {
 							"$ref": "#/definitions/CCs"
 						},
+						"maySupportBasicCC": {
+							"const": false
+						},
 						"specific": {
 							"type": "object",
 							"patternProperties": {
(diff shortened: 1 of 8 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6528, opened 2023-11-30
fix(config): override user code count for Yale ZW2 locks to expose admin code

AlCalzone:
for https://github.com/zwave-js/node-zwave-js/discussions/6527

waiting on confirmation


AlCalzone:
@zwave-js-bot automerge

PR #6535, opened 2023-12-01
feat(config): add Safe Grow NSG-AB-02 Z-Wave Plus Smart Outlet Plug

waters0:
The Safe Grow NSG-AB-02 Z-Wave Plus Smart Outlet Plug is a [widely available](https://www.homedepot.com/p/Safe-Grow-Z-Wave-Plus-Smart-Outlet-Plug-Pack-of-2-SG-AB-02/303011873) smart plug. It has manufacturer id 0x013c, which suggests it's made by Philio. It appears to function similarly to a Philio PAN-15 without the power monitoring features, and configuration parameters 1-4 are therefore not supported.

As far as I can tell, neither Safe Grow nor Philio provides a manual for this device. I've got one operating on my Home Assistant instance using this configuration file, which is based on the node-zwave-gs pan-15-nes [configuration file](https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x013c/pan15-1-nes.json) and the PAN15-1/-2 [manual](https://az188720.vo.msecnd.net/product-downloads/PAN15-1-2manual-201512092_2.pdf) . I confirmed that the parameters I removed are not supported (i.e. state cannot be queried) and that the remaining parameters actually work.

### DIFF
--- a/packages/config/config/devices/0x013c/nsg-ab-02.json
+++ b/packages/config/config/devices/0x013c/nsg-ab-02.json
@@ -1,0 +1,123 @@
+{
+	"manufacturer": "Safe Grow",
+	"manufacturerId": "0x013c",
+	"label": "NSG-AB-02",
+	"description": "Smart Outlet Plug",
+	"devices": [
+		{
+			"productType": "0x0001",
+			"productId": "0x002f"
+		}
+	],
+	"firmwareVersion": {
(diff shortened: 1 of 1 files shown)


AlCalzone (approved):
Thanks, I did a few changes for consistency

PR #6542, opened 2023-12-06
feat(config): add ZVIDAR Z-TRV-V01 thermostatic valve

liollury:
I add support for thermostatic valve from ZVIDAR. I found manufacturer ID 0x045a from zwavejs server logs, productId from zwave alliance listing https://products.z-wavealliance.org/products/4819. I'm not sure which product type it is.

Parameters come from product documentation.


AlCalzone (approved):
Thanks, I made a few changs for consistency

PR #6569, opened 2023-12-27
feat(config): parameter update for Zooz Zen16 v2.0

CodeOriginal:
adds parameters available to the Zooz Zen16 v2.0 from [Zooz's Documentation](https://www.support.getzooz.com/kb/article/471-zen16-multirelay-advanced-settings/) relating to the switch inputs which are missing from the current config files.

### DIFF
--- a/packages/config/config/devices/0x027a/templates/zooz_template.json
+++ b/packages/config/config/devices/0x027a/templates/zooz_template.json
@@ -909,6 +909,78 @@
 			{
 				"label": "Garage door (momentary mode for Z-Wave control)",
 				"value": 3
+			},
+			{
+				"label": "Leak alarm (water sensor)",
+				"value": 4
+			},
+			{
+				"label": "Heat alarm",
+				"value": 5
+			},
+			{
+				"label": "Motion alert",
+				"value": 6
+			},
+			{
+				"label": "Open/close alert (door sensor)",
+				"value": 7
+			},
+			{
+				"label": "CO alarm",
+				"value": 8
+			},
+			{
+				"label": "CO2 alarm",
+				"value": 9
+			},
+			{
+				"label": "On/off report (dry contact switch/sensor)",
+				"value": 10
+			},
+			{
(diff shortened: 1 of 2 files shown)


packages/config/config/devices/0x027a/templates/zooz_template.json
```diff
@@ -909,6 +909,78 @@
 			{
 				"label": "Garage door (momentary mode for Z-Wave control)",
 				"value": 3
+			},
```
AlCalzone, line 912: This should be a new template, since those input types are only available in revision 2 (firmware 2.0+) of the switch.

packages/config/config/devices/0x027a/zen16.json
```diff
@@ -74,6 +74,13 @@
 			"$import": "templates/zooz_template.json#zen16_switch_type_v2",
 			"label": "Switch 1: Type"
 		},
+		{
+			"#": "25",
+			"$if": "firmwareVersion >= 1.2",
```
AlCalzone, line 79:
```suggestion
			"$if": "firmwareVersion >= 2.0",
```
Same for the other new parameters.

packages/config/config/devices/0x027a/zen16.json
```diff
@@ -74,6 +74,13 @@
...
+			"$import": "templates/zooz_template.json#zen16_switch_invert_v2",
+			"label": "Switch 1: Invert Value",
+			"description": "Invert the reported values for your selected switch type in Sw1 terminals (select the same value as you chose for parameter 2 to invert the values for open and closed circuit)."
```
AlCalzone, line 82: We try to keep descriptions as short as possible and avoid repeating the label in other words. Only use this if necessary to add additional information, e.g.
```suggestion
			"description": "Select the same value as in parameter 2 to invert the values for open and closed circuit."
```

Same for params 26/27

AlCalzone (changes requested):

AlCalzone:
This has been done elsewhere

PR #6581, opened 2024-01-01
fix(config): change MH9-CO2 Temperature Reporting Threshold step size to 0.1

damige:
Device supports a temperature unit of 0.1. This change reflects that.

This change has been deployed on a Home Assistant Yellow by manually editing the json and works as expected.

For reference this was discussed here: https://github.com/home-assistant/core/issues/106807

### DIFF
--- a/packages/config/config/devices/0x015f/mh9-co2.json
+++ b/packages/config/config/devices/0x015f/mh9-co2.json
@@ -80,7 +80,7 @@
 			"$if": "firmwareVersion >= 2.4",
 			"label": "Temperature Reporting Threshold",
 			"valueSize": 1,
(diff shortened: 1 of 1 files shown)


AlCalzone:
Was this changed in a newer firmware or did we have it wrong from the start?

damige:
> Was this changed in a newer firmware or did we have it wrong from the start?

I cannot answer this with 100% certainty. What i can say is that i have used this device for 2 years previously on openhab with 0.1 unit precision.

Also the physical display of the device reports 0.1 in increments.
image

This makes me believe the value was wrong from the start.

AlCalzone (approved):

Issue #6521, opened 2023-11-29
🚧 Maintenance Notification 🚧


AlCalzone:
# I'll be on parental leave until April 1st, 2024
_starting December 1st, 2023_

![parental_leave](https://github.com/zwave-js/node-zwave-js/assets/17641229/bbba961a-8561-4e4e-90fb-3a3f5adccb25)
_(artistic representation)_

**This means ...**
... I won't respond to issues, unless I need to kill some time
... PRs adding new features won't get merged (don't want to break things)
... PRs contributing bugfixes will have to wait until I find some time reviewing them (no guarantees!)
... PRs concerning configuration files will have to wait for someone knowledgeable to find some time reviewing them
... patch releases can be done by @robertsLando if necessary

**In case of emergency** (and I need to fix it), you can likely reach me on the [Z-Wave JS Discord](https://discord.gg/HFqcyFNfWd)

---

Take care and see you in 2024
_&ndash; AlCalzone_

PR #6583, opened 2024-01-02
fix(config): support Comet parameters properly

Mart124:
Hello,

This PR fixes / enhances the Eurotronic Comet Z-Wave in various ways.

### 1. Manufacturer specific mode (1st commit)

I have Eurotronic CometZ-Wave devices, which report like this :
https://pastebin.com/raw/zbF6cpfY

As for the Eurotronic Spirit devices, the _Manufacturer mode_ is [supported / documented](https://eurotronic.org/wp-content/uploads/2024/06/bedienungsanleitung-comet-z-wave-en.pdf), but not reported by the device.

```
        "4-64-0-mode": {
          "endpoint": 0,
          "commandClass": 64,
          "commandClassName": "Thermostat Mode",
          "property": "mode",
          "propertyName": "mode",
          "ccVersion": 3,
          "metadata": {
            "type": "number",
            "readable": true,
            "writeable": true,
            "label": "Thermostat mode",
            "min": 0,
            "max": 255,
            "states": {
              "0": "Off",
              "1": "Heat",
              "11": "Energy heat",
              "15": "Full power"
            },
            "stateful": true,
            "secret": false
          },
          "value": 1
        },
```

Interestingly enough is the _Full power_ mode, reported by the device, but not documented by Eurotronic.
Device's firmware may have been updated since the documentation...

Let's then apply the same modification as for the Spirit device :
https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/devices/0x0148/spirit.json

Result :

image

Note that when enabling the _Manufacturer specific_ mode in Home Assistant through its default / builtin thermostat, it works, but also reports the following warning :
_Failed to call service climate/set_hvac_mode. expected HVACMode or one of 'off', 'heat', 'cool', 'heat_cool', 'auto', 'dry', 'fan_only' for dictionary value @ data['hvac_mode']_

When setting the _Manufacturer specific_ mode manually (through Home Assistant Developer Tools / zwave_js.set_value), the warning does not trigger. Should then be related to the default / builtin thermostat itself.

### 2. Supervision (2nd commit)

There's a report [here](https://github.com/home-assistant/core/issues/103254) indicating that parameters can't be set.
As found on other devices, seems like Comet device does not like supervision. Let's then disable it.
It then fixes https://github.com/home-assistant/core/issues/103254.
**Should a report be made to Eurotronic about supervision ?**

### 3. Valve opening (3rd commit)

Some devices report valve opening parameter, some don't, may be related to firmware version (or to template used, I'm not sure here...).
But even when not initially reported by the device, it is supported.
Let's then enable this useful option.

Result (enabled after 6:00) :

image

Thank you very much 👍


Mart124:
When enabling the _Manufacturer specific_ mode in Home Assistant through its default thermostat, it works, but also reports the following warning :
_Failed to call service climate/set_hvac_mode. expected HVACMode or one of 'off', 'heat', 'cool', 'heat_cool', 'auto', 'dry', 'fan_only' for dictionary value @ data['hvac_mode']_

When setting the _Manufacturer specific_ mode manually (through Home Assistant Developer Tools / zwave_js.set_value), the warning does not trigger. Should then be related to the default thermostat itself.

Mart124:
@AlCalzone would really be glad to see this merged into next release, if possible of course 😃
Thank you very much 👍

gnipp:
Please tell me, how I may test it. I would like to add a review.
(Path to a descriptions is sufficient - I simply do not know, how to put the code on my Home Assistant instance.)
Thank you.

Mart124:
Unfortunately there's no "easy" solution.
One could be to migrate from Z-Wave JS to Z-Wave JS UI and to create your custom / own template file based on this PR.

AlCalzone:
I'll review when I find the time --> https://github.com/zwave-js/node-zwave-js/issues/6521

gnipp:
I hope, your maintenance issue #6521 was successful so far. There is no better adventure than that. All the best for you and your family!
My I however dare to ask if I can support somehow?

packages/config/config/devices/0x0148/cometz_700.json
```diff
@@ -80,6 +80,17 @@
 				}
 			]
 		},
+		{
+			// Some devices report this param, some don't, may be related to firmware version.
+			// But even when not reported by the device, it is supported.
+			"#": "6",
+			"label": "Valve Opening Percentage Report",
```
AlCalzone: Is this a threshold, meaning if this is set to 5%, the device reports every change over 5%?

packages/config/config/devices/0x0148/cometz_700.json
```diff
@@ -80,6 +80,17 @@
...
+			"valueSize": 1,
+			"unit": "%",
+			"minValue": 0,
+			"maxValue": 100,
+			"defaultValue": 0
```
AlCalzone: Does 0 mean **Disabled**?

AlCalzone (commented):
@Mart124 thanks for the PR. I hope you're still around, because I have two small questions.

AlCalzone:
Ok from what I've gathered, my assumptions are correct.

AlCalzone (approved):

AlCalzone:
Thanks, and thanks for the patience!

packages/config/config/devices/0x0148/cometz_700.json
Mart124: Yes

packages/config/config/devices/0x0148/cometz_700.json
Mart124: and yes 👍

Mart124:
Many thanks @AlCalzone for your review 👍

Issue #6584, opened 2024-01-02
Missing device configuration: Building 36 Technologies ADC-SWM150 (ValueError)


Anto79-ops:
### Checklist

- [X] It is not in the [configuration DB](https://devices.zwave-js.io/?search=3)

- [X] It was not [merged recently](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Apr+is%3Aclosed) or has a [pending PR](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Aopen+is%3Apr)

- [X] The device was interviewed completely and has no missing or `undefined` IDs

### Which device is missing?

Building36 Smart Water Valve & Meter ADC-SWM150

### Manufacturer ID

0x0190

### Product Type

0x0007-0x0001

### Product ID

0x0001

### Firmware Version

1.30

### Do you have a manual?

problem is when temperature of water goes below 5oC a temperature alarm is given but its not handled correctly in z-wave js

```
Logger: homeassistant
Source: components/sensor/__init__.py:623
First occurred: 12:16:58 PM (1 occurrences)
Last logged: 12:16:58 PM

Error doing job: Task exception was never retrieved
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 507, in async_add_entities
    await asyncio.gather(*tasks)
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 752, in _async_add_entity
    await entity.add_to_platform_finish()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1278, in add_to_platform_finish
    self.async_write_ha_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 941, in async_write_ha_state
    self._async_write_ha_state()
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 1062, in _async_write_ha_state
    state, attr, capabilities, shadowed_attr = self.__async_calculate_state()
                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 999, in __async_calculate_state
    state = self._stringify_state(available)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 947, in _stringify_state
    if (state := self.state) is None:
[...]

PR #6606, opened 2024-01-12
fix: allow State as a valid HA option for 6 States with No Data State Parameter options

apella12:
PR description here

Not being a HA user and a beginner in ZWave JS, this is highly likely not correct, but I can handle that.

This is a potential hack to address #6584.  For the 6 states that allow the "No Data" option as state parameters, the device mfr might not provide a state parameter (optional per the ZW spec) and in this case ZWave JS sends the state, not the state parameter.  This throws an HA validation error.  The intent of this PR is to add the "state" to the "state parameter" list so that validation will succeed. Conveniently all the "states" are higher than any Zwave "state parameter" options, so the device will never send that "state parameter".  It is only to inform HA that the value is allowed and means no data was provided.

### DIFF
--- a/packages/config/config/notifications.json
+++ b/packages/config/config/notifications.json
@@ -300,7 +300,8 @@
 								"0x01": "No data",
 								"0x02": "Below low threshold",
 								"0x03": "Above high threshold",
-								"0x04": "Max"
+								"0x04": "Max",
+								"0x06": "No data"
 							}
 						}
 					}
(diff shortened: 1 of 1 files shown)


AlCalzone:
I'm afraid the approach wasn't correct. When there are possible states, we want the value to be one of those, not introduce duplicate states. #6719 solves the issue that way.

apella12:
no worries. Just thought I would try for the learning.

PR #6610, opened 2024-01-17
feat(config): added Wave devices 

QubinoHelp:
added Wave devices i4, i4DC, Plug UK, Plug US, PRO 1, PRO 2, PRO 2, PRO 2PM
added wave_template

### DIFF
--- a/packages/config/config/devices/0x0460/qnpl-001X12.json
+++ b/packages/config/config/devices/0x0460/qnpl-001X12.json
@@ -1,0 +1,105 @@
+{
+	"manufacturer": "Shelly Europe",
+	"manufacturerId": "0x0460",
+	"label": "QNPL-001X12",
+	"description": "Wave Plug UK",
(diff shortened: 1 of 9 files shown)


AlCalzone:
Thanks for this. I'll try to look at this when I find some time, as I'm currently on parental leave.

apella12:
Not the official reviewer, but it looks like the lint-zwave errors are from not using the lower case hex (i.e. 008a, not 008A, etc.) in five files.

Also three files (001x16, 0a1p16, and 0a1x16) are duplicates? (same TYPE:ID and all versions) that could be a problem.  Might want to check or combine into one with all the labels

QubinoHelp:
@apella12 
Thank you for the review, I will check the files and fix the issues you mentioned.

QubinoHelp:
There was mistakes, will make a new pull request. Closing this one.

PR #6620, opened 2024-01-21
fix: use separate S2 keys for ZWLR

AlCalzone:
The specification requires to use separate S2 keys for Z-Wave Long Range, so this is what we do now. They can be specified using the new `securityKeysLongRange` property of the `ZWaveOptions`.


AlCalzone:
@zwave-js-bot automerge

PR #6623, opened 2024-01-22
fix(config): support MCO Home MH-S412 parameters properly

laurentgillet:
### Checklist
 It was not **merged recently** nor has a **pending PR**

**Device**
Mco Home Technology Co. LTD MH-S412

**Manufacturer ID**
0x015f

**Product Type**
0x4121

**Product ID**
0x5102

**Is the device listed on the Z-Wave Alliance website?**
No

**Do you have a manual?**
Yes

http://mcohome.com/index.php?c=Front/DownDetail&a=downloadFujian&name=MH-S412&path=L2NvbWRhdGEvOTUzMDEvcHJvZHVjdGZ1amlhbi8yMDIxMDIwMzE3Mjc0NjYwMWE2YzkyYjIxMTQucGRm

### DIFF
--- a/packages/config/config/devices/0x015f/mh-s412-5102.json
+++ b/packages/config/config/devices/0x015f/mh-s412-5102.json
@@ -13,6 +13,12 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
+	"metadata": {
+		"inclusion": "Hold any key for >8s OR 3 fast clicks",
+		"exclusion": "Hold any key for >8s OR 3 fast clicks",
+		"reset": "20 fast clicks on any key",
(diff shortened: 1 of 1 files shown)


AlCalzone:
Just to be sure - do you have this exact device and can confirm the parameters?
Because there also seems to be a 700 series (?) version with product ID 0x7102, but the manual on the website is not clear for which generation it is.

laurentgillet:
Absolutely! I've confirmed this particular device works correctly with the proposed parameters in this PR. I have the device physically and tested the changes locally.

As a side note, I previously used this device with Jeedom and noticed a regression in behavior after switching to ZWaveJS. This prompted me to investigate and submit this PR.

AlCalzone (approved):
Thanks for confirming. I cleaned up a bit for consistency.

PR #6625, opened 2024-01-22
fix: hasDeviceConfigChanged should be false for the controller

AlCalzone:
for https://github.com/zwave-js/zwave-js-ui/discussions/3544


AlCalzone:
@zwave-js-bot automerge

PR #6629, opened 2024-01-22
fix(config): disable Supervision for Alfred DB1 Digital Deadbolt Lock

AlCalzone:
The manufacturer has identified a bug where it stays awake after receiving a Supervision Get command with "request updates" set to true, which quickly drains the battery.


AlCalzone:
@zwave-js-bot automerge

PR #6630, opened 2024-01-22
fix(config): update Zooz ZEN30 to latest revisions

### LINKED ISSUE #6624: Zooz ZEN30: update config file to match newest firmwares
From Zooz:
We do have a ZEN30 800LR 4.10 firmware and ZEN30 700 3.20 firmware - we have added ramp rates and change single tap and double tap behavior - to match ZEN72 and ZEN77.

change log: https://www.support.getzooz.com/kb/article/393-zen30-double-switch-change-log/
parameters: https://www.support.getzooz.com/kb/article/468-zen30-double-switch-advanced-settings/

apella12:
PR description here;
Had to assume minimum level for this device was 1.5 (I'm assuming that 1.05 gets read as 1.5 per normal convention) The current file seemed to make that assumption too as param 22 was missing, 24, 25. and 26 were already added. I did not understand some of the "If" firmware statements, so changed them to how I understand it.

Quite a mess of firmware revisions. Params 28, 29 and 30 were added for 1.11, missing from 2.0, added back in 2.11, dropped for 3.0 and finally added for good in 3.10.  It should be close, so take a look if you have anytime for device configs.

Would close #6624


AlCalzone (approved):
I've taken the liberty to align the conditional stuff with how I normally do (and document) it

apella12:
No problem.  Was just trying to help (it was labeled : Help Wanted).  Actually kind of forgot about it

PR #6633, opened 2024-01-23
feat(config): added 11 Shelly Qubino Wave devices

QubinoHelp:
added Wave devices i4, i4DC, Plug UK, Plug US, PRO 1, PRO 2, PRO 2, PRO 2PM
added wave_template

fixed errors from previous pull request [feat(config): added Wave devices ](https://github.com/zwave-js/node-zwave-js/pull/6610)#6610


apella12:
I'm sure you saw the failed tests.  Most are because the "label" is above the "import".
`For consistency, config param properties should follow the order "#", "$if", "$import", "label", "description", "valueSize", "unit", "minValue", "maxValue", "defaultValue", "unsigned", "readOnly", "writeOnly", "allowManualEntry", "options"  @zwave-js/consistent-device-config-property-order`

QubinoHelp:
@apella12 Yes, I'm fixing that right now.

Is there any way to re-run the tests without doing a new pull? Sorry for the question not so skilled in programming and git
got it  :)

apella12:
@QubinoHelp Is there any way to re-run the tests without doing a new pull? No that I'm aware of.  (I'm assuming you mean after a new commit is added to the PR?)  You don't have to delete and start over like you did, just keep adding commits to fix the issues.

I don't know how you are doing this, but you should use VSC, it should show the errors as you go.

QubinoHelp:
I use the VSC I don't have set check all of the code, need more time to set that too  :)

apella12:
Last effort looks close.  I think it is just step 2 (zwave lint has passed) and the bottom on this posting to clean up the files.
[https://zwave-js.github.io/node-zwave-js/#/config-files/contributing-files](url)

apella12:
Last effort looks close. I think it is just step 2 (zwave lint has passed) at the bottom on this posting to clean up the files.
https://zwave-js.github.io/node-zwave-js/#/config-files/contributing-files

duylong:
Hi,

I tested the configuration for Wave I4. My device is well recognized, however I cannot make an association with a socket to turn it on/off, "Unable to add Node xx to Group xx of Node xx". Does it work for you? 

Tested with an association "On/Off [SW1]" in "Root Endpoint" for "Node Endpoint" and "No Endpoint" for "Target Endpoint".

QubinoHelp:
Hi @duylong Set it like this:
Leave the i4 Parameter 1 - 4 at default
Choose Root and there the input from which you want to send the On Off command, associate it with the device no endpoint
 
image

Just done the test and it is working to me, pay attention that both of the Z-Wave devices are included with the same security level.

duylong:
I had integrated it in secure mode, thanks for the tip, it's perfect now.

duylong:
For information, the association of a Wave i4 in push button mode (momentary switch) with a Wave 1 configured in push button mode (momentary switch) (with an Auto-Off at 1 second) does not seem to work. I have to click several times to get my garage door to react.

Key pressed send an BASIC_SET ON and key released send BASIC_SET OFF ? or Key pressed send an BASIC_SET ON and key pressed again send an BASIC_SET OFF ?

Maybe my usage is not suited to the Wave i4 and I need to use a Wave 1 instead.
In any case I prefer to let you know in case there are improvements to make this pull requests validated.

QubinoHelp:
Hi @duylong 

Did you open also a ticket on the Shelly support portal, since I get the same question there  ;)

If not, below the answer.

The association is linked to the inputs and will only activate upon pressing the corresponding input button. Sending the command "ON" triggers Wave 1 to turn on and after a set amount of time, it will turn off. Although this is reported to the gateway, it is not reported back to the device that sent the command (i4), so it is unaware that Wave 1 is now off. The i4 believes the last command sent was "ON" and expects the next command to be "OFF". Therefore, pressing the i4 a second time will not affect Wave 1 as it will receive the "OFF" command, and the next press will send the "ON" command again.  

Luckily this have nothing to do with the configuration files I've made the pull request, anyhow thanks for letting us know.

AlCalzone (approved):
Thanks! I cleaned up the templates a bit, removed some unnecessary fields and restored the QNSH-001P10 file which I had cleaned up when you added it.

PR #6648, opened 2024-01-30
fix(config): correct Zooz ZEN1x timer config params

### LINKED ISSUE #6647: Cannot select zero as a configuration value in UI, but can manually type in and send a zero
### Checklist

### Deploy method

Snap

### Z-Wave JS UI version

9.7.1.04fdaf5

### ZwaveJS version

12.4.1

### Describe the bug

When using the up/down arrows on a configuration value, can select any number from 1-65535.  Should be able to select any number from 0-65535.  (top bound is untested/unverified, bug is about lower bound.)

### To Reproduce

1. Include Zooz zen14
2. Open configuration v4
3. select option 2, auto-turn-off timer duration (outlet 1)
4. see value is zero
5. hit up, option is now 1 (though unsent)
6. hit down, option **will not** decrement back to 0
7. click on text box, type in 0
8. hit send button, receive confirmation that zero was sent and set

### Expected behavior

Zero should be selectable if it's a valid value (as it is in the case of a zen14).

### Additional context

Whoops, off by 1!

apella12:
PR description here
Closes #6647

### DIFF
--- a/packages/config/config/devices/0x027a/templates/zooz_template.json
+++ b/packages/config/config/devices/0x027a/templates/zooz_template.json
@@ -1,8 +1,7 @@
 {
-	// Auto-On/Off-Timers for ZEN3x and ZEN7x
-	"auto_timer_base_0x_3x_7x": {
+	// Auto-On/Off-Timers for ZEN0x, ZEN1x (except ZEN15), ZEN3x and ZEN7x
+	"auto_timer_base_0x_1x_3x_7x_nounit": {
 		"valueSize": 4,
-		"unit": "minutes",
 		"minValue": 0,
 		"maxValue": 65535,
 		"defaultValue": 0,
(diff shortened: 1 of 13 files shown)


AlCalzone:
Looks like Zooz actually uses the same timer configuration for the Zen1x series as they do for the 0x, 3x and 7x series - except the Zen15 which is different. Such a mess...

AlCalzone (approved):
Ok with my changes

TheKorn2:
Correct me if I'm wrong, but this doesn't seem to be quite correct for a zen14. The change makes multiple references to config parameter 15, 17, etc. , which don't exist for a zen14.  (Example, line 137 of the change.)  AFAIK zen14's parameters stop at 8, and the units are not configurable.

https://www.getzooz.com/downloads/zooz-outdoor-double-plug-zen14-manual.pdf

apella12:
@TheKorn2 
You are fine. Instead of a surgical one-line fix to your device in my PR the template was modified for most all the devices of the 1x series.  You have to look carefully to find the changes for your device using the heading.  Line 137 was for Zen16.  Only params 2-5 were changed for your device to reflect a new imported template that allows for "0".

AlCalzone:
Correct. I noticed that the other timer-related parameters for the Zen14 were even wronger and then I noticed that it shared most of the definitions with the other Zen1x series devices, so I cleaned them up too.

This is the relevant part for ZEN14:
https://github.com/zwave-js/node-zwave-js/pull/6648/files#diff-761aff99adad23adfcde190a6015f8454ff440010df1927e1d018d6cfbbfcf83L39-L55

TheKorn2:
Ah! OK Cool.

PR #6651, opened 2024-02-01
feat: add Zniffer support

AlCalzone:
This PR adds support for communicating with a Zniffer. This is done via a new "driver" available via the `Zniffer` class.

TODO:

- [x] code cleanup
- [x] option for selecting a default region
- [x] proper support for the different 500 series frequencies
- [x] support for parsing FLiRS frames
- [x] storing and exporting captured frames
- [x] documentation
- [x] expose CRC errors
- [x] disable strict validation (Entry Control CC etc.)
- [x] investigate "broken" frames that confuse the parser


docs/api/zniffer.md
```diff
@@ -0,0 +1,347 @@
+# Zniffer
+
+A Zniffer is a Z-Wave controller with special firmware that can capture Z-Wave traffic from any nearby Z-Wave device. This is useful for troubleshooting, as it allows you to see exactly what is being sent and received by the devices in your network.
+
```
marcus-j-davies: I suspect many implementations will start to pay more attention to this area, as such, maybe add some brief outline on how to obtain the FW (i.e links) and flash a sniffer to the stick?

In the past - I needed to register an account on silabs to obtain the tools (programmer, Sniffer FW) not sure if that process has been simplified somewhat since the gecko release on GitHub.

I cant remember if this was accurate or not - but this was coined as being a permanent flash, and you couldn't re-flash/revert to the controller API FW - I seemed to be able to (series 500), but if it is, might be worth stating that also.

AlCalzone: added a section about that

AlCalzone:
@zwave-js-bot automerge

PR #6654, opened 2024-02-02
fix(config): use Color Switch V2 for Inovelli LZW42

Zarklord:
In the Inovelli Official Hubitat driver for the LZW42, they use Color Switch V2 commands ([source](https://github.com/InovelliUSA/Hubitat/blob/master/Drivers/inovelli-bulb-multi-color-lzw42.src/inovelli-bulb-multi-color-lzw42.groovy#L338)).
I also own these lightbulbs, and I can confirm they do indeed respond to the V2 version of set, which takes a dimming duration.

I am not super familiar with how to write these config files, so if this is the wrong way to force it to use Color Switch V2, please point me in the correct direction, so that I can fix it.

### DIFF
--- a/packages/config/config/devices/0x031e/lzw42.json
+++ b/packages/config/config/devices/0x031e/lzw42.json
@@ -68,7 +68,7 @@
 		"manual": "https://products.z-wavealliance.org/ProductManual/File?folder=&filename=product_documents/3613/LZW42%20Manual.pdf"
 	},
 	"compat": {
-		// The device supports Color Switch V2, despite the Version query saying
+		// The device supports Color Switch V2, despite only reporting support for V1
 		"commandClasses": {
 			"add": {
 				"Color Switch": {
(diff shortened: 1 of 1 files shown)


AlCalzone:
If the device supports V2 of the Color Switch CC, it should tell Z-Wave JS when asked for the version. Z-Wave JS automatically uses V2 of the CC when the device supports it. The dimming duration is then either "default", or the given one.

To verify, please make a [driver log](https://zwave-js.github.io/zwave-js-ui/#/troubleshooting/generating-logs?id=driver-logs) of a re-interview, loglevel `debug` and attach it here as a file (drag & drop into the text field).

In 99% of the cases, there's no need to hardcode CC versions like in other drivers.

Zarklord:
[zwavejs_2024-02-02.log](https://github.com/zwave-js/node-zwave-js/files/14143081/zwavejs_2024-02-02.log)

Here's the log, surprise surprise, it's reporting Color Switch V1, unfortunately this isn't one of those 99% cases...

packages/config/config/devices/0x031e/lzw42.json
```diff
@@ -66,5 +66,16 @@
 		"exclusion": "Put your hub in exclusion mode and turn the power to the bulb on.  Your bulb will blink twice (2x) indicating it's in exclusion mode.  When exclusion is successful, it will blink one more time (1x) to confirm.  Your hub should say that the device is excluded.",
 		"reset": "You may power on/off the bulb 6x (between 0.5-2 seconds each time) or use a certified controller to remove the device from your network to factory default.  Only use this procedure in the event that the network primary controller is missing or otherwise inoperable.  Your bulb will flash twice to confirm factory reset.",
 		"manual": "https://products.z-wavealliance.org/ProductManual/File?folder=&filename=product_documents/3613/LZW42%20Manual.pdf"
+	},
+	"compat": {
+		// The device supports Color Switch V2, despite the Version query saying
```
AlCalzone: Heh. In this case what you did should work, but requires a re-interview to take effect.
Can you complete the comment please?
```suggestion
		// The device supports Color Switch V2, despite only reporting support for V1
```

AlCalzone (changes requested):

packages/config/config/devices/0x031e/lzw42.json
Zarklord: I'm gonna be honest, I don't know where the last half of the comment went when I commited that...

AlCalzone (approved):

PR #6685, opened 2024-02-22
fix: termostat modes switch corrected based on paragraph 7.8 Thermostat Mode of the manual

jakubbaginski:


jakubbaginski (commented):
fix: switching the modes should be done similarly to Sprint.

jakubbaginski:
fix: according to specification device modes switching should be done similarly to Sprint version

AlCalzone:
Why close?

PR #6703, opened 2024-03-24
fix(config): correct group 3 label for GE/Enbrighten 26931/ZW4006

vexofp:
Unlike other GE/Enbrighten switches, this one does not actually support double-tap. Both groups 2 and 3 behave identically (Basic Set for single tap). 😞 

The manual says as much, and confirmed through testing with my device:
image

I just set the labels to be the same; not sure if there might be a better alternative?


AlCalzone:
I don't think there's anything better :/

AlCalzone (approved):

PR #6718, opened 2024-04-03
fix: support channel 3 and use measured noise floor for SNR margin calculation if possible

AlCalzone:
This is needed to perform range tests and diagnostics with Z-Wave Long Range


AlCalzone:
@zwave-js-bot automerge

PR #6759, opened 2024-04-11
fix: consider max LR payload size in firmware updates

AlCalzone:
This effectively increases the fragment size and reduces the update time by a factor of 5 🤯

[Screencast from 11.04.2024 09:52:42.webm](https://github.com/zwave-js/node-zwave-js/assets/17641229/833bf5f6-b1fc-44f3-bd76-0a32fb55f25c)


AlCalzone:
@zwave-js-bot automerge

PR #6761, opened 2024-04-11
fix(config): versioning logic for parameter 26 of Zooz ZEN72

teliseo:
This now matches correct logic of ZEN77. ZEN32 also updated correspondingly to remove redundant compares, without changing logic.

My symptom was that parameter 26 wasn’t recognized for a ZEN72 800LR running firmware v3.30.0.


AlCalzone (approved):
Agree about the ZEN72 changes. ZEN32 was technically correct, although needlessly complicated.

PR #6765, opened 2024-04-12
fix: handle synthetic multicast `SupervisionResult` in `sendCommand`

### LINKED ISSUE #6696: Invalid response from S2 multicast setValue command
### Discussed in https://github.com/zwave-js/node-zwave-js/discussions/6695

<sup>Originally posted by **@kpine** March 13, 2024</sup>
### Checklist

### Describe the issue

What is happening?

An invalid result is returned for a S2 multicast setValue command. The result is the object `{"nodeId":255,"ccId":"Supervision","ccCommand":"0x02"}` but it should be a `SetValueResult` instead.

The invalid result causes exceptions when submitted via HA's Z-Wave integration, as it expects a `SetValueResult` return value. See https://github.com/home-assistant/core/issues/111685.

The method should return `SetValueResult`, example: `{"status":254}`.

I used ZUI driver code to exclude HA as the problem, and could reproduce. Run this driver code, where the multicast targets are included with Security S2:

```javascript
const { logger } = this

const virtualNode = driver.controller.getMulticastGroup([39, 43]);
const result = await virtualNode.setValue(
    {"commandClass": 38, "property": "targetValue"},
    24,
);

logger.info(`Response: ${JSON.stringify(result)}`);
```

You will see a log message like:

```text
2024-03-12 21:48:21.340 INFO Z-WAVE: Response: {"nodeId":255,"ccId":"Supervision","ccCommand":"0x02"}
```

Anything else we should know?

It's not a problem when the multicast targets are included with no security. With the same driver code I see:

```text
2024-03-12 22:39:47.110 INFO Z-WAVE: Response: {"status":254}
```

In the attached log file I used nodes 39 and 43 (S2 Auth) for the invalid response, and 2 and 14 (no security) for the working one.

### Software versions

zwave-js-ui: 9.9.0.09b107a
zwave-js: 12.4.4

### Device information

N/A

### Checklist

### Upload Logfile

[zwave_js.log](https://github.com/zwave-js/node-zwave-js/files/14582870/zwave_js.log)
</div>

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/issues/6696

The problem was that the S2 multicast message generator creates a synthetic SupervisionResult as the response, but `sendCommand` did not handle this because the original message was not sent supervised.

### DIFF
--- a/packages/zwave-js/src/lib/driver/Driver.ts
+++ b/packages/zwave-js/src/lib/driver/Driver.ts
@@ -5654,8 +5654,22 @@ ${handlers.length} left`,
 		}
 
 		// Fall back to non-supervised commands
+		const result = await this.sendCommandInternal(command, options);
+
+		// When sending S2 multicast commands to supporting nodes, the singlecast followups
+		// may use supervision. In this case, the multicast message generator returns a
+		// synthetic SupervisionCCReport.
+		// sendCommand is supposed to return a SupervisionResult though.
+		if (
+			options?.s2MulticastGroupId != undefined
+			&& result instanceof SupervisionCCReport
+		) {
+			// @ts-expect-error TS doesn't know we've narrowed the return type to match
+			return result.toSupervisionResult();
+		}
+
 		// @ts-expect-error TS doesn't know we've narrowed the return type to match
-		return this.sendCommandInternal(command, options);
+		return result;
 	}
 
 	/** @internal */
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot automerge

bharvey88:
also fixes https://github.com/home-assistant/core/issues/111685 right?

AlCalzone:
AFAIK yes

PR #6773, opened 2024-04-16
feat: rework compat flags for Basic CC mapping

AlCalzone:
Over time we have collected a couple of compat flags related to handing received Basic CC Set/Report commands. Those became a bit conflated, mutually exclusive and less flexible than we need. As of this PR, the flags `disableBasicMapping`, `enableBasicSetMapping`, and `treatBasicSetAsEvent` have been merged into two flags, each dedicated to one of those commands:

`mapBasicReport` defines how `Basic CC Report`s are handled:

- `false`: treat the report verbatim without mapping
- `"auto"` **(default)**: Depending on the device type (Binary Switch, Multilevel Switch, or Binary Sensor), the command is mapped to the corresponding report for that device type. If no matching mapping is found, the command is treated verbatim without mapping.
- `"Binary Sensor"`: Regardless of the device type, the command is treated like a `Binary Sensor CC::Report`.

`mapBasicSet` defines how `Basic CC Set`s are handled:

- `"report"` **(default)**: The command is treated like a `Basic CC::Report`, but the **target value** is used as the **current value**.
- `"auto"`: Depending on the device type (Binary Switch, Multilevel Switch, or Binary Sensor), the command is mapped to the corresponding report for that device type. If no matching mapping is found, the command is treated like a `Basic CC::Report`, but the **target value** is used as the **current value**.
- `"event"`: Emit a `value event` for the Basic `"event"` property.
- `"Binary Sensor"`: Regardless of the device type, the command is treated like a `Binary Sensor CC::Report`.

### DIFF
--- a/.github/workflows/zwave-js-bot_comment.yml
+++ b/.github/workflows/zwave-js-bot_comment.yml
@@ -477,7 +477,7 @@ jobs:
   # Adds a compat flag to an existing file when an authorized person posts
   # @zwave-js-bot add compat flag to <deviceId>
   # ```
-  # enableBasicSetMapping: true
+  # mapBasicSet: "auto"
   # ```
   add-compat-flag:
     if: |
--- a/.vscode/deviceSnippets.code-snippets
+++ b/.vscode/deviceSnippets.code-snippets
@@ -135,7 +135,7 @@
 		"body": [
 			"\"compat\": {",
 			"\t$LINE_COMMENT This device incorrectly ${1:[problem]}",
-			"\t\"${2|disableBasicMapping,disableStrictEntryControlDataValidation,disableStrictMeasurementValidation,forceNotificationIdleReset,mapRootReportsToEndpoint,preserveRootApplicationCCValueIDs,skipConfigurationNameQuery,skipConfigurationInfoQuery,treatMultilevelSwitchSetAsEvent,treatSetAsReport,treatDestinationEndpointAsSource|}\": true",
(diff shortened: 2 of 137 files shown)


AlCalzone:
In addition, the Aeotec Multisensor 6 config file has been updated so the Basic Sets sent by default are always mapped to the binary sensor value.

AlCalzone:
@zwave-js-bot automerge

nmpu:
Is there any good reason to create 'double-tap' association groups for 'central-scene' devices? I had deleted those associations and of course they come back when I reinterview in **Z-Wave JS UI**.

AlCalzone:
Not really. If you can tell me which devices are affected, we can remove the auto-association for them.

nmpu:
```
The following support central scene. Automatic double-click association is just redundant traffic:

UltraPro 39351 / ZW3010
Jasco 46562 / ZW4008
Enbrighten 55258 / ZW4002

However, it's possible that some automations depend on the legacy trigger. This is classic see-saw.
```

```
Jasco 14318 / ZW4005 supports group 2 & 3 association even though it does not appear to be documented.

I added the following to mimic other similar devices:

"endpoints": { "0": { "associations": { "3": { "label": "Double Tap", "maxNodes": 5, "isLifeline": true }}}},
"compat": { "mapBasicSet": "event" },

I've personally never seen any Z-Wave Plus switch that doesn't support double-tap.
```

nmpu:
FYI: These config file changes break **Jasco 26931/ZW4006** and **26932/ZW3008** motion devices in **Home Assistant**. These devices previously communicated via 37-**1** and 38-**1**. After reinterview, they now communicate via 37-**0** and 38-**0**. Obviously, any custom event automations will fail. Existing entities will become _unavailable_ and new duplicate entities will be created. It took me a while to figure out what happened and clean up the mess. I don't expect any action at his point, but it seems likely this may affect other devices.

PR #6774, opened 2024-04-17
refactor: split QR module into frontend and Node.js code

AlCalzone:


AlCalzone:
This has been done in v15

PR #6777, opened 2024-04-19
fix: force NVM migration when source has shifted app version layout

### LINKED ISSUE #6775: NVM restore: detect incorrectly formatted application version on 800 series
The 800 series application version NVM file `0x41000` is formatted differently on the Zooz stick than on the firmwares I can flash onto the devkit:

```
Zooz:   00 03 13 07
Devkit: 03 13 07 00
        ┌─ ┌─ ┌─ ┌─
        │  │  │  ??
        │  │  └ MAJOR
        │  └ MINOR
        └ PATCH
```

Since this is stored as little endian, the actual values used in code are `0x07130300` for the Zooz and `0x00071303` for the "official" firmware. When the migration routine for 7.20.0 -> 7.21.0 compares that value with `APP_VERSION_7_21_0` (`0x00071500`), the app version on the Zooz is bigger, so the migration does not trigger, causing the RF settings to be corrupted after a firmware upgrade or restore on a controller running 7.21.0 or higher.

We should detect this situation, fix the application version in NVM, and not rely on the firmware built-in migration in this case.

AlCalzone:
fixes #6775

### DIFF
--- a/packages/nvmedit/src/convert.ts
+++ b/packages/nvmedit/src/convert.ts
@@ -74,6 +74,7 @@ import {
 } from "./files/ApplicationNameFile";
 import {
 	type NVM3Objects,
+	type NVM3Pages,
 	type NVMMeta,
 	encodeNVM,
 	getNVMMeta,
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6782, opened 2024-04-22
HomeSeer products configuration changes

spudwebb:
This PR fix some issues in some existing config files for HomeSeer products and add a new config file for a new device (PS100)


AlCalzone:
FWIW you didn't have to make a new PR, this would have been enough:

grafik

PR #6787, opened 2024-04-24
fix: ignore SmartStart requests with missing keys for granted security classes

### LINKED ISSUE #6771: Ignore SmartStart inclusion requests when S2 keys are not configured
### Discussed in https://github.com/zwave-js/node-zwave-js/discussions/6644

<sup>Originally posted by **@kpine** January 26, 2024</sup>
### Checklist

### Describe the issue

If S2 keys are not configured and a user includes a node with SmartStart, the S2 bootstrapping fails (presumably, but I didn't see the logs). This problem was reported by a user in HA Discord who reported the "infinite" inclusion using SmartStart, and the problem was solved by adding S2 keys, which were previously not set.

According to one of the spec docs [1], on S2 bootstrapping failure:

> The joining node may continue with SmartStart learn mode until successful SmartStart inclusion or re-try at least up to 2 times and give up.

Kind of a weird phrasing, but it sounds like the device can either try forever, or try for a minimum of 2 times and then give up. It seems this device (ZEN04) tries forever.

So I guess it doesn't it doesn't make much sense to attempt an inclusion without S2 keys as SmartStart requires S2. I did not find anything in the spec docs that alludes to this use case.

[1] Section "4.5.4.3.4 Unsuccessful SmartStart inclusion" in Z-Wave and Z-Wave Long Range Network Layer Specification 2.0.

### Software versions

_No response_

### Device information

_No response_

### Checklist

- [ ] The log includes the problematic interaction with the device _(if applicable)_.

- [ ] I provided the node ID of the problematic device _(if applicable)_.

### Upload Logfile

None provided by user.</div>

AlCalzone:
Attempting to include a node via SmartStart will now log an error instead of including the node, failing S2 bootstrapping and starting over. Also, `provisionSmartStartNode` will now filter out invalid security classes, like S2 Unauthenticated and S0 for Z-Wave Long Range.

Fixes: https://github.com/zwave-js/node-zwave-js/issues/6771

### DIFF
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -799,6 +799,16 @@ export class ZWaveController
 		// And that the entry contains valid data
 		assertProvisioningEntry(entry);
 
+		// Discard any invalid security classes that may have been granted. This can happen
+		// when switching the protocol to ZWLR for a device that requests S2 Unauthenticated
+		// for Z-Wave Classic.
+		if (entry.protocol === Protocols.ZWaveLongRange) {
+			entry.securityClasses = entry.securityClasses.filter((sc) =>
+				sc === SecurityClass.S2_AccessControl
+				|| sc === SecurityClass.S2_Authenticated
+			);
+		}
+
 		const provisioningList = [...this.provisioningList];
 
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6795, opened 2024-04-26
fix: parse removal messages for ZWLR nodes correctly

AlCalzone:
16-bit node IDs weren't parsed correctly when removing a node, causing Z-Wave JS to think that the node is still around until restart.


AlCalzone:
@zwave-js-bot automerge

PR #6788, opened 2024-04-24
fix: only rely on NIF to determine CC support

AlCalzone:
fixes: #6784

PR #6796, opened 2024-04-26
fix: re-introduce mandatory CCs to DeviceClass as deprecated

AlCalzone:
https://github.com/zwave-js/node-zwave-js/pull/6788 was a breaking change for Z-Wave JS Server and HA. This PR re-introduces the mandatory CC fields in `DeviceClass`, but as deprecated.

@raman325 FYI, these can be removed from the server


AlCalzone:
@zwave-js-bot automerge

PR #6824, opened 2024-05-08
feat: add methods to get/set max. LR powerlevel, add driver option to automatically configure it

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6836, opened 2024-05-12
feat: add method to return Zniffer capture as Buffer

raman325:
If we want to support Zniffer's in the server, we need the `Buffer` for the capture instead of outputting to a file. This enables that to occur


AlCalzone (dismissed):
Thanks, I changed the method name to be a bit more obvious.

AlCalzone (approved):

AlCalzone:
@zwave-js-bot automerge

PR #6841, opened 2024-05-13
feat: cache Long Range RF settings, support channel selection, add channel to driver options

AlCalzone:
This PR adds functions to manage ZWLR channels, and exposes the channel selection as a driver option. Also the channel info and max LR power are now cached as a property on the controller instance.


AlCalzone:
@zwave-js-bot automerge

PR #6843, opened 2024-05-13
feat: automatically prefer LR-capable RF regions over their non-LR counterparts

AlCalzone:
With this PR, Z-Wave JS automatically prefers LR-capable RF regions over their non-LR counterparts, e.g. `USA` -> `USA (Long Range)`, during startup and when setting the region with `setRFRegion`.

Although there should be no downside to using the LR capable region, this behavior can be disabled by setting the `rf.upgradeToLRRegion` driver option to `false`.

**NOTE:** https://github.com/zwave-js/node-zwave-js/pull/6845 renames the driver option to `preferLRRegion`

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -824,6 +824,8 @@ export enum RFRegion {
 }
 ```
 
+> [!NOTE] Long Range capable regions are automatically preferred over their non-LR counterparts. This behavior can be disabled by setting the driver option `rf.upgradeToLRRegion` to `false`.
+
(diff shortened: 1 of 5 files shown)


AlCalzone:
@robertsLando when implementing this, please verify that the RF region dropdown gets updated.
image
This is currently not the case when setting or refreshing the region. I have to reload the page for this.

AlCalzone:
@zwave-js-bot automerge

PR #6848, opened 2024-05-14
fix: add `destroy` method to Zniffer to free serial port

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6849, opened 2024-05-14
fix: parse ZW inclusion frames from ZWLR devices, recognize ZWLR protocol frames, include CC in routed frames, add Zniffer active property

AlCalzone:
This PR adds a bit of polishing for Zniffer and some bugfixes:

- The `active` property can be used to determine whether Zniffer is currently running
- ZWLR protocol frames are recognized (but not parsed in detail) instead of logging an error
- ZW Classic inclusion frames from ZWLR devices are now parsed instead of erroring
- The `payload` property for routed frames contains the parsed CC now


AlCalzone:
@zwave-js-bot automerge

PR #6852, opened 2024-05-14
feat: expose captured Zniffer frames, raw data and add method to clear capture

AlCalzone:
The `"frame"` and `"corrupted frame"` events now include the raw frame data (without Zniffer protocol overhead) as their 2nd parameter. Previously captured frames can be read using the `capturedFrames` property, and cleared using the `clearCapturedFrames()` method.


AlCalzone:
@zwave-js-bot automerge

PR #6855, opened 2024-05-16
feat(config): parameter update for Zooz Zen16 v2.0 and v2.10

gmjck:
Support for Zen16 FW 2.0 and 2.1. Added new parameters (Reverse Reported Input Value, params 25-27; Input Trigger Duration, params 28-30; Fixed Input Actions, param 31) and updated defined option values for Switch Types (params 2-4) and DC Motor Mode (param 24). Moved some definitions to Zooz template as they are now common between the Zen16 and Zen17 relays. Based on Zooz Zen16 [Advanced Settings](https://www.support.getzooz.com/kb/article/471-zen16-multirelay-advanced-settings/) documentation and [2.10 Release Notes](https://www.support.getzooz.com/kb/article/356-zen16-multirelay-change-log/).

### DIFF
--- a/packages/config/config/devices/0x027a/templates/zooz_template.json
+++ b/packages/config/config/devices/0x027a/templates/zooz_template.json
@@ -954,6 +954,110 @@
 			}
 		]
 	},
+	// zen16 and zen17 relay devices now (Zen16 FW >=2.0, Zen17 FW >= 1.04) share common definition for switch type (and value inversion) and input trigger duration
+	"multi_universal_relay_switch_type": {
+		"valueSize": 1,
+		"defaultValue": 2,
+		"unsigned": true,
+		"allowManualEntry": false,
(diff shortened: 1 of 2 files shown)


AlCalzone (commented):
Looks good, I've asked Zooz for clarification about parameter 31 though. I think the current naming is confusing.

gmjck:
Yeah if they have a better internal name or are amenable to one (Separate Relay On/Off Control?) it'd be a lot clearer than the title given in the current documentation. Also wonder if they're planning a Zen17 FW release with this parameter & the definition should move to the template file.

AlCalzone:
> Also wonder if they're planning a Zen17 FW release with this parameter & the definition should move to the template file.

If they do, we'll do it then.

AlCalzone (approved):

PR #6863, opened 2024-05-21
feat: more Zniffer polishing

AlCalzone:
This PR adds:
- a special frame type to distinguish `Broadcast` frames (= non-routed singlecast with broadcast Node ID)
- an optional parameter for `znifferProtocolDataRateToString` to omit the protocol name
- an option to limit the number of captured frames kept in memory


AlCalzone:
@zwave-js-bot automerge

PR #6870, opened 2024-05-24
fix(config): override Central Scene CC version for Springs Window Fashions BRZ

spudwebb:
fixes  #6869


AlCalzone (approved):

deviantintegral:
Is there anything special needed to get a re-interview to work? I have three remotes, with one that was purchased and setup a year after the first two. The oldest ones keep showing a need to re-interview (even after repeatedly waking up the remotes), while the newest one doesn't show needing to re-interview at all.

image

AlCalzone:
Possible reasons for this:
- The device isn't fully interviewed yet
- The hash that's used to store with which device config it was interviewed isn't stored

In any case, I'd just re-interview it anyways

deviantintegral:
It turns out to wake up the remote is not just pressing an up / down button, but pressing the pairing button once on the back. Then, after the re-interview is done another single press puts the remote back to sleep.

PR #6871, opened 2024-05-26
fix(config): correct label for Remote 3-Way Switch parameter on Zooz ZEN32

tsuckow:
Update label for parameter 26


AlCalzone (approved):
I don't know why this was even overridden locally - the template already has the correct label.

PR #6876, opened 2024-05-29
feat: add driver option to configure vendor-specific constants

### LINKED ISSUE #6854: Customize manufacturer ID, product type, product ID, hardware ID, icon types for responses

AlCalzone:
This PR adds a way to configure the values of some constants used in responses to queries from other nodes.

@raman325 we may need support in the server for this, although `updateOptions` should handle it already.
@spudwebb this might interest you too

fixes: #6854

### DIFF
--- a/packages/cc/src/cc/VersionCC.ts
+++ b/packages/cc/src/cc/VersionCC.ts
@@ -724,8 +724,7 @@ export class VersionCCReport extends VersionCC {
 				.split(".")
 				.map((n) => parseInt(n))
 				.slice(0, 2),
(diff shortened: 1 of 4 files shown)


spudwebb:
@AlCalzone , not sure I understand...
Why, when and how should I use that? Could you give me a use case example?

AlCalzone:
For certification (assuming you plan on certifying with Z-Wave JS), Z-Wave JS needs to know which manufacturer ID etc. to use for responses to some queries. This gives you a way to use yours.

AlCalzone:
@zwave-js-bot automerge

PR #6890, opened 2024-06-04
fix: abort S2 bootstrapping when CSA is requested

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #6906, opened 2024-06-07
feat: add `ZWaveNode.createDump()` method

AlCalzone:
This PR adds the `createDump` method to the `ZWaveNode` class which allows exporting all known information about that node as a JSON object, including internal values which are not exposed to applications otherwise.


AlCalzone:
@zwave-js-bot automerge

PR #6907, opened 2024-06-07
feat(mock-server): support node dumps as input

AlCalzone:
With this PR, the `mock-server` can parse node dumps created by `node.createDump()`. This can either be done programmatically using the `createMockNodeOptionsFromDump` method, or by passing a file with extension `.dump` as the mock server config, either a standalone file or in a directory with mocks.

As a followup we should create more CC-specific capabilities from CC values. I don't have a useful dump handy right now.


AlCalzone:
@zwave-js-bot automerge

PR #6915, opened 2024-06-11
feat: export `NodeDump` type

raman325:
This type was not exported and therefore couldn't be used in applications consuming the library


AlCalzone (approved):
Thanks - just to be sure, please don't add any strict validation for this type on the HA side :)
I'm taking the liberty to change it as needed.

raman325:
it's really just a typing issue in Z-Wave JS Server. I'll have to figure out how to expose this in the library and will keep that in mind

PR #6918, opened 2024-06-11
fix: add compat flag to always encode Set-type commands using target node's CC version

### LINKED ISSUE #6917: TKBHome TZ67 does not understand Multilevel Switch commands encoded with V2 (including duration byte)

AlCalzone:
fixes: #6917


AlCalzone:
@zwave-js-bot automerge

PR #6926, opened 2024-06-14
chore: release v13

### LINKED ISSUE #6627: getNodeNeighbors returns nonsense for LR nodes
It shouldn't return anything, ZWLR has no mesh. Also neighbors should be removed from diagnostics. And the no. of neighbors should not be considered in calculating the route health for LR.

### LINKED ISSUE #6772: Indicate why adding an association is not allowed or failed
Hi,

I recently tried to add an association between 2 devices that did not have the same level of security. I got the errors below:

error1

error2

Once you know the problem it's easy, but if you don't, the message doesn't help.

Is it possible to improve the error message?

### LINKED ISSUE #6794: Supervised Meter CC Reset does not update values
Even if we're not using the value API, a successful Meter CC Reset should update the values locally, because no unsolicited report is expected.

### LINKED ISSUE #6823: Refactor toLogEntry not to need an application host, refactor non-device configs from JSON to TS, remove from ConfigManager

AlCalzone:
This PR is meant to collect the changes that will go into the next major release.

Fixes: #6823
fixes: https://github.com/zwave-js/node-zwave-js/issues/6794
fixes: https://github.com/zwave-js/node-zwave-js/issues/6772
fixes: https://github.com/zwave-js/node-zwave-js/issues/6627

**Migration guide:** https://github.com/zwave-js/node-zwave-js/blob/v13/docs/getting-started/migrating-to-v13.md


AlCalzone:
@zwave-js-bot automerge

AlCalzone:
rebased manually

PR #6927, opened 2024-06-14
refactor: remove hard `ZWaveApplicationHost` dependency from `CommandClass.toLogEntry()`

AlCalzone:
Now that `toLogEntry` no longer needs a `ConfigManager` instance, we can remove the hard dependency on `ZWaveApplicationHost`, allowing Zniffer to call it without passing a host.
The argument has been made optional and changed to the new `ZWaveValueHost` interface, which is just the portion from `ZWaveApplicationHost` that deals with value DBs. This allows the `MeterCCGet` command to still resolve the meter scale if the host is passed. Other CCs don't really need it.


AlCalzone:
@zwave-js-bot automerge

PR #6928, opened 2024-06-14
refactor: remove deprecated things

AlCalzone:
This PR removes the following deprecated things from the codebase:

- The first (`secure: boolean`) parameter of the `"inclusion started"` event. This has been replaced with the second parameter (`strategy: InclusionStrategy`), which was unfortunately undocumented, except in types. To determine whether an inclusion is supposed to be secure, check if `strategy !== InclusionStrategy.Insecure`.
- The `noBulkSupport` property of `ConfigurationMetadata`. There is no replacement, as it was never meant to be user-facing
- The `enableSoftReset` property of the `ZWaveOptions`. Use `features.softReset` instead.
- The `mandatorySupportedCCs` and `mandatorControlledCCs` properties of the `DeviceClass` class. There is no replacement.


AlCalzone:
@zwave-js-bot automerge

PR #6935, opened 2024-06-16
refactor: replace `Controller.isAssociationAllowed` with `checkAssociation`

### LINKED ISSUE #6772: Indicate why adding an association is not allowed or failed

AlCalzone:
This PR replaces the `Controller.isAssociationAllowed` method with the new `checkAssociation` method that returns the reason why an association is not allowed:
```ts
public checkAssociation(
	source: AssociationAddress,
	group: number,
	destination: AssociationAddress,
): AssociationCheckResult;
```
with
```ts
enum AssociationCheckResult {
	OK = 0x01,
	/** The association is forbidden, because the destination is a ZWLR node. ZWLR does not support direct communication between end devices. */
	Forbidden_DestinationIsLongRange,
	/** The association is forbidden, because the source is a ZWLR node. ZWLR does not support direct communication between end devices. */
	Forbidden_SourceIsLongRange,
	/** The association is forbidden, because a node cannot be associated with itself. */
	Forbidden_SelfAssociation,
	/** The association is forbidden, because the source node's CC versions require the source and destination node to have the same (highest) security class. */
	Forbidden_SecurityClassMismatch,
	/** The association is forbidden, because the source node's CC versions require the source node to have the key for the destination node's highest security class. */
	Forbidden_DestinationSecurityClassNotGranted,
	/** The association is forbidden, because none of the CCs the source node sends are supported by the destination. */
	Forbidden_NoSupportedCCs,
}
```

The old behavior can be restored by checking whether the returned result equals `AssociationCheckResult.OK`.

The errors thrown by `addAssociations` in case an association is not allowed (`code: ZWaveErrorCodes.AssociationCC_NotAllowed`) now contain an array with all forbidden destinations in the error's `context` property, each with an added `checkResult` property which contains the reason why the association is forbidden:
```ts
{
	checkResult: AssociationCheckResult;
	nodeId: number;
	endpoint?: number | undefined;
}[]
```

fixes: https://github.com/zwave-js/node-zwave-js/issues/6772

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -696,7 +696,7 @@ getAllAssociations(nodeId: number): ReadonlyObjectKeyMap<
 	ReadonlyMap<number, readonly AssociationAddress[]>
 >;
 
-isAssociationAllowed(source: AssociationAddress, group: number, destination: AssociationAddress): boolean;
+checkAssociation(source: AssociationAddress, group: number, destination: AssociationAddress): AssociationCheckResult;
 
 addAssociations(source: AssociationAddress, group: number, destinations: AssociationAddress[]): Promise<void>;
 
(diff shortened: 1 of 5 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6936, opened 2024-06-16
fix: do not assign routes after adding an association to a LR node

AlCalzone:
This just results in unnecessary warnings being logged.


AlCalzone:
@zwave-js-bot automerge

PR #6944, opened 2024-06-17
fix: don't ping dead nodes before attempting communication

AlCalzone:
With this PR, we no longer ping dead nodes before attempting the actual communication.
It didn't really achieve anything, as the ping still needs to be sent to the device. In case of failure, nothing is gained. In case of success, we had to send an additional command.


AlCalzone:
@zwave-js-bot automerge

PR #6945, opened 2024-06-17
fix(zniffer): hop parsing and route display for inbound frames

### LINKED ISSUE #6941: Zniffer: Routed ACK / Routed Error / AckDirect don't include the current hop
Also, we should show that this is an error, but neither Z-UI nor the logs do.

AlCalzone:
Turns out that inbound hops are encoded differently than we expected. Also we did not show the failed hop in the logs.
Also we now consistently show the route for outgoing and incoming frames in the same order, even though the source and destination nodes are swapped in the incoming frames.

fixes: https://github.com/zwave-js/node-zwave-js/issues/6941


AlCalzone:
@zwave-js-bot automerge

PR #6949, opened 2024-06-20
chore: support for Yarn 4

AlCalzone:
This PR switches our package manager to Yarn 4. We no longer vendor the yarn executable, which means [corepack](https://github.com/nodejs/corepack) is now required.


AlCalzone:
@zwave-js-bot automerge

PR #6954, opened 2024-06-23
feat: enable hardware watchdog on 700/800 series controllers

AlCalzone:
I was told that Z/IP Gateway enables the hardware watchdog on 700/800 series controllers. Unlike on 500 series, we don't have to feed it ourselves. This should help with the lockup issue, as the controller is automatically restarted when it fails to feed the watchdog.

This feature is enabled by default, unless the `ZWAVEJS_DISABLE_WATCHDOG` env variable is set, or the `features.watchdog` driver option is set to `false`.


AlCalzone:
@zwave-js-bot automerge

PR #6957, opened 2024-06-24
feat: add method to query supported RF regions

AlCalzone:
With the upcoming support for EU Long Range, the list of supported regions can differ between different 700/800 series controllers.

This PR adds a method `getSupportedRFRegions`, which can be used to dynamically determine which regions to show to a user.

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -828,6 +828,14 @@ export enum RFRegion {
 
 > [!ATTENTION] Not all controllers support configuring the RF region. These methods will throw if they are not supported
 
+To determine which regions are supported by the current controller, use the following method:
+
+```ts
+getSupportedRFRegions(filterSubsets: boolean): MaybeNotKnown<readonly RFRegion[]>
+```
+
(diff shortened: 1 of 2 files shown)


AlCalzone:
@robertsLando Z-Wave JS UI should use this, like we did with Zniffer.

AlCalzone:
@zwave-js-bot automerge

robertsLando:
> like we did with Zniffer.

What did we do for Zniffer? I actually parse them statically:

https://github.com/zwave-js/zwave-js-ui/blob/f386c4f94c4727578919c2277a81f47b62ed6155/src/lib/items.js#L4-L10

https://github.com/zwave-js/zwave-js-ui/blob/f386c4f94c4727578919c2277a81f47b62ed6155/src/lib/items.js#L12-L18

AlCalzone:
I thought we talked about that. Zniffer should definitely have a dynamic frequency selection, because they are totally different for 500 series based Zniffers:
grafik

PR #6958, opened 2024-06-24
feat: add notification variable for tilt

### LINKED ISSUE #23: Improve handling of Door/Window state
Z-Wave JS exposes two semi-duplicate values for the **Door/Window state** notification value:

| Value sent from device   | Door state                      | Door state (simple) |
|--------------------------|---------------------------------|---------------------|
| open                     | open (22)                       | open (22)           |
| open in regular position | open in regular position (5632) | open (22)           |
| open in tilt position    | open in tilt position (5633)    | open (22)           |
| closed                   | closed (23)                     | closed (23)         |

Unfortunately it is impossible to detect whether a device supports the "open in ... position" states, or just supports the CC version that includes them. As a result, the 4-state sensor is exposed as 3 or 4 binary sensor in HA, in addition to the 2-state sensor which adds another binary sensor.

This should be simplified into two binary sensors: open / close and tilted / not tilted.

AlCalzone:
This PR adds a notification variable to detect whether a window/door is tilted or not. Since there's no way to detect whether a device actually supports tilt, we only create that value once a report with the tilt flag has been received. Only after that we keep updating it.

fixes: https://github.com/zwave-js/certification-backlog/issues/23

### DIFF
--- a/packages/cc/src/cc/NotificationCC.ts
+++ b/packages/cc/src/cc/NotificationCC.ts
@@ -119,6 +119,29 @@ export const NotificationCCValues = Object.freeze({
 				autoCreate: shouldAutoCreateSimpleDoorSensorValue,
 			} as const,
 		),
+
+		// Binary tilt value extracted from the Door state variable.
+		...V.staticPropertyAndKeyWithName(
+			"doorTiltState",
+			"Access Control",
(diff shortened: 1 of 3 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #6959, opened 2024-06-24
fix: create metadata for unknown notification events for all CC versions

AlCalzone:
... not just v2


AlCalzone:
@zwave-js-bot automerge

PR #6967, opened 2024-06-26
fix: abort interview in more cases when node is dead

AlCalzone:
There were some situations outside of the interview of "normal" CCs where a failure would not abort the interview. This was especially annoying when the failure was due to the node making the controller become unresponsive, as every attempt would take ages.
With this PR, we now abort in these situations too.


AlCalzone:
@zwave-js-bot automerge

PR #6968, opened 2024-06-26
fix: reset controller again when transmit stalls after reset, don't block by enabling watchdog

AlCalzone:
If the repeated communication attempt after recovering the controller with a soft-reset also made the controller become unresponsive, the controller would stay unresponsive until some other action would reset it again.

With this PR, we try to recover the controller in this situation again. Also, enabling the watchdog after a planned soft-reset now no longer blocks everything.


AlCalzone:
@zwave-js-bot automerge

PR #6984, opened 2024-07-08
refactor: handle Basic CC interview separately, be very explicit when the CC is considered supported

AlCalzone:
Previously, Basic CC was interviewed as part of the normal CC interview machinery, but with added modifications inbetween steps. This made determining whether the CC is (supposed to be) supported very fragile and lead to confusing heuristics.

We now interview the Basic CC for the root device and all endpoints at the very end of the interview, and determine then whether the CC is supported and/or controlled.
In addition, `currentValue` is no longer exposed when the CC is controlled and Basic Sets are mapped to a different CC.


AlCalzone:
@zwave-js-bot automerge

PR #6990, opened 2024-07-09
fix(config): disable Supervision for Zooz ZSE11

### LINKED ISSUE #6974: Zooz ZSE11: Disable Supervision
It turns out that the device does not accept supervision-encapsulated Wakeup Interval Set commands (or rather responds with Fail). This results in no Wakeup notifications being sent.

AlCalzone:
fixes: #6974


AlCalzone:
@zwave-js-bot automerge

PR #7001, opened 2024-07-11
fix: only mark Basic CC as controlled if compat flags are set

AlCalzone:
Apparently the changes from 12.12.3 unintentionally exposed many Basic CC currentValue entities in cases where the CC should not be supported. With this PR, this should only happen if opted in with a Basic CC mapping compat flag.


AlCalzone:
@zwave-js-bot automerge

PR #7008, opened 2024-07-15
fix: forget supported CCs of endpoints during re-interview

AlCalzone:
It was found that the list of supported CCs of endpoints would survive a re-interview, thus preserving incorrect information.


AlCalzone:
@zwave-js-bot automerge

PR #7010, opened 2024-07-15
fix: do not mark Basic CC as supported if in secure command list

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #7012, opened 2024-07-15
fix: set version for Basic CC if Version CC is not supported

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #7023, opened 2024-07-17
fix: unknown and unused specific device class

AlCalzone:
TS was right, once again.


AlCalzone:
@zwave-js-bot automerge

PR #7030, opened 2024-07-19
fix: don't verify delivery of S2 frames in link reliability check

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #7041, opened 2024-07-23
fix: devcontainer setup

alexschneider:
Devcontainer setup needed some updates to work with current development tree.


AlCalzone (approved):
Thanks!

PR #7042, opened 2024-07-23
fix: multicast groups for LR nodes

### LINKED ISSUE #7035: ZWLR Multicast Command Affects More Nodes Than Targeted
### Discussed in https://github.com/zwave-js/node-zwave-js/discussions/7033

<sup>Originally posted by **@alexschneider** July 20, 2024</sup>
### Checklist

### Describe the issue

What is happening?
When sending a multicast command, all nodes on the network that match the command class appear to be effected, even though only a subset were targeted.

Only the targeted nodes effected. 

Send multicast command to a portion of nodes on the network. 

Anything else we should know?

### Software versions

zwave-js-ui: 9.16.1
zwave-js: 13.0.1
home id: 3806735750
home hex: 0xe2e62d86

### Device information

Manufacturer: ZVIDAR
Model name: Z-CM-V01
Node ID: 256-265

(Most recent attempt to do multicast were nodes 263-265 but there are other attempts too)

### Checklist

### Upload Logfile

[zwavejs_current.log](https://github.com/user-attachments/files/16318142/zwavejs_current.log)
</div>

alexschneider:
Fixes #7035

### DIFF
--- a/packages/core/src/security/Manager2.test.ts
+++ b/packages/core/src/security/Manager2.test.ts
@@ -191,6 +191,21 @@ test("createMulticastGroup() -> should return a different group ID for a differe
 	t.not(group1, group2);
 });
 
+test("createMulticastGroup() -> should return a different group ID for a different node set for LR nodes", (t) => {
+	const man = new SecurityManager2();
+	dummyInit(man);
+	const group1 = man.createMulticastGroup(
+		[260, 261, 262],
+		SecurityClass.S2_Authenticated,
+	);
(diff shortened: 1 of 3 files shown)


AlCalzone (changes requested):
This doesn't work for 2 reasons:
1. Changing `encodeNodeBitmask` breaks code that expects the bitmask to be 29 bytes long
2. Using 1000-character strings as keys does not sound like a good idea

I'll make some modifications.

AlCalzone (approved):

PR #7051, opened 2024-07-29
fix(zniffer): do not query region info on Zniffers before FW 2.55

AlCalzone:
Not sure in which version the `GetFrequencyInfo` command was implemented, but 2.55 definitely supports it. Before that, just assume the regions are the same ones as in 2.55.


AlCalzone:
@zwave-js-bot automerge

AlCalzone:
@zwave-js-bot pack this

PR #7058, opened 2024-07-30
fix: preserve granted security classes when switching protocols

### LINKED ISSUE #7055: Smart Start Provisioning Entry Settings Change During Protocol Switch
### Checklist

### Deploy method

Home Assistant Add-on

### Z-Wave JS UI version

9.14.5

### ZwaveJS version

12.12.3

### Describe the bug

Whenever a smart start provisioning entry is changed from `Z-WAVE` to `Z-WAVE LONG RANGE` and then back, the state of the `S2 Unauthenticated` and `S0 Legacy` checkboxes will be unchecked even if they were checked by default and not previously changed by the user.

### To Reproduce

1. Scan the QR for a device that supports LR
2. Note the state of the mentioned checkboxes (ensure at least one is checked)
3. Switch the provisioning entry to LR
4. Switch the provisioning entry back
5. Note the state of the mentioned checkboxes (observe they are no longer checked)

### Expected behavior

Presumably optimal behavior would be to leave the checkbox values alone when switching between protocols, ignoring the inapplicable checkboxes on the backend.  This way, they will consistently contain their "correct" (default or user selected) value regardless of prior circumstances.  Given that, whether inapplicable checkboxes are hidden, disabled, or left in a modifiable state even though they won't have any effect with the selected protocol becomes a design decision.

### Additional context

I originally observed this with both mentioned checkboxes in a previous version where I believe the checkboxes disappeared when switching to LR and reappeared unchecked when switching back.  I more recently observed it on this version with a device that did not have `S0 Unauthenticated` checked by default and believe the checkboxes unchecked vs disappearing when I switched to LR.  If both recollections are accurate, then it is feasible (though presumably unlikely) that changes to this feature could mean that the bug now only applies to `S2 Unauthenticated`.

AlCalzone:
fixes: #7055

### DIFF
--- a/packages/cc/src/cc/index.ts
+++ b/packages/cc/src/cc/index.ts
@@ -140,9 +140,11 @@ export type { ClockCCSetOptions } from "./ClockCC";
 export { ClockCC, ClockCCGet, ClockCCReport, ClockCCSet } from "./ClockCC";
 export type {
 	ColorSwitchCCGetOptions,
+	ColorSwitchCCReportOptions,
 	ColorSwitchCCSetOptions,
 	ColorSwitchCCStartLevelChangeOptions,
 	ColorSwitchCCStopLevelChangeOptions,
+	ColorSwitchCCSupportedReportOptions,
 } from "./ColorSwitchCC";
 export {
 	ColorSwitchCC,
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #7062, opened 2024-07-30
fix: version of Humidity Control Mode CC is 1, not 2

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #7070, opened 2024-08-01
fix: abort S2 bootstrapping when `KEXSetEcho` has reserved bits set

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #7078, opened 2024-08-02
feat(cc): bump version of Association and MCA CC

AlCalzone:
There were no changes on the controller side, and we already send unsolicited commands using the highest security class of the target


AlCalzone:
@zwave-js-bot automerge

PR #7079, opened 2024-08-02
feat: support Firmware Update Meta Data CC v8

### LINKED ISSUE #7075: Add support for Firmware Update CC v8

AlCalzone:
This PR adds support for resuming aborted firmware updates and transferring firmware update data without encryption to reduce overhead. This requires end devices implementing version 8 of the firmware update CC.

The methods to perform firmware updates with or without the update service can now specify whether those features should be used if supported through an additional `options` parameter:
```ts
export interface FirmwareUpdateOptions {
	/**
	 * Whether a previous attempt to update this node's firmware should be resumed (if supported).
	 */
	resume?: boolean;
	/**
	 * Whether the firmware data should be transferred without encryption (if supported).
	 * This can massively reduce the time needed.
	 */
	nonSecureTransfer?: boolean;
}
```

fixes: #7075


AlCalzone:
@zwave-js-bot automerge

PR #7080, opened 2024-08-02
fix: try to handle non-implemented CCs

AlCalzone:
With this PR, non-implemented CCs will no longer be dropped immediately. Instead possible awaiters are invoked first, in case the application is waiting for them.


AlCalzone:
@zwave-js-bot automerge

PR #7092, opened 2024-08-10
fix(config): fix for condtional parameters for zen16

ksmith173:
Fixed conditional parameter checks per https://www.support.getzooz.com/kb/article/356-zen16-multirelay-change-log/

Conditional checks incorrectly assume 1.1, 1.2, and 1.3 for items that are actually 1.0.1, 1.0.2, and 1.0.3. Note: the link above lists them as 1.01, 1.02, and 1.03. My devices are on the latest firmware (1.20) for hardware version 1 which is reported as 1.2. As such, with the current config file I cannot access parameters (21-24) that I should be able to.

### DIFF
--- a/packages/config/config/devices/0x027a/zen16.json
+++ b/packages/config/config/devices/0x027a/zen16.json
@@ -64,13 +64,13 @@
 		},
 		{
 			"#": "2",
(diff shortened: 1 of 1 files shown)


ksmith173:
Disregard.... user error

PR #7114, opened 2024-08-15
feat: implement 32-bit NVM operations

AlCalzone:
This PR adds support for the extended NVM operations command that uses 32-bit addresses.


AlCalzone:
@zwave-js-bot automerge

PR #7115, opened 2024-08-15
fix(config): remove endpoint workaround for Zooz ZEN30, FW 3.20+

AlCalzone:
According to Zooz, this issue is also fixed in FW 3.20


AlCalzone:
@zwave-js-bot automerge

PR #6751, opened 2024-04-10
feat: add support for EU Long Range

AlCalzone:
This can be tested now on 800 series devkits using Simplicity Studio and the new **Simplicity SDK** from Silicon Labs in version 7.22.1.
Alternatively, the latest firmware versions of Z-Wave.me hardware also support EU LR.

PR #7118, opened 2024-08-16
feat: query supported RF regions and their info

AlCalzone:
Newer firmwares have support for querying which regions are supported. This is especially useful for knowing whether EU_LR is supported.

Enables: #6751


AlCalzone:
@zwave-js-bot automerge

PR #7122, opened 2024-08-23
fix(config): add alarm mapping for Schlage lock CKPD FE599

spudwebb:
Added alarm mapping for the Schlage lock FE549 because it was not reporting any notifications. 
Maybe some other Schlage locks need the same mapping but I'm not sure of which ones.

### DIFF
--- a/packages/config/config/devices/0x003b/fe599nx.json
+++ b/packages/config/config/devices/0x003b/fe599nx.json
@@ -92,5 +92,18 @@
 			"defaultValue": 255,
 			"unsigned": true
 		}
-	]
+	],
+	"compat": {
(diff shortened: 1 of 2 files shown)


AlCalzone (approved):

ibddot:
What was the reason for adding the Keypad unlock mapping? For my setup the mapping creates unexpected behaviors with this lock since the physical lock auto re-locks after ~10 seconds but that action is not captured anywhere and so zwave thinks the lock is still unlocked when it actually is locked.

I created a discussion on this but haven't gotten any input there yet: https://github.com/zwave-js/node-zwave-js/discussions/7511

Issue #7096, opened 2024-08-12
Take over network from existing controller


AlCalzone:
To make migration from legacy controllers easier, especially when NVM backup/restore is not an option, consider the following flow:

- Start learn mode, join existing network. As part of this, existing nodes' info is transferred to Z-Wave JS
- Ask user to turn off existing controller
- Read NVM, modify NVM to set own node ID to the node ID of the previous SUC
- Write modified NVM
- Restart to apply modified NVM
- Maybe (investigate?): Change own role to SUC/SIS

TomMalfrere:
I'm following...
I have the following setup 
- Raspberri Pi 3B+ rev1.3
- Raspian Bullseye 32bit OS
- AEOTEC Gen5 ZW090 USB controller 
- z-wave-js UI in a docker
- MQTT (mosquitto) in a docker
- Domoticz (not in a docker)

I'm have issues lately with my AEOTEC Gen5 ZW090 USB-stick that becomes unresponsive after 2-3 days.
A reboot (ZWave stick remains powered and plugin) fixes the problem.

I would like to move to
- Raspberri Pi 3B+ rev1.4
- Raspian Bookworn 64bit OS
- RAZBerry 
- z-wave-js UI in a docker
- MQTT (mosquitto) in a docker
- Domoticz (in a docker)

How can I replace the Z-Wave controller in my network but make sure my devices are reported the same in MQTT ?
This is a live system with 41 z-wave devices.

AlCalzone:
If your Z-Stick is on firmware 1.2 you can just take an NVM backup and restore that on a new controller using Z-Wave JS UI.

TomMalfrere:
> If your Z-Stick is on firmware 1.2 you can just take an NVM backup and restore that on a new controller using Z-Wave JS UI.

  "productLabel": "ZW090",
  "productDescription": "Z‐Stick Gen5 USB Controller",
  "manufacturer": "AEON Labs",
  "firmwareVersion": "1.0",
  "sdkVersion": "6.51.2",
  "protocolVersion": 3,

AlCalzone:
It's possible that you can upgrade with the firmware updater from Aeotec (search for Z-Stick 5 firmware 1.2).

If not then this feature would be for you.

TomMalfrere:
I tried updating the firmware this weekend. It did not work.
It appears my stick is running since september 2017 so it was likely manufactured before 2018...

[...]

Issue #1070, opened 2020-10-22
Support controller replication etc.


AlCalzone:
According to INS13954-7, several more steps and different timeouts are involved when including a controller. We currently do the bare minimum, but should check out what to do when including a controller (and what that SIS/SUC stuff means 😅 )

robertsLando:
http://www.dtvrc.com/2019/02/06/difference-between-suc-and-sis-controller/

AlCalzone:
Well, that is better than the sad attempt at an explanation that can be found in the specs.

AlCalzone:
Nope and honestly it's not at the top of my to do list.

eliasisrael:
I understand. Lots to do and many requests. I'm interested to help if I can, but my hours available for coding are limited these days. Assuming I could do something, would that be a help?

eliasisrael:
OK, let me try to learn the code base and see if I can plausibly spend enough time to be useful. After that, we can talk about things.

kizza42:
Just chiming in on this issue as well. 
I have a different [Z-Wave hub](https://products.z-wavealliance.org/products/2720?selectedFrequencyId=-1) (I'm on the team developing it) as my primary controller and added a Aeotec USB Stick as a secondary controller, for the most part, things work well together with HA/ZWaveJS. I'm a bit jealous because you guys do few things a bit better than our implementation!! 
But I noticed the other day that new nodes added to my primary controller aren't being passed through to HA/ZWaveJS.
If I add items via HA/ZWaveJS they do pass back to my primary controller however...

Hedda:
Any updates on ZWave JS device driver feature request for secondary Z-Wave Controller (Z-Wave secondary controller) support?

https://github.com/zwave-js/node-zwave-js/discussions/1901

Or does that feature request need a separate feature request or is that the same as supporting Z-Wave controller replication?

PR #7135, opened 2024-08-29
feat: support learn mode

### LINKED ISSUE #6428: Support Learn Mode to become a secondary controller
### Discussed in https://github.com/zwave-js/node-zwave-js/discussions/6427

<sup>Originally posted by **@thomaswitt** October 19, 2023</sup>
### Description

I bought an Aeotec Z-Stick 7 for my RPi 4 and hoped it could be used as a secondary controller for my Homey Pro system (as the Homey Aeotec App supports the Z-Stick Gen 5 officially). There's even some [official documentation](https://aeotec.freshdesk.com/support/solutions/articles/6000263528-z-stick-gen-5-as-secondary-controller-in-homey) on the Aeotec page for that. I wanted to gain three functionalities by using Z-Wave JS UI with that:

1. Seamless MQTT integration - so that all actions of all my devices can be broadcasted to my MQTT broker.
2. Firmware Updates for Z Wave Devices (as Homey doesn't support Firmware Updates)
3. Getting easy API access to send commands to Z Wave Devices (e.g. turning on/off a light via an API call easily)

Unfortunately, Z-Wave JS UI - as I understand - doesn't support learn mode at the moment and therefore I can't include it as a device/secondary controller in my existing network.

### Do you have a solution in mind?

I would suggest to add support for Z-Wave UI JS for being a secondary controller.

### Describe alternatives you've considered

There are non as I don't want to drop Homey as my primary system.

### Additional context

_No response_</div>

AlCalzone:
This PR adds support for joining another network using S2 and S0 and being a secondary controller

fixes: #6428

enables https://github.com/zwave-js/node-zwave-js/issues/7096
enables #1070


AlCalzone:
@zwave-js-bot automerge

iskiselev:
Should FAQ be also updated now? https://github.com/zwave-js/node-zwave-js/blob/master/docs/getting-started/faq.md or is it still considered "Not supported"?

AlCalzone:
It's not fully supported yet - the other half of the Inclusion Controller CC implementation is missing. But I guess it's good enough to update the FAQ.

brandond:
Does this mean I can run a second instance of zwave-js-ui elsewhere in my environment, with another stick. and join it to the network as a range extender? It's not clear from the docs how to get it to join an existing network on startup.

AlCalzone:
No, a secondary controller does not have the same functionality as the primary. For example with some devices you can not have them report to more than one controller.

What you may be thinking of (multiple instances cooperating) would require some backend work. It may come one day, but no guarantees.

brandond:
Right but even if I didn't want to have it report to the secondary controller, or do any management from that node, could I at least get it as far as joining the network to extend the range of the mesh? I'm confused as to what the current limited support actually allows for.

AlCalzone:
I don't think there's support for this in Z-Wave JS UI yet.

> I'm confused as to what the current limited support actually allows for.

Mainly in preparation for https://github.com/zwave-js/node-zwave-js/issues/7096

/cc @robertsLando

robertsLando:
No support in UI yet but you could still try using it with driver function

brandond:
Do you have any pointers on how to do that? I'm assuming I'd need to run the driver headless and pass it the keys and such?

robertsLando:
@brandond Never tried those new APIs so I have no clue but if you want to understand how to work with custom driver functions you can look at docs here: https://zwave-js.github.io/zwave-js-ui/#/usage/driver_function

37Rb:
I'm interested in experimenting with this as well. I'm using the [Node-RED Z-Wave JS](https://flows.nodered.org/node/node-red-contrib-zwave-js) package. I don't expect to be able to do it from there but can I use learn mode via code in the driver used by Node-RED and then use the UI after it's included?

AlCalzone:
@37Rb what do you expect to do when Z-Wave JS is a secondary controller?
Right now it "works", but it probably does not behave as expected - i.e. no automatic device interview and when you trigger it manually, it will assign a wrong lifeline association.

37Rb:
I'm trying to discover what's possible to do. I want to add a z-stick (Zooz ZST39 LR) to an existing network and use Node-RED to do as much as possible with it. The goal is to enhance the existing network to do automations that aren't possible due to limitations on the primary controller. I want to add Z-Wave to [this tool](https://support.suretyhome.com/t/surety-iq-mate/32721) and see how much mileage I can get.

marcus-j-davies:
@37Rb 

if you’re willing to todo some testing - I can start the ground work in the NR module?

It can then be done in the UI

37Rb:
I'm happy to do testing on it. That would be easier than trying to modify the driver instance myself!

marcus-j-davies:
I'll ping you when something is ready to test.

PR #7173, opened 2024-09-17
fix: do not deep-clone old log config in `updateOptions`

AlCalzone:
This PR fixes an issue where passing a custom log transport to `updateOptions` would cause a call stack overflow.


AlCalzone:
@zwave-js-bot automerge

PR #7174, opened 2024-09-17
fix: identification of primary controller role on older controllers

AlCalzone:
This should fix https://github.com/home-assistant/core/issues/125843

Apparently some older controllers don't have the **real primary** flag set, even when they are the primary controller.


AlCalzone:
@zwave-js-bot automerge

PR #7178, opened 2024-09-19
fix(config): correct max. value of SKU parameters for Kwikset locks

ccutrer:
Without this, when using ZWave-JS-UI, openHAB is complaining that the value my locks are reporting is not in range 0 to 0.


AlCalzone:
Should those parameters be readonly? Sounds like they are, but they are currently not marked that way.

AlCalzone (dismissed):

ccutrer:
You're correct, they're read only. I just tried to update it in ZWave JS UI, and it reverted back after a second.

AlCalzone (approved):

PR #7181, opened 2024-09-20
feat: log debugging information for S0 encryption in dev mode

### LINKED ISSUE #7162: Include S0 plaintext in debug logs when `NODE_ENV=development`
### Discussed in https://github.com/zwave-js/node-zwave-js/discussions/7157

<sup>Originally posted by **@spudwebb** September 12, 2024</sup>
Is there an easy way to decode the serial data that one can see in the log file at debug level, when the node use S0 or S2?

Context:  I'm investigating an issue with a lock (Kwikset 910) that is sending a door lock logging report with an "empty" status. Since it is marked as empty zwavejs does not parse any other information from the message sent. I suspect this lock to send a valid report in which only the status is wrong, because the same device was working with another driver. So, I would like to confirm this by manually parsing the message to check what data (timestamp, event type, user code) it contains.

```
2024-09-12T14:35:21.667Z SERIAL « 0x012f00a8000001004c2198810452043a03a036e03c87addd9941566e85b4ee9e7 (49 bytes)
                                  cfe98cc8ffbb27a074c3100b2007f7fd8
2024-09-12T14:35:21.668Z SERIAL » [ACK]                                                                   (0x06)
2024-09-12T14:35:21.671Z DRIVER « [Node 076] [REQ] [BridgeApplicationCommand]
                                  │ RSSI: -78 dBm
                                  └─[SecurityCCCommandEncapsulation]
                                    │ sequenced: false
                                    └─[DoorLockLoggingCCRecordReport]
                                        record #: 6 (empty)
```</div>

AlCalzone:
fixes: #7162

### DIFF
--- a/packages/cc/src/cc/SecurityCC.ts
+++ b/packages/cc/src/cc/SecurityCC.ts
@@ -656,6 +656,11 @@ export class SecurityCCCommandEncapsulation extends SecurityCC {
 			this.secondFrame = !!(frameControl & 0b10_0000);
 
 			this.decryptedCCBytes = frameControlAndDecryptedCC.subarray(1);
+
+			// Remember for debugging purposes
+			this.authData = authData;
+			this.authCode = authCode;
+			this.iv = iv;
 		} else {
 			this.encapsulated = options.encapsulated;
 			options.encapsulated.encapsulatingCC = this as any;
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #7195, opened 2024-09-27
feat(config): add fingerprint to Remotec ZXT-800.json

jbarreiros:
The ZXT-800 device configuration already exists. This PR adds an additional device to the `devices` array.

I pulled the `productType` and `productId` from https://products.z-wavealliance.org/products/4788 and confirmed via the "Download Diagnostics" button in Home Assistant.

### DIFF
--- a/packages/config/config/devices/0x5254/zxt-800.json
+++ b/packages/config/config/devices/0x5254/zxt-800.json
@@ -8,6 +8,11 @@
 			"productType": "0x0004",
 			"productId": "0x8492",
 			"zwaveAllianceId": 4788
+		},
+		{
+			"productType": "0x0100",
+			"productId": "0x8493",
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x5254/zxt-800.json
```diff
@@ -8,6 +8,11 @@
 			"productType": "0x0004",
 			"productId": "0x8492",
 			"zwaveAllianceId": 4788
```
AlCalzone: This ID belongs to the entry you just added. Not sure why it is assigned to the existing one.
```suggestion
			"productId": "0x8492"
```

AlCalzone (changes requested):

AlCalzone (approved):

PR #7193, opened 2024-09-26
feat: implement task scheduler

AlCalzone:
Enables: https://github.com/zwave-js/node-zwave-js/issues/3707

Issue #3707, opened 2021-11-14
[Tracking] Migration to task scheduler


AlCalzone:
As discussed on Slack, there are issues with SmartStart inclusion when too many things are happening at once. For example multiple interviews, plus automatic inclusion of nodes. It seems that the active communication with nodes is interfering with the passive communication that's happening under the hood.

This is somewhat parallel to #903 and #884, but we could avoid this by grouping communication in batches (some of which can be paused) and scheduling them. We currently only queue individual transactions, leading to seemingly chaotic queues during busy interview times. By scheduling larger batches, interviews of individual nodes wouldn't be interrupted by others.
User-initiated communication would not be affected.

Use cases:
- [x] Inclusion
- [x] Exclusion
- [ ] Security bootstrapping (not pausable, other than the times where user interaction is happening)
- [ ] Node interview (pausable for high-priority batches)
- [x] Remove failed nodes
- [x] Replace failed nodes
- [x] Network healing #7196
- [ ] Health checks
- [x] OTA updates #7256
- [x] dropping everything for switching to bootloader #7255
- [x] dropping everything for hard-resets #7255

PR #7196, opened 2024-09-27
feat: rework route rebuilding to use task scheduler

AlCalzone:
With this PR, rebuilding routes of individual nodes or the entire network uses the new task scheduler introduced in #7193.

This means that `beginRebuildingRoutes` can now be called while individual node rebuilds are pending, and `rebuildNodeRoutes` can be called for multiple nodes without having to wait for the previous rebuild to finish.

Overlapping tasks are reused, so if `rebuildNodeRoutes` is has queued a task to rebuild routes for one node, a concurrent `beginRebuildingRoutes` will wait for the queued task instead of queued another one.

solves one part of https://github.com/zwave-js/node-zwave-js/issues/3707


AlCalzone:
@zwave-js-bot automerge

PR #7252, opened 2024-10-08
feat: skip rebuilding routes for nodes with priority return routes

### LINKED ISSUE #7147: Rebuild routes: Add option to skip nodes with priority routes
### Discussed in https://github.com/zwave-js/node-zwave-js/discussions/7146

<sup>Originally posted by **@duylong** September  4, 2024</sup>
from time to time I redo my routes following a change in my network. In this action, we can ignore the modules on battery, it would be nice to be able not to overwrite the custom routes.</div>

AlCalzone:
With this PR, full-network route rebuilding will skip nodes with priority return routes (end node or SUC) by default, unless the new `deletePriorityReturnRoutes` option is set to `true`.

Note that rebuilding routes for a single node will still delete existing priority return routes to end nodes and the SUC. It is recommended to first check if priority return routes are known to exist using `getPriorityReturnRoutesCached` and `getPrioritySUCReturnRouteCached` and asking for confirmation before proceeding.

fixes: #7147

### DIFF
--- a/docs/api/controller.md
+++ b/docs/api/controller.md
@@ -428,6 +428,8 @@ async rebuildNodeRoutes(nodeId: number): Promise<boolean>
 
 Rebuilds routes for a single alive node in the network, updating the neighbor list and assigning fresh routes to association targets. The returned promise resolves to `true` if the process was completed, or `false` if it was unsuccessful.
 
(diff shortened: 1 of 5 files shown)


AlCalzone:
@zwave-js-bot automerge

Issue #2397, opened 2021-04-18
Support Leviton scene/zone controllers


j9brown:
**Is your feature request related to a problem? Please describe.**

Now that OpenZWave is deprecated in Home Assistant, I've started looking at device support that I'll need to migrate over to node-zwave-js.  This issue concerns Leviton devices (of which I have many) such as VRCS2, VRCS4, VRCZ2, VRCZ4, VRR15, VRS15, etc.

These devices use some custom and less common command classes for various features.  OpenZWave never supported them but I was able to work around the limitations by sending raw Zwave commands and parsing responses (not ideal).  I'd love to implement a better solution for node-zwave-js but need some guidance.

Features needed:
- Indicator LED setting (each LED can be individually set to green/amber/red to indicate status)
- Configuring scene groups
- Handling scene group activations
- Handling scene up/down level changes
- Quirks: switches don't always send BASIC_SET when toggled at the wall but they do send SCENE_ACTUATOR_CONF_REPORT which can be used as a trigger to refresh the value

**Describe the solution you'd like**

I basically need some help to plan the changes needed to support these features.  I'm happy to write the code or provide documentation or test cases to delegate to someone else more familiar with this project.

For LED setting, we'll need a custom command-class.  How should vendor-specific classes be implemented?

For scene activations, we'll need some method to configure the device's scene groups and receive activations so that the button presses can be handled correctly.  This setting uses a different set of messages from the usual configuration parameters or association groups so how should it work?

**Describe alternatives you've considered**

If there's a way to send/receive raw Zwave messages then I could keep using my own code which generates/parses them but it wouldn't be of much use to anyone else.

**Additional context**

Here are some technical details from a patch request I sent to OpenZwave but didn't commit (because it would have required deeper changes to the project to implement properly): https://github.com/j9brown/open-zwave/tree/master/config/leviton

I reverse engineered most of this using a Zniffer.

For example, this explains some quirks of VRCZ4.

```
<!-- Indicator Lights -->
  <!-- Use these messages to override the indicator lights.
       91 00 1D 0D 01 00 00 : Reset LEDs to locally controlled operation (default behavior).
       91 00 1D 0D 01 FF xx : Set LEDs by OR-ing together the following bit patterns.
         00000000: button 1 off
         00000001: button 1 green
         00010000: button 1 red
         00010001: button 1 amber
         00000000: button 2 off
         00000010: button 2 green
         00100000: button 2 red
         00100010: button 2 amber
[...]

PR #7253, opened 2024-10-08
feat: add `node info received` event

### LINKED ISSUE #6867: Expose the receipt of a NIF as an event
_Originally posted by @alexeip0 in https://github.com/zwave-js/node-zwave-js/issues/2397#issuecomment-2119320177_

I believe the same can be achieved with HA if we had ability to trigger off of this event:
```
2024-05-16T07:06:37.496Z SERIAL « 0x0115004984050f020100852d7c778273867291ef2b2cf1                    (23 bytes)
2024-05-16T07:06:37.498Z SERIAL » [ACK]                                                                   (0x06)
2024-05-16T07:06:37.500Z DRIVER « [Node 005] [REQ] [ApplicationUpdateRequest] payload: 0x050f020100852d7c778273867291ef2b2c
2024-05-16T07:06:37.501Z CNTRLR « [Node 005] Received updated node info
2024-05-16T07:06:37.502Z CNTRLR   [Node 005] Node does not send unsolicited updates; refreshing actuator and sensor values...
```
HA observes it when a controller button is pressed, even if controllers are configured for direct association to other loads. The event is always the same, so it doesn't tell you which button was pressed, but the scripts' logic can be changed to deal with that by refreshing loads.

AlCalzone:
This PR adds the new `node info received` event to signal to applications that a node has unexpectedly sent a NIF. This can be used to automate value refreshes on some older devices.

fixes: #6867

### DIFF
--- a/docs/api/node.md
+++ b/docs/api/node.md
@@ -1581,3 +1581,11 @@ interface RouteStatistics {
 	routeFailedBetween?: [number, number];
 }
 ```
+
+### `"node info received"`
+
+The node has sent a node information frame (NIF) Z-Wave JS did not expect. This can be caused by the user pushing a button on the device. Some older devices also send a NIF to notify the controller that their status has changed. The callback only references the node itself:
+
+```ts
+(node: ZWaveNode) => void
+```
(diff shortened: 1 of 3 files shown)


AlCalzone:
@raman325 would be great to have a way in HA to react to this, probably an event entity or so. See https://github.com/zwave-js/node-zwave-js/issues/2397#issuecomment-2119320177 for a use case.

AlCalzone:
@zwave-js-bot automerge

PR #7255, opened 2024-10-08
fix: remove all tasks when hard-resetting or entering bootloader

AlCalzone:
This does not solve anything yet, but it will once everything was migrated to use the scheduler.


AlCalzone:
@zwave-js-bot automerge

PR #7270, opened 2024-10-10
feat: implement Multilevel Switch mocks, add default state for Binary Switch mocks

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #7271, opened 2024-10-10
feat: add Thermostat Setback mocks, fix state encoding, remove CC values

AlCalzone:
While implementing the mocks for Thermostat Setback CC, I noticed that the CC values don't really make sense. They cannot be used to set the setback type or state, they don't have useful states, and the setback state does not expose the special values. So they are now removed.


AlCalzone:
@zwave-js-bot automerge

PR #7272, opened 2024-10-11
feat: implement playing tones in Sound Switch CC mocks, parse more switch-type dumps

AlCalzone:
With this PR, "playing" sounds on mocked sirens is now possible. Also extends the parsing of dump files for Binary/Multilevel Switch capabilities and Sound Switch capabilities.


AlCalzone:
@zwave-js-bot automerge

PR #7287, opened 2024-10-17
feat(config): add new parameters 17 and 18 for HeatIt TF016_TF021 FW 1.92

nesheimt:
Added support for new parameter(17 and 18) from v1.92

New datasheet attached

[HeatIt_1v9.pdf](https://github.com/user-attachments/files/17413128/HeatIt_1v9.pdf)

### DIFF
--- a/packages/config/config/devices/0x019b/tf016_tf021.json
+++ b/packages/config/config/devices/0x019b/tf016_tf021.json
@@ -232,7 +232,13 @@
 			"unit": "10 seconds",
 			"minValue": 0,
 			"maxValue": 8640,
-			"defaultValue": 6
+			"defaultValue": 6,
+			"options": [
+				{
+					"label": "Disable",
+					"value": 0
+				}
+			]
 		},
 		{
 			"#": "18",
--- a/packages/config/config/devices/0x019b/tf016_tf021.json
+++ b/packages/config/config/devices/0x019b/tf016_tf021.json
@@ -223,6 +223,27 @@
 			"minValue": 50,
 			"maxValue": 400,
 			"defaultValue": 210
+		},
+		{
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x019b/tf016_tf021.json
```diff
@@ -223,6 +223,26 @@
 			"minValue": 50,
 			"maxValue": 400,
 			"defaultValue": 210
+		},
+		{
+			"#": "17",
+			"$if": "firmwareVersion >= 1.92",
```
AlCalzone, line 229: Is this really firmware 1.92 and not 1.9? The manual you linked says "1.9"

AlCalzone (changes requested):

packages/config/config/devices/0x019b/tf016_tf021.json
nesheimt: Hi. v1.92 is the version that reads out from the device(the manufacturer also refers to v1.92 when in contact with their service desk), snip from wave js us below:
image
But there are no v1.91 so 1.9 would also work in the "config code".

AlCalzone: Ok then 1.92 is fine. Seems like another case of marketing versioning.

nesheimt: Good! Should I press "Resolve conversation"(or something else) for this change to proceed?

AlCalzone: No need. I'll finalize the PR once I'm back on the computer

AlCalzone (approved):

PR #7296, opened 2024-10-19
feat(config): add ZVIDAR WM25L Smartwings Smart Motor

xuexiaofei:
feat(config): add ZVIDAR WM25L Smartwings Smart Motor


xuexiaofei (commented):
fix(config): update Update the configuration file format error

PR #7306, opened 2024-10-22
refactor: decouple Serial API Message parsing and creation

AlCalzone:
This PR follows up on https://github.com/zwave-js/node-zwave-js/pull/7288, but this time for the Serial API message implementations.

The `Message` constructor signature was changed so it exclusively deals with constructing a message instance. Parsing binary payloads now happens in the static `from()` methods.

---

### List of breaking API changes:
- The `[MessageName]Options` interfaces no longer extend another interface and now exclusively contain the message-specific properties required for constructing that particular instance
- Message-specific constructors now accept a single options object of type `[MessageName]Options & MessageBaseOptions`.
- The `MessageDeserializationOptions` interface and the `gotDeserializationOptions` method no longer exist
- The `DeserializingMessageConstructor` interface no longer exists
- The `MessageCreationOptions` type has been renamed to `MessageOptions`
- Instead, message instances parse their specific payload using the new static `from` method, which takes a pre-parsed `MessageRaw` (message type, function type, binary payload) and the `MessageParsingContext`. All Message implementations with a custom constructor are expected to implement the `from` method aswell.
To parse an arbitrary message buffer use `Message.parse(buffer, context)`.
- Several Message implementations had their constructor options type reworked slightly for correctness. In some cases it is now necessary to pass properties that were previously inferred.
- The static `Message` methods `extractPayload`, `getConstructor`, `getMessageLength` and `isComplete` were removed.
- The `Message.options` property was removed
- The `TXReport` object no longer has the `numRepeaters` property, as this was implied by the length of the `repeaterNodeIds` array.
- The `ICommandClassContainer` interface was replaced with the new `ContainsCC` interface, which indicates that something contains a deserialized CC instance.
- The `Driver.computeNetCCPayloadSize` method now requires that the passed message instance contains a deserialized CC instance.
- For messages that contain a serialized CC instance, `Message.parse` no longer deserializes it automatically. This has to be done in a separate step.


AlCalzone:
@zwave-js-bot automerge

PR #7314, opened 2024-10-24
feat(mock-server): support add and remove mode

AlCalzone:
With this PR, the controller simulated by the mock-server can be put into inclusion and exclusion mode. No nodes will be found, but at least there won't be an error when trying to start those modes.


AlCalzone:
@zwave-js-bot automerge

PR #7317, opened 2024-10-24
revert: add ZVIDAR ZW882 Z-Wave Gateway

AlCalzone:
Seems the fingerprint was the same as for the ZW881

Reverts zwave-js/node-zwave-js#7310

@xuexiaofei do you know which product ID is the correct one?


AlCalzone:
@zwave-js-bot automerge

xuexiaofei:
> Seems the fingerprint was the same as for the ZW881
> 
> Reverts #7310
> 
> @xuexiaofei do you know which product ID is the correct one?

The  ZW881 is correct.

AlCalzone:
Which one should be used for the 882 then?

PR #7322, opened 2024-10-25
fix: correct unit of Meter CC values

AlCalzone:
found in https://github.com/zwave-js/zwave-js-ui/issues/3956#issuecomment-2436977873


AlCalzone:
@zwave-js-bot automerge

PR #7326, opened 2024-10-28
refactor: replace `wait` with `setTimeout` from `node:timers/promises`

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

AlCalzone:
Actually... it's better not to depend on the Node.js implementation everywhere

PR #7327, opened 2024-10-28
fix: detect bootloader in more difficult cases

AlCalzone:
This is a backport from #7325


AlCalzone:
@zwave-js-bot automerge

PR #5317, opened 2023-01-10
feat(config): add parameters for new Zooz ZEN72/74/77 firmwares, fix ZEN05 param 1

alvinchen1:
Revised https://github.com/zwave-js/node-zwave-js/pull/5156.

~~Partially addresses issue https://github.com/zwave-js/node-zwave-js/issues/4786~~
Added parameter 27, 28, 29 to zen77.
Added parameter 30 to zen72, zen74 and zen77 hardware v1.
Fix zooz zen05 param 1 https://github.com/zwave-js/node-zwave-js/issues/5248

Still no resolution for hardware versions where firmware drops below version 10.x (i.e. version 2.0, 3.0) differentiation see https://github.com/zwave-js/node-zwave-js/issues/5248 and https://github.com/zwave-js/node-zwave-js/pull/5244

But still provides updated configurations for zooz devices where firmware versions don't decrease vs. increase.

fixes: #4786 
fixes: #5248

PR #7328, opened 2024-10-28
fix(config): update Zooz ZEN74 to firmware 2.10

dhgoldberg:
Zooz ZEN74 hardware version 2.0 has been released since #5317 

ref: https://www.support.getzooz.com/kb/article/1051-zen74-toggle-dimmer-change-log/


AlCalzone (approved):
Thanks!

PR #7332, opened 2024-10-29
refactor: replace Node.js Buffer with Uint8Array or portable replacement

AlCalzone:
Since the beginning of Z-Wave JS, we've been using Node.js's `Buffer` class to manipulate binary data. This works fine, but is not portable, and prevents us from exploring compatibility with other runtimes, or even doing things in the browser, e.g. for flashing controllers or modifying NVM contents.

Following [Sindre Sorhus's example](https://sindresorhus.com/blog/goodbye-nodejs-buffer), this PR replaces the use of `Buffer`s with `Uint8Array`s where applicable.
In input positions where Z-Wave JS previously accepted `Buffer`s, this is backwards compatible, as `Buffer` is a subclass of `Uint8Array`.
In output positions however, this is a breaking change. Where applications are affected by this, check if `Buffer` methods like `readUInt32BE` etc. are actually needed, or if changing the expected type to `Uint8Array` would be enough. This is usually the case when just passing the binary data around, or accessing its content by index.

In some cases, Z-Wave JS now uses an almost drop-in replacement for Node.js's `Buffer`, the new `Bytes` class exported from `@zwave-js/shared`. This is a portable subclass of `Uint8Array` with some additions to make its API more (but not 100%) compatible with `Buffer`. It supports most of the `Buffer` functionality like `from()`, `concat()`, `toString()` (with limitations), `read/write[U]Int...LE/BE`.

In both cases, if your application really needs a `Buffer` instance, you can simply convert `Uint8Array` or `Bytes` to a `Buffer` using `Buffer.from(...)`.

### DIFF
--- a/docs/getting-started/migrating/v14.md
+++ b/docs/getting-started/migrating/v14.md
@@ -6,7 +6,24 @@ The `ZWaveHost` and `ZWaveApplicationHost` interfaces have been replaced by mult
 
 Furthermore, `Message` and `CommandClass` implementations are no longer bound to a specific host instance. Instead, their methods that need access to host functionality (like value DBs, home ID, device configuration, etc.) now receive a method-specific context object. Parsing of those instances no longer happens in the constructor, but in a separate `from` method.
 
-All in all, this release contains a huge list of breaking changes, but most of those should not affect any application, unless very low-level APIs are frequently used.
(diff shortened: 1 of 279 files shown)


AlCalzone:
@zwave-js-bot pack this

AlCalzone:
@zwave-js-bot automerge

PR #7336, opened 2024-10-30
refactor: rework `Security2Extension` to new parsing pattern

AlCalzone:
also use the decorator utilities


AlCalzone:
@zwave-js-bot automerge

PR #7344, opened 2024-10-30
fix: compile `zwave-js` version into source

AlCalzone:
This prevents us from having to look up the package.json at runtime for determining the version.


AlCalzone:
@zwave-js-bot automerge

PR #7345, opened 2024-10-30
refactor: remove `__dirname` from production code

AlCalzone:
This PR removes a few instances where we were relying on the CommonJS builtin `__dirname` to determine the location of the current source file on disk.

This changes two paths:
1. The default logfile location is now in the cwd of the current process, rather than the main entry point
2. The cache files of the flasher utility are now in a subdirectory of the cwd of the current process.
3. The cache files of the driver itself have been changed by #7344 already


AlCalzone:
@zwave-js-bot automerge

PR #7347, opened 2024-10-30
fix: do not format LogicParser.ts using `yarn fmt`

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #7360, opened 2024-11-04
refactor: migrate legacy decorators to ES decorators

AlCalzone:
Adding support for ESM resulted in some issues with the legacy decorators in transpiled TS code. Unfortunately, field decorators can no longer be used for reflection purposes with ES decorators, so they had to be moved to class decorators.


AlCalzone:
@zwave-js-bot automerge

PR #7365, opened 2024-11-05
refactor: only install config updates in external config directory

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #7366, opened 2024-11-05
fix: parse negative setback state consistently

### LINKED ISSUE #7364: Wrong decoding of negative setback values in `Climate Control Schedule CC`
### Discussed in https://github.com/zwave-js/node-zwave-js/discussions/7362

<sup>Originally posted by **@bueeler** November  5, 2024</sup>
### Checklist

### Describe the issue

I am using MT02650 TRV and use "Climate Control Schedule CC" command class to adjust the TRV set point based on input from room thermostat.
After sending negative setback values through function setOverride() from "Climate Control Schedule CC" command class, the value for "Override state" is not correctly shown in Z-Wave JS UI:
grafik

I would expect that negative setback values are correctly displayed as negative number in Z-Wave JS UI (as it is done for positive values). e.g. value "252" from screenshot above shall be displayed as "-0.4".

Issue seems caused by missing handling for negative numbers in function decodeSetbackState():
https://github.com/zwave-js/node-zwave-js/blob/master/packages/cc/src/lib/serializers.ts#L31
The specification defines the value as signed byte:
grafik
but the implementation treats it as "number".
All negative values where val>=128 (0x80) are handled as "special state", instead handled as negative setback state.

### Software versions

zwave-js-ui: 9.25.0.38e48a8
zwave-js: 13.10.1

It should be unrelated to controller

### Device information

Manufacturer: Devolo
Model name: ...
Node ID: e.g. 008, 005 ...

### Checklist

### Upload Logfile

[zwave-js-ui-store.zip](https://github.com/user-attachments/files/17629519/zwave-js-ui-store.zip)
</div>

AlCalzone:
fixes: #7364

### DIFF
--- a/packages/cc/src/cc/ClimateControlScheduleCC.ts
+++ b/packages/cc/src/cc/ClimateControlScheduleCC.ts
@@ -483,8 +483,9 @@ export class ClimateControlScheduleCCOverrideReport
 	): ClimateControlScheduleCCOverrideReport {
 		validatePayload(raw.payload.length >= 2);
 		const overrideType: ScheduleOverrideType = raw.payload[0] & 0b11;
-		const overrideState: SetbackState = decodeSetbackState(raw.payload[1])
-			|| raw.payload[1];
+		const overrideState: SetbackState = decodeSetbackState(raw.payload, 1)
+			// If we receive an unknown setback state, return the raw value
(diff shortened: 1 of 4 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #7368, opened 2024-11-05
fix: fallback to Europe when setting region to default

AlCalzone:
Setting the "default" region is not possible. Controllers are supposed to default to the EU region, so we now do just that.


AlCalzone:
@zwave-js-bot automerge

PR #7369, opened 2024-11-05
feat: allow specifying RF region for OTA firmware updates if unknown

### LINKED ISSUE #6891: Allow Manual Region Specification for automatic firmware updates on older controllers
I noticed that I was not being offered an update which should have been available (for a Zooz ZSE43). I believe the issue is that I do not have a region specified by my ZWave stick, so the update cannot match.

In cases like this, could it be possible to manually specify a region (for older sticks)?

Also, just to check that my assumption is correct, where would I see this region, if my stick did allow retrieving it?

As a sidenote, zwave-js/firmware-updates#82 would be useful here, as at first glance, it simply looked like my device was up to date, when in reality, the update check could not be properly executed.

### LINKED ISSUE #7323: Ota firmware not found
### Checklist

- [x] I have checked the troubleshooting section and my problem is not described there.
- [x] I have read the changelog and my problem is not mentioned there.

### Deploy method

Docker

### Z-Wave JS UI version

9.24.0.5531846

### ZwaveJS version

13.9.1

### Describe the bug

OTA updates shows no firmware available.

grafik

Device has 10.27, available is 10.30.
[Firmware is there](https://github.com/zwave-js/firmware-updates/blob/d6f781569543597c42cee014d218d7f238798388/firmwares/shelly/qnsw-002P16.json#L17)

### To Reproduce

Open OTA Updates, nothing there

### Expected behavior

Ota update is available

### Additional context

_No response_

AlCalzone:
This PR expands the options for `Controller.getAvailableFirmwareUpdates` and adds a new `rfRegion` property that will be used as a fallback in case the controller's RF region cannot be determined.

This allows using the OTA update service with older controllers that have no support for querying or changing the RF region. This feature should be used with caution, as specifying the wrong region may result in bricking the device.

fixes: #6891
fixes: https://github.com/zwave-js/node-zwave-js/issues/7323

### DIFF
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -8207,7 +8207,7 @@ export class ZWaveController
 					productType,
 					productId,
 					firmwareVersion,
-					rfRegion: this.rfRegion,
+					rfRegion: this.rfRegion ?? options?.rfRegion,
 				},
 				{
 					userAgent: this.driver.getUserAgentStringWithComponents(
@@ -8250,6 +8250,7 @@ export class ZWaveController
 	): Promise<void> {
 		if (
 			deviceId.rfRegion !== undefined
+			&& this.rfRegion !== NOT_KNOWN
 			&& deviceId.rfRegion !== this.rfRegion
 		) {
 			throw new ZWaveError(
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #7374, opened 2024-11-05
feat: add "browser" condition to main entrypoint for bundlers

AlCalzone:
With this PR, we begin adding a "browser" condition to the main entrypoint in "exports". This can be used by bundlers like esbuild to access a browser-safe subset of each package's functionality. Over time we will phase out the `/safe` entrypoints in favor of this.


AlCalzone:
@zwave-js-bot pack this

AlCalzone:
@zwave-js-bot automerge

PR #7393, opened 2024-11-12
fix(config): ignore setpoint range for Ecolink TBZ500

AlCalzone:
fixes: https://github.com/zwave-js/node-zwave-js/discussions/7384


AlCalzone:
@zwave-js-bot automerge

PR #7394, opened 2024-11-12
refactor: reduce dependency on node internals

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #7405, opened 2024-11-14
fix(config): correct config JSON syntax for Zooz ZSE70 800LR 

hceuterpe:
Updated the initial imported config file for the new ZooZ ZSE70 800LR outdoor sensor with meaningful descriptions, along with removing blank entries that failed linting.

Due to the new parameter types of this sensor, and slight variations compared to existing templates, I was only able to find a single meaningful template to include, and since it was just one, deliberately didn't include any templates.

### DIFF
--- a/packages/config/config/devices/0x027a/zse70_800lr.json
+++ b/packages/config/config/devices/0x027a/zse70_800lr.json
@@ -29,37 +29,32 @@
 		{
 			"#": "1",
 			"label": "Motion Sensitivity",
-			"description": "0",
+			"description": "Adjust motion sensitivity where 8 is the most sensitive setting.",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 8,
-			"defaultValue": 6,
-			"unsigned": true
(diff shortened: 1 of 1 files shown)


apella12:
The description should use uppercase and lowercase normally, except for the first character which must be lowercase.
`correct config JSON syntax for Zooz ZSE70 800LR`

hceuterpe:
> The description should use uppercase and lowercase normally, except for the first character which must be lowercase. `correct config JSON syntax for Zooz ZSE70 800LR`

Good grief.  Thanks for pointing that out. :D

hceuterpe:
> The description should use uppercase and lowercase normally, except for the first character which must be lowercase. `correct config JSON syntax for Zooz ZSE70 800LR`

Good grief. Thanks for pointing that out. :D

tofurky:
tested this on my new zse70, seems to work fine, thanks.

AlCalzone:
I found https://github.com/zwave-js/node-zwave-js/pull/7496 first, so the work will continue there.

PR #7428, opened 2024-11-21
refactor: add portable bindings for the filesystem

AlCalzone:
This PR adds portable FS bindings that will later be used to support different platforms than Node.js
As a side effect, the `serial` package is now in charge of listing serial ports.

Followup tasks:
- eliminate `fs-extra` from `jsonl-db`, replace with portable bindings
- replace `execa` with portable bindings where supported
- eliminate `fs-extra` from `proper-lockfile`, replace with portable bindings


AlCalzone:
@zwave-js-bot automerge

PR #7435, opened 2024-11-22
fix: argument validation of parameters with defaults, Map and Set

AlCalzone:
fixes the issue reported in https://github.com/zwave-js/node-zwave-js/discussions/7430#discussioncomment-11350284


AlCalzone:
@zwave-js-bot automerge

PR #7437, opened 2024-11-22
refactor: use DOM compatible EventTarget to replace Node.js's EventEmitter

AlCalzone:
...except for the `serial` package which needs to be converted to Web streams first.

This comes with the added benefit that attaching an `error` handler to the driver instance is no longer necessary.


AlCalzone:
@zwave-js-bot automerge

PR #7455, opened 2024-12-04
fix(config): add fingerprint to Yale YRL210

gawdygoda:
Added a product id for the original YRL210 Lever Door Lock, the generation BEFORE Azure series came out.

PR description here


gawdygoda:
FWIW, this is tested and proven to work. it adds additional Yale specific functionality like event notification when state changes happen directly on the lock, ability to change YRL specific parameters, etc.

See: https://community.home-assistant.io/t/zwave-js-ui-config-db-override/807450/5

AlCalzone (approved):

PR #7460, opened 2024-12-04
refactor: replace `xstate` with a simple built-in state machine

AlCalzone:
This PR introduces a custom state machine implementation, which is much less powerful than `xstate` but at the same time much less complex. It now replaces `xstate` where that was previously used (sending frames, node status, receiving transport service frames).


AlCalzone:
@zwave-js-bot automerge

PR #7473, opened 2024-12-09
feat: add missing parameters for the MCO MH-C221 shutter

dlm31:
Added all parameters for MH-C221 shutter device


dlm31:
Tested with own devices, and worked.

Issue #5157, opened 2022-10-12
Zooz Outdoor Motion Sensor ZSE29 - SupervisionCCReport fails - cannot change device configuration values


robertvorthman:
### Is your problem within Home Assistant (Core or Z-Wave JS Integration)?

NO, my problem is NOT within Home Assistant or the ZWave JS integration

### Is your problem within Z-Wave JS UI (formerly ZwaveJS2MQTT)?

NO, my problem is NOT within Z-Wave JS UI

### Checklist

- [X] I have checked the [troubleshooting section](https://zwave-js.github.io/node-zwave-js/#/troubleshooting/index) and my problem is **not** described there.

- [X] I have read the [changelog](https://github.com/zwave-js/node-zwave-js/blob/master/CHANGELOG.md) and my problem was **not** mentioned there.

### Describe the bug

The Zooz Outdoor Motion sensor, ZSE29, does not appear to support supervision command class.  Attempting to update any value, such as setting motion sensitivity from 10 to 9, fails and the value returns to the previous value.  I see an error about SupervisionCCReport in the logs when attempting to update values.

```
2022-10-12T01:17:32.939Z DRIVER « [Node 091] [REQ] [BridgeApplicationCommand]
│ RSSI: -71 dBm
└─[Security2CCMessageEncapsulation]
│ sequence number: 188
└─[SupervisionCCReport]
session id: 4
more updates follow: false
status: Fail
duration: 0s
2022-10-12T01:17:32.951Z CNTRLR [Node 091] [setValue] result of SET_VALUE API call for ConfigurationCCAPI: (Su
pervisionResult)
status: Fail
2022-10-12T01:17:32.952Z CNTRLR [Node 091] [setValue] not updating value
```

This issue seems similar to issue #5139 and pull #5008.

### Device information

Manufacturer: Zooz
Model name: ZSE29
Node ID in your network: 92

### How are you using `node-zwave-js`?

- [ ] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker image (latest)
- [ ] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker image (dev)
- [ ] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker manually built _(please specify branches)_
- [ ] `ioBroker.zwave2` adapter _(please specify version)_
- [X] `HomeAssistant zwave_js` integration  _(please specify version)_
- [ ] `pkg`
[...]

PR #7475, opened 2024-12-09
fix(config): disable Supervision for Everspring SP817 Motion Sensor

angainor:
disable Supervision for Everspring SP817 Motion Sensor: setting configuration parameters always fails.

similar to issue discussed in https://github.com/zwave-js/node-zwave-js/issues/5157


angainor:
@AlCalzone Do you think this will be merged? I got my sensor working after I installed this custom device file.

AlCalzone (approved):

AlCalzone:
Thanks!

PR #7478, opened 2024-12-10
feat: add missing parameters for the MCO MH-C221 shutter

dlm31:
PR description here


AlCalzone:
the CI failure should be fixable by running `yarn fmt` and committing the changed file.

PR #7479, opened 2024-12-10
refactor: replace `got` with `ky`

AlCalzone:
`got` is pretty heavy for what it does and it is not portable. Since Node.js now natively supports `fetch`, we can swap to a light wrapper instead.


AlCalzone:
@zwave-js-bot automerge

PR #7483, opened 2024-12-12
fix: avoid cross-link issues in devcontainer and gitpod

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #7486, opened 2024-12-12
refactor: allow switching out the DB bindings

AlCalzone:
This PR adds an abstraction for the underlying storage of the network and value caches. This allows using other storage backends than `jsonl-db`, for example a light wrapper around `window.localStorage`.


AlCalzone:
@zwave-js-bot automerge

PR #7510, opened 2024-12-22
feat(config): add Aeotec ZWA046 Home Energy Meter 8

jc2a:
I recently installed the US 2P version of the Aeotec Home Energy Meter 8. Without a config file the parameter table is unwieldy. I reviewed the spec sheets for the EU 1P and 3P meters as well in an attempt to generate a config for this meter.

Product manuals attached to Z-Wave Alliance pages have parameter details for each of the models.

- US 2-Phase, https://products.z-wavealliance.org/products/5070?selectedFrequencyId=2
- EU 3-Phase, https://products.z-wavealliance.org/products/5113?selectedFrequencyId=1
- EU 1-Phase, https://products.z-wavealliance.org/products/5069?selectedFrequencyId=1

Compared to previous versions, this one separates parameters for consumption and production reports which the spec sheet calls import and export. Most notably, parameters 101-103 are similar to ZW053 and represent import reports, but 104-106 are added for export reports.

The report masks in parameters 101-106 differ between the US 2P model and the EU 3P and 1P models. I have only been able to test with the 2P model.

Attached below is a screenshot of the default parameter table of the US 2P model with this config loaded.

Screenshot 2024-12-21 at 21-15-04 Z-Wave JS UI


conflipper:
Would be awesome if this could get added into the Config DB. I ended up using the custom config file that jc2a created until this is done; and that seems to be working no issues.

conflipper:
@AlCalzone have you had a chance to review this?

AlCalzone (dismissed):
Sorry for the delay.

I didn't particularly like the unnecessary params for unused bits. Looking at the engineering sheets, it looked like the values were filled from MSB to LSB, where the whole HEM was controlled by the least significant used byte. Modeling that with partial parameters is awkward, so I ended up splitting the file into 3.

AlCalzone (approved):

PR #7523, opened 2024-12-28
feat(config): update to TKB Home TZ88

Cobis-Knight:
update to TKB Home TZ88. Added missing configuration parameters and corrected a typo on parameter 3.


AlCalzone (approved):
Thanks! The file was due for aligning with our style guide - I've done that as part of this PR.

PR #7542, opened 2025-01-11
feat(config): update Zooz ZEN17 800LR to firmware 2.0

hceuterpe:
Update with param 28 newly added with the 800LR variants.  Clarified Z-wave Alliance info and manual links for 800LR version.


hceuterpe:
@AlCalzone I preemptively committed another update with review requests similar to my other pull requests.  BTW, looks like this updated pull is failing the same way with the node.js 18 erroneous error.

AlCalzone (approved):

PR #7545, opened 2025-01-12
feat(config): support Intermatic PE653 v3.1 with water temperature reading

function1st:
# Pull Request

## Title

feat(config): support Intermatic PE653 v3.1 with water temperature reading

## Description

This pull request adds support for the Intermatic PE653 device running firmware v3.1, tested using a Raspberry Pi 5 and a RaZberry 7 Z-Wave serial device. The patch updates both the config file and the command class code so that the water temperature reading is parsed correctly. In current testing, the UI label may display this measurement as "Air Temperature," but it is effectively the pool water temperature for the PE653.

## Changes

1. Updated manufacturer-proprietary command class to parse and store the water temperature from the PE653 v3.1 firmware payload.  
   - @IntermaticPE653CC.ts  
2. Modified config file for manufacturer 0x0072 to define the “waterTemperature” property as a readable number.  
   - @pe653.json  

## Testing

• Tested the modified config and command class on a device running firmware v3.1.  
• Confirmed that the water temperature is read and displayed (labeled as “Air Temperature” in the UI, but functionally correct).  
• Verified no regression occurs with legacy firmware or other devices.

### DIFF
--- a/packages/config/config/devices/0x0072/pe653.json
+++ b/packages/config/config/devices/0x0072/pe653.json
@@ -1,17 +1,38 @@
 {
-	"manufacturer": "Intermatic",
-	"manufacturerId": "0x0005",
-	"label": "PE653",
-	"description": "Pool Control",
+	"manufacturer": "Intermatic 0x0072 WaterTemp5",
+	"manufacturerId": "0x0072",
+	"label": "PE653 0x0072 WaterTemp5",
+	"description": "Pool Control 0x0072 WaterTemp5",
 	"devices": [
 		{
-			"productType": "0x5045",
-			"productId": "0x0653"
+			"productType": "0x0500",
+			"productId": "0x7205"
 		}
 	],
 	"firmwareVersion": {
 		"min": "0.0",
 		"max": "255.255"
+	},
+	"proprietary": {
+		"manufacturerProprietaryConfig": {
+			"manufacturerId": 114,
+			"commandClasses": {
+				"waterTemperature": {
+					"type": "number",
+					"unit": "°F",
+					"min": 0,
+					"max": 100,
+					"readOnly": true,
+					"valueChangeOptions": ["always"],
+					"stateless": false,
+					"isFromConfig": true,
+					"ccSpecific": {
+						"propertyName": "waterTemperature",
+						"propertyKey": "waterTemperature"
+					}
+				}
+			}
+		}
 	},
 	"paramInformation": [
 		{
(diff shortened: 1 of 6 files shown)


packages/transformers/src/validateArgs/transformDecorators.ts
AlCalzone, line 1: Can you explain what this file has to do with your changes?

packages/intermatic-pe653/src/IntermaticPE653CC.ts
AlCalzone, line 1: These implementations cannot be here. If they are needed, they must be in the `packages/cc/src/cc/manufacturerProprietary` folder where the other one is located too. You can call the one using the legacy manufacturer ID `IntermaticPE653LegacyCC` or so to avoid naming collisions.
However it does look like this code isn't actually used? As is, it wouldn't build.

packages/config/config/devices/0x0072/pe653.json
AlCalzone, line 1: I think this file should have been copied, not moved. After all there are still users with the old version.

packages/config/config/devices/0x0072/pe653.json
```diff
@@ -1,18 +1,39 @@
 {
-	"manufacturer": "Intermatic",
-	"manufacturerId": "0x0005",
-	"label": "PE653",
-	"description": "Pool Control",
+	"manufacturer": "Intermatic 0x0072 WaterTemp",
+	"manufacturerId": "0x0072",
+	"label": "PE653 0x0072 WaterTemp",
+	"description": "Pool Control 0x0072 WaterTemp",
```
AlCalzone, line 5: Just edit the manufacturer ID, keep the rest as-is. Unless the device is marketed under a new name?

packages/config/config/devices/0x0072/pe653.json
```diff
@@ -1,18 +1,39 @@
...
+					"stateless": false,
+					"isFromConfig": true,
+					"ccSpecific": {
+						"propertyName": "waterTemperature",
+						"propertyKey": "waterTemperature"
+					}
+				}
+			}
+		}
+	},
```
AlCalzone, line 36: This seems unnecessary

packages/config/config/devices/0x0072/pe653.json
```diff
@@ -481,72 +502,64 @@
...
-			// changes but the value is always 0 so treat it as an event.
 			"mapBasicSet": "event"
 		},
 		{
 			"commandClasses": {
-				// Force use of Multi Channel CC V1 despite the device reporting V2
 				"add": {
 					"Multi Channel": {
 						"isSupported": true,
 						"version": 1
```
AlCalzone, line 538: Keep the comments. They explain why each compat flag exists.

packages/config/config/devices/0x0072/pe653.json
```diff
@@ -481,72 +502,64 @@
...
+					},
+					"Manufacturer Proprietary": {
+						"isSupported": true,
+						"version": 1,
+						"optional": false
 					}
```
AlCalzone, line 544: Can you share an interview log? I'd expect this to be advertised as supported.

packages/config/config/devices/0x0072/pe653.json
```diff
@@ -481,72 +502,64 @@
...
-				// The response to the setpoint query is off by one bit: https://github.com/zwave-js/node-zwave-js/issues/5335
-				"Thermostat Setpoint": [
-					{
-						"method": "getSupportedSetpointTypes",
-						"result": [
-							1, // Heating
-							7 // Furnace
-						]
-					}
-				]
```
AlCalzone, line 635: Is this issue fixed in this device version?

packages/config/config/devices/0x0072/pe653.json
```diff
@@ -481,72 +502,64 @@
...
+					"version": 1
+				},
+				"0x91": {
+					"isSupported": true,
+					"version": 1,
+					"optional": false
+				}
+			}
+		}
+	}
```
AlCalzone, line 563: Likewise, I'd like to see the interview log to know if this is necessary.

AlCalzone (changes requested):
Thanks for this PR. I have a few preliminary remarks - didn't look at everything in depth, as this is still a bit all over the place (understandably, since implementing a proprietary CC isn't documented).

AlCalzone:
Additionally, do you have some documentation for that proprietary protocol? I wonder what else is in there.

PR #7561, opened 2025-01-17
fix: preserve all endpoints for Fibaro FGFS101, FW 26.26

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #7575, opened 2025-01-25
feat(config): add params for Enbrighten 58436

ncdc:
Add params for the Enbrighten 58436 In-Wall Smart Switch. These are the same as the 58433, so I copied from there. The only real changes I made were to the inclusion/exclusion instructions, as 58436 is a toggle and not a dimmer switch.


ncdc:
Nevermind 😄. I created this because one of my 58436s wasn't showing the parameters in HomeAssistant, but after re-interviewing it, they showed up. Sorry for the noise!

PR #7586, opened 2025-01-31
feat: it runs in the browser!!!

AlCalzone:
With this PR, the browser bundle finally succeeds and we can continue to playing whack-a-mole with runtime errors.


AlCalzone:
...and the runtime errors are gone

AlCalzone:
grafik

PR #7588, opened 2025-02-01
fix: update ZSE44 based on latest docs

kristof-mattei:
Based on: https://www.support.getzooz.com/kb/article/853-zse44-temperature-humidity-xs-sensor-advanced-settings/


AlCalzone (approved):
Thanks! From what I understood from the documentation, the definition for param 1 was not quite correct (removed in 1.20, added back in 1.30 with different range, restored to 1.20 version in 2.0). I've updated that.

PR #7600, opened 2025-02-06
feat(config): update N4002 to correct parameters and other information

The00Dustin:
This corrects the N4002 parameters, the changes match those listed on the updated manual which can be found at https://newonesmarthome.com/pages/download (the manual previously linked at zwavealliance.org has parameters for a switch or something, it isn't right).  These changes were tested against an N4012 which, detects as an N4002 prior to this config update.  According to the manufacturer's customer support department, "Our technical department reported that only the appearance and technology of N4002 and N4012 have been upgraded. The actual product ID and function have not changed."  While I don't have the older device to verify the product IDs (or firmware versions) to verify (or differentiate), my testing of the N4012 and the parameters in the N4012 manual that customer service provided support this claim (when compared to the N4002 manual in the link above).  Here is the N4012 manual for reference:
[N4012.pdf](https://github.com/user-attachments/files/18695366/N4012.pdf)

The following changes may or may not meet project standards and should be reviewed:

- "label" ("Product code" in ZUI) - I added " / N4012" since the device could be either of these.  I am not sure if that is the best way to prevent future issues from confused users ("I have an N4002 but it says N4012" or vice versa).
- "manual" (the manual download link in ZUI) - I originally removed this since the manual it links to was wrong, but I added it back with the link above once I found that, because it allows the user to download either manual (and because I wasn't sure how to extract the actual download URL from the file service they are using)

The remaining changes to the parameters have been sanity tested and should be sound.  I will also note that a 13th parameter is detected during interviews on my N4012, but as it is not documented in either manual and may or may not exist on the N4002, so I thought it would be better to leave it out of the config file.


The00Dustin:
Attaching the N4002 manual currently available from the link above for reference because customer service provided a newer version missing part of the parameter 9 information and indicated they would be updating the manuals on the site.  They also previously indicated that updates are not being made to the N4002 manual anymore because it is out of production, so this is the best source for confirming that parameter remained consistent without that device in hand.
[N4002Manual.pdf](https://github.com/user-attachments/files/18705683/N4002Manual.pdf)

AlCalzone (approved):
Thanks! I made a minor modification for consistency.

PR #7608, opened 2025-02-11
feat: support network-level security / Security S2 v2

### LINKED ISSUE #6005: Add support for Network-Level security (S2 v2)
> Network Layer Security or NLS is a feature embedded within the Security 2 Command Class and
provides a secure alternative for performing certain lower level mesh network protocol operations
previously not addressed by S2 Security.

> A SIS or Primary Controller supporting NLS MUST accept protocol commands [...] from an NLS-enabled node: only when communication is using the highest common Security Class shared between the SIS or Primary Controller and the sending node

It is currently unclear how this should work for commands the host application normally doesn't see like `Assign Return Route`, `Set SUC ACK`, etc., as S2 keys are not known to the controller atm.

Will have to wait for a Security S2 v2 capable device to test this out.

AlCalzone:
fixes: #6005


AlCalzone:
TODO:
- [x] Handle Request Protocol CC Encryption command

AlCalzone:
As of the last PR, Z-Wave JS can include simulated devices with NLS and perform protocol-level encryption for the simulated controller.
The Z-Wave stack draft has a few issues though where it either deviates from the specification, or is plain buggy. We'll have to figure that out first before merging, since I modified the stack for testing.

PR #7621, opened 2025-02-18
fix: handle proxy inclusion when NIF and Initiate command are switched

AlCalzone:
During CTT tests, we first receive the Initiate command, and then the NIF. This change makes sure that we handle both cases.


AlCalzone:
@zwave-js-bot automerge

PR #7628, opened 2025-02-19
feat: end device CLI support

AlCalzone:
This PR adds support for communicating with the text-based CLI exposed by some SoC end device firmwares. To do so, the new `driver.cli` property can be used. Since the Z-Wave module we're communicating with can now be something else than the bootloader or a controller firmware, a breaking change is required:

- The `driver.isInBootloader()` method has been removed. Use the new `driver.mode` property instead, which allows distinguishing between controller firmware, end device CLI firmware and the bootloader.


AlCalzone:
@zwave-js-bot pack this

PR #7629, opened 2025-02-21
feat(config): add new Zooz ZEN32 parameter 27

marksieczkowski:
Firmware 2.40 introduced a new parameter - 27.  This parameter appears to be unique to the Zen32 so I did not create a new template for it

Reference:
- https://www.support.getzooz.com/kb/article/1158-zooz-ota-firmware-files/
- https://www.support.getzooz.com/kb/article/608-zen32-scene-controller-advanced-settings/


AlCalzone (approved):
Thanks, I've shortened the label and description a bit to match our style guide

PR #7641, opened 2025-02-27
feat(config): add Shelly Door/Window Sensor, Wave Plug S, Wave PRO Dimmer 1PM/2PM

QubinoHelp:
Wave Plug S, Wave Pro Dimmer 1PM, Wave Pro Dimmer 2PM, Wave Door/Window sensor, add new templates.

PR description here


packages/config/config/devices/0x0460/templates/wave_template.json
```diff
@@ -164,7 +164,7 @@
 			},
 			{
 				"label": "Factory reset",
-				"value": 1431655765
+				"value": 1
 			}
```
AlCalzone, line 192: This template is used in many, if not all of your device files. Was the value wrong the entire time or does the new value only apply to new devices? In that case we may need to introduce a separate template for the new files only.

packages/config/config/devices/0x0460/templates/wave_template.json
```diff
@@ -174,6 +174,81 @@
...
+		"options": [
+			{
+				"label": "inactive",
+				"value": 0
+			},
+			{
+				"label": "active",
+				"value": 1
+			}
+		]
```
AlCalzone: We have a "master" template for these kinds of parameters:
```
~/templates/master_template.json#base_enable_disable
```

Please use that instead.

AlCalzone (changes requested):

packages/config/config/devices/0x0460/templates/wave_template.json
QubinoHelp: It was wrong all the time, at first was intended to be 143.... but once we start to add the parameter we chose to implement 1 instead of the 143...

packages/config/config/devices/0x0460/templates/wave_template.json
QubinoHelp: ok noted I will change that

QubinoHelp: I did the requested changes.

AlCalzone (approved):

AlCalzone:
Thanks - I've made some minor changes for consistency.

PR #7661, opened 2025-03-06
fix: harden CLI detection

AlCalzone:
Our current logic didn't work in the case that the end device CLI had just started up. In this case it won't respond to a NAK with a NAK, but with the `> ` prompt string.
This lead to `detectMode` returning `SerialAPI` although `driver.cli` was defined.


AlCalzone:
@zwave-js-bot automerge

PR #7667, opened 2025-03-10
fix: treat EU_LR as EU for FW update service calls

AlCalzone:
fixes an issue where no firmware updates would show as available when the region is set to EU_LR.


AlCalzone:
@zwave-js-bot automerge

PR #7677, opened 2025-03-12
feat(config): add alarmType 132 mapping for Yale YRD4x0 locks

dlgoodr:
Adds the correct notification events for Yale Locks when alarmType 132 fires.


AlCalzone:
Which locks does this apply to? Right now this only adds a template that can be used by device configuration files.

dlgoodr:
Sorry, belongs to the yrd4x0 series.

AlCalzone (approved):
Thanks!

PR #7679, opened 2025-03-13
refactor: split Node class into more mixins, move CC handlers out

AlCalzone:
This PR continues the Node class refactoring I started before the v15 release. It moves some more functionality into specific "mixin" classes and puts each CC handler into its own module.


AlCalzone:
@zwave-js-bot automerge

PR #7682, opened 2025-03-14
feat: add API to query supported `notification` events of a node

AlCalzone:
similar to `getDefinedValueIDs`, the new `getSupportedNotificationEvents` method allows discovering which specific events will be emitted through the `notification` event.


AlCalzone:
@zwave-js-bot automerge

PR #7687, opened 2025-03-21
fix: ensure `Indicator` is defined in the esbuild bundle

AlCalzone:
for https://github.com/zwave-js/zwave-js-ui/issues/4162


AlCalzone:
@zwave-js-bot automerge

PR #7713, opened 2025-04-01
feat(config): add param 29 (load sense) to HomePro ZDP100

lamharry:
As per manual pulled from :
https://www.manualslib.com/manual/1323496/Homepro-Zdp100.html?page=3#manual


AlCalzone (approved):
Ok with my changes. No need for a separate template only to add a label.

PR #7721, opened 2025-04-05
fix: return payload from `sendAndReceiveData` method of Manufacturer Proprietary CC

philh30:
Addresses the issue below where the sendAndReceiveData method of Manufacturer Proprietary no longer returns the payload.

https://github.com/zwave-js/zwave-js-ui/issues/4181


AlCalzone (approved):
Thanks! The changes to constructor options were not entirely correct, so I fixed that. Also I added a regression test to make sure this keeps working.

PR #7722, opened 2025-04-07
fix: preserve `host` bindings while updating driver options

AlCalzone:
For: https://github.com/zwave-js/zwave-js-ui/issues/4164


AlCalzone:
@zwave-js-bot automerge

PR #7729, opened 2025-04-10
fix: respect remove endpoints compat flag in Multi Channel V1 interview

philh30:
Incorporates the Remove Endpoints compat flag into the Multi Channel V1 interview. Currently, the flag only impacts Multi Channel V1 devices when set to "*".


AlCalzone (approved):
And thanks again!

PR #7731, opened 2025-04-12
feat(config): update Inovelli VZW31-SN to FW 1.04

The00Dustin:
Current changes in this PR:

- Change labels of parameters 53 & 54 for better compatibility as HomeAssistant configuration entities.
- Add parameters 32 & 33 for firmware v1.02 and higher.
- Add parameters 23, 24, 120, and 130-134 for beta firmware v1.04.
- Correct typo in metadata.

I have already collaborated with Inovelli on these changes, but am tagging @InovelliUSA for additional review / approval in case it is necessary and because certain added parameters aren't currently in the documentation I could find for this device and weren't explicitly discussed.

### DIFF
--- a/packages/config/config/devices/0x031e/vzw31-sn.json
+++ b/packages/config/config/devices/0x031e/vzw31-sn.json
@@ -23,7 +23,7 @@
 					"isLifeline": true
 				},
 				"2": {
-					"label": "Basic Switch Set",
+					"label": "Basic Set",
 					"maxNodes": 10
 				},
 				"3": {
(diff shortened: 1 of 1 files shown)


The00Dustin:
Two things worth consideration:

Parameter 120 appears to be hardcoded to cycle through multilevel values 0, 25, 50, and 99 while parameters 131-133 have customizable multilevel set values that don't apply to parameter 120.  As such, I wonder if the value descriptions in parameter 120 should be adjusted to prevent potential confusion.

Based on "Ability to disable the relay when in OnOff + Full wave mode" in the [firmware changelog](https://help.inovelli.com/en/articles/8505159-red-series-2-1-switch-firmware-changelog), I wonder if there is an existing parameter that needs another labeled value added.

The00Dustin:
I have updated the names and descriptions for parameters 130-134.  I removed the description for parameter 130 because I wasn't sure whether something like "Control group 7 using config button" would add value as desired by zwave-js maintainers or add confusion (i.e. by implying that group 7 can otherwise be controlled some other way).  I believe it may also be valuable to add an `associations` section to the config file to rename group 7, but figure that should be a separate commit.

The00Dustin:
I'm not sure how, but part of my previous commit was missing, so the commit to add the associations also finished changes described in the prior commit.  Group 7 is only available in firmware v1.4+, so I tested this config by re-interviewing switches on v1.4 and older versions.  The switch with an older firmware did not show group 7 in ZUI, so I think the association config is good.

InovelliUSA:
This looks good to me. Thanks @The00Dustin for the help!

AlCalzone (approved):
Thanks a lot! I fixed some minor issues I noticed.

PR #7732, opened 2025-04-12
fix: wait for complete transmission, even when no ACK is requested

AlCalzone:
It turns out that the assumption "no ACK = no callback" is incorrect. By waiting for the callback, we ensure that we're not flooding the TX queue of the radio when transmitting a non-acknowledged frame takes longer than normal.

Might help with https://github.com/home-assistant/addons/issues/3982 if the failing transmissions are the ones directly following a transmitted Supervision Report (no ACK)


AlCalzone:
@zwave-js-bot automerge

PR #7746, opened 2025-04-23
feat: add options to omit optional data during NVM migration

AlCalzone:
When migrating between different controller hardware, at least deleting the known routes can make sense, especially if the hardware has vastly different RF characteristics.

This PR adds the possibility to do that, along with options to strip other optional information.


AlCalzone:
@zwave-js-bot automerge

PR #7748, opened 2025-04-24
fix: more resilient recovery from disconnected TCP serial ports

AlCalzone:
This should help with detecting and recovering disconnected TCP serial connections more reliably.


AlCalzone:
@zwave-js-bot automerge

PR #7749, opened 2025-04-24
fix: do not delete battery temperature unit if value is unknown

AlCalzone:
Found while investigating https://github.com/home-assistant/core/issues/143480


AlCalzone:
@zwave-js-bot automerge

PR #7755, opened 2025-04-25
fix: purge old node-specific values when node is found

### LINKED ISSUE #7692: Ensure node values are removed when removing a node
There was a report that including a node after wrapping the node IDs back to 1 could reuse the stale data from the database for the node that used to have this ID. This can be avoided by clearing all existing entries for a node ID when a new node with that ID gets included.

We should make sure though that removing a node (no matter how) actually clears the corresponding entries from the value DB. Right now this does not happen.

AlCalzone:
fixes: #7692


AlCalzone:
@zwave-js-bot automerge

PR #7762, opened 2025-04-26
fix: work around `import_serial.isAnySendDataResponse` is not a function

AlCalzone:
There's something broken with how the re-exports in `@zwave-js/serial`'s index file are compiled in downstream projects. Avoiding this particular re-export should work around the issue for now.


AlCalzone:
@zwave-js-bot automerge

Issue #7764, opened 2025-04-26
Investigate: SDK 7.23.1/2 lockup on the third communication attempt with unreachable device


AlCalzone:
To reproduce:
- Use controller with SDK 7.23.1/2
- Successfully communicate with an always listening device (here: GreenWave GWPN1)
- Cut power to the device
- Attempt communication (SendDataBridge) with the device, 3+ application level retries, transmit options ACK+AutoRoute+Explore
- On the 3rd attempt, SendDataBridge commands stop getting acknowledged
- Waiting 100ms before retrying avoids the issue

AlCalzone:
To clarify what *application level retry* means:

I execute a `SendDataBridge` command with the specified transmit options. Immediately after the callback comes back with status `NoAck`, I retry the command. On the 3rd attempt, the controller does not ACK the Serial API command anymore.

Soft-reset recovers the controller communication.

AlCalzone:
Not a stack issue. Fixed by: https://github.com/AlCalzone/nc-zwave-esp-bridge/pull/7

PR #7766, opened 2025-04-26
fix: avoid controller lockup on retransmit after `NoACK`

AlCalzone:
Works around #7764 by waiting a short time before retrying the command.


AlCalzone:
@zwave-js-bot automerge

PR #7771, opened 2025-04-29
refactor: hide node-specific runtime imports from bundlers

AlCalzone:
With this change, it is possible to import `zwave-js` from the frontend without requiring node-specific APIs to be available.


AlCalzone:
@zwave-js-bot automerge

PR #7772, opened 2025-04-29
refactor: mark modules with `sideEffect: false` to aid bundlers

AlCalzone:
...except for the one file where reflection decorators are defined. All others should be fine being marked as side-effect free.


AlCalzone:
@zwave-js-bot automerge

PR #7787, opened 2025-05-05
feat: re-create controller instance instead of destroying driver after certain actions

AlCalzone:
Historically, Z-Wave JS relied on the application to create a new driver instance after certain invasive actions like NVM restore and OTW firmware updates. Depending on how the application connects to the driver instance, this can be awkward to handle.

As of this PR, the driver is no longer destroyed as the result of these "routine" actions. Instead, all timers, queues, etc. are cleaned up internally and a new `Controller` instance is created. 

> [!IMPORTANT]  
> The `driver ready` event now always implies that a new controller instance was created. Applications MUST ensure that their controller event handlers are set up afterwards, as the old ones will no longer work.


AlCalzone:
@zwave-js-bot automerge

PR #7795, opened 2025-05-08
fix: option `s2VerifyDelivery` not working

### LINKED ISSUE #7793: S2 NonceReport shortly after SET-type command is not considered part of the transaction
### Discussed in https://github.com/zwave-js/zwave-js/discussions/7720

<sup>Originally posted by **hceuterpe** April  4, 2025</sup>
### Checklist

### Describe the issue

What is happening?

Ever since Z-wave JS UI 9.33.1 (last known good version is 9.33.0), I have noticed zwave js either missing commands, or dropping commands sent sequentially.  In this case light switch LED color syncing with light status via HomeAssistant automations.  I have tried version 10.1.4, and the issue persists.  Rolling back to version 9.33.0 and the issue goes away.

Commands not missed.

Simply turning a switch on/off, even slowly.  

Anything else we should know?

### Software versions

Driver (zwave-js): ...
15.0.4
Z-Wave JS UI: ...
10.1.4
Home Assistant Z-Wave Integration: ...

Home Assistant Z-Wave JS Addon: ...

ioBroker.zwave2 Adapter: ...

If you are using something non-standard, tell us here: ...

ZOOZ ZST39 Z-Wave Long Range USB Stick

### Device information

Numerous switches with LED, and state control automations in HA.

### Checklist

### Upload Logfile

[Missed_LED_ZWave_Debug.txt](https://github.com/user-attachments/files/19609897/Missed_LED_ZWave_Debug.txt)

Specific node ID that missed commands in logfile is 114.  This switch is also roughly 3 meters from the controller, with a single drywall inbetween.
</div>

AlCalzone:
fixes: https://github.com/zwave-js/zwave-js/issues/7793

Since the device in the issue had Supervision disabled, the S2 delivery verification should have handled the collision, but it didn't.
The refactorings done in v15 broke the `s2VerifyDelivery` because the "unexpected" S2 nonce was now discarded before dynamic handlers were executed. Simply moving the default Nonce handler fixes this.

### DIFF
--- a/packages/zwave-js/src/lib/driver/Driver.ts
+++ b/packages/zwave-js/src/lib/driver/Driver.ts
@@ -5527,9 +5527,7 @@ ${handlers.length} left`,
 			if (msg.command instanceof Security2CCNonceGet) {
 				return this.handleSecurity2NonceGet(node);
 			}
-			if (msg.command instanceof Security2CCNonceReport) {
-				return this.handleSecurity2NonceReport(node, msg.command);
-			}
+			// Nonce Report is handled further down, as we might have dynamic handlers for it
 			if (msg.command instanceof Security2CCCommandsSupportedGet) {
 				return this.handleSecurity2CommandsSupportedGet(
 					node,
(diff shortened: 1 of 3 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #7800, opened 2025-05-09
fix: use local time for logging to file

### LINKED ISSUE #4088: Make Log Files Log with Correct Time Zone Instead of UTC
Logging appears to be in UTC. This makes troubleshooting more complicated when inspecting logs as the issue being tracked (for example, lining up when a node goes Unavailable in HA which is in the correct time zone), doesn't align with zwave-js logs.

Have logging match up with the zwave-js instance's time zone. In my case, the Docker Compose environment variable is set correctly, but zwave-js's logging doesn't respect that environment variable.

None that I can think of.

AlCalzone:
fixes: #4088

### DIFF
--- a/packages/core/src/bindings/log/browser.ts
+++ b/packages/core/src/bindings/log/browser.ts
@@ -14,6 +14,7 @@ import {
 	type LogContext,
 	type LogFactory,
 	type ZWaveLogInfo,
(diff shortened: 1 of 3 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #7805, opened 2025-05-11
feat(config): add PE653 endpoints for VSP speeds and P5043ME pool/spa mode

philh30:
The Intermatic PE653 reports 5 endpoints during the interview, which correspond to its 5 relays. This pull request adds endpoints 16-19 and 39. This pull request relates to some of the discussion in #4588, but won't close that issue as the thermostat setpoint scale remains to be addressed.

Endpoints 16-19 control the 4 variable speed pump (VSP) speeds. Only one of these endpoints can be turned on at a time (switching one on changes the pump to that speed and turns all others off), and turning any of them off turns off the pump. I've tested the VSP speeds with this config file and they work properly, though polling is required to keep the reported states in sync with the device when switching between speeds (e.g. when switching from speed 3 to speed 4 by turning endpoint 19 on, the device doesn't report that endpoint 18 has been turned off unless polled).

Endpoint 39 is used to switch between pool and spa modes when the P5043ME expansion module is connected. Without the expansion module, circuit 4 is used to switch pool/spa mode. The P5043ME requires firmware 3.4+. I don't have the expansion module, so this hasn't been tested.

I'm not aware of a use for endpoints 6-15 or 20-38. There's no official documentation of any endpoints from the manufacturer, so everything that's known is from zniffing or trial and error.

### DIFF
--- a/packages/config/config/devices/0x0005/pe653.json
+++ b/packages/config/config/devices/0x0005/pe653.json
@@ -472,11 +472,47 @@
 			"label": "Variable Speed Pump Speed 4 Schedule 3"
 		}
 	],
-	"associations": {
+	"endpoints": {
+		"0": {
+			"label": "Root",
+			"associations": {
+				"1": {
+					"label": "Lifeline",
+					"maxNodes": 5,
+					"isLifeline": true
+				}
+			}
+		},
 		"1": {
-			"label": "Lifeline",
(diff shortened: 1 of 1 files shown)


AlCalzone (approved):
Wow, this file is almost turning sentient

PR #7816, opened 2025-05-13
fix: preserve status of awake nodes when interview starts

AlCalzone:
Fixes an issue where sleeping nodes that report being awake after a firmware update will cause their interview to start, but pause immediately after the ProtocolInfo stage is completed.


AlCalzone:
@zwave-js-bot automerge

PR #7400, opened 2024-11-13
feat(config): add HomeSys HomeMech-2001/2

aweddell:
Add new device config for HomeSys HomeMech-2001 and HomeMech-2002 dimmers

Issue #7050, opened 2024-07-28
Missing device configuration: homeSYS (GroupSYS) HomeMech-2001 Dimmer


austwhite:
### Checklist

- [X] It is not in the [configuration DB](https://devices.zwave-js.io/?search=3)

- [X] It was not [merged recently](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Apr+is%3Aclosed) or has a [pending PR](https://github.com/zwave-js/node-zwave-js/pulls?q=is%3Aopen+is%3Apr)

- [X] The device was interviewed completely and has no missing or `undefined` IDs

### Which device is missing?

homeSYS (GroupSYS) HomeMech-2001 Dimmer

### Manufacturer ID

816

### Product Type

514

### Product ID

53276

### Firmware Version

1.38

### Is the device listed on the Z-Wave Alliance website?

https://products.z-wavealliance.org/products/3331?selectedFrequencyId=1

### Do you have a manual?

[HomeMech-2002 Instruction.pdf](https://github.com/user-attachments/files/16402231/HomeMech-2002.Instruction.pdf)

I found a manual for the HomeMech-2002 which is the push button version of the HomeMech-2001.  I believe they are virtually identical, aside from the 2001 having the rotary dimmer and the 2002 having a push button.

I couldn't find the config for the homeSys HomeMech-2001 or 2002 dimmers in the Z-Wave JS listing.   I hope I have posted this correctly.   Not sure if these are rebranded, I couldn't find a listing on the z-wave alliance pages for this particular dimmer.

austwhite:
I was doing some research and found this device may also be listed as Shenzhen Sunricher Technology Limited,   Model: SR-ZV2835PAC for Australia/NZ Frequency or for EU frequency, the SR-ZV2835RAC

austwhite:
Updated Z-wave certification link with information provided by homeSYS
https://products.z-wavealliance.org/products/3331?selectedFrequencyId=1

AlCalzone:
@zwave-js-bot import config 3331 from zwa

philh30:
The fingerprint listed - `0x0330/0x0202/0xD01C` - was added in #7400. Recommend closing this issue.

PR #7817, opened 2025-05-13
feat(config): add association group for Vesternet VES-ZW-DIM-001

philh30:
This pull request will revert almost all of the changes from the automatic import by zwave-js-bot. The device being imported is already in the database, and the current config file appears to match the manual. The only thing I see missing is association group 1.

The import was done in response to #7050. However, that issue referenced a zwave alliance page that did not match the fingerprint in question - `0x0330/0x0202/0xD01C`. That issue can be closed as support for that fingerprint was added in #7400.


philh30:
The lint error is coming from zwave-js-bot's edits to `packages/config/src/LogicParser.ts` in the original pull request.

PR #7819, opened 2025-05-15
fix: retry on ACK timeout before soft-resetting, increase timeout

AlCalzone:
fixes: https://github.com/zwave-js/zwave-js/discussions/7782

It seems like the functionality to retry after an ACK timeout went missing somewhere between v14 and v15. This PR restores it and adds a test to ensure it stays in.


AlCalzone:
@zwave-js-bot automerge

PR #7821, opened 2025-05-15
feat: allow the application to disable support for specific CCs

AlCalzone:
Z-Wave JS normally uses all Command Classes it implements and responds to version queries for all of them.
However, some Command Classes come with certification requirements that might not be fulfilled by the application.

To allow for deploying products based on Z-Wave JS without having to add (mostly) UI support for all supported CCs, this PR adds a driver option `features.disableCommandClasses` where an application can specify which CCs it will not support.
Z-Wave JS will then act as if it does not implement those CCs.

Note that the list of CCs that can be disabled is only a subset of all CCs - encapsulation CCs and some mandatory CCs cannot be disabled.

### DIFF
--- a/docs/api/CCs/Version.md
+++ b/docs/api/CCs/Version.md
@@ -27,7 +27,7 @@ async getCCVersion(
 ### `reportCCVersion`
 
 ```ts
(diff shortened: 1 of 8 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #7839, opened 2025-05-19
fix: device ID check during OTA updates

AlCalzone:
whoops :)


AlCalzone:
@zwave-js-bot automerge

PR #7840, opened 2025-05-19
feat: support OTW updates via the firmware update service

AlCalzone:
With this PR, the `driver.firmwareUpdateOTW` method now also accepts a `FirmwareUpdateInfo` as returned by the `getAvailableFirmwareUpdates` method. In that case, downloading and verifying the firmware update is before attempting the update, just like with OTA updates.


AlCalzone:
@zwave-js-bot automerge

PR #7842, opened 2025-05-19
fix: do not consider default region to be Europe for FW updates

### LINKED ISSUE #7818: RFRegion `Default (EU)` should not be considered "europe" for the FW update service
The EU in the name is misleading. This is only EU if not set as a token, but it can be any region really.

AlCalzone:
fixes: #7818


AlCalzone:
@zwave-js-bot automerge

PR #7846, opened 2025-05-20
fix: work around missing protocol version file in NVM backed up from SDK `7.23.0` and `.1`

### LINKED ISSUE #7843: NVM migration: work around 7.23.0/1 NVM files missing the protocol file format info
NVMs backed up from controllers running SDK 7.23.0 and 7.23.1 are missing the protocol version file `0x50000`, which is used to infer the layout of some NVM objects. We need to detect this somehow and work around it.

AlCalzone:
fixes: #7843


AlCalzone:
@zwave-js-bot automerge

PR #7848, opened 2025-05-23
fix: force Scene Activation CC as supported on FGD-212

Matsuo3rd:
https://github.com/zwave-js/zwave-js/pull/1848 was not effectively superseded by #1849. Using compat flag properly advertises "Scene Activation v1" in node's supported CC. Successfully tested with FGD-212 both firmware 3.4 and 3.5.


apella12:
For an OH user with three fibaro devices #7983 I modified device jsons for a potential PR, but see that you have one of the devices covered. Should I leave this device out or do you want to combine them? Let me know

Separately, I noticed that 1) you have the config after metadata and it should be before. 2) Also from the docs version default is 1, so that shouldn't be needed.
[fgd212.json](https://github.com/user-attachments/files/21939515/fgd212.json)

Matsuo3rd:
@apella12: let’s make it simple and combine PRs into one. Thanks for your great work!

spudwebb:
I see the same issue with FGD-212.
Why does this PR has not been approved yet?

PR #7862, opened 2025-06-02
feat: apply auto powerlevels on every actual region change

AlCalzone:
Follow-up PR to #7853

It didn't really make sense that auto powerlevel settings were only applied during startup, but not manual region changes. With this PR, every actual region change (e.g. EU -> US_LR, but not US -> US_LR) will result in the powerlevel settings to be updated, if the corresponding driver option is set to `"auto"`.


AlCalzone:
@zwave-js-bot automerge

PR #7870, opened 2025-06-04
fix(config): remove unlock mapping for Schlage lock FE599

ibddot:
The unlock mapping creates unexpected behaviors with this lock since the physical lock auto re-locks after ~10 seconds but that action is not captured anywhere and so zwave thinks the lock is still unlocked when it actually is locked.

I created a discussion on this but haven't gotten any input there yet: https://github.com/zwave-js/zwave-js/discussions/7511

Also commented on original pull request and did not get any feedback there how the unlock mapping was helpful: https://github.com/zwave-js/zwave-js/pull/7122


AlCalzone (approved):
Agreed, the current behavior isn't helpful if it leads to a wrong status being reported to applications.

PR #7878, opened 2025-06-12
docs: explain when joining another network is not permitted

### LINKED ISSUE #7876: Docs update : Secondary Controller
As discussed.

Playing around with Secondary Controller mode in Node RED (Great Success BTW)
I pulled out an old controller I had, with Nodes on it.

And ran into the `Forbidden` issue.
I think it maybe worth, hinting in the docs,  the controller needs to be in a Reset state - before it can act as a slave controller, this is so the `Primary` flag gets wiped, and will go into network Learn Mode

AlCalzone:
fixes: #7876


AlCalzone:
@zwave-js-bot automerge

PR #7883, opened 2025-06-13
ci: classify logfiles using AI

AlCalzone:
Our logfile check has historically been hit or miss. Whenever a user renamed the file it would break.

With this PR, we now check the contents of the file and classify it with a simple AI model to better help the users upload the correct ones.


AlCalzone:
@zwave-js-bot automerge

PR #7889, opened 2025-06-17
feat: support parsing Zniffer frames from file

AlCalzone:
This is almost done, just one more edge case needs to be handled: The official Zniffer application writes received serial messages to disk as they are received, even if they are partial commands or partial data frames:
```
# outgoing command:
1f95cd4d13addd888103000000230500fe
# one incoming command, split across 2 serial messages
e9e2cd4d13addd88010100000023fe
3b0ace4d13addd8801020000000500fe
```


AlCalzone:
@zwave-js-bot pack this

PR #7900, opened 2025-06-24
fix: initialize LWR statistics for LR nodes

AlCalzone:
With this PR, we automatically populate the LWR statistics for LR nodes on startup/creation. This info never changes, so it makes sense to do this here.

Related to https://github.com/zwave-js/zwave-js-ui/issues/4160#issuecomment-2999123911


AlCalzone:
@zwave-js-bot automerge

PR #7918, opened 2025-06-26
fix: wait indefinitely during inclusion/exclusion, stop on task drop

### LINKED ISSUE #7914: Fix waiting for status reports during inclusion/exclusion
The `awaitStatusReport` calls currently use a timeout for the `waitForMessage` calls, which is not ideal. Also they do not get cleaned up/aborted when the tasks are dropped, or inclusion/exclusion is stopped.

AlCalzone:
With this PR, we again wait indefinitely for status reports during inclusion/exclusion/replace operations. When the tasks are dropped for some reason, we now stop waiting.

TODO:

- [x] Test
- [x] Drop inclusion/exclusion tasks when user wants to stop the process

fixes: https://github.com/zwave-js/zwave-js/issues/7914

### DIFF
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -2136,6 +2136,7 @@ export class ZWaveController
 		options: InclusionOptionsInternal,
 	): TaskBuilder<void> {
 		const self = this;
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #7935, opened 2025-06-30
fix: pass join network user callbacks to `expectSecurityBootstrapS2`

AlCalzone:
...so we actually show the user the DSK


AlCalzone:
@zwave-js-bot automerge

PR #7936, opened 2025-06-30
fix: omit LR nodes from rebuild routes progress

AlCalzone:
Reported in https://github.com/zwave-js/zwave-js/discussions/7907 - it does not make sense to report the status for LR nodes for 2 reasons:
1. LR nodes do not route
2. Theoretically we could end up emitting events with 4000 entries, most of which are irrelevant.


AlCalzone:
@zwave-js-bot automerge

PR #7951, opened 2025-07-01
fix: import brand name and correct param descriptions from ZWA

AlCalzone:
This PR fixes a couple of issues with our ZWA import script where:
- the brand name gets ignored
- the device description does not get used, but instead the device name (which is sometimes empty)
- config parameter descriptions are the literal "0" instead of the actual description


AlCalzone:
@zwave-js-bot automerge

PR #7957, opened 2025-07-02
fix(config): rename Antenna to Adapter for ZWA-2

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #7959, opened 2025-07-03
refactor: revise config file hashes, auto-apply some metadata on startup

AlCalzone:
With this PR, we no longer store actual hashes of the config file contents, but a versioned, compressed subset of the information that should generally be unchanged. On startup, we upgrade old versions of the hash to the new format if we determine them to be unchanged.

Labels and descriptions are now omitted from hashes, so renamed config parameters will no longer cause the "re-interview required" info. Instead, the existing metadata is updated with the strings from otherwise unchanged config files on every startup.


AlCalzone:
@zwave-js-bot automerge

PR #7962, opened 2025-07-05
feat(config): add Fantem FT117 range extender

jefft4:
Add Fantem FT117 range extender (rebadged Aeotec ZW117)
File pass lint checks in GitHub online.


AlCalzone (approved):
Thanks!

PR #7963, opened 2025-07-08
fix: query actual value immediately after successful supervised `Multilevel Switch Set` with value 255

### LINKED ISSUE #5124: After successful supervised MultilevelSwitchCCSet (255) query the actual value quicker
```
2022-09-26T14:38:44.203Z DRIVER « [Node 004] [REQ] [ApplicationCommand]
│ source: 1
│ destination: 0
└─[SupervisionCCReport]
session id: 19
more updates follow: false
status: Success
duration: 0s
```

We don't need to wait 5 seconds after such a report.

Starting point:
`packages/cc/src/cc/MultilevelSwitchCC.ts` --> `SET_VALUE_HOOKS`

PhilipTrauner:
fixes #5124

previously a poll was scheduled after the `transitionDuration`.
scheduling a poll for a later  is not neccessary, because `verifyChanges` is only run *after* the value has already changed.

### DIFF
--- a/packages/cc/src/cc/MultilevelSwitchCC.ts
+++ b/packages/cc/src/cc/MultilevelSwitchCC.ts
@@ -10,6 +10,7 @@ import {
 	type MessageRecord,
 	NOT_KNOWN,
 	type SupervisionResult,
+	SupervisionStatus,
 	ValueMetadata,
 	type WithAddress,
 	maybeUnknownToString,
@@ -376,7 +377,7 @@ export class MultilevelSwitchCCAPI extends CCAPI {
 	protected [SET_VALUE_HOOKS]: SetValueImplementationHooksFactory = (
 		{ property },
 		value,
-		_options,
+		options,
 	) => {
 		// Enable restoring the previous non-zero value
 		if (property === "restorePrevious") {
@@ -385,6 +386,7 @@ export class MultilevelSwitchCCAPI extends CCAPI {
 		}
 
 		if (property === "targetValue") {
+			const duration = Duration.from(options?.transitionDuration);
 			const currentValueValueId = MultilevelSwitchCCValues.currentValue
 				.endpoint(
 					this.endpoint.index,
@@ -459,15 +461,34 @@ export class MultilevelSwitchCCAPI extends CCAPI {
 					// If we don't know the actual value, we need to verify the change, regardless of the supervision result
 					return value === 255;
 				},
-				verifyChanges: async () => {
+				verifyChanges: async (result) => {
 					if (
-						this.isSinglecast()
 						// We generally don't want to poll for multicasts because of how much traffic it can cause
 						// However, when setting the value 255 (ON), we don't know the actual state
-						|| (this.isMulticast() && value === 255)
+						!(this.isSinglecast()
+							|| (this.isMulticast() && value === 255))
 					) {
-						// After successful supervised set the value has already changed, no need to wait
-						await this.pollValue!.call(this, currentValueValueId);
+						return;
+					}
+
+					switch (result?.status) {
+						case SupervisionStatus.Success:
+						case SupervisionStatus.Fail:
+							await this.pollValue!.call(
+								this,
+								currentValueValueId,
+							);
+							break;
+						default:
+							(this as this).schedulePoll(
+								currentValueValueId,
+								value === 255 ? undefined : value,
+								{
+									duration: result?.remainingDuration
+										?? duration,
+								},
+							);
+							break;
 					}
 				},
 			};
--- a/packages/cc/src/cc/MultilevelSwitchCC.ts
+++ b/packages/cc/src/cc/MultilevelSwitchCC.ts
@@ -376,7 +376,7 @@ export class MultilevelSwitchCCAPI extends CCAPI {
 	protected [SET_VALUE_HOOKS]: SetValueImplementationHooksFactory = (
 		{ property },
 		value,
-		options,
+		_options,
 	) => {
 		// Enable restoring the previous non-zero value
 		if (property === "restorePrevious") {
@@ -385,7 +385,6 @@ export class MultilevelSwitchCCAPI extends CCAPI {
 		}
 
 		if (property === "targetValue") {
-			const duration = Duration.from(options?.transitionDuration);
 			const currentValueValueId = MultilevelSwitchCCValues.currentValue
 				.endpoint(
 					this.endpoint.index,
@@ -460,19 +459,15 @@ export class MultilevelSwitchCCAPI extends CCAPI {
 					// If we don't know the actual value, we need to verify the change, regardless of the supervision result
 					return value === 255;
 				},
-				verifyChanges: () => {
+				verifyChanges: async () => {
 					if (
 						this.isSinglecast()
 						// We generally don't want to poll for multicasts because of how much traffic it can cause
 						// However, when setting the value 255 (ON), we don't know the actual state
 						|| (this.isMulticast() && value === 255)
 					) {
-						// We query currentValue instead of targetValue to make sure that unsolicited updates cancel the scheduled poll
-						(this as this).schedulePoll(
-							currentValueValueId,
-							value === 255 ? undefined : value,
-							{ duration },
-						);
+						// After successful supervised set the value has already changed, no need to wait
+						await this.pollValue!.call(this, currentValueValueId);
 					}
 				},
 			};
--- a/packages/cc/src/cc/MultilevelSwitchCC.ts
+++ b/packages/cc/src/cc/MultilevelSwitchCC.ts
@@ -474,12 +474,10 @@ export class MultilevelSwitchCCAPI extends CCAPI {
 					switch (result?.status) {
 						case SupervisionStatus.Success:
 						case SupervisionStatus.Fail:
-							await this.pollValue!.call(
-								this,
-								currentValueValueId,
-							);
+							await this.pollValue!(currentValueValueId);
 							break;
-						default:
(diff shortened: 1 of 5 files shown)


packages/zwave-js/src/lib/test/driver/setValueSupervision255Verify.test.ts
```diff
@@ -0,0 +1,69 @@
...
+	MultilevelSwitchCCValues,
+} from "@zwave-js/cc";
+import { CommandClasses, type MaybeUnknown } from "@zwave-js/core";
+import type { MockNodeBehavior } from "@zwave-js/testing";
+import { integrationTest } from "../integrationTestSuite.js";
+
+integrationTest(
+	"successful supervised setValue(255): expect verify GET",
+	{
+		nodeCapabilities: {
```
AlCalzone, line 24: This test does not work as you intend to. Our mocks currently don't automatically respond to supervised requests, so that behavior needs to be set up in the test itself.

You can take a look at `packages/zwave-js/src/lib/test/driver/setValueSuccessfulSupervisionNoGet.test.ts` for an example of how to respond with a success report.

AlCalzone: Also, if you add `debug: true,` before the `nodeCapabilities`, you can see what's actually happening. In this case:
```
2025-07-09 08:19:17.267 DRIVER » [Node 002] [REQ] [SendDataBridge]
                                 │ source node id:   1
                                 │ transmit options: 0x25
                                 │ callback id:      13
                                 └─[SupervisionCCGet]
                                   │ session id:      29
                                   │ request updates: true
                                   └─[MultilevelSwitchCCSet]
                                       target value: 255

...

2025-07-09 08:19:18.314 CNTRLR   [Node 002] Timed out while waiting for a response from the node (ZW0201)
```

packages/cc/src/cc/MultilevelSwitchCC.ts
```diff
@@ -460,19 +459,15 @@ export class MultilevelSwitchCCAPI extends CCAPI {
...
 						|| (this.isMulticast() && value === 255)
 					) {
-						// We query currentValue instead of targetValue to make sure that unsolicited updates cancel the scheduled poll
-						(this as this).schedulePoll(
-							currentValueValueId,
-							value === 255 ? undefined : value,
-							{ duration },
-						);
+						// After successful supervised set the value has already changed, no need to wait
+						await this.pollValue!.call(this, currentValueValueId);
```
AlCalzone: This is a bit too simple and breaks existing behavior for non-supervised commands. You can reproduce that by commenting out Supervision support in the test and adding
```
{ transitionDuration: "5s" }
```
as the third parameter to the `node.setValue` call.

In the logs you'll then see the Set command is sent with a transition duration of 5s, so we expect the change to be done only after 5s. However the verification is now done immediately:
```
2025-07-09 08:26:02.086 DRIVER » [Node 002] [REQ] [SendDataBridge]
                                 │ source node id:   1
                                 │ transmit options: 0x25
                                 │ callback id:      13
                                 └─[MultilevelSwitchCCSet]
                                     target value: 255
                                     duration:     5s

...

2025-07-09 08:26:02.111 DRIVER » [Node 002] [REQ] [SendDataBridge]
                                 │ source node id:   1
                                 │ transmit options: 0x25
                                 │ callback id:      14
                                 └─[MultilevelSwitchCCGet]
```

You need to check first if the command was supervised and successful (there is a helper method for that), and only then poll immediately.

AlCalzone: Maybe it helps to take a look at the first 3 flows in https://github.com/zwave-js/zwave-js/discussions/7801, just until "encrypted commands" to understand how this fits together.

AlCalzone (changes requested):

packages/cc/src/cc/MultilevelSwitchCC.ts
PhilipTrauner: > This is a bit too simple and breaks existing behavior for non-supervised commands. You can reproduce that by commenting out Supervision support in the test and adding
> 
> ```
> { transitionDuration: "5s" }
> ```
> 
> as the third parameter to the `node.setValue` call.
> 
> In the logs you'll then see the Set command is sent with a transition duration of 5s, so we expect the change to be done only after 5s. However the verification is now done immediately:
 
so `transitionDuration` should be taken into account in `verifyChanges`, but the [current fallback](https://github.com/zwave-js/zwave-js/blob/570fa667e41853fc594a7ded53972ae354649b99/packages/zwave-js/src/lib/node/Node.ts#L518) should not be applied anymore, right? that's where the [5 seconds](https://github.com/zwave-js/zwave-js/blob/7f04bb602bded4ca0d8edb4ae60e319f438e99a8/packages/zwave-js/src/lib/driver/Driver.ts#L320) came from after all.

> ```
> 2025-07-09 08:26:02.086 DRIVER » [Node 002] [REQ] [SendDataBridge]
>                                  │ source node id:   1
>                                  │ transmit options: 0x25
>                                  │ callback id:      13
>                                  └─[MultilevelSwitchCCSet]
>                                      target value: 255
>                                      duration:     5s
> 
> ...
> 
> 2025-07-09 08:26:02.111 DRIVER » [Node 002] [REQ] [SendDataBridge]
>                                  │ source node id:   1
>                                  │ transmit options: 0x25
>                                  │ callback id:      14
>                                  └─[MultilevelSwitchCCGet]
> ```
> 
> You need to check first if the command was supervised and successful (there is a helper method for that), and only then poll immediately.

in order to use [that utility function](https://github.com/zwave-js/zwave-js/blob/570fa667e41853fc594a7ded53972ae354649b99/packages/core/src/definitions/Supervision.ts#L34) in [`verifyChanges`](https://github.com/zwave-js/zwave-js/blob/570fa667e41853fc594a7ded53972ae354649b99/packages/cc/src/cc/MultilevelSwitchCC.ts#L463) i'd have to pass along the [result obtained from the `setValue` call](https://github.com/zwave-js/zwave-js/blob/570fa667e41853fc594a7ded53972ae354649b99/packages/zwave-js/src/lib/node/Node.ts#L555), right?
if that would indeed be necessary, what's your take on deferring [the call to `verifyChanges`](https://github.com/zwave-js/zwave-js/blob/570fa667e41853fc594a7ded53972ae354649b99/packages/zwave-js/src/lib/node/Node.ts#L660) in `ZWaveNode.setValue`, instead of handling this in individual command classes?

AlCalzone: We need to keep the logic for non-supervised commands, so the current fallback should also stay in place.

> what's your take on deferring [the call to verifyChanges](https://github.com/zwave-js/zwave-js/blob/570fa667e41853fc594a7ded53972ae354649b99/packages/zwave-js/src/lib/node/Node.ts#L660) in ZWaveNode.setValue, instead of handling this in individual command classes?

Not sure I'm understanding this question. The `verifyChanges` hooks are only called from `ZWaveNode.setValue` and `VirtualNode.setValue`, but their implementations differ between command classes due to subtle differences.

Passing along the `result` to `verifyChanges` seems good to me.

PhilipTrauner: by deferring i meant waiting in `ZWaveNode.setValue` before calling `verifyChanges`, instead of waiting in `verifyChanges` itself.

doesn't matter though, passing along `result` is fine with me!

packages/cc/src/cc/MultilevelSwitchCC.ts
```diff
@@ -459,15 +461,34 @@ export class MultilevelSwitchCCAPI extends CCAPI {
...
+							);
+							break;
+						default:
+							(this as this).schedulePoll(
+								currentValueValueId,
+								value === 255 ? undefined : value,
+								{
+									duration: result?.remainingDuration
+										?? duration,
+								},
```
PhilipTrauner, line 487: will adjust the test momentarily, just wanted to get clarification if `result.remainingDuration` should be applied only when the `result.status` is `SupervisionStatus.Working`, or in other statuses as well?

packages/cc/src/cc/MultilevelSwitchCC.ts
AlCalzone: Oh, I'll have to think about that and all its edge cases. Let's stick with passing along the result for now though.

packages/cc/src/cc/MultilevelSwitchCC.ts
AlCalzone: I think it makes sense to poll immediately when the supervision status is `Success` or `Fail`.

When it is `Working`, `remainingDuration` should be defined and non-zero. When it is unsupervised (`undefined`), `duration` should be defined and non-zero.

packages/zwave-js/src/lib/test/driver/setValueSupervision255Verify.test.ts
```diff
@@ -53,7 +131,20 @@ integrationTest(
 				const currentValue = node.getValue(
 					MultilevelSwitchCCValues.currentValue.id,
 				);
-				t.expect(currentValue, "should be initial value").toBe(20);
+				t.expect(currentValue).toBe(20);
+			}
+
+			await node.setValue(MultilevelSwitchCCValues.targetValue.id, 255, {
+				transitionDuration: "1s",
+			});
```
PhilipTrauner, line 139: added to exercise the `result?.remainingDuration` path

AlCalzone (approved):
Thanks a lot! There were a few very minor changes, but I did them myself to not keep you busy for longer.

PR #7965, opened 2025-07-10
fix: detect and fix incorrect default +20/+20 powerlevels

AlCalzone:
It was noticed that in some cases a controller firmware may default to the +20/+20 dBm powerlevels hardcoded in the Z-Wave controller sample application.

With this PR we now detect this (in auto-powerlevel mode) and apply correct/legal powerlevels, even without a region change.


AlCalzone:
@zwave-js-bot automerge

PR #7971, opened 2025-07-13
fix(config): missing and invalid partial config parameters for Aeotec Water Sensors

barna-bash:
###  Turn off LED indicator on water leakage detection
I was not able to turn off the LED indicator for water leak detection on an Aeotec Water Sensor 7 (ZWA018) exclusively because the '0x01' hex value was not included the device's config file as partial parameter.

### Turn off sleep/wake up and tamper LED
However the LED can be turned of entirely for all events setting the configuration parameter (3) value to 0 but the sleep/awake and tamper led indicators neither could be turned off separately due to invalid bitmaps presented in device's config.

I am not sure if the partial parameter keys were changed by a recent firmware update but the manufacturer's documentation I found already included the keys I added to the config.

[ZWA018 docs](https://manual.zwave.eu/backend/make.php?lang=en&sku=AEOEZWA018&cert=&type=mini#:~:text=Parameter%203%3A%20Visual%20LED%20Indications)
[ZWA019 docs](https://manual.zwave.eu/backend/make.php?lang=en&sku=AEOEZWA019&cert=&type=mini#:~:text=Parameter%203%3A%20Visual%20LED%20Indications)

> Parameter 3: Visual LED Indications
> This parameter defines when LEDs will indicate events. Disabling all indications may extend battery life. 
> Value 0 means - No Indications.
> Size: 1 Byte, Default Value: 7 (values 1 + 2 + 4 summarized)
> 
> Setting	Description
> 1	Water Leakage Status Change
> 2	Wake Up
> 4	Device Tampering

#### Water Leak LED in Manufacturer config
Additionally I added a new value to the Aeotec manufacturer config json to have a standard labeling of water leak detection LED indicator states.

### Devices used for testing
Please note that this was tested on a ZWA018 (fw version 2.3) but but I also updated the similar Pro version's config (ZWA019) according to it's documentation.

[inclusion_zwa018.log](https://github.com/user-attachments/files/21204217/inclusion_zwa018.log)

### DIFF
--- a/packages/config/config/devices/0x0086/templates/aeotec_template.json
+++ b/packages/config/config/devices/0x0086/templates/aeotec_template.json
@@ -3226,6 +3226,11 @@
 		"label": "LED Indicator: Open/Close Change",
 		"defaultValue": 1
 	},
+	"led_indicator_water_leak": {
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "LED Indicator: Water Leak",
+		"defaultValue": 1
+	},
 	"led_indicator_wake": {
 		"$import": "~/templates/master_template.json#base_enable_disable",
 		"label": "LED Indicator: Wake Up",
--- a/packages/config/config/devices/0x0371/zwa018.json
+++ b/packages/config/config/devices/0x0371/zwa018.json
@@ -38,12 +38,17 @@
 	},
 	"paramInformation": [
 		{
-			"#": "3[0x04]",
+			"#": "3[0x01]",
+			"$import": "~/0x0086/templates/aeotec_template.json#led_indicator_water_leak"
+		},
+		{
+			"#": "3[0x02]",
 			"$import": "~/0x0086/templates/aeotec_template.json#led_indicator_wake"
 		},
 		{
-			"#": "3[0x10]",
-			"$import": "~/0x0086/templates/aeotec_template.json#led_indicator_tamper"
+			"#": "3[0x04]",
+			"$import": "~/0x0086/templates/aeotec_template.json#led_indicator_tamper",
+			"defaultValue": 1
 		},
 		{
 			"#": "4",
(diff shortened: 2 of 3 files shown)


barna-bash:
#### Off-topic
During my tests I faced a problem which does not let user **set** partial parameters if it is not inlcuded in the device config.
Either after changing the config file I had to exclude and the re-include the node to purge the related config ValueDB.

This is the related key and value in ValueDB: 
```
[
  "{\"nodeId\":5,\"commandClass\":112,\"endpoint\":0,\"property\":3,\"propertyKey\":2}",
  {
    type: "number",
    readable: true,
    writeable: true,
    format: 3,
    valueSize: 1,
    requiresReInclusion: false,
    allowManualEntry: true,
    isAdvanced: false,
    noBulkSupport: false,
    isFromConfig: false,
    min: 0,
    max: 1,
    default: 1,
  },
]
```
Since the **isFromConfig** value was _false_ I could not exclusively update the value either after I added my changs to the device specific JSON config.

### Expected behaviour
1. Include device
     - [0x02] partial parameter is not included in config so the **isFromConfig** is stored as false in the ValueDB
Partial parameter in device config is updated
2. [0x02] Was added to the config
     - I expect the ValueDB for related node(s) is updated accordingly when the driver.start() is called: **isFromConfig: true**

### Actual workflow
1. Include device
2. Device config updated
3. Exclude device
4. Include device
    - Device gets assigned to new nodeId which results in loss of some configuration entries

I wonder if there is there any other ways to sync the ValueDB with the device configuration file if the config file was updated. This would be needed to allow users use up-to-date device partial config without a need to remove a node.

If there is no such sync function I would like to contribute on the development.

AlCalzone:
> I wonder if there is there any other ways to sync the ValueDB with the device configuration file if the config file was updated

Yes, re-interviewing.

PR #7975, opened 2025-07-14
fix(config): minor fixes to Zooz ZEN27 and ZEN30

AlCalzone:
Let's delay this until there is significant use of 15.9.0


AlCalzone:
Superseded by https://github.com/zwave-js/zwave-js/pull/8899

PR #7984, opened 2025-07-16
feat: convert Battery CC `isLow` value to notification event

AlCalzone:
After re-reading the corresponding specs, it became clear that "battery low" is not meant as a status to replace the current battery level, but as a warning notification.

This PR removes the `isLow` property in favor of a new possible argument to the node's or endpoint's `"notification"` event. A low battery report will now trigger that event instead of changing the current battery level.

### DIFF
--- a/docs/api/CCs/Battery.md
+++ b/docs/api/CCs/Battery.md
@@ -7,7 +7,7 @@
 ### `get`
 
 ```ts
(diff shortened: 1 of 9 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #7988, opened 2025-07-20
feat: add rate of incoming messages/second for nodes

### LINKED ISSUE #7841: Keep track of rate of incoming messages per second
This could be done using an exponential moving average, similar to what we already do for the background RSSI. Quick mockup in Python - thanks Copilot :D

```python
import time

class MessageRateTracker:
    def __init__(self, averaging_period=60):
        self.averaging_period = averaging_period  # Zeitraum in Sekunden
        self.rate = 0  # Aktuelle Nachrichtenrate
        self.last_update_time = time.time()  # Letzter Aktualisierungszeitpunkt

    def update(self, message_count=1):
        current_time = time.time()
        elapsed_time = current_time - self.last_update_time

        if elapsed_time > 0:
            # Berechne den Gewichtungsfaktor für den gleitenden Durchschnitt
            alpha = 1 - (1 / self.averaging_period)
            decay = alpha ** elapsed_time

            # Aktualisiere die Rate mit dem exponentiellen gleitenden Durchschnitt
            self.rate = self.rate * decay + message_count * (1 - decay)

            # Aktualisiere den letzten Aktualisierungszeitpunkt
            self.last_update_time = current_time

    def get_rate(self):
        return self.rate
```

hugo-vrijswijk:
Keeps track of the rate of incoming messages for each node, averaged over the last minute.

Implemented using a new `RateTracker` class. Uses `setTimeout` to recalculate and clean up if no messages are received for a while. To prevent `Timeout` overload, the rate is recalculated at most every 2 seconds, even if messages are received more frequently.

I'm fairly confident in the `RateTracker` implementation. It is (I think) fully tested. I'm not too familiar with the rest of the codebase, but I think I am handling the nodeCache and statistics properly. For _when_ to record an incoming message, I've mostly looked at the `lastSeen` statistic, but it is possible I've missed some places or messages are counted twice in some cases. Review carefully :).


packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -6351,6 +6354,7 @@ ${handlers.length} left`,
 					node.updateRTT(msg);
 					// Update last seen state
 					node.lastSeen = new Date();
+					node.updateRate();
```
AlCalzone: We shouldn't update the rate here. This codepath is executed when an outgoing command is acknowledged. We're only interested in the rate of incoming commands, which should be low after the interview is completed.

packages/zwave-js/src/lib/driver/NetworkCache.ts
```diff
@@ -97,6 +97,7 @@ export const cacheKeys = {
 			defaultVolume: `${nodeBaseKey}defaultVolume`,
 			lastSeen: `${nodeBaseKey}lastSeen`,
 			deviceConfigHash: `${nodeBaseKey}deviceConfigHash`,
+			rate: `${nodeBaseKey}rate`,
```
AlCalzone: I also wouldn't cache that information. It will self-populate again whenever there is communication.

`lastSeen` is cached because we use it to decide in which order nodes are pinged after startup.

packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -398,6 +402,11 @@ export class ZWaveNode extends ZWaveNodeMixins implements QuerySecurityClasses {
 		}));
 	}
 
+	/** Average rate of incoming messages per second, in the last minute.. */
+	public get rate(): number {
+		return this.driver.cacheGet(cacheKeys.node(this.id).rate) || 0;
```
AlCalzone: See comment above.

packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -3655,6 +3664,26 @@ ${formatRouteHealthCheckSummary(this.id, otherNode.id, summary)}`,
...
 
+	/**
+	 * Updates the rate of incoming messages for this node
+	 * @internal
+	 */
+	public updateRate(): void {
+		if (!this.rateTracker) {
+			// Callback for when the rate is updated, also update cache and statistics
+			const onUpdateStatistics = (rate: number) => {
+				this.driver.cacheSet(cacheKeys.node(this.id).rate, rate);
```
AlCalzone: See comment above.

AlCalzone (changes requested):
I'll try to test this soon, but the PR looks pretty solid in general. Just a few notes:

packages/zwave-js/src/lib/driver/NetworkCache.ts
hugo-vrijswijk: Done, removed

packages/zwave-js/src/lib/node/Node.ts
hugo-vrijswijk: Removed, now uses a direct call to the rateTracker, or 0

packages/zwave-js/src/lib/node/Node.ts
hugo-vrijswijk: Callback only updates statistics now

AlCalzone (dismissed):
Thanks! I'll mark this as a draft until I've verified with real devices that it works as intended.

PR #8004, opened 2025-07-23
fix: disable optimistic value updates for slow device classes

### LINKED ISSUE #3892: Distinguish between fast and slow device classes when doing optimistic value updates
We should distinguish between devices that do fast and device that do slow transitions and only do optimistic updates for fast ones.
This way, UIs could give immediate feedback about the state of lights but it would still be possible to wait for the slower ones like roller shutters.

Fast would be:
- Lights without a specified transition time
- Lights with a short transition time (< 5s for example)
- Binary switches, etc.

Slow would be:
- Motor control
- Window Covering

fmauNeko:
Allows specifying per device class whether optimistic updates are supported or not (defaults to supported to avoid breaking any existing behavior).

Closes #3892


packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -636,9 +636,15 @@ export class ZWaveNode extends ZWaveNodeMixins implements QuerySecurityClasses {
 
 				const shouldUpdateOptimistically =
 					api.isSetValueOptimistic(valueId)
+					// Check if the device class supports optimistic value updates
+					&& (endpointInstance.deviceClass?.specific
+						?.supportsOptimisticValueUpdate
+						?? endpointInstance.deviceClass?.generic
+							?.supportsOptimisticValueUpdate
+						?? true)
```
AlCalzone: This seems redundant. If `deviceClass` is defined, so is `deviceClass.specific` and its `supportsOptimisticValueUpdate` property has also been initialized.


```suggestion
					&& (endpointInstance.deviceClass?.specific.supportsOptimisticValueUpdate
						?? true)
```

packages/core/src/test/deviceClasses.test.ts
AlCalzone, line 1: Move this to the `registries` folder and call it `DeviceClasses.test.ts` please so it matches the file it tests.

packages/zwave-js/src/lib/test/optimisticValueUpdateLogic.test.ts
AlCalzone, line 1: This entire test seems redundant with the other tests you added. I would actually only keep the integration tests and the ones for the registry definitions.

AlCalzone (changes requested):
Thanks! I've got a few remarks.

AlCalzone:
Thanks! I'm going to take a deeper look at the integration test itself when I'm back in front of a bigger screen. I'll probably fix what I find myself though to not keep you busy any longer. Looks okay at a first glance but there might be some details I missed.

AlCalzone (dismissed):

AlCalzone:
Thanks again. Took me a while, but I finally was able to test this with a motorized blind controller

AlCalzone (approved):

PR #8005, opened 2025-07-23
feat(config): prepare Inovelli VZW31-SN for future firmware upgrade

The00Dustin:
This PR contains multiple value and description fixes in addition to laying the groundwork for a future firmware upgrade.  There are also some parameter name corrections based on more accurately describing functionality and one name correction to match existing documentation.

Tagging @InovelliUSA in case explicit approval is required.

### DIFF
--- a/packages/config/config/devices/0x031e/vzw31-sn.json
+++ b/packages/config/config/devices/0x031e/vzw31-sn.json
@@ -771,7 +771,7 @@
 		{
 			"#": "131",
 			"$if": "firmwareVersion >= 1.4",
-			"label": "Group 7: Low Level",
+			"label": "Group 7: Level 1",
 			"description": "Low level to use when group 7 cycles levels",
 			"valueSize": 1,
 			"minValue": 1,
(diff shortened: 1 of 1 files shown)


The00Dustin:
In case it's due to some change, and even though all checks passed, I should mention that the codespaces problems tab says "Comments are not permitted in JSON" for lines 386-388, but I did not remove them since they were already there.

packages/config/config/devices/0x031e/vzw31-sn.json
```diff
@@ -207,12 +207,13 @@
...
 			"#": "15",
 			"$import": "templates/inovelli_templates.json#level_0-99_100_previous",
-			"label": "State After Power Restored",
+			"label": "Level After Power Restored",
 			"defaultValue": 100
 		},
 		{
 			"#": "17",
 			"label": "Load Level Indicator Timeout",
+			"description": "Allowable range: 0-11",
```
AlCalzone:
```suggestion
```

This was redundant given that there is no gap in the range.

AlCalzone (changes requested):

AlCalzone (approved):

AlCalzone:
Thanks!

PR #8010, opened 2025-07-26
fix(config): remove non-existent parameter 107 for Shelly Wave Plus S

nberlee:
According to the documentation there is no such thing as parameter 107. 

I think this was a leftover from copying the Shelly Wave i4? (@QubinoHelp)

Documentation: https://www.shelly.com/blogs/documentation/shelly-wave-plug-s-eu


AlCalzone (approved):
Seems reasonable, thanks!

PR #8013, opened 2025-07-26
feat(config): update Aeotec Trisensor 8 to firmware 2.8.4

nberlee:
See https://aeotec.freshdesk.com/support/solutions/articles/6000276825-update-trisensor-8-to-v2-8-4 
pdf: https://aeotec.freshdesk.com/helpdesk/attachments/6210241735

### DIFF
--- a/packages/config/config/devices/0x0086/templates/aeotec_template.json
+++ b/packages/config/config/devices/0x0086/templates/aeotec_template.json
@@ -122,7 +122,7 @@
 				"value": 5
 			},
 			{
-				"label": "Puple",
+				"label": "Purple",
 				"value": 6
 			},
 			{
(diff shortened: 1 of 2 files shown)


AlCalzone (changes requested):
From the linked article it seems like the firmware only adds new parameters. We shouldn't split the file for that. Please add the parameters to the original file instead with a condition
```
"$if": "firmwareVersion >= 2.8.4",
```
as the first thing in the new parameters.

Also, we don't generally keep the verbose manufacturer descriptions if they don't add significant value in addition to the label.

nberlee:
@AlCalzone fixed.

The if statement makes sense, I didn't found that on my own. And I think I removed most verbose descriptions I added.

AlCalzone:
For reference, the corresponding documentation is here: https://zwave-js.github.io/zwave-js/#/config-files/conditional-settings
You could go quite crazy with it if you wanted - not that it is always a good thing.

AlCalzone (dismissed):
Thanks, I made a few more changes to get this across the finish line.

AlCalzone (approved):

PR #8021, opened 2025-07-30
feat(zwave-js): skip interview if S2 failed

### LINKED ISSUE #7011: Skip node interview if inclusion with Security S2 failed
When I am including devices sometimes S2 Security fails (device took too long to respond, wrong pin, etc) and the device is included with no security.

Can you add an option to set in the inclusion process for Security only, if S2 (or S0) fails the whole inclusion fails.

What I currently have to do is wait for the device discovery to finish, then I can run an exclusion.  Once eclusded I can try the inclusion again.

**Additional context**

renaiku:
Fixes [#7011](https://github.com/zwave-js/zwave-js/issues/7011)

[This answer](https://github.com/zwave-js/zwave-js/issues/7011#issuecomment-2228010088) to the issue guided me to the solution.

### Problem:
A node should not be interviewed when security S2 fails

### Solution:
Adding **InclusionStrategy** in **InclusionResult** allows to check for low security of S2 strategy just before the interview.


AlCalzone (changes requested):
Hmm, this is not ideal.

a) The `InclusionResult` is used by applications, and those applications pass in the inclusion strategy when starting the inclusion. So this would essentially pass unnecessary info back

b) It does not consider the interviews that happen when Z-Wave JS restarts, loads its network information from cache and some devices have not fully been interviewed.

renaiku:
> a) The `InclusionResult` is used by applications, and those applications pass in the inclusion strategy when starting the inclusion. So this would essentially pass unnecessary info back

Oh I see. Would moving the skip condition before the "node added" event would be better then ?

```ts
// Skip the interview if S2 failed
if (result.lowSecurity
  && opts.strategy == InclusionStrategy.Security_S2) return;

this.emit("node added", newNode, result);
```

> b) It does not consider the interviews that happen when Z-Wave JS restarts, loads its network information from cache and some devices have not fully been interviewed.

Can you point me where this happen please ? I would like to look into it.

AlCalzone:
> Would moving the skip condition before the "node added" event would be better then ?

Not really. The event is needed for applications. I think it would make more sense to store this as an attribute of the node that also gets persisted to the cache.

> Can you point me where this happen please ? I would like to look into it.

See the `initializeControllerAndNodes` method in `Driver.ts` - pretty far at the end of that method.

AlCalzone (changes requested):
This is better I think. But did you forget the part of the code that actually skips the interview if this flag is set?

renaiku:
Yes sorry @AlCalzone, I 100% forgot 🥲

If it's ok to put the skip there in the Driver.ts file, I can quickly go to a computer with git on it and update the PR.
```typescript
    /** This is called when a new node has been added to the network */
	private onNodeAdded(node: ZWaveNode): void {
		this.addNodeEventHandlers(node);

		if (this._options.interview?.disableOnNodeAdded) return;
		if (this._options.testingHooks?.skipNodeInterview) return;

		if (node.failedS2) return; // <===== SKIP ADDED HERE

		// Interview the node
		// don't await the interview, because it may take a very long time
		// if a node is asleep
		void this.interviewNodeInternal(node);
	}
```

EDIT: pushed it

AlCalzone:
Superseded by https://github.com/zwave-js/zwave-js/pull/8162

PR #8039, opened 2025-08-05
feat: implement tools to convert Z-Wave logs back to structured, filtered logs

AlCalzone:
I've been playing with enabling AI-assisted log file analyzing. A major issue for this is the sheer size of our logs. They also contain a lot text that is useful for debugging, but fills up the context of AI models with noise.

This PR implements some tools to automatically convert formatted Z-Wave JS logs back to a structured representation, and filters logs that are not interesting for automatic analysis.


AlCalzone:
This has been pulled out into https://github.com/zwave-js/log-analyzer/

PR #8042, opened 2025-08-06
feat(config): bring Inovelli VZW32-SN up to date with latest firmware changes and restore parity with VZW31-SN

The00Dustin:
This update:

- Reverts some unapproved name changes that  slipped through the cracks
- Improves some parameter names and descriptions
- Corrects some default values and ranges to match the firmware (without conditionals for pre-production versions)
- Adds some new parameters from the latest firmware (again, without conditionals for pre-production versions)
- Adds a description to VZW31-SN that was previously overlooked in #8005

Tagging @InovelliUSA in case explicit approval is required.


The00Dustin:
Forgot to mention I tried to bypass lint using codespace for some parameters that load preset XYZ values where the case suggestion surely wouldn't make sense; that may explain failure even though codespace showed passing...

The00Dustin:
Undid the change to VZW31-SN, it was previously added and reverted, but I didn't notice that when looking at the merged PR.

The00Dustin:
Added a new parameter that should also show up in VZW31-SN on future firmware.

The00Dustin:
Added another new parameter that is VZW-32SN specific and "disabled" some other VZW-32SN parameters that shouldn't be enabled until a future TBD firmware version.

packages/config/config/devices/0x031e/vzw32-sn.json
```diff
@@ -872,6 +895,65 @@
 			"unsigned": true,
 			"readOnly": true
 		},
+		{
+			"#": "116[0xff]",
+			"$import": "templates/inovelli_templates.json#mmwave_person_in_zone",
```
AlCalzone, line 888: This import target does not exist.

packages/config/config/devices/0x031e/vzw32-sn.json
```diff
@@ -1002,6 +1084,26 @@
...
+		{
+			"#": "239",
+			"label": "SPI Flash Status",
+			"description": "0 = SPI flash fail; 1 = SPI flash success",
+			"valueSize": 1,
+			"minValue": 0,
+			"maxValue": 1,
+			"defaultValue": 0,
+			"readOnly": true
+		},
```
AlCalzone: Convert the description to options please.

packages/config/config/devices/0x031e/vzw32-sn.json
```diff
@@ -1002,6 +1084,26 @@
...
+		{
+			"#": "240",
+			"label": "Factory Reset",
+			"description": "Set to 120935937 to perform a factory reset",
```
AlCalzone, line 1096: This parameter should also have two options: 0 for idle and 120935937 for factory reset.

packages/config/config/devices/0x031e/vzw32-sn.json
```diff
@@ -706,40 +723,44 @@
 			"unsigned": true
 		},
 		{
-			"#": "109[0xff000000]",
-			"label": "After Hours Time Range Start Hour",
-			"description": "The UTC hour where After Hours mode should start",
+			"#": "109[0xff]",
+			"label": "Future Use",
```
AlCalzone, line 715: Remove please. We have the option to get/set arbitrary config parameters - no need to document ones that should not be used.

AlCalzone (changes requested):

packages/config/config/devices/0x031e/vzw32-sn.json
The00Dustin: These parameters exist in the firmware, but aren't working as expected.  The goal here is to discourage unguided use until the features are working as expected (similar to https://github.com/zwave-js/zwave-js/issues/6273#issuecomment-1714585107 but temporary).  If removed from the config, they will still show up in the interview and will not be read only.  I couldn't find a way to hide the parameters from view when they show up in the interview.  I also tried making it one read only parameter, but the interview overrode that as well.  Did I overlook a way to hide these, is my config a viable alternative option, or do I need to suggest Inovelli have their firmware engineer exclude these from the interview until testing is completed?

packages/config/config/devices/0x031e/vzw32-sn.json
The00Dustin: The goal here is to prevent accidental factory resets.  Causing this to be a drop down option instead of requiring the user to type or copy and paste the value required to trigger a reset would negate that goal.

packages/config/config/devices/0x031e/vzw32-sn.json
The00Dustin: Pushed a commit to resolve this.

packages/config/config/devices/0x031e/vzw32-sn.json
The00Dustin: You said convert, but if I delete the description, the information is lost since it is a read only parameter.
As is:
Screenshot 2025-09-04 at 08 59 00
After requested change (add options, but remove description):
Screenshot 2025-09-04 at 08 51 48
Does this need the options (that don't appear to do anything) as well as the description, are options only the correct config with a ZUI bug introduced in v11, or should it stay as is?

packages/config/config/devices/0x031e/vzw32-sn.json
The00Dustin: Not sure negate is the right word, and not trying to be terse, just looking for additional input here, operating under the assumption that Inovelli won't like the suggested change.  Alternatives might include removing the description so that users must refer to documentation to use the parameter and/or moving the number from the description to the factory reset help.

packages/config/config/devices/0x031e/vzw32-sn.json
AlCalzone: Ohh, I see. That's fine for now then. We'll need to add a way to effectively hide config parameters from users.

packages/config/config/devices/0x031e/vzw32-sn.json
AlCalzone: I understand the intent. This might need a more general discussion, because what I proposed is how we handle it for many other devices.
Forcing a confirmation of sorts could be useful for these "destructive" parameters.

AlCalzone: In that case, let's keep it like you have it for now.

packages/config/config/devices/0x031e/vzw32-sn.json
AlCalzone: can you try adding `allowManualEntry: false`?
This might need a change in Z-UI then, as the intention is for options to also serve as labels for specific values, even when readonly.

The00Dustin: Adding `allowManualEntry: false` didn't work. Based on comparison with other parameters, I also removed minValue, maxValue, and defaultValue, but it still didn't work.  Re-interview solved it.  Changing it further and restarting updated as expected.  I've been feeling like I'm not prompted for re-interview when I should be in some cases, but I haven't reported, this is another example of that.

The00Dustin: I probably also should have mentioned that I've pushed the commit for this...

AlCalzone (approved):

PR #8052, opened 2025-08-12
fix: reset controller recovery phase after successful transmission

AlCalzone:
Fixes the issue reported in https://github.com/zwave-js/zwave-js/discussions/8016#discussioncomment-14079490


AlCalzone:
@zwave-js-bot automerge

PR #8087, opened 2025-08-19
fix: logic to fallback to stored slow refresh capability

### LINKED ISSUE #8066: Logic Group ZHC5010 Wall Switch - Long press not working after Z-wave JS version 0.9.0
### Discussed in https://github.com/zwave-js/zwave-js/discussions/8041

<sup>Originally posted by **2mashoffmann** August  6, 2025</sup>
### Checklist

### Describe the issue

What is happening?
When long pressing the Logic Group ZHC5010 wall switch, Z-Wave JS send "KeyHeldDown" and immediately after send "KeyReleased", and not when the button is released. This problem started after upgrading from Z-wave JS 0.9.0 to 0.10.0. Same problem in the new version 0.20.0. When downgrading to Z-wave JS 0.9.0 everything works fine again, and I only get "KeyReleased" when I release the button on the wall switch.

"KeyReleased" send when I release the button.

  1. Long Press the wall switch

I have successfully re-interview the problematic device when logging was enabled, but I can't find it in the log.

### Software versions

Driver (zwave-js): 15.10.0

Home Assistant Z-Wave JS Addon: Yes, ver. 0.20.0

AEON Labs ZW090 (Z‐Stick Gen5) firmware 1.2

### Device information

Manufacturer: Logic Group
Model name: ZHC5010
Node ID: 024

### Checklist

### Upload Logfile

[home-assistant_zwave_js_2025-08-06T09-10-02.353Z.log](https://github.com/user-attachments/files/21615039/home-assistant_zwave_js_2025-08-06T09-10-02.353Z.log)
</div>

## Analysis:

The device does not officially support slow refresh, but it also does not report every 200ms. The specs mention this case:

> A controller SHOULD apply an adaptive approach based on the reception of the Key Released Noti-
fication. Initially, the controller SHOULD time out if not receiving any Key Held Down Notification
refresh after 400ms and consider this to be a Key Up Notification. If, however, the controller subse-
quently receives a Key Released Notification, the controller SHOULD consider the sending node to
be operating with the Slow Refresh capability enabled

but we don't currently follow that recommendation. In the case where we receive a KeyReleased notification after we originally forced the key to be released, we should change the information we know about the node:
- supportsSlowRefresh -> true
- slowRefresh -> true

and log an appropriate message.

AlCalzone:
The problem was that the `slowRefresh` field on KeyReleased Central Scene notifications is always defined, but it is `false`. This lead to us not evaluating the cached/fixed value for that field.

fixes: https://github.com/zwave-js/zwave-js/issues/8066

### DIFF
--- a/packages/zwave-js/src/lib/node/CCHandlers/CentralSceneCC.ts
+++ b/packages/zwave-js/src/lib/node/CCHandlers/CentralSceneCC.ts
@@ -98,7 +98,8 @@ export function handleCentralSceneNotification(
 		// If the node does not advertise support for the slow refresh capability, we might still be dealing with a
 		// slow refresh node. We use the stored value for fallback behavior
 		const slowRefresh = command.slowRefresh
-			?? node.valueDB.getValue<boolean>(slowRefreshValueId);
+			// Prefer the stored value, even if the command claims slowRefresh == false
+			|| node.valueDB.getValue<boolean>(slowRefreshValueId);
 		store.keyHeldDownContext = {
 			sceneNumber: command.sceneNumber,
 			// Unref'ing long running timers allows the process to exit mid-timeout


AlCalzone:
@zwave-js-bot automerge

PR #8143, opened 2025-09-05
feat: support creating mixed LR and non-LR multicast groups

### LINKED ISSUE #7073: Support creating mixed LR and non-LR multicast groups
Like with mixed security, this should internally be split into two groups, one for each protocol.

Edvinas9:
Aim of this PR is to implement a new feature that is described in this issue: https://github.com/zwave-js/zwave-js/issues/7073

:exclamation:  This is **work in progress** still, but due to lack of time I'm not sure if I'll be able to finish it up, so just leaving here as a reference for whoever is gonna pick it up later if I won't do it myself.

Fixes: #7073


AlCalzone (changes requested):
Thanks, I can take it from here. For future reference, I'd flatten the grouping logic from nested groups (protocol -> security class) to a flat group of (protocol + security class), which should make the rest a little bit easier to follow.

Other than that, the approach looks solid.

AlCalzone:
Confirmed working. Thanks for the head start!

AlCalzone (approved):

PR #8155, opened 2025-09-09
fix(serial): hide TX report fields when transmitStatus is Fail

### LINKED ISSUE #8051: When transmitStatus is Fail, none of the other fields are populated
```
2025-08-12 07:48:51.956 DRIVER « [REQ] [SendDataBridge]
                                   callback id:            80
                                   transmit status:        Fail, took 0 ms
                                   routing attempts:       0
                                   protocol & route speed: Unknown (0x00)
                                   routing scheme:         Idle
                                   ACK RSSI:               0 dBm
                                   ACK channel no.:        0
                                   TX channel no.:         0
```

...we should ignore them.

winterborn:
- Only show TX report fields in logs when `transmitStatus` is `OK` or `NoAck`
- Prevents misleading log output showing invalid routing/ACK data on failures  
- Add unit tests to verify correct behavior for `TransmitStatus` values

Fixes #8051

### DIFF
--- a/packages/serial/src/serialapi/transport/SendDataBridgeMessages.test.ts
+++ b/packages/serial/src/serialapi/transport/SendDataBridgeMessages.test.ts
@@ -1,0 +1,206 @@
+import { TransmitStatus } from "@zwave-js/core";
+import { describe, expect, test } from "vitest";
+import { SendDataBridgeRequestTransmitReport } from "./SendDataBridgeMessages.js";
+
+describe("SendDataBridgeRequestTransmitReport", () => {
+	test("should not show TX report fields when transmitStatus is Fail", () => {
+		const rawMessage = new Uint8Array([
+			80,
+			TransmitStatus.Fail,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+		]);
+		const report = SendDataBridgeRequestTransmitReport.from(
+			{ payload: rawMessage } as any,
+			{} as any,
+		);
+
+		expect(report.transmitStatus).toBe(TransmitStatus.Fail);
+		expect(report.callbackId).toBe(80);
+
+		const logEntry = report.toLogEntry();
+		expect(logEntry.message).toStrictEqual({
+			"callback id": 80,
+			"transmit status": "Fail",
+		});
+	});
+
+	test("should show TX report fields when transmitStatus is OK", () => {
+		const rawMessage = new Uint8Array([
+			80,
+			TransmitStatus.OK,
+			10,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+		]);
+		const report = SendDataBridgeRequestTransmitReport.from(
+			{ payload: rawMessage } as any,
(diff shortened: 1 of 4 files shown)


packages/serial/src/serialapi/transport/SendDataBridgeMessages.ts
```diff
@@ -310,9 +310,13 @@ export class SendDataBridgeRequestTransmitReport
 				"transmit status":
 					getEnumMemberName(TransmitStatus, this.transmitStatus)
 					+ (this.txReport
+							&& (this.transmitStatus === TransmitStatus.OK
+								|| this.transmitStatus === TransmitStatus.NoAck)
```
AlCalzone: I think we should keep the txTicks in the log message always. There are some edge cases with older controllers which respond with a `Fail` status after a while - probable due to errors in route resolution or so. This is functionally different from the case where the controller immediately responds with `Fail`.

packages/serial/src/serialapi/transport/SendDataMessages.ts
```diff
@@ -298,9 +298,13 @@ export class SendDataRequestTransmitReport extends SendDataRequestBase
 				"transmit status":
 					getEnumMemberName(TransmitStatus, this.transmitStatus)
 					+ (this.txReport
+							&& (this.transmitStatus === TransmitStatus.OK
+								|| this.transmitStatus === TransmitStatus.NoAck)
```
AlCalzone: dito

packages/serial/src/serialapi/transport/SendDataBridgeMessages.test.ts
```diff
@@ -0,0 +1,206 @@
...
+			0,
+			0,
+			0,
+			0,
+			0,
+		]);
+		const report = SendDataBridgeRequestTransmitReport.from(
+			{ payload: rawMessage } as any,
+			{} as any,
+		);
```
AlCalzone: I feel like it would be easier to understand to use the `SendDataBridgeRequestTransmitReport` constructor, like so:
```ts
const report = new SendDataBridgeRequestTransmitReport({
  callbackId: ...,
  transmitStatus: TransmitStatus.Fail,
  // ... other fields
});
```

same for the other test cases.

AlCalzone (changes requested):
Thanks! I have a couple of remarks.

packages/serial/src/serialapi/transport/SendDataBridgeMessages.ts
winterborn: Sorry for the delay, changes are [here](https://github.com/zwave-js/zwave-js/pull/8155/commits/294a835c8d98f77448311373331bc8bb9b7ba4bf).

packages/serial/src/serialapi/transport/SendDataMessages.ts
winterborn: Sorry for the delay, changes are [here](https://github.com/zwave-js/zwave-js/pull/8155/commits/294a835c8d98f77448311373331bc8bb9b7ba4bf).

packages/serial/src/serialapi/transport/SendDataBridgeMessages.test.ts
winterborn: Sorry for the delay, changes are [here](https://github.com/zwave-js/zwave-js/pull/8155/commits/35ab9827c056fe05ae6d89bc079b07ef726d41d6).

AlCalzone (approved):
Thanks!

I just added the txReport field to the `Fail` tests to ensure that everything but the duration is actually omitted. With that change, this is good to merge.

PR #8157, opened 2025-09-11
feat: support checking for all firmware updates at once

### LINKED ISSUE #82: Make it possible to distinguish between `no updates available` and `no updates supported`
Today, if you query the firmware update service, you will only see updates if there are firmware updates available for the device in question. When you don't see any updates, however, it's hard to know whether that means that the device IS supported by the update service but has no updates available or the device ISN'T supported by the update service and therefore has no updates available.

One suggestion @AlCalzone had on Discord was to return a `null` instead of an empty list to indicate that the device is not supported by the service. Not sure if this needs to go into the service itself or into the caller, but starting here for tracking purposes

AlCalzone:
This PR adds a new method `getAllAvailableFirmwareUpdates` to the `Controller` class, which checks for firmware updates for all identified devices in a single request. This also adds the possibility to detect which devices are not known to the firmware update service, since they will be omitted from the returned `Map`.

fixes: https://github.com/zwave-js/firmware-updates/issues/82


AlCalzone:
@zwave-js-bot automerge

PR #8161, opened 2025-09-13
feat(hidden-parameters): implement support for hidden device configuration parameters

### LINKED ISSUE #8138: Add a way to hide existing config parameters
Right now if they are undocumented, they show up after the interview if the device reports them. We need a way to hide "unwanted" ones.

oniseun:
PR description here

Add support for hiding device configuration parameters from user interfaces
and API outputs. This addresses issue #8138 by allowing device manufacturers
and contributors to mark parameters as hidden to improve user experience.

- Add hidden field support to device configuration JSON schema
- Update ConditionalParamInformation class to handle hidden property
- Update ConfigurationCC to completely skip hidden parameters during loading
- Add documentation in docs/config-files/hidden-parameters.md
- Update file-format.md to reference hidden parameter feature
- Fix execa import issue in maintenance/test.mts

Hidden parameters are now completely invisible to applications - no metadata 
is created and no values are persisted. This allows hiding factory test 
parameters, calibration data, debug settings, and other parameters that 
should not be exposed to end users.

Closes #8138


packages/cc/src/cc/ConfigurationCC.ts
```diff
@@ -1591,6 +1591,7 @@ alters capabilities: ${!!properties.altersCapabilities}`;
 				label: info.label,
 				description: info.description,
 				isFromConfig: true,
+				hidden: info.hidden,
```
AlCalzone: I'd go a step further and not create metadata or persist values for hidden parameters at all. By only setting the flag we shift the burden to applications.

packages/core/src/values/Metadata.ts
```diff
@@ -179,6 +179,7 @@ export interface ConfigurationMetadata extends ValueMetadataAny {
 	states?: Record<number, string>;
 	allowManualEntry?: boolean;
 	isFromConfig?: boolean;
+	hidden?: boolean;
```
AlCalzone: This property does not need to be here if we hide the values entirely.

AlCalzone (changes requested):
Thanks, this is a good start. It does not fully do what the issue intended though.

oniseun:
> Thanks, this is a good start. It does not fully do what the issue intended though.

Hey @AlCalzone! Thanks for the feedback, and sorry for taking a while to get back to you. I've updated the handling of hidden parameters so instead of just marking them as hidden, the system now completely ignores them during config loading.

What changed:

- Removed hidden property from ConfigurationMetadata
- Updated ConfigurationCC to skip hidden params entirely with continue
- Updated docs to reflect the new behavior

 Does this approach work better for what you had in mind?

AlCalzone:
Did you close this on purpose?

AlCalzone:
I've now merged your work in https://github.com/zwave-js/zwave-js/pull/8449

PR #8162, opened 2025-09-13
feat: skip node interview after S2 bootstrapping failure

### LINKED ISSUE #7011: Skip node interview if inclusion with Security S2 failed

greguz:
### Why

fixes: #7011

### What

The interview process should be skipped when a node is expected to use S2 and fails the secure inclusion:

1. The info about when S2 is expected (`InclusionOptions.strategy`) is only available during inclusion.
2. Once the node is restored from cache, the original `strategy` is lost.
3. There are even other methods that can trigger the interview process (like the `onNodeAlive` hook).

I added a cached flag `skipInterview` for possible future use (maybe for S0 or S3) and directly blocked the `interviewInternal` method. Also added a `warn` log.


greguz:
I'm still figuring out how to implement at least one basic test.

greguz:
Added a test for the new cached value. If anyone can confirm that the solution is what we’re expecting, I’d be more than happy to ask for some sample code to mock an S2 inclusion and end the PR.

packages/zwave-js/src/lib/node/mixins/05_Security.ts
```diff
@@ -77,4 +77,15 @@ export abstract class NodeSecurityMixin extends NetworkRoleMixin
 		// If we don't have the info for every security class, we don't know the highest one yet
 		return missingSome ? NOT_KNOWN : SecurityClass.None;
 	}
+
+	/**
+	 * Should be `true` when an S2-capable node fails to be included when S2 is expected.
+	 */
+	public get skipInterview(): boolean {
```
AlCalzone: I think `failedS2Bootstrapping` or something like that would capture the intention better.

packages/zwave-js/src/lib/test/driver/s2FailureDisablesInterview.test.ts
```diff
@@ -0,0 +1,60 @@
...
+	const oldNode = new ZWaveNode(2, driver);
+	expect(oldNode.skipInterview).toBe(false);
+	oldNode.skipInterview = true;
+	expect(oldNode.skipInterview).toBe(true);
+
+	const newNode = new ZWaveNode(2, driver);
+	expect(newNode.skipInterview).toBe(true);
+});
+
+// TODO: Make the node fail S2 and then check if skipInterview is true and also ensure that "interview completed" is NOT emitted
```
AlCalzone: A test for this is going to be challenging, especially if you don't have too much experience with the low level Z-Wave details.
The correct approach would be an `integrationTest(...)` based test, but you'd have to define the end device behavior up to a certain point.

I'd leave this out for now - I'll test manually with an actual device and maybe write a test if there is a need for it.

packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -948,10 +948,19 @@ export class ZWaveNode extends ZWaveNodeMixins implements QuerySecurityClasses {
...
 		}
 
+		if (this.skipInterview) {
+			this.driver.controllerLog.logNode(
+				this.id,
+				"node cannot be inteviewed",
+				"warn",
+			);
+			return true;
+		}
```
AlCalzone: I think it would make more sense to move this check to the start of `Driver.interviewNodeInternal(...)`

AlCalzone (changes requested):
Thanks, just a few minor revisions please.

AlCalzone (commented):
Oh, I forgot to mention - there are a few other cases where S2 bootstrapping happens and can fail. Make sure to handle them all.

packages/zwave-js/src/lib/test/driver/s2FailureDisablesInterview.test.ts
greguz: Saw the `integrationTest` suite. I'll take a good look at some code that uses the `MockNodeBehavior` feature. I should be able to push an update in a couple of days at most.

greguz:
> Oh, I forgot to mention - there are a few other cases where S2 bootstrapping happens and can fail. Make sure to handle them all.

Yeah, you're right, thanks!

greguz:
I played a bit with `integrationTest` and `defineBehavior`: I accept your kind offer to manually test the S2 bootstrap failure :)

Let me know if I missed something.

AlCalzone (dismissed):
Thanks again. I was finally able to add the integration test now that we can simulate device inclusion with S2.

AlCalzone (approved):

PR #8164, opened 2025-09-15
feat: add recommendedValue field to the parameter config

### LINKED ISSUE #6239: Define recommended config parameter settings
... and optionally apply them as part of the interview (opt-in), or add an alert to https://github.com/zwave-js/node-zwave-js/issues/2909

_originally brought up in [#6235](https://github.com/zwave-js/node-zwave-js/discussions/6235)_

eshurakov:
Fixes #6239 

This PR introduces a new optional field, `recommendedValue`, to the parameter configuration. This field suggests a parameter value that may offer better performance or behavior than the factory default and is expected to benefit most users.

The `recommendedValue` is applied automatically during the device interview only if the `applyConfigurationRecommendedValues` driver option is enabled (disabled by default) and the current value is still set to the default. This ensures the feature remains opt-in and prevents unexpected parameter changes.

New tests are added to verify that the recommended values are applied correctly (`Node.recommendedValues.test.ts`)


docs/config-files/file-format.md
eshurakov, line 1: I've only added `recommendedValue` row to the table, rest of the changes are from the formatter.

eshurakov:
I've also looked into `ConfigurationMetadata`, but I'm not certain whether `recommendedValue` needs to be added there and persisted, or if it should remain only in the parameter configuration.

packages/zwave-js/src/lib/node/Node.ts
```diff
@@ -1050,6 +1058,100 @@ export class ZWaveNode extends ZWaveNodeMixins implements QuerySecurityClasses {
 		this.driver.controllerLog.interviewStage(this);
 	}
 
+	/**
+	 * Applies all recommended values from the device config to this node
+	 * Applies only if the current value is still the default value and
+	 * the recommended value is different
+	 * This is called at the end of the interview
+	 */
+	private async applyRecommendedValues(): Promise<void> {
```
AlCalzone: I would move this entire method into the `ConfigurationCC` class in `packages/cc/src/cc/ConfigurationCC.ts`. Call it after the call to `refreshValues` in the `interview` method.

This has the advantage that you can simply call `setBulk` with the desired param/bitmask/value combinations and let the existing code figure out which values to update and how.

AlCalzone (changes requested):
Thank you - very thorough. Just one remark.

Regarding your other question
> I've also looked into ConfigurationMetadata, but I'm not certain whether recommendedValue needs to be added there and persisted, or if it should remain only in the parameter configuration.

I think it would make sense. That way, applications can highlight recommended values.

packages/zwave-js/src/lib/node/Node.ts
eshurakov: Good point! I will refactor to use setBulk and move the method to `ConfigurationCC.ts`.

Thank you!

eshurakov: I've moved the `applyRecommendedValues` method into the `ConfigurationCC` class, it looks much better!  

However, I've run into an issue that I'm not sure how to solve due to my limited knowledge of the system. After calling `setBulk`, the values returned by `node.getValue` are still the old ones. I checked the `setBulk` method and noticed that it doesn't update values after a successful call. On the other hand, `node.setValue` includes logic to verify changes, which I believe is why my test worked before moving the code into `ConfigurationCC` and using `setBulk`.  

For now, I've added a `refreshValues` call after `setBulk`, but this feels more like a workaround than the right solution.

https://github.com/zwave-js/zwave-js/pull/8164/files#diff-f87a57b236158621d7d0cecb4dd33e9133425d8ee799d402169b50b99bf67dacR1360-R1361

eshurakov:
👋  @AlCalzone, thank you for the review and for the feedback! I've addressed it, but had another question that I described in my comment here: https://github.com/zwave-js/zwave-js/pull/8164#discussion_r2356555037

Appreciate any guidance.

packages/zwave-js/src/lib/node/Node.ts
AlCalzone: The API methods are meant to do what they say and nothing unexpected. Allmost all magic is done in higher layers.

I've checked the `setBulk` implementation and it is a bit messy for relying on the possibly supervised execution result, because it also handles the case where the actual `BulkSet` command isn't supported.

I think refreshing the values regardless of the outcome is fine, but it should be done with `getBulk` instead of `refreshValues` so only the affected parameters are actually refreshed and not possibly 200+.

AlCalzone:
Thanks - I've responded to the comment thread.

packages/zwave-js/src/lib/node/Node.ts
eshurakov: That makes sense! I've updated the method to use `getBulk` instead of `refreshValues`.

AlCalzone (dismissed):
Thanks! I'm going to test this with an actual device before merging, but this looks good.

AlCalzone (dismissed):

AlCalzone (approved):

PR #8165, opened 2025-09-15
fix(zniffer): parse multiple frames from ZLF, support attachments

AlCalzone:
While trying to investigate a Zniffer trace, I noticed some issues with our ZLF parsing:
- ZLF "lines" can contain multiple ZnifferMessages, but we only handled fragmented ones
- The length field is actually a 32-bit integer, we only parsed the first byte of it
- Attachments were not supported

This PR fixes all of the above, with the exception of only parsing network key attachments, but not the other kinds.


AlCalzone:
@zwave-js-bot automerge

PR #8170, opened 2025-09-16
feat(cc): add ability to mark config parameters as destructive

### LINKED ISSUE #8137: Add a way to mark config parameters as destructive, forcing confirmation before setting them
This could be useful for factory reset parameters.

winterborn:
- Mark configuration parameters as `destructive: true` in device configs
- Require explicit `{ confirm: true }` option when setting destructive parameters
- Throw `ConfigurationCC_ConfirmationRequired` error for unconfirmed destructive operations
- Add comprehensive unit and integration tests for validation behavior

Implements #8137


AlCalzone:
> * Require explicit `{ confirm: true }` option when setting destructive parameters

I understand why you did this, but this is a breaking API change. I'd prefer if we just add the flag as an information to applications and let them decide how to proceed with destructive parameters.

AlCalzone (approved):

PR #8181, opened 2025-09-17
fix: assume Security CC support during S0 node replacement

### LINKED ISSUE #8037: S0 bootstrapping is not performed when replacing a failed node
### Discussed in https://github.com/zwave-js/zwave-js/discussions/8036

<sup>Originally posted by **memphi2** August  4, 2025</sup>
### Checklist

### Describe the issue

What is happening?
I am currently unable to (re)include any devices with S0 security. This affects various Fibaro devices (FGS213, FGS223, etc.), which I am certain should support this feature. The log is contradictory here:

First:
[Node 009] Security S0 bootstrapping failed, the node was not granted the S0 security class: Node 9 does not support the Command Class Security! (ZW0302)
But then:
[Node 009]   supports CC Security (0x98) in version 1

Include /replace node with S0 security

Anything else we should know?

### Software versions

Driver (zwave-js): 10.9.0

Z-Wave JS UI: 15.9.0

Home Assistant Z-Wave JS Addon: 4.7.0

ZAC93 SDK 7.23.2

### Device information

Manufacturer: Fibaro
Model name: FGS213
Node ID: 009

### Checklist

### Upload Logfile

[zwavejs_2025-08-04.log](https://github.com/user-attachments/files/21584938/zwavejs_2025-08-04.log)
[zwave-js-ui-store.zip](https://github.com/user-attachments/files/21584854/zwave-js-ui-store.zip)
</div>

davidsgullick:
fix(controller): assume Security CC support during S0 node replacement

When replacing a failed node with S0 security, the controller doesn't receive
a NIF (Node Information Frame) to determine which Command Classes are supported.
This causes the secureBootstrapS0 method to fail because it can't verify that the Security CC is supported.

Setting the assumeSupported parameter to true during the secureBootstrapS0 method during node replacement enables successful node replacement. I think nodes that don't support this will gracefully fail afterwards anyway

Closes https://github.com/zwave-js/zwave-js/issues/8037


AlCalzone (changes requested):
This change looks like it would fix the linked issue, but the tests don't really add much value I think. Let's remove the tests.

AlCalzone (dismissed):

AlCalzone:
Turns out we also needed to change the other assumption here. Figuring that out required an actual device though.

AlCalzone (approved):

PR #8187, opened 2025-09-23
fix(config): correct parameter size for factory reset of Shelly Wave devices

janbajc:
Adjusted the factory_reset parameter to use the proper value size so that the configuration aligns with device specifications

### DIFF
--- a/packages/config/config/devices/0x0460/templates/wave_template.json
+++ b/packages/config/config/devices/0x0460/templates/wave_template.json
@@ -195,7 +195,7 @@
 	"factory_reset": {
 		"label": "Factory Reset",
 		"description": "Reset to factory default settings and remove from the Z-Wave network.",
-		"valueSize": 4,
+		"valueSize": 1,
 		"defaultValue": 0,
 		"allowManualEntry": false,
 		"options": [


AlCalzone:
I'm going to need confirmation from @QubinoHelp for this. This change affects the template and by extension all of these device config files:
```
qpsw-0A1X16.json
qnsn-0D24X.json
qmem-0A1PC16.json
qmsw-0A1X8.json
qmsw-0A1P8.json
qndw-002C.json
qnsn-0A24X.json
qnpl-001X16.json
qlmo-003F.json
qnpl-0A112.json
qnsw-001X16.json
qnsw-001P16.json
qlht-OU2Z.json
qnsw-002P16.json
qpsw-0A2X16.json
qpsw-0A1P16.json
qpsw-0A3X16.json
qnpl-001X12.json
qpsw-0A2P16.json
qpdm-0A2P01.json
```
Is the size of the factory reset parameter wrong in all of them?

janbajc:
Yes. The factory reset parameter size used to be defined as 4 bytes, but in later firmware releases (and newer device models) it was changed to 1. @QubinoHelp will confirm.

AlCalzone:
Oh. I just saw your name. Nevermind the confirmation.

If this changed after certain firmware versions, we should probably have two versions of this and use [conditional parameters](https://zwave-js.github.io/zwave-js/#/config-files/conditional-settings) to apply the correct one for each firmware version

QubinoHelp:
@AlCalzone That's correct the initial idea was to use 4 bytes which was implemented maybe in one or two first firmwares, later on this was changed to 1 byte.

QubinoHelp:
> Oh. I just saw your name. Nevermind the confirmation.
> 
> If this changed after certain firmware versions, we should probably have two versions of this and use [conditional parameters](https://zwave-js.github.io/zwave-js/#/config-files/conditional-settings) to apply the correct one for each firmware version

I will have to check if any of the firmware has that and will start to do the conditional parameters for that and also because we will start to add the detached mode and other features in the newer firmware updates.

janbajc:
If I may, I’d suggest we proceed with implementing this change for now. It correctly covers the majority of our devices, most released firmware versions, and all upcoming ones. This would serve as a practical interim solution until we implement conditional parameters.

AlCalzone (approved):
Fair enough

PR #8193, opened 2025-09-25
feat: add API to enable more frequent background RSSI measurements

### LINKED ISSUE #8186: Idea: High frequency background RSSI measurements (temporary)
We should consider adding an API to temporarily enable higher-frequency background RSSI measurements. This could help with finding a good location for a controller.

djwmarcx:
**Why**

closes: https://github.com/zwave-js/zwave-js/issues/8186

**What**

This PR introduces a new High-Frequency RSSI Polling Mode. In this mode, background RSSI measurements are performed at shorter, configurable intervals.

This mode can be enabled or disabled via an API call. It is temporary by default to avoid accidentally leaving it enabled.

Additionally, i've added a test file (actually, is not really a test, but a tool) to monitor Background RSSI in realtime, leveraging also the new HF Mode. 

https://github.com/user-attachments/assets/e4c85a3f-82fe-404c-99c7-f71ce50f8e5a

**Summary**:
* Added High-Frequency RSSI Polling Mode, which reduces the polling time to 2 seconds
* New API to enable/disable HF RSSI mode, with configurable duration.
* New rssi-monitor terminal tool -> yarn ts test/rssi-monitor.ts [serial-port]

Would you mind taking a general look @AlCalzone??
* Do I need to generate the docs manually? 
* In general, I'm completely new to the project, please tell me if something seems off!

Thank You!

### DIFF
--- a/packages/zwave-js/src/lib/driver/Driver.ts
+++ b/packages/zwave-js/src/lib/driver/Driver.ts
@@ -781,8 +781,6 @@ export interface DriverEventCallbacks extends PrefixedNodeEvents {
 	"firmware update finished": (
 		result: OTWFirmwareUpdateResult,
 	) => void;
-	"rssi hf status": (enabled: boolean, timeout?: number) => void;
-
 	error: (err: Error) => void;
 }
 
@@ -9347,8 +9345,6 @@ integrity: ${update.integrity}`;
 			);
 		}
 
-		this.emit("rssi hf status", true, timeoutMs);
-
 		this.poolBackgroundRSSIIHFTimeoutMs = timeoutMs;
 		this.handleQueueIdleChange(this.queueIdle);
 
@@ -9361,8 +9357,6 @@ integrity: ${update.integrity}`;
 	 * This will restore the default behavior of polling the background RSSI only once every 30 seconds.
 	 */
 	public disableBackgroundRSSIHFMode(): void {
-		this.emit("rssi hf status", false);
-
 		this.clearBackgroundRSSIHFTimer();
 		this.poolBackgroundRSSIIHFTimeoutMs = undefined;
 	}
--- a/packages/zwave-js/src/lib/driver/Driver.ts
+++ b/packages/zwave-js/src/lib/driver/Driver.ts
@@ -5602,12 +5602,12 @@ ${handlers.length} left`,
 			) {
 				// The message we're currently sending is still in progress but expects this message in response.
 				// Remember the message there.
-					this.controllerLog.logNode(msg.getNodeId()!, {
-						message:
+				this.controllerLog.logNode(msg.getNodeId()!, {
+					message:
 						`received expected response prematurely, remembering it...`,
-						level: "verbose",
-						direction: "inbound",
-					});
+					level: "verbose",
+					direction: "inbound",
+				});
 				currentMessage.prematureNodeUpdate = msg;
 				return;
 			}
@@ -9229,6 +9229,12 @@ integrity: ${update.integrity}`;
 	public enableFrequentRSSIMonitoring(
 		durationMs: number,
 	): void {
+		if (durationMs < 10000 || durationMs > 3600 * 1000) {
+			throw new ZWaveError(
+				`The duration must be between 10 seconds and one hour!`,
+				ZWaveErrorCodes.Argument_Invalid,
+			);
+		}
 		this.hfBackgroundRSSIEndTimestamp = Date.now() + durationMs;
 		// Restart a running timer to poll as soon as possible
 		if (this.pollBackgroundRSSITimer) {
--- a/packages/zwave-js/src/lib/driver/Driver.ts
+++ b/packages/zwave-js/src/lib/driver/Driver.ts
@@ -9226,6 +9226,7 @@ integrity: ${update.integrity}`;
 		this.pollBackgroundRSSITimer = undefined;
 	}
 
+	/** Enable frequent RSSI monitoring for the given amount of milliseconds. During this time, the background RSSI will be measured every 2 seconds. */
 	public enableFrequentRSSIMonitoring(
 		durationMs: number,
 	): void {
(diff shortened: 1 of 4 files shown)


packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -316,6 +316,8 @@ const defaultOptions: ZWaveOptions = {
 		sendDataAbort: 20000, // If a controller takes over 20 seconds to reach a node, it's probably not going to happen
 		sendDataCallback: 30000, // INS13954 defines this to be 65000 ms, but waiting that long causes issues with reporting devices
 		sendToSleep: 250, // The default should be enough time for applications to react to devices waking up
+		pollBackgroundRSSIMin: 5000, // 5 seconds
+		pollBackgroundRSSIMax: 30000, // 30 seconds
```
AlCalzone: I don't think we need to make this configurable. Let's just use a fixed timeout of 1 or 2 seconds for now.
If the triggering is still done by the background task, it will automatically get interrupted if there is any other activity, so I'm not worried about congestion in any way.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -748,6 +781,8 @@ export interface DriverEventCallbacks extends PrefixedNodeEvents {
 	"firmware update finished": (
 		result: OTWFirmwareUpdateResult,
 	) => void;
+	"rssi hf status": (enabled: boolean, timeout?: number) => void;
+
```
AlCalzone: I also don't think we need this.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -9174,40 +9210,163 @@ integrity: ${update.integrity}`;
...
-						await this.controller.getBackgroundRSSI();
-					} catch {
-						// ignore errors
-					}
-				}, timeout).unref();
+				// If high frequency mode is enabled, use HF timeout instead
+				if (this.poolBackgroundRSSIIHFTimeoutMs !== undefined) {
+					timeoutMs = this.poolBackgroundRSSIIHFTimeoutMs;
+				}
+				this.setBackgroundRSSITimer(timeoutMs);
```
AlCalzone: I'd keep the original logic, except for using the fixed HF timeout instead of the calculation we do now.

AlCalzone (changes requested):
Thanks a lot! I'd actually keep this a bit simpler with a fixed interval, see comments below.

As for the utility - just keep it where it is, that's perfect.

AlCalzone:
Side note: Background noises between -80 and -90 dBm are very high for 800 series controllers like the one in your video. Not sure if you did that for demonstration purposes, but if not, I'd look into that. This high-frequency RSSI polling feature seems ideal for such an analysis :D

packages/zwave-js/src/lib/driver/Driver.ts
djwmarcx: Just to clarify — your suggestion is to remove the varible HF interval entirely, correct? So:

* The API call to activate HF would only require durationMs to automatically disable it.
* Also, remove the min/max options from the timeout options and always use 2 seconds.

If that’s the case, is keeping the regular polling interval at 30 seconds still OK, as it was before?

djwmarcx:
> Side note: Background noises between -80 and -90 dBm are very high for 800 series controllers like the one in your video. Not sure if you did that for demonstration purposes, but if not, I'd look into that. This high-frequency RSSI polling feature seems ideal for such an analysis :D

Apart for the "foil paper trick" you see in the video to artificially reduce the noise, everything else is just regular noise. After your comment, I was moving the antenna within the 1-meter cable range... it only reaches 100 if I place it on the floor. Which levels are considered normal?

packages/zwave-js/src/lib/driver/Driver.ts
djwmarcx: I went ahead and performed these changes, hopefully it's what you requested :)

AlCalzone: That was the idea, yes.
Keep the default polling interval so it triggers 5s after the queues are idle, then every 30s.
While the high frequency polling is active, trigger every 2 seconds.

AlCalzone:
> After your comment, I was moving the antenna within the 1-meter cable range... it only reaches 100 if I place it on the floor. Which levels are considered normal?

800 series has a sensitivity of around -110 dBm, so I'd consider -100..-110 good.

packages/zwave-js/src/lib/driver/Driver.ts
djwmarcx: Great. We are almost there.

How important is to keep those 5 seconds? Would it be possible to reduce it to 2 seconds? Due to the Math.max logic we have in place, if we set that specific value to 5, anything below that will always be raised to 5. So the HF interval can't be lower than that.

I can try to make some changes, but I don't want to complicate things unnecessarily. 

Actually, I don't know at first how to track which poll iteration is the first. Each iteration changes the state to busy and then idle again, which initiates the next round, right? So all iterations are "the first".

I'm missing something maybe? What's your opinion?

Thank you!

packages/zwave-js/src/lib/driver/Driver.ts
djwmarcx: removed

packages/zwave-js/src/lib/driver/Driver.ts
AlCalzone: > Each iteration changes the state to busy and then idle again, which initiates the next round, right? So all iterations are "the first".

Yes, each iteration triggers itself by queuing a command, after which the queue may go to idle.

> Due to the Math.max logic we have in place, if we set that specific value to 5, anything below that will always be raised to 5

We can just skip the Math.max if the HF mode is active.

djwmarcx: Thanks!

djwmarcx:
I think it's done. I need to review the documentation to determine if I need to add an explanation for this feature.

AlCalzone:
Thanks. A short mention in the docs would be great - doesn't have to be long.

The rest looks good - I'll play with it a bit before merging.

AlCalzone (approved):
Thanks again. I made a few simplifications to the timer logic and added documentation.

PR #8199, opened 2025-09-26
fix(config): add fingerprint `0x0811:0x23a9` to "Kwikset HC620"

dreysolano:
Add missing fingerprint for HC620 lock. I updated my config locally in Z-Wave JS UI and the locks now display accurate manufacturer-related data as expected.

### DIFF
--- a/packages/config/config/devices/0x0090/hc620.json
+++ b/packages/config/config/devices/0x0090/hc620.json
@@ -21,6 +21,11 @@
 		},
 		{
 			"productType": "0x0811",
+			"productId": "0x23a9",
+			"zwaveAllianceId": 4184
+		},
+		{
+			"productType": "0x0811",
 			"productId": "0x00a9",
 			"zwaveAllianceId": 5046
 		}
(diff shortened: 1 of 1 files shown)


dreysolano:
@AlCalzone This is a very small, almost identical change to #4173, just a different `productId`

packages/config/config/devices/0x0090/hc620.json
```diff
@@ -19,6 +19,11 @@
 			"productId": "0x23a8",
 			"zwaveAllianceId": 4184
 		},
+		{
+			"productType": "0x0811",
+			"productId": "0x23a9",
+			"zwaveAllianceId": 4184
```
AlCalzone:
```suggestion
			"productId": "0x23a9"
```

AlCalzone (changes requested):

AlCalzone (approved):

packages/config/config/devices/0x0090/hc620.json
dreysolano: @AlCalzone for my education, why was the `zwaveAllianceId` removed from the new entry? Because it's already referenced on the prior `0x23a8` entry?

AlCalzone: Yes. The Z-Wave Alliance website only lists a single set of fingerprints per products, so there cannot be two fingerprints with the same `zwaveAllianceId`.

PR #8207, opened 2025-09-30
feat(config): add productId 0x0083 to Shelly Wave Shutter

nberlee:
After beta update https://github.com/QubinoHelp/Shelly_Wave_FW_OTA/commit/ceff5ca86e17f1d651bb7c2a1e793d733612aa1e the productId switches to 83

### DIFF
--- a/packages/config/config/devices/0x0460/qnsh-001P10.json
+++ b/packages/config/config/devices/0x0460/qnsh-001P10.json
@@ -10,6 +10,10 @@
 		},
 		{
 			"productType": "0x0003",
+			"productId": "0x0083"
+		},
+		{
+			"productType": "0x0003",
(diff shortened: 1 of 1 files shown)


nberlee:
kindly mention so  @QubinoHelp can review if this actually was intended by the beta

QubinoHelp:
Hi @nberlee,

Thank you for your effort and for notifying us.

Unfortunately, the new firmware for the Wave Shutters was built using an updated parameter numbering. Because of this, the previous numbering is no longer valid. We will therefore need to prepare and release a new configuration file that matches the updated Shutter firmware.

nberlee:
ok clear. Merging it in the current is therefor not an option

PR #8339, opened 2025-10-07
chore: bump `jsonl-db` dependency

AlCalzone:
This transitively updates `alcalzone-shared` so bundlers no longer complain about some ESM weirdness.
Also, it gets rid of `fs-extra`, which hopefully helps with that one pesky memory leak one user was having.


AlCalzone:
@zwave-js-bot automerge

PR #8340, opened 2025-10-07
chore: update release-script dependency

AlCalzone:
This transitively updates the last remaining instance of `@alcalzone/shared-utils` to a version that's compatible with ESM.


AlCalzone:
@zwave-js-bot automerge

PR #8343, opened 2025-10-08
fix: remove browser-unsafe serial export

AlCalzone:
This export was not needed and caused warnings from bundlers.


AlCalzone:
@zwave-js-bot automerge

PR #8344, opened 2025-10-08
feat: mock secure inclusion of nodes and fix S0 auto-inclusion

### LINKED ISSUE #8346: Door Lock included with no security when using default inclusion strategy
### Discussed in https://github.com/zwave-js/zwave-js/discussions/8205

<sup>Originally posted by **spudwebb** September 30, 2025</sup>
### Checklist

### Describe the issue

When including a Kwikset Door Lock that supports S0 but not S2 using  the default inclusion strategy, the lock is added with no security at all which makes it unusable because the useful CC are not discovered.

I was expecting ZWaveJS to try to include it using S0 as its specific device class is "Secure Keypad Door Lock" which requires security.

From my quick analysis of the log file and ZWave JS code, it looks like ZWave JS doesn't know it is a "Secure Keypad Door Lock" until too late, thus it does not try to add it using S0.  Is that possible? Is it fixable?

Note that if I use `strategy=S0` or` strategy=Default` with `force security` checked, it works as expected.  But most users don't know they have to use that, and since the inclusion with no security does not report any error they are confused.

### Software versions

Driver (zwave-js): 15.12.0

Z-Wave JS UI:  11.2.1

HomeSeer smartstick  G3

### Device information

Manufacturer: Kwikset
Model name: SmartCode 888
Node ID: 178

### Checklist

### Upload Logfile

[zwavejs_2025-09-30_doorlock_S0.log](https://github.com/user-attachments/files/22624651/zwavejs_2025-09-30_doorlock_S0.log)
</div>

AlCalzone:
This PR allows us to run tests for secure (S0) inclusion of devices. This made it possible to reproduce the underlying issue of #8346

Fixes: #8346

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/packages/cc/src/cc/SecurityCC.ts
+++ b/packages/cc/src/cc/SecurityCC.ts
@@ -919,6 +919,17 @@ export class SecurityCCSchemeReport extends SecurityCC {
 @CCCommand(SecurityCommand.SchemeGet)
 @expectedCCResponse(SecurityCCSchemeReport)
 export class SecurityCCSchemeGet extends SecurityCC {
+	public static from(
+		raw: CCRaw,
+		ctx: CCParsingContext,
+	): SecurityCCSchemeGet {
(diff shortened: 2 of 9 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #8357, opened 2025-10-15
fix(config): add parameters 3, 40 and 84 for Enbrighten 55258, firmware version 5.51

jascoproducts:
Added parameters 3, 40 and 84 for firmware version 5.51

### DIFF
--- a/packages/config/config/devices/0x0063/55258_zw4002.json
+++ b/packages/config/config/devices/0x0063/55258_zw4002.json
@@ -32,8 +32,33 @@
 	},
 	"paramInformation": [
 		{
+			"#": "3",
+			"$import": "~/templates/master_template.json#led_indicator_four_options"
+		},
+		{
 			"#": "4",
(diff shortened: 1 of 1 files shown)


AlCalzone (approved):
Thanks. Can you please approve https://github.com/zwave-js/firmware-updates/pull/216 too if the changes are okay?

packages/config/config/devices/0x0063/55258_zw4002.json
```diff
@@ -31,9 +31,34 @@
 		}
 	},
 	"paramInformation": [
+		{
+			"#": "3",
+			"$import": "~/templates/master_template.json#led_indicator_four_options"
```
steverep, line 36: @jascoproducts you used the 4-option template here which does not match your documentation in https://github.com/jascoproducts/firmware/issues/50#issuecomment-3406906633.  I can fix this in #8359 if you can confirm the 3-option inverted template is actually the correct one.

PR #8359, opened 2025-10-16
fix(config): add version conditionals and correct parameters for Enbrighten 55258

steverep:
1. Corrects #8357 to add conditionals on firmware version (@jascoproducts please correct me if I'm wrong here)
2. Fixes broken link to manual
3. I also believe the LED indicator template is incorrect, so I'll hold this as draft until that is confirmed

### DIFF
--- a/packages/config/config/devices/0x0063/55258_zw4002.json
+++ b/packages/config/config/devices/0x0063/55258_zw4002.json
@@ -33,6 +33,7 @@
 	"paramInformation": [
 		{
 			"#": "3",
+			"$if": "firmwareVersion >= 5.51",
 			"$import": "~/templates/master_template.json#led_indicator_four_options"
 		},
 		{
@@ -41,6 +42,7 @@
 		},
 		{
 			"#": "40",
+			"$if": "firmwareVersion >= 5.51",
 			"label": "Fan Speed Control",
 			"valueSize": 1,
 			"defaultValue": 0,
@@ -58,6 +60,7 @@
 		},
 		{
 			"#": "84",
+			"$if": "firmwareVersion >= 5.51",
 			"$import": "templates/jasco_template.json#factory_default"
 		}
 	],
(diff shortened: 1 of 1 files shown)


jascoproducts:
@steverep this is cool, didn't know you could add conditional params by firmware version. Current LED indicator template is correct; we had forgotten to put valid value 3 in the parameters table when we [replied to you yesterday](https://github.com/jascoproducts/firmware/issues/50#issuecomment-3406906633). We've updated that comment

steverep:
Thanks for confirming.  Note the template is also inverted for values 0 and 1 from your comment, so may want to adjust that as well.

jascoproducts:
@steverep actually yeah, upon reviewing the documentation it looks like the original comment was correct. Parameter 3 is just a bit different for this specific device. Can you revise the config to match the original reply? We edited it back.

steverep:
Yes I can fix here.  Your original comment matches the `led_indicator_three_options_inverted` template, except that the default is unconventionally 1 instead of 0.

@AlCalzone can I use that template and then override the default, or should I just create a new template?

jascoproducts:
@steverep yes, valid values 1 and 0 on this device are flipped from the template. This is going to be the only device like this so no template necessary, I think.

packages/config/config/devices/0x0063/55258_zw4002.json
```diff
@@ -33,14 +33,17 @@
 	"paramInformation": [
 		{
 			"#": "3",
-			"$import": "~/templates/master_template.json#led_indicator_four_options"
+			"$if": "firmwareVersion >= 5.51",
+			"$import": "~/templates/master_template.json#led_indicator_three_options_inverted",
+			"defaultValue": 1
```
steverep, line 38: I assume here the import is inline and normal JSON rules are applicable, but LMK

steverep:
Parameter 19 shows up via interview but is missing from the config, so I also added that copied from the 58446 (700 series model).

packages/config/config/devices/0x0063/55258_zw4002.json
steverep: NM, I see this is done all over the place now 👍🏻

AlCalzone:
@steverep is this ready to review?

AlCalzone (approved):
Looks good. Let me know if you are planning additional changes or if I should merge.

steverep:
Done with this one. You can merge.

PR #8366, opened 2025-10-20
fix(config): correct user code slot bit field for BE468ZP/BE469/BE469ZP

AlCalzone:
fixes: https://github.com/zwave-js/zwave-js/discussions/8360


AlCalzone:
@zwave-js-bot automerge

PR #8367, opened 2025-10-20
chore: fix "author config file" copilot prompt

AlCalzone:
The tool name definitions in VSCode were updated. This PR fixes that, uses a newer model and makes sure the MCP server is built during repo `bootstrap`.


AlCalzone:
@zwave-js-bot automerge

PR #8382, opened 2025-10-24
fix: rename driver option for auto config param setting

AlCalzone:
`applyRecommendedConfigParamValues` feels a bit more natural than `applyConfigurationRecommendedValues`.


AlCalzone:
@zwave-js-bot automerge

PR #8393, opened 2025-10-29
feat(config): add Shelly Wave Shutter, update Wave 1, Wave 1PM, Wave 2PM, Wave Plug S

QubinoHelp:
added Wave Shutter with new parameters,
added the parameter for Detached Mode and Remote Reboot for Wave 1, Wave 1PM, Wave 2PM,
added Wave Plug s Outdoor config


QubinoHelp:
Error: Parameter #76 is invalid: min/maxValue is incompatible with valueSize 1 (min = -128, max = 127).
 Don't have this specified in the code, don't know where is the problem.
 
 Error: Parameter #79 is invalid: The option value 0 must be in the range 3...50!
  In this parameter there's the range from 3 - 50 selectable and the 0 the values 1 and 2 are not valid entries
 
 Error: Parameter #81 is invalid: The option value 65000 must be in the range 1...32000!
  In this parameter similar to the above the range selectable is from 1 - 32000, and 65000 the in between values are not valid.

packages/config/config/devices/0x0460/templates/wave_template.json
```diff
@@ -360,6 +360,74 @@
...
+		]
+	},
+		"motor_limit_detection": {
+		"label": "Motor Limit Detection",
+		"description": "Power threshold used to detect a limit switch. Increase this value if the device falsely detects limit switches.",
+		"valueSize": 1,
+		"minValue": 0,
+		"maxValue": 255,
+		"defaultValue": 1,
+		"options": [
```
AlCalzone, line 391:
```suggestion
		"unsigned": true,
		"options": [
```
should fix the error about invalid min/max value

packages/config/config/devices/0x0460/templates/wave_template.json
```diff
@@ -360,6 +360,74 @@
...
+			{
+				"label": "Auto calibration",
+				"value": 1
+			}
+		]
+	},
+		"delay_limit_detection": {
+		"label": "Limit Detection Delay Time",
+		"description": "Defines the delay before motor power is read after a relay switches ON",
+		"unit": "0,1 sec",
```
AlCalzone:
```suggestion
		"unit": "0.1 seconds",
```

packages/config/config/devices/0x0460/templates/wave_template.json
```diff
@@ -360,6 +360,74 @@
...
+			},
+			{
+				"label": "Auto calibration",
+				"value": 1
+			}
+		]
+	},
+		"delay_limit_detection": {
+		"label": "Limit Detection Delay Time",
+		"description": "Defines the delay before motor power is read after a relay switches ON",
```
AlCalzone:
```suggestion
		"description": "Defines the delay before motor power is read after a relay switches ON. Allowable range: 0, 3-50.",
```

packages/config/config/devices/0x0460/templates/wave_template.json
```diff
@@ -360,6 +360,74 @@
...
+		"unit": "0,1 sec",
+		"valueSize": 1,
+		"minValue": 3,
```
AlCalzone:
```suggestion
		"minValue": 0,
```
We currently have no good way to declare gaps in the valid value range. As a workaround, we use the description field (see comment above) to explain those.

packages/config/config/devices/0x0460/templates/wave_template.json
```diff
@@ -360,6 +360,74 @@
...
+		"options": [
+			{
+				"label": "Automatic time",
+				"value": 0
+			}
+		]
+	},
+		"max_moving_time": {
+		"label": "Shutter Max Moving Time",
+		"description": "Defines the maximum time the motor can move before stopping",
```
AlCalzone:
```suggestion
		"description": "Defines the maximum time the motor can move before stopping. Allowable range: 1-32000, 65000",
```

packages/config/config/devices/0x0460/templates/wave_template.json
```diff
@@ -360,6 +360,74 @@
...
+		"unit": "seconds",
+		"valueSize": 2,
+		"minValue": 1,
+		"maxValue": 32000,
+		"defaultValue": 120,
```
AlCalzone, line 424: This parameter may also need

```suggestion
		"defaultValue": 120,
		"unsigned": true,
```
if the checks complain about incompatible min/max value after the change.

AlCalzone (changes requested):
Please also run `yarn fmt` when you're done to fix the indentation in the template file.

AlCalzone (approved):

PR #8405, opened 2025-11-03
fix(config): update parameters for Shelly Wave 1PM firmware 13.02

reneboer:
With firmware 13.02 the Shelly Wave 1PM has the new parameter 7 (Detached Mode) and they removed parameters 91, 92, 93 and 94. In the proposed template the changes are implemented conditionally so it will work with older firmware versions correctly.

Details from Shelly can be found at https://github.com/QubinoHelp/Shelly_Wave_FW_OTA/tree/main/Wave_1PM.

### DIFF
--- a/packages/config/config/devices/0x0460/qnsw-001P16.json
+++ b/packages/config/config/devices/0x0460/qnsw-001P16.json
@@ -43,7 +43,7 @@
 		},
 		{
 			"#": "17",
-			"$import": "templates/wave_template.json#detached_mode",
+			"$import": "~/templates/master_template.json#state_after_power_failure_prev_off",
 			"label": "O1: State After Power Failure"
 		},
 		{
--- a/packages/config/config/devices/0x0460/qnsw-001P16.json
+++ b/packages/config/config/devices/0x0460/qnsw-001P16.json
@@ -36,6 +36,12 @@
 			"label": "SW1 Switch Type"
 		},
 		{
+			"#": "7",
+			"$if": "firmwareVersion > 13.0",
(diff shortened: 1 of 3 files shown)


packages/config/config/devices/0x0460/qnsw-001P16.json
```diff
@@ -35,9 +35,15 @@
...
+			"#": "7",
+			"$if": "firmwareVersion > 13.0",
+			"$import": "templates/wave_template.json#detached_mode",
+			"label": "Detach O1 Output"
+		},
 		{
 			"#": "17",
-			"$import": "~/templates/master_template.json#state_after_power_failure_prev_off",
+			"$import": "templates/wave_template.json#detached_mode",
 			"label": "O1: State After Power Failure"
```
AlCalzone, line 47: Something is wrong here. Either the label or the import?

AlCalzone (changes requested):

packages/config/config/devices/0x0460/qnsw-001P16.json
reneboer: Hi, That change on line 46 is in error. Parameter #17 should not be changed, I missed that in my test sorry. I corrected the line to its original value.

reneboer (commented):
The proposed import uses a better definition for the detached mode option values.

reneboer:
Creating new request without confusion on error on parameter 17.

AlCalzone:
I don't think closing was needed. All commits are going to be squashed anyways.

PR #8435, opened 2025-11-16
fix(zooz_template): change auto timer default to 0 for ZEN04 Smart Plug

marleyjaffe:
Updated to match values for ZEN04 Smart Plug 
[ ZEN04 Smart Plug Advanced Settings ](https://www.support.getzooz.com/kb/article/1114-zen04-smart-plug-advanced-settings/)

Not sure if the previous defaults are accurate. If they are then maybe this should be a product specific change.

### DIFF
--- a/packages/config/config/devices/0x027a/templates/zooz_template.json
+++ b/packages/config/config/devices/0x027a/templates/zooz_template.json
@@ -142,9 +142,9 @@
 	"auto_timer_duration_base": {
 		"valueSize": 4,
 		"unit": "minutes",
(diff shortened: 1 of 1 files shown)


AlCalzone (changes requested):
The derived `auto_off_timer_duration` and `auto_on_timer_duration` templates are also used by ZEN21, ZEN22, ... ZEN27 and at least for the ZEN21 and 22 it is accurate - I didn't check all the others:
grafik

So this should probably be a device-specific fix.

PR #8436, opened 2025-11-16
feat(cc): implement `refreshValues` for Window Covering CC

vexofp:
Previously, the "Refresh" button for Window Coverings was a no-op. This PR implements `WindowCoveringCC.refreshValues()` so that we actually query the device to refresh the values.

Tested with an iblinds v3 window covering.


AlCalzone (changes requested):
Commonly if we have a refreshValues implementation, we try to use it in the `interview` method. You can take a look at some other CCs for reference.

Please remove the existing querying in the `interview` method in favor of a call to `refreshValues`.

AlCalzone (approved):
OK with my changes

PR #8450, opened 2025-11-25
chore: fix commit hooks

AlCalzone:
Turns out they weren't actually in use...


AlCalzone:
@zwave-js-bot automerge

PR #8451, opened 2025-11-25
chore: run Oxlint on precommit

AlCalzone:
This is fast enough so it doesn't interrupt the workflow


AlCalzone:
@zwave-js-bot automerge

PR #8459, opened 2025-11-26
chore: update CC exports generator to void every single statement

AlCalzone:
Copilot has been complaining about this in pull requests. I don't mind either way so let's make it shut up.


AlCalzone:
@zwave-js-bot automerge

PR #8460, opened 2025-11-27
feat(cc): implement Active Schedule CC

AlCalzone:


packages/cc/src/lib/_Types.ts
```diff
@@ -11,6 +11,91 @@ import {
...
+	YearDayScheduleReport = 0x08,
+	DailyRepeatingScheduleSet = 0x09,
+	DailyRepeatingScheduleGet = 0x0a,
+	DailyRepeatingScheduleReport = 0x0b,
+}
+
+export enum ActiveScheduleSetAction {
+	Erase,
+	Modify,
+}
```
AlCalzone, line 31: when defining enums from spec values, be explicit: `Erase = 0x00`, `Modify = 0x01`. Apply throughout.

packages/testing/src/CCSpecificCapabilities.ts
```diff
@@ -24,6 +24,17 @@ import type {
 	MaybeUnknown,
 } from "@zwave-js/core";
 
+export interface ActiveScheduleCCCapabilities {
+	targets: Map<
+		CommandClasses | number,
```
AlCalzone: The `| number` seems to make no sense.

packages/zwave-js/src/lib/node/mockCCBehaviors/ActiveSchedule.ts
```diff
@@ -0,0 +1,462 @@
...
+
+function eraseSchedules(
+	self: MockNode,
+	capabilities: ActiveScheduleCCCapabilities,
+	kind: ActiveScheduleScheduleKind,
+	targetCC: number,
+	targetId: number,
+): void {
+	// CC:00A4.01.06.11.007: A Schedule Slot ID value of zero during an Erase
+	// Set Action signifies a batch erase operation.
```
AlCalzone: The full spec text for this requirement is
> A Schedule Slot ID value of zero during an Erase Set Action signifies a batch erase operation. Assuming
that the target is otherwise valid, the operation MUST execute based on the target values outlined in
Table 2.18

These comments MUST include the actual requirement, which is the sentence containing MUST/SHALL NOT/SHOULD/...

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2272 @@
...
+			this.commandOptions,
+		);
+
+		return result?.enabled;
+	}
+
+	@validateArgs()
+	public async setYearDaySchedule(
+		slot: ActiveScheduleSlotId,
+		schedule?: ActiveScheduleYearDaySchedule,
```
AlCalzone: On signatures like these we should be explicit. Check similar CCAPIs for consistency, whether they split set/delete, or have a combined signature like this one.
If we keep it combined, we should "require" the caller to pass `undefined` to delete a schedule.

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2272 @@
...
+	}
+
+	/**
+	 * Requests the year day schedule stored in the given slot.
+	 * A slot ID of 0 requests the first occupied slot for the given target.
+	 *
+	 * @returns The schedule together with the slot it is actually stored in
+	 * and the ID of the next occupied slot (0 if there is none),
+	 * `false` if the requested slot is empty,
+	 * or `undefined` if the node did not respond.
```
AlCalzone, line 546: Do we use this pattern _anywhere_ else? This feels unusual.

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2272 @@
...
+				};
+			}
+			return false;
+		}
+	}
+
+	@validateArgs({ strictEnums: true })
+	public async setDailyRepeatingSchedule(
+		slot: ActiveScheduleSlotId,
+		schedule?: ActiveScheduleDailyRepeatingSchedule,
```
AlCalzone: See previous comment on split/combined set/delete.

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2272 @@
...
+	}
+
+	/**
+	 * Requests the daily repeating schedule stored in the given slot.
+	 * A slot ID of 0 requests the first occupied slot for the given target.
+	 *
+	 * @returns The schedule together with the slot it is actually stored in
+	 * and the ID of the next occupied slot (0 if there is none),
+	 * `false` if the requested slot is empty,
+	 * or `undefined` if the node did not respond.
```
AlCalzone, line 671: See previous comment on this pattern.

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2272 @@
...
+			const logLines: string[] = [
+				"received supported target CCs:",
+			];
+			for (const [targetCC, cap] of caps.targetCapabilities) {
+				logLines.push(
+					`  ${
+						getEnumMemberName(CommandClasses, targetCC)
+					}: ${cap.numSupportedTargets} targets, ${cap.numYearDaySlotsPerTarget} year day slots, ${cap.numDailyRepeatingSlotsPerTarget} daily repeating slots`,
+				);
+			}
```
AlCalzone, line 766: We should format this similar to the Schedule Entry Lock CC logs.

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2272 @@
...
+				endpoint: this.endpointIndex,
+				message: logLines.join("\n"),
+				direction: "inbound",
+			});
+
+			// Per the figure "Active Schedule Command Class Interview", the
+			// schedules of Target CCs with exactly one Target are interviewed
+			// now with Target ID 1. Schedules of Target CCs with more than one
+			// Target are interviewed during their respective Target CC
+			// interview.
```
AlCalzone: Far too verbose.

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2272 @@
...
+		// receiving node.
+		validatePayload(numTargetCCs >= 1);
+
+		const targetCapabilities = new Map<
+			CommandClasses,
+			Omit<ActiveScheduleTargetCapabilities, "targetCC">
+		>();
+		let offset = 1;
+
+		// Parse target CCs (can be extended CCs, 1 or 2 bytes each)
```
AlCalzone: Unnecessary comment, this is obvious.

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2272 @@
...
+		for (let i = 0; i < numTargetCCs; i++) {
+			const { ccId, bytesRead } = parseCCId(raw.payload, offset);
+			targetCCs.push(ccId);
+			offset += bytesRead;
+		}
+
+		// After the CCs, we expect 3 arrays of 16-bit values (6 bytes per target CC)
+		validatePayload(raw.payload.length >= offset + numTargetCCs * 6);
+
+		// Parse numSupportedTargets for all target CCs
```
AlCalzone: These comments are okay - Z-Wave isn't always consistent in how it orders fields (struct-of-arrays vs array-of-structs style).

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2272 @@
...
+				ActiveScheduleCCValues.dailyRepeatingSchedule(
+					targetCC,
+					targetId,
+					slotId,
+				).endpoint(endpoint.index),
+			);
+	}
+}
+
+// =============================================================================
```
AlCalzone: Get rid of these dividers.

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2272 @@
...
+			offset += encodeDate(this.startDate!, this.payload, offset);
+			offset += encodeDate(this.stopDate!, this.payload, offset);
+			this.payload[offset++] = metadataLength & 0b111;
+			if (this.metadata) {
+				this.payload.set(this.metadata, offset);
+			}
+			offset += metadataLength;
+		}
+
+		this.payload = this.payload.subarray(0, offset);
```
AlCalzone, line 1672: This is not necessary, is it? We calculate the length in advance.

AlCalzone: If there are other instances of this pattern, fix them too.

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2272 @@
...
+	// CC:00A4.01.00.11.018: Schedule Metadata MUST be less than or equal to
+	// seven (7) bytes in length.
+	if (metadata && metadata.length > 7) {
+		throw new ZWaveError(
+			"The schedule metadata must be at most 7 bytes long",
+			ZWaveErrorCodes.Argument_Invalid,
+		);
+	}
+}
+
```
AlCalzone, line 198: Don't some of these helpers already exist in shared modules?

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2272 @@
...
+export const ActiveScheduleCCValues = V.defineCCValues(
+	CommandClasses["Active Schedule"],
+	{
+		...V.staticProperty("supportedTargetCCs", undefined, {
+			internal: true,
+		}),
+		...V.dynamicPropertyAndKeyWithName(
+			"targetCapabilities",
+			"targetCapabilities",
```
AlCalzone, line 248: I'm not 100% certain right now but if the name of the value is the same as its "key", `dynamicPropertyAndKey` should allow us to avoid this repetition.

packages/cc/src/lib/_Types.ts
AlCalzone: Done, also for `ActiveScheduleScheduleKind`.

packages/testing/src/CCSpecificCapabilities.ts
AlCalzone: Removed. The map is now typed with the shared `ActiveScheduleTargetCapabilities` type.

packages/zwave-js/src/lib/node/mockCCBehaviors/ActiveSchedule.ts
AlCalzone: Done. Also verified all other CC:/CL: citations on the branch against the spec and fixed 12 more truncated or paraphrased quotes.

packages/cc/src/cc/ActiveScheduleCC.ts
AlCalzone: Schedule Entry Lock CC uses the same combined signature (`setWeekDaySchedule(slot, schedule?)`). Kept it combined, but the parameter is now `ActiveScheduleYearDaySchedule | undefined` instead of optional, so erasing requires passing `undefined` explicitly.

packages/cc/src/cc/ActiveScheduleCC.ts
AlCalzone: Schedule Entry Lock uses the same pattern for its cached getters (`MaybeNotKnown<ScheduleEntryLockWeekDaySchedule | false>`). Its API getters return `undefined` for both empty slot and missing response, but here that would be ambiguous: a slot-0 query on a target without schedules returns a valid "empty" report, which callers need to tell apart from a timeout. Kept.

packages/cc/src/cc/ActiveScheduleCC.ts
AlCalzone: Same as for the year day variant: kept combined, `schedule` now requires an explicit `undefined` to erase.

packages/cc/src/cc/ActiveScheduleCC.ts
AlCalzone: See the year day getter: `false` matches the cached-value convention from Schedule Entry Lock and distinguishes an empty slot from a missing response.

packages/cc/src/cc/ActiveScheduleCC.ts
AlCalzone: Done, now formatted like the Schedule Entry Lock interview logs.

packages/cc/src/cc/ActiveScheduleCC.ts
AlCalzone: Shortened.

packages/cc/src/cc/ActiveScheduleCC.ts
AlCalzone: Removed, along with its `Write target CCs` counterpart in serialize().

packages/cc/src/cc/ActiveScheduleCC.ts
AlCalzone: Removed.

packages/cc/src/cc/ActiveScheduleCC.ts
AlCalzone: It was needed because the CC id encodes as 1 or 2 bytes, but the sizes are now computed exactly via a new `ccIdLength()` helper in core (also reused by `encodeCCList`), so all the trailing trims are gone.

AlCalzone: All instances in this file are gone — every serialize() now allocates the exact payload size.

packages/cc/src/cc/ActiveScheduleCC.ts
AlCalzone: No — Schedule Entry Lock uses 1-byte year offsets and Time Parameters includes seconds, so no shared codec exists for this 6-byte date format. Only `formatDate`/`formatTime` exist in shared, and they are already used.

packages/cc/src/cc/ActiveScheduleCC.ts
AlCalzone: There is no `dynamicPropertyAndKey`, only the `...WithName` variants, so the repetition has to stay.

packages/cc/src/cc/ActiveScheduleCC.ts
AlCalzone: Update: reverted the exact-size refactor as discussed — the max-size allocation + `subarray` trim stays, now with a comment explaining that the final length depends on whether the CC id encodes as 1 or 2 bytes.

packages/cc/src/lib/_Types.ts
```diff
@@ -1284,19 +1368,11 @@ export interface ScheduleEntryLockSlotId {
...
-	Sunday = 0x00,
-	Monday = 0x01,
-	Tuesday = 0x02,
-	Wednesday = 0x03,
-	Thursday = 0x04,
-	Friday = 0x05,
-	Saturday = 0x06,
-}
+/** @deprecated Use {@link ScheduleWeekday} instead */
+export import ScheduleEntryLockWeekday = ScheduleWeekday;
```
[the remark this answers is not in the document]
AlCalzone: We should just use the new ScheduleWeekday enum throughout, mark this a breaking change and target the `next` branch.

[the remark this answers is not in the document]
AlCalzone: Fixed in 48a0e21fc — removed the deprecated alias entirely and migrated all usages to `ScheduleWeekday`.

packages/cc/src/cc/ActiveScheduleCC.ts
```diff
@@ -0,0 +1,2248 @@
...
+			ScheduleWeekday.Sunday,
+		);
+		const { timespan, bytesRead: timespanBytes } = parseTimespan(
+			raw.payload,
+			offset,
+		);
+		offset += timespanBytes;
+		validatePayload(raw.payload.length >= offset + 1);
+		// CC:00A4.01.06.11.009: These bits MUST be set to 0 by a sending node
+		// and MUST be ignored by a receiving node.
```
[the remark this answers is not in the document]
AlCalzone: Fixed in 48a0e21fc — corrected the reference to CC:00A4.01.09.11.009.

AlCalzone:
@zwave-js-bot automerge

PR #8495, opened 2025-12-13
feat(config): add Nexa ZPR-111 parameters

FredrikFornstad:
Adding parameters for Nexa ZPR-111 smart switch. Note that Nexa have published a configuration file on their webpage, but it is incorrect (as of Dec 13th, 2025). [Nexa ZPR-111](https://nexa.se/smarta-hem/z-wave/zpr-111-2)

The correct parameters are the same as 0x0258/nas-wr01ze. However, since Nexa specifies that ZPR-111 can only handle 2300 W, I have set the default values for power cut-off at 11A and alarm at 10A, just to be sure.

ZPR-111 only reports voltage when the relay is on (reports 0 Volt when the relay is off). I guess that is a feature...

### DIFF
--- a/packages/config/config/devices/0x0268/zpr111.json
+++ b/packages/config/config/devices/0x0268/zpr111.json
@@ -13,6 +13,82 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
+	"paramInformation": [
+		{
+			"#": "1",
+			"$import": "~/templates/master_template.json#state_after_power_failure_off_prev"
+		},
+		{
+			"#": "2",
+			"$import": "~/templates/master_template.json#base_enable_disable",
+			"label": "Button",
+			"defaultValue": 1
+		},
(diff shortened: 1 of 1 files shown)


FredrikFornstad:
@AlCalzone would you mind review this pull request?

With the above parameters the device works as it should (apart from having the same bugs as the Shenzhen Neo plug including the "random" toggle of the most significant bit in the meter value being reported).

For anyone interested: I have contacted Nexa AB in Sweden and asked them to correct their faulty configuration information on their webpage and received a response more than a month ago that they were to look into it. Obviously someone have mixed up the configuration parameters between the nas-wr01z and the nas-wr01ze device, but I get a feeling the people at Nexa AB do not care.

AlCalzone (approved):
Thanks - I just added a small thing we implemented since you posted this PR.

PR #8496, opened 2025-12-15
fix: ignore `ApplicationUpdateRequestNodeAdded` during inclusion

AlCalzone:
For some reason this command can be received while we're including a node. In that case we need to ignore it or it gets mis-interpreted as a secondary controller notifying us of the inclusion, causing security bootstrapping to fail


AlCalzone:
@zwave-js-bot automerge

PR #8512, opened 2025-12-31
feat(config): hide VZW32-SN test parameter from production use

The00Dustin:
This paramater was previously marked read-only.  This change depends on #8492.


AlCalzone (commented):
Is this dependent on a specific firmware version or is this all pre-release testing anyways?

The00Dustin:
There is not yet any firmware with this feature confirmed working as expected.  A new PR with `$if` can be submitted once a firmware version exists where this feature is confirmed functional.

AlCalzone (approved):

PR #8521, opened 2026-01-02
fix(config): label endpoints of Shelly Wave Shutter

AlCalzone:
as discussed in https://github.com/zwave-js/zwave-js-ui/issues/3881#issuecomment-3672146787


AlCalzone:
@zwave-js-bot automerge

PR #8525, opened 2026-01-02
fix: clear user codes on initial interview unless `queryAllUserCodes` driver option is set

AlCalzone:
During certification testing it was found that the current behavior is not compliant. During the initial interview, we're supposed to either delete all codes, or query them (if the checksum isn't 0).

Now we do.

### DIFF
--- a/packages/cc/src/cc/UserCodeCC.ts
+++ b/packages/cc/src/cc/UserCodeCC.ts
@@ -940,9 +940,10 @@ export class UserCodeCC extends CommandClass {
 		}
 
 		// Synchronize user codes and settings
-		if (ctx.getInterviewOptions()?.queryAllUserCodes) {
-			await this.refreshValues(ctx);
-		}
+		await this.refreshValues(
+			ctx,
+			ctx.getInterviewOptions()?.queryAllUserCodes ?? false,
+		);
 
 		// Remember that the interview is complete
 		this.setInterviewComplete(ctx, true);
@@ -950,6 +951,17 @@ export class UserCodeCC extends CommandClass {
 
 	public async refreshValues(
 		ctx: RefreshValuesContext,
+	): Promise<void>;
+
+	/** @internal */
+	public async refreshValues(
+		ctx: RefreshValuesContext,
+		queryAllUserCodes: boolean,
+	): Promise<void>;
+
+	public async refreshValues(
+		ctx: RefreshValuesContext,
+		queryAllUserCodes: boolean = false,
 	): Promise<void> {
 		const node = this.getNode(ctx)!;
 		const endpoint = this.getEndpoint(ctx)!;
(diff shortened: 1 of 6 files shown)


AlCalzone:
@zwave-js-bot automerge

claudiucismaru-vsp:
Hello,

From my understanding in the AWG document, this is not a MUST or MUST NOT (according to their definitions) requirement. For me it sounds more like a recommendation. From the text I understand that the controlling node should not automatically delete any user code in the interview process, unless is in the initial interview (and not in other re interviewings). It doesn't mention that the controlling node MUST do this operation, but instead is preventing it happening. 

As a result should I understand that Zwave JS took the decision to delete all the users at the interview time? Is it important if we can rely on this specific behavior for the application development.

AlCalzone:
This change was prompted by a recent test run with the Certification Test Tool (CTT), which flagged the old behavior as non-compliant.

We used to not query any codes, but you could opt in.
We had some complaints in the past that including a v1 lock took very long and put huge strain on the battery, due to the sheer amount of commands that were being exchanged. Which is why querying was disabled by default, but again this is not compliant.

---

This section in the specs is worded a bit strangely, but I understand it differently from you - and that's how the CTT has the tests implemented too. Keep in mind that these requirements are exceptions to the mandatory interview procedure that's shown in Figure 6.27: User Code Command Class interview.
This procedure requires us to query all user codes by default.

> CL:0063.01.21.03.1 For a node controlling version 1:
> It is OPTIONAL to send a User Code Get Command for every User Identifier to any supporting
node if the controlling node issues a User Code Set Command (User ID = 0, User ID Status = 0) to delete the user codes below User ID 256.

--> When controlling a v1 node, the controlling node may skip querying all user codes during the initial interview if it instead deletes all user codes.

> CL:0063.01.21.02.3 For a node controlling version 2 or newer:
> It is OPTIONAL to send an Extended User Code Get Command for every User Identifier to a
node supporting version 2 or newer if:
– The controlling node requested the checksum and it is set to 0, or
– The controlling node issues an Extended User Code Set Command (User ID = 0, User ID
Status = 0) to delete all user codes, or
– The supporting node reports that no more User Identifiers are set in the Extended User
Code Report Command with the Next User Identifier field.

--> When controlling a v2 node, we have more options. Namely we can abort querying early if we find that there are no more user codes to follow - or if the checksum tells us there are no user codes at all. However there is a possibility that we would need to query thousands of user codes (up to 65535 to be exact).
Again the alternative to this is deleting all codes - which makes sense from a consistency POV.

---

If there is interest in querying codes instead, I can consider making this configurable again.

claudiucismaru-vsp:
Hey,

Thanks for the clarifications. Indeed, I missed the MUST part of the interview diagram. I though that one is a recommended flow, not a mandatory one. Now it makes a lot of sense.

The idea is that was pretty complicated from the application perspective to deal with non queried user code in case you sent a duplicate one as it could have pointed to a user id that you don't know anything about it. If you start with an empty lock or you do a all user query, things are more simple to implement.

Querying the lock instead of wiping it out is a use case for us, but wiping the lock at inclusion time it's a trade of for generating additional network traffic / consuming batteries as we have systems in place to repopulate the lock with the correct codes.

All I wanted to know is I can rely on the wiping out the lock at the inclusion time as the default behavior.

PR #8554, opened 2026-01-15
fix: avoid false-positive in `hasDeviceConfigChanged`

AlCalzone:
This PR fixes a false positive for `hasDeviceConfigChanged` after updating from a version < 15.18.


AlCalzone:
@zwave-js-bot automerge

PR #8555, opened 2026-01-15
fix: handle `allowed` field in device config hash comparison

AlCalzone:
Followup to #8547 that prevents the internal `min/maxValue` to `allowed` normalization to falsely trigger the device config changed diagnostic.


AlCalzone:
@zwave-js-bot automerge

PR #8556, opened 2026-01-15
feat(config): complete device config for Contec/Da Vinci Touch Panel (FW 10.10)

contec4ever:
Summary
This PR updates the device config for Contec / Da Vinci Touch Panel (FW 10.10)
(IDs: 0x025d / 0x0200 / 0x0100).

Why
The existing “Da Vinci Switch Default” config is incomplete (exposes only a partial set of parameters).
This PR adds the missing configuration parameters and improves parameter metadata based on the manufacturer manual FW 10.10.

What’s included
- Configuration parameters up to no. 61 (labels, ranges/defaults, dropdown options where relevant).
- Association group labels for root and endpoints aligned with the manual tables.

Tested
Validated on real hardware via local override and re-interview in Z-Wave JS UI.


contec4ever:
All required checks are passing (lint, lint-zwave, tests).

Evidence:
- Current default DB entry shows only “Configuration Parameters (29)” (see screenshot).
- Manufacturer manual (FW 10.10) documents additional parameters (e.g., no 57–61 short press times) (see screenshot).

Validated on real hardware via local override + re-interview in Z-Wave JS UI.
IDs: 0x025d / 0x0200 / 0x0100
NEW_PARAMETERS
OLD_PARAMETERS

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -1,8 +1,8 @@
 {
-	"manufacturer": "Avadesign Technology Co.",
+	"manufacturer": "Contec / Da Vinci",
 	"manufacturerId": "0x025d",
-	"label": "Da Vinci Switch Default",
-	"description": "",
+	"label": "Da Vinci Touch Panel (v10.10)",
+	"description": "Da Vinci Touch Panel",
```
AlCalzone: We shouldn't mention the firmware version in the device label if possible.
Also repeating the name in the label and description isn't ideal. How about this (not perfect, but avoids some repetition)?


```suggestion
	"label": "Da Vinci Touch",
	"description": "Wall Controller",
```

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
 		"min": "0.0",
 		"max": "255.255"
```
AlCalzone, line 14: Are any of the config parameters firmware dependent? Or can we assume they all exist in all devices out in the wild?

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
...
 	},
+	"endpoints": {
+		"1": {
+			"label": "Endpoint 1 (Button 1)",
+			"associationGroups": {
```
AlCalzone: Is this device Z-Wave Plus? If so it may not be necessary to define all association groups here.

AlCalzone: If we need to, the correct property name would be
```suggestion
			"associations": {
```

and the same for the other endpoints.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
...
+					"label": "Multilevel report (SWITCH_MULTILEVEL_REPORT)",
+					"maxNodes": 5
+				}
+			}
+		}
+	},
 	"paramInformation": [
 		{
-			"#": "1",
+			"#": 1,
```
AlCalzone: Although parameters are numbers, we encode them as strings

```suggestion
			"#": "1",
```

...and the same for all other parameters.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
...
+				},
+				{
+					"label": "Reserved",
+					"value": 7
+				},
+				{
+					"label": "Dimmer",
+					"value": 8
+				}
+			]
```
AlCalzone: It looks like this parameter is the same for all endpoints. We can cut down on the repetition using local `import`s by importing the definition of parameter 1

```suggestion
			"$import": "#paramInformation/1",
			"label": "EP 2 Type"
```

Again, same for all other "EP ... Type" parameters.

And the same principle applies to all the other repeated parameters like Partner ID

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
...
 		},
 		{
-			"#": "3",
+			"#": 3,
 			"label": "EP 1 Blind Duration",
 			"valueSize": 1,
 			"minValue": 0,
 			"maxValue": 127,
 			"defaultValue": 0
 		},
```
AlCalzone: This parameter (and all other similar ones) should have a `unit`

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
...
+				},
+				{
+					"label": "",
+					"value": 6
+				},
+				{
+					"label": "Max (7)",
+					"value": 7
+				}
+			]
```
AlCalzone:
```suggestion
			"minValue": 0,
			"maxValue": 3
```

Same for the OFF param

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
...
+		},
+		{
+			"#": 48,
+			"label": "Device Toggle or Pulse Output",
```
AlCalzone:
```suggestion
			"label": "Output Mode",
```
makes more sense as a label IMO

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
...
+				{
+					"label": "Seconds",
+					"value": 40
+				},
+				{
+					"label": "Always on",
+					"value": 255
+				}
+			]
+		},
```
AlCalzone: Are no other values than 0, 10, 20, 30, 40 and 255 allowed or can the intermediate values be configured too?

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
...
+				},
+				{
+					"label": "Level 6",
+					"value": 6
+				},
+				{
+					"label": "Least sensitive",
+					"value": 7
+				}
+			]
```
AlCalzone: We shouldn't have `allowManualEntry: false` or options here. Just set minValue/maxValue accordingly and mention in the description that 0 is most sensitive and 7 is least sensitive.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
...
+				},
+				{
+					"label": "3.3 s",
+					"value": 33
+				},
+				{
+					"label": "4.4 s",
+					"value": 44
+				}
+			]
```
AlCalzone:
```suggestion
			"unit": "0.1s",
			"options": [
				{
					"label": "Always long",
					"value": 0
				}
			]
```

and remove the unit from the parameter label.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
...
+		},
+		{
+			"#": 53,
+			"label": "Button 1 Short Press Time (10 Ms Units)",
```
AlCalzone: See previous parameter

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
...
+				},
+				{
+					"label": "1000 ms",
+					"value": 100
+				}
+			]
+		},
+		{
+			"#": 54,
+			"label": "Button 2 Short Press Time (10 Ms Units)",
```
AlCalzone: Import the Button 1 definition and update the label like I showed above. Same for the other following parameters.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -13,238 +13,2758 @@
...
+					"value": 90
+				},
+				{
+					"label": "1000 ms",
+					"value": 100
+				}
+			]
+		}
+	],
+	"associationGroups": {
```
AlCalzone:
```suggestion
	"associations": {
```

AlCalzone (changes requested):
Thanks a lot! We need to make some modifications though.

contec4ever:
Thanks for the review — addressed the comments in a batch update.

Changes:
- Updated device label/description (removed FW from label, reduced repetition).
- Removed endpoint association definitions (Z-Wave Plus / AGI should cover them).
- Encoded parameter numbers as strings.
- Added/normalized units (e.g. seconds, 0.1s, 0.01s) and removed units from parameter labels.
- Updated params 49/50 per suggestions (49 supports intermediate values via min/max; 50 uses min/max + description instead of options).
- Renamed param 48 label to “Output Mode”.
- Kept repeated parameters explicit for now. I attempted local `$import` but it caused “multiple definitions” errors with the array-based `paramInformation` format. Happy to refactor using the preferred `$import` style if there’s a recommended working example for this format.

All required checks are passing.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: I built this config based on the vendor manual for FW 10.10 and validated it on real hardware running FW 10.10. I’m not aware of firmware-dependent differences in the parameter set, but I haven’t been able to verify older firmware revisions in the field.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: Applied in the latest commit — removed the FW version from the device label and reduced repetition (label: “Da Vinci Touch”, description: “Wall Controller”). Thanks!

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: Yes, it’s Z-Wave Plus. To avoid duplicating what the device reports via AGI, I removed the explicit endpoint association definitions from the config. Thanks for pointing out the correct property name (`associations`) as well.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: Applied in the latest commit — all parameter numbers are encoded as strings now (e.g. "#": "1"). Thanks!

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: I tried refactoring the repeated EP parameters using local `$import`, but with the current array-based `paramInformation` format this caused “multiple definitions / malformed config” errors during linting. To keep the config valid and CI green, I kept the repeated parameter definitions explicit for now.

If there is a recommended working example of `$import` for array-based `paramInformation` in this repo (or the exact `$import` syntax you expect here), I’m happy to refactor the repeated EP Type/Partner/… params accordingly in a follow-up.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: Applied — added units to Blind Duration and the other time-based parameters (using the standardized units, e.g. `seconds`, `0.1s`, `0.01s` as applicable).

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: Applied — switched the LED intensity parameters to plain min/max ranges (0–7) without predefined options (same for the OFF intensity parameter). Thanks!

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: Applied — removed options/allowManualEntry for Touch Sensitivity and replaced with a simple range (min 0 / max 7) plus a description noting 0 is most sensitive and 7 is least sensitive. Thanks!

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: Applied — renamed the parameter label to “Output Mode”. Thanks!

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: Intermediate values are supported. According to the manual, 10–40 is a seconds range (0 = always off, 255 = always on). I updated the parameter to keep the full min/max range and kept the options list only for the common values (0/10/20/30/40/255).

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: Applied — moved the time scale into the `unit` field and removed the unit from the parameter label for Long Press Time (and the related timing parameters as well). Thanks!

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: Applied — handled the Short Press Time parameters the same way as the previous one (moved the time scale into `unit` and removed the unit from the label). Thanks!

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: I kept the short-press time parameters explicit for now. I attempted local `$import` refactors in this array-based `paramInformation` format, but it caused “multiple definitions / malformed config” lint errors. Happy to refactor these (and the other repeated params) using the preferred `$import` approach if there’s a recommended working example for this format in the repo.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: Applied via a different approach: since it’s Z-Wave Plus, I removed the explicit endpoint association definitions entirely to avoid duplicating what the device reports via AGI. Thanks for pointing out that the correct property name would be `associations` if we ever need to define them here.

contec4ever:
Reminder,
I wanted to check when our file will be reviewed?
Thanks

AlCalzone:
Hi, I haven't forgotten, but I was on vacation last week and I'm still catching up.

contec4ever:
Hi,

Thanks for the update


Best regards,
***@***.***
GIORA KOHAVI
C.T.O
***@***.******@***.***>
CONTEC intelligent housing
Office: + 972-4-8555522 / Cel: +972-54-7700067
17 Hamesila st'Nesher, ISRAEL
http://www.contec-touch.com




מאת: AlCalzone ***@***.***>
נשלח: יום ג 27 ינואר 2026 16:26
אל: zwave-js/zwave-js‏ ***@***.***>
עותק: giora kohavi‏ ***@***.***>; Author‏ ***@***.***>
נושא: Re: [zwave-js/zwave-js] feat(config): complete configuration parameters and association groups for Contec/Da Vinci Touch Panel (FW 10.10) (PR #8556)

[https://avatars.githubusercontent.com/u/17641229?s=20&v=4]AlCalzone left a comment (zwave-js/zwave-js#8556)<https://github.com/zwave-js/zwave-js/pull/8556#issuecomment-3805521930>

Hi, I haven't forgotten, but I was on vacation last week and I'm still catching up.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -42,209 +272,493 @@
...
+					"value": 22
+				},
+				{
+					"label": "3.3 s",
+					"value": 33
+				},
+				{
+					"label": "4.4 s",
+					"value": 44
+				}
```
AlCalzone: We shouldn't list those explicit values. What's the maximum for this parameter?

packages/config/config/devices/0x025d/da_vinci_switch_default.json
```diff
@@ -42,209 +272,493 @@
...
+				},
+				{
+					"label": "900 ms",
+					"value": 90
+				},
+				{
+					"label": "1000 ms",
+					"value": 100
+				}
+			]
```
AlCalzone: We shouldn't list those explicit values. What's the maximum for this parameter?

AlCalzone (changes requested):
I've done the refactoring to use local imports for you. Make sure to pull the branch first if you're working on it locally.

There are now just 2 parameters left we need to clean up, but I'm not sure what the correct definition is, see below:

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: yes,  list those explicit values,
The maximum for this parameter is 1000 ms

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: yes, list those explicit values,
The  maximum for this parameter 4.4 s

AlCalzone: Help me understand why they need to be listed. We typically don't do that here, unless they are the only values that are allowed to be set. Given that the default of 20 (2.0s) is not in there, this does not seem to be the case.

My suggestion would be to have "Always long" be a predefined option and simply set 44 (4.4s) as the maximum.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
AlCalzone: Help me understand why they need to be listed. We typically don't do that here, unless they are the only values that are allowed to be set. Is that the case or can users set any value between 0 and 100?

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: understood. Our initial intent with the explicit list was to provide a few common presets that are easy to pick in the UI and reduce user error when configuring short vs long press timings (we’ve seen integrators pick values that unintentionally invert the relationship). That said, I agree the config DB should primarily reflect the device’s documented capabilities, not enforce UX guardrails.
Per the v10.10 manual this parameter is a continuous range: 0 is a special value, otherwise 1–250 in 0.1s units (max 25.0s). We updated the config to reflect that with minValue: 0, maxValue: 250, unit: 0.1s, and unsigned: true. We also removed the explicit preset list and kept only a single predefined option for the special value (0 = "Always long"). Any additional “friendly presets” or cross-field guardrails will be implemented in our installer/UI instead of the config DB.

packages/config/config/devices/0x025d/da_vinci_switch_default.json
contec4ever: Same reasoning : the explicit list was meant as “recommended presets” for integrators/end users to avoid getting lost in too many choices and to reduce common misconfiguration (e.g choosing a long-press time shorter than the short-press time). However, I agree that the device config DB shouldn’t try to solve that UX problem by constraining the valid values ,it should describe what the device supports.
According to the v10.10 manual these short-press timing parameters are also a continuous range: 0 is a special value, otherwise 1–250 in 0.01s units (max 2.5s). We updated the baseline definition accordingly (minValue: 0, maxValue: 250, unit: 0.01s, unsigned: true) and refactored the per-button parameters to reuse that definition via local imports. We removed the long list of discrete options and kept only a single predefined option for the special value (0 = "Disabled"). Any preset “recommended values” and validations (e.g warning if long < short) will be handled in our installer/UI.

AlCalzone (approved):

AlCalzone:
Thanks again!

PR #8600, opened 2026-02-06
feat: allow annotating the semantic purpose of config parameters

### LINKED ISSUE #8440: Idea: semantically tag config parameters
That would allow Z-Wave JS to detect what a config parameter means and make suggestions, like increasing the reporting interval if there are lots of devices in the network which only support time based reporting

AlCalzone:
fixes: #8440


AlCalzone:
@zwave-js-bot automerge

PR #8607, opened 2026-02-11
feat: support gaps in value range of arbitrary CC values

### LINKED ISSUE #8448: Unable to set wakeup interval to 0 using the `wakeUpInterval` value.
Requires adding some mechanism to define gaps in the allowed value range.

AlCalzone:
Certain CC values, especially the wakeup interval, have gaps in their value ranges (0 = off, plus a stepped range). Until now this was not modeled properly. This PR adds support for a new `allowed` property for `ValueMetadata`, which is modeled after the solution for config parameters introduced in #8547.

For backwards compatibility, the `min` and `max` are still provided and define the envelope of all allowed values. This value range may have gaps of disallowed values though.

Applications should prefer the new `allowed` property where it exists if they need to do client-side validation.

### DIFF
--- a/docs/api/valueid.md
+++ b/docs/api/valueid.md
@@ -88,6 +88,7 @@ interface ValueMetadataNumeric extends ValueMetadataAny {
 	min?: number;
 	max?: number;
 	steps?: number;
+	allowed?: readonly AllowedValue[];
 	default?: number;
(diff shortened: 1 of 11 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #8624, opened 2026-02-18
fix: serialization of name and location in NodeNamingAndLocationCC

dfederm:
The `serialize()` methods in `NodeNamingAndLocationCCNameSet` and `NodeNamingAndLocationCCLocationSet` wrote the name/location buffer at offset 0, overwriting the encoding byte (Char. Presentation field). The Z-Wave spec defines byte 0 as the encoding indicator and bytes 1–16 as the character data.

The corresponding `from()` parsers already correctly read the encoding from `payload[0]` and the text from `payload.subarray(1)`, so any round-trip through serialize -> parse would lose the encoding byte and corrupt the first character.

Fixed by writing the name/location buffer at offset 1 in both methods.


AlCalzone (approved):
Thanks. The way the payload length was determined was also wrong, so I took this chance to correct it too.

PR #8625, opened 2026-02-18
fix(thermostat): correct operating state report by masking reserved bits

dfederm:
`ThermostatOperatingStateCCReport.from()` assigned the full first payload byte as the operating state. The Z-Wave spec defines only bits 0–3 as the Operating State field; bits 4–7 are reserved. If a device sets any reserved bits, the parsed state would be an invalid enum value.

Fixed by masking the byte with `& 0x0f`.

### DIFF
--- a/packages/cc/src/cc/ThermostatOperatingStateCC.ts
+++ b/packages/cc/src/cc/ThermostatOperatingStateCC.ts
@@ -185,7 +185,7 @@ export class ThermostatOperatingStateCCReport
 		ctx: CCParsingContext,
 	): ThermostatOperatingStateCCReport {
 		validatePayload(raw.payload.length >= 1);
-		const state: ThermostatOperatingState = raw.payload[0];
+		const state: ThermostatOperatingState = raw.payload[0] & 0x0f;
 
 		return new this({
 			nodeId: ctx.sourceNodeId,


AlCalzone (changes requested):
Version 2 of the CC defines the full byte as the operating state.
image

An argument could be made for discarding commands with an operating state >= `0x0B` but not for the proposed change.

dfederm:
Ah true it does explicitly say they should be discarded, not just ignore the extra bits. Also not totally sure where I got 0xf from, need to check which spec I'm looking at (it's harder to find the definitive docs than I'd like).

Not sure it's worth it then, but it really depends on precedent in the rest of the repo. Do you typically do checks for valid values, or do you just hope for the best (forward compat?)?

AlCalzone:
Technically that is a certification requirement so it does make sense to check.

I'm not sure though if it's actually tested or enforced at the moment.

dfederm:
Another wrinkle:
https://github.com/zwave-js/zwave-js/blob/e221ba55038238b67db51f3afd14c7f682ab75fe/packages/cc/src/lib/_Types.ts#L1459-L1467

So V2 for this CC isn't implemented. So should the v2 values be allowed to be returned?

dfederm:
Also, foudn where I got `0x0f`. V1 has:

image

(Weirdly, I also see a reference to "Table 138" but can't see to find said table)

So _technically_ this change is correct for V1, and only V1 is currently implemented.

However, I'm not super convinced this masking is actually helpful in light of V2 existing (don't know how I didn't see V2 at all before).

Probably the most correct behavior is to diverge behavior for V1 and V2. For V1, the mask is required (reserved area "must be ignored by a receiving node") and then validate against the valid V1 values and if not in that set, drop the report entirely. And then for V2, check against the V2 values and also drop if not in that set. And then also implement the other V2 commands =P

dfederm:
Closing in favor of #8630

PR #8627, opened 2026-02-19
test: extend Window Covering mocks, support transitioning

AlCalzone:
This PR adds better support for mocking Window Covering CC and adds generic transition support to Window Covering CC and Multilevel Switch CC mocks.


AlCalzone:
@zwave-js-bot automerge

PR #8630, opened 2026-02-20
feat(cc): implement Thermostat Operating State CC V2

dfederm:
Implements V2 of the Thermostat Operating State CC

This includes these 4 new V2 commands:
- **Logging Supported Get/Report**: Bitmask-based query of which operating state logs the device supports
- **Logging Get/Report**: Bitmask-based request for specific operating state logs which include per-state usage data (hours/minutes for today and yesterday) in 5-byte entries with a "reports to follow" field

This change also includes a small bug fix for the V1 Report. See #8625 for the full discussion, but the gist is that v1 defines the upper nibble as reserved and it must be ignored, so for V1 the value is masked to ensure that ignoring. I also added value validation since the spec says reserved values must be ignored.

There's also a small fix around `if (state)` -> `if (state != undefined)` so that `Idle` (0x00) state is logged correctly

Added unit tests for *all* commands (serialization, deserialization, reserved bit masking)

Disclosure: This change was heavily assisted by AI, although I did make manual tweaks as needed and thoroughly reviewed it myself before asking others to do so.


packages/cc/src/cc/ThermostatOperatingStateCC.ts
```diff
@@ -193,7 +357,33 @@ export class ThermostatOperatingStateCCReport
...
+			validatePayload.withReason(
+				`Invalid V1 operating state ${num2hex(this.state)}`,
+			)(VALID_V1_STATES.has(this.state));
+		} else {
+			// V2 spec: Operating State is 8 bits,
+			// reserved values MUST be ignored
+			validatePayload.withReason(
+				`Invalid V2 operating state ${num2hex(this.state)}`,
+			)(VALID_V2_STATES.has(this.state));
+		}
```
AlCalzone: I don't know where it says that in the specs - it is somewhere obscure, but:
CCs in Z-Wave are generally defined to be forwards compatible (with 1-2 notable exceptions). The specs for a CC version 2 define how a device implementing version 2 has to encode and interpret commands it receives and sends, no matter the version of the other party.

This means we don't have to do version-specific validation here. For us (implementing version 2), checking that `this.state <= ThermostatOperatingState["3rd Stage Aux Heat"]` while **parsing** is sufficient.
Yes, there could be a possibility that a v1 device sends a report that would be invalid in version 1, but that also means that device is not compliant.

packages/cc/src/cc/ThermostatOperatingStateCC.ts
```diff
@@ -112,16 +206,51 @@ export class ThermostatOperatingStateCC extends CommandClass {
...
+				const logMessage =
+					`received supported operating state logging types: ${
+						supportedLoggingTypes
+							.map((t) =>
+								getEnumMemberName(
+									ThermostatOperatingState,
+									t,
+								)
+							)
+							.join(", ")
```
AlCalzone: This may be long, let's break it into individual lines with a middot as a prefix. See this example:
```ts
				const logMessage = "received supported sensor types:\n"
					+ sensorTypes
						.map((t) => getSensorName(t))
						.map((name) => `· ${name}`)
						.join("\n");
				ctx.logNode(node.id, {
					endpoint: this.endpointIndex,
					message: logMessage,
					direction: "inbound",
				});
```

packages/cc/src/cc/ThermostatOperatingStateCC.ts
```diff
@@ -100,6 +148,52 @@ export class ThermostatOperatingStateCCAPI extends PhysicalCCAPI {
 		);
 		return response?.state;
 	}
+
+	public async getLoggingSupported(): Promise<
```
AlCalzone: We typically switch the order of words in these API names to be more natural and align them with the realted type names:
```
getSupportedLoggingTypes
```

packages/cc/src/cc/ThermostatOperatingStateCC.ts
```diff
@@ -51,9 +59,46 @@ export const ThermostatOperatingStateCCValues = V.defineCCValues(
...
+		...V.dynamicPropertyAndKeyWithName(
+			"operatingStateLogging",
+			"operatingStateLogging",
+			(state: ThermostatOperatingState) => state,
+			({ property, propertyKey }) =>
+				property === "operatingStateLogging"
+				&& typeof propertyKey === "number",
+			undefined,
+			{ internal: true, minVersion: 2 },
+		),
```
AlCalzone: I don't think this data should be persisted at all. Lets keep this purely an API that communicates with the device.

packages/cc/src/cc/ThermostatOperatingStateCC.ts
```diff
@@ -158,6 +286,44 @@ export class ThermostatOperatingStateCC extends CommandClass {
...
+					[...supportedLoggingTypes],
+				);
+				if (loggingData) {
+					ctx.logNode(node.id, {
+						endpoint: this.endpointIndex,
+						message:
+							`received thermostat operating state logging data for ${loggingData.length} state(s)`,
+						direction: "inbound",
+					});
+				}
```
AlCalzone: Likewise, I don't think querying the operation history during the interview is needed. Applications can query this on demand using the API

packages/cc/src/cc/ThermostatOperatingStateCC.ts
```diff
@@ -208,3 +398,229 @@ export class ThermostatOperatingStateCCReport
...
+
+		return new this({
+			nodeId: ctx.sourceNodeId,
+			reportsToFollow,
+			loggingData,
+		});
+	}
+
+	public readonly reportsToFollow: number;
+	public readonly loggingData: ThermostatOperatingStateLoggingData[];
```
AlCalzone: I think this should be a `Map<ThermostatOperatingState, ...>` similar to how we do it for ConfigurationCCBulkReport

packages/cc/src/cc/ThermostatOperatingStateCC.ts
```diff
@@ -208,3 +398,229 @@ export class ThermostatOperatingStateCCReport
...
+		}
+
+		return new this({
+			nodeId: ctx.sourceNodeId,
+			reportsToFollow,
+			loggingData,
+		});
+	}
+
+	public readonly reportsToFollow: number;
```
AlCalzone: Since this CC is using the reportsToFollow pattern for long reports, we should also add functionality to wait for them and merge them. You can see an example in AssociationCCReport. Keywords: getPartialCCSessionId / expectMoreMessages / mergePartialCCs

AlCalzone (changes requested):
Thanks - I didn't realize we hadn't implemented version 2. I've got a few remarks.

AlCalzone (approved):

AlCalzone:
Thanks - took me a while to get back to this 😇

PR #8642, opened 2026-02-26
feat(config): standardize "Control Associated Device" triggers across 7x series Zooz switches

gallagherrchris:
Added "Association Reports" parameter to all 7 series Zooz switches.
This also exposes all values that are possible for this parameter.

Confirm documentation:
https://www.support.getzooz.com/kb/article/535-zen71-on-off-switch-advanced-settings/
https://www.support.getzooz.com/kb/article/538-zen72-dimmer-switch-advanced-settings/
https://www.support.getzooz.com/kb/article/639-zen73-on-off-toggle-switch-700-advanced-settings/
https://www.support.getzooz.com/kb/article/648-zen74-toggle-dimmer-700-advanced-settings/
https://www.support.getzooz.com/kb/article/1942-zen75-heavy-duty-switch-advanced-settings/
https://www.support.getzooz.com/kb/article/549-zen76-s2-on-off-switch-advanced-settings/
https://www.support.getzooz.com/kb/article/550-zen77-s2-dimmer-advanced-settings/


packages/config/config/devices/0x027a/zen71.json
```diff
@@ -54,36 +54,9 @@
...
-			"description": "Determine whether a trigger of this type should prompt a status change report to associated devices.",
-			"defaultValue": 1
-		},
-		{
-			"#": "7[0x08]",
-			"$if": "firmwareVersion >= 10.20 || firmwareVersion >= 2.20 && firmwareVersion < 10.0",
-			"$import": "~/templates/master_template.json#base_enable_disable",
-			"label": "Send Status Change Report: Timer",
-			"description": "Determine whether a trigger of this type should prompt a status change report to associated devices.",
-			"defaultValue": 1
```
AlCalzone, line 86: This split was done on purpose to make it easier to find which kinds of reports should be sent. We should add it to the other devices where this was missing, not undo this.

https://zwave-js.github.io/zwave-js/#/config-files/partial-parameters

AlCalzone (changes requested):

packages/config/config/devices/0x027a/zen71.json
gallagherrchris: It's not clear why these were split out. The text can certainly be updated/changed, but the main issue (aside from outright missing) is that they don't currently allow for all options available. Instead of duplicating with partial values, this change aims to centralize and provide full access without duplication

gallagherrchris: Spent some more time digging into the bitmask and partial implementation. Your suggestion of keeping the current implementation allows for the proper combination of all values because we are displaying each bit as a separate parameter and allowing the user to combine. It still feels strange to have a single parameter "Association Reports" displayed to the user in a fragmented manner. I will update the PR to follow the convention here and still try to consolidate these values in a more reusable way.

AlCalzone: > It still feels strange to have a single parameter "Association Reports" displayed to the user in a fragmented manner.

I would say it feels strange to have a single parameter with 16 options that essentially just turns 4 separate things on or off.

gallagherrchris:
@AlCalzone Can you give another look over the proposed changes?

packages/config/config/devices/0x027a/zen71.json
gallagherrchris: Hey @AlCalzone , I'd appreciate a re-review of the changes. The style is in-line with what you proposed above.

AlCalzone:
I've further clarified the parameters. We're not actually dealing with reports to the associated devices, but controlling commands.
Also the actual groups vary between the different switches. Some only have group 2, some 2 and 3.

AlCalzone (approved):

PR #8644, opened 2026-02-27
fix: cancel reader instead of closing port when closing web serial sink

AlCalzone:
This PR fixes a race condition that could cause an error while closing the web serial binding. We no longer close the port, but instead cancel and release the reader.


AlCalzone:
@zwave-js-bot automerge

PR #8662, opened 2026-03-04
fix: respect pollTime requirements for value auto-refreshes

### LINKED ISSUE #8659: Respect polling requirements from AWG specs
Section 8.2.6 mentions several requirements for automatically polling devices outside of user interaction, all of which include wait timers. We do not respect those, especially not in the code that refreshes potentially stale sensor readings.

Likely course of action:
- Use the `MessagePriority.Poll` for those, if not already done.
- Handle all polling in a coroutine that deals with the required waiting

AlCalzone:
Communication MUST be considered polling if a controlling device autonomously sends requests to
one or more nodes in a repeating fashion to monitor nodes or to get information from nodes. This
applies to any combination of commands.

There was actually one case where we didn't respect this requirement: automatic refreshes of sensor values after a certain amount of time.

This PR changes the logic for transactions with priority `Poll` to respect the specification requirements regarding polling. Now each of those can only be started if the last transaction was at least `pollTime` seconds ago, or `pollTime + CommandTime` when the last transaction was a poll.
Technically we wait even longer because we measure the time until the transaction is completed, which also includes the end node's time to respond.


AlCalzone:
@zwave-js-bot automerge

claudiucismaru:
Is this supposed to affect the Node's pollValue as well as refreshValues? I see a big time gap when I do pollValue until the promise resolves and the event is fired. This wasn't in 15.21.x. Is there any other way to refresh a value ID from a node, other than pollValue?

AlCalzone:
It's not supposed to - this is an oversight. Will post a fix shortly.

PR #8668, opened 2026-03-06
fix: emit value update with `UNKNOWN_STATE` when clearing notification variable

AlCalzone:
Removing the value was too much.


AlCalzone:
@zwave-js-bot automerge

PR #8626, opened 2026-02-18
feat(config): hide VZW32-SN future use parameter bits from production display

The00Dustin:
Depends on #8598

PR #8683, opened 2026-03-12
feat(config): add VZW30-SN config

The00Dustin:
To prevent conflicts, this is built on top of #8626 while depending on #8598


The00Dustin:
In case it isn't otherwise immediately obvious, the longer than usual option 0 descriptions in parameters 140-145 are due to the fact that disabling any one of each set of 3 there effectively disables all 3 by disabling the reporting.

AlCalzone (approved):

PR #8696, opened 2026-03-20
chore: enforce a minimal dependency age of 7 days

AlCalzone:
This should make us less likely to become a victim of supply chain attacks


AlCalzone:
@zwave-js-bot automerge

PR #8708, opened 2026-03-22
fix(config): add new product type for Shelly Wave PM Mini, FW 12.0

### LINKED ISSUE #8687: Missing device configuration: Shelly Wave PM Mini
### Checklist

- [x] It is not in the [configuration DB](https://devices.zwave-js.io/?search=3)

- [x] It was not [merged recently](https://github.com/zwave-js/zwave-js/pulls?q=is%3Apr+is%3Aclosed) or has a [pending PR](https://github.com/zwave-js/zwave-js/pulls?q=is%3Aopen+is%3Apr)

- [x] The device was interviewed completely and has no missing or `undefined` IDs

### Which device is missing?

Shelly Wave PM Mini (QMEM-0A1PC16). Note: The device itself is already in the database under Product Type 0x0007, but the v12.0.0 firmware update changed the device's Product Type to 0x0004, causing it to show up as Unknown.

### Manufacturer ID

0x0460

### Product Type

0x0004

### Product ID

0x0081

### Firmware Version

12.0.0

The device is already in the Z-Wave JS database here: https://devices.zwave-js.io/?jumpTo=0x0460:0x0007:0x0081:10.6.0 - It just needs the new 0x0004 Product?

### Do you have a manual?

_No response_

After updating my Shelly Wave PM Mini from firmware v10.6.0 to v12.0.0 via an OTA update, the device changed its `productType` identifier. 

Because of this change, the device is now showing up as `Unknown product 0x0081` instead of using the existing Wave PM Mini configuration. I have already fully re-interviewed the node, and the new Product Type persists.

* **Previous IDs (Firmware 10.6.0):** `0x0460 : 0x0007 : 0x0081`
* **New IDs (Firmware 12.0.0):** `0x0460 : 0x0004 : 0x0081`

Dooleweerdt:
It changed from 0x0007 to 0x0004 in latest FW update (12.0.0).

ProductType changed from 0x0007 to 0x0004 in latest FW update (12.0.0).

Fixes: #8687

### DIFF
--- a/packages/config/config/devices/0x0460/qmem-0A1PC16.json
+++ b/packages/config/config/devices/0x0460/qmem-0A1PC16.json
@@ -5,7 +5,7 @@
 	"description": "Wave PM Mini",
 	"devices": [
 		{
-			"productType": "0x0007",
+			"productType": "0x0004",
 			"productId": "0x0081",
 			"zwaveAllianceId": 5078
 		}
--- a/packages/config/config/devices/0x0460/qmem-0A1PC16.json
+++ b/packages/config/config/devices/0x0460/qmem-0A1PC16.json
@@ -5,12 +5,12 @@
 	"description": "Wave PM Mini",
 	"devices": [
 		{
-			// Firmware version < 12.0
+			// Firmware version < 12.0 and 12.0 <
 			"productType": "0x0007",
 			"productId": "0x0081"
 		},
 		{
-			// Firmware version >= 12.0
+			// Firmware version = 12.0
 			"productType": "0x0004",
 			"productId": "0x0081"
 		}
--- a/packages/config/config/devices/0x0460/qmem-0A1PC16.json
+++ b/packages/config/config/devices/0x0460/qmem-0A1PC16.json
@@ -5,12 +5,13 @@
 	"description": "Wave PM Mini",
 	"devices": [
 		{
-			// Firmware version < 12.0 and 12.0 <
 			"productType": "0x0007",
 			"productId": "0x0081"
 		},
 		{
-			// Firmware version = 12.0
+			// In firmware version 12.0, the product type was mistakenly
+			// changed to 0x0004:
+			// https://github.com/QubinoHelp/Shelly_Wave_FW_OTA/issues/8#issuecomment-4153013931
 			"productType": "0x0004",
 			"productId": "0x0081"
 		}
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x0460/qmem-0A1PC16.json
```diff
@@ -5,7 +5,7 @@
 	"description": "Wave PM Mini",
 	"devices": [
 		{
-			"productType": "0x0007",
+			"productType": "0x0004",
```
AlCalzone, line 15: In this case we should add a second entry here, not edit the existing one.
Otherwise this config would no longer apply to older firmware versions.

AlCalzone (changes requested):

packages/config/config/devices/0x0460/qmem-0A1PC16.json
Dooleweerdt: I have now added a second entry instead.
I'm still in doubt if it is an issue from Shelly - I'll try and ask them.

AlCalzone: Shall I hold off on merging?

Dooleweerdt: I don't think it will do any harm to merge it ;-)
I'm still in doubt where/how the bug came up in the first place, as I don't see any mention of changed ID's in the release notes from Shelly...
I'll re-post here (or in a new PR) if I get an answer from them.

danhead: I've just updated a PM Mini to 12.0.0 and it's now an unknown device, I wish I'd seen this PR before I clicked update! Thanks for fixing, much appreciated.

Dooleweerdt: Apparently there is a FW version 13.01 which has this fix, according to the response from Shelly support.
I'll  make an PR in the zwave-js/firmware-update repo to push it out on OTA - if my local update succeeds...

Dooleweerdt: Confirmed! The FW version 13.1 fixes the problem. I'll make a PR to the zwave-js/firmware-update repository.
I don't think this PR should go in as it looks now - at least the comments should be changed.

nifoc:
Not sure how this can happen, but apparently the change was a mistake by Shelly: https://github.com/QubinoHelp/Shelly_Wave_FW_OTA/issues/8#issuecomment-4153013931

packages/config/config/devices/0x0460/qmem-0A1PC16.json
AlCalzone: I'd say we merge this for users of the 12.x firmware and update the comment.

AlCalzone (approved):

PR #8711, opened 2026-03-23
fix: update `targetValue` optimistically, even for slow device classes

### LINKED ISSUE #8710: `targetValue` not updated optimistically on devices with slow device classes
When controlling a device with a slow device class (e.g. Motor Control Class B) through a Command Class that distinguishes between targetValue and currentValue and when using Supervision, the targetValue is not updated optimistically after the command was accepted (supervision status `Working`). This leads to weird quirks in applications, like the one reported in https://github.com/home-assistant/core/issues/165718

AlCalzone:
fixes: #8710


AlCalzone:
@zwave-js-bot automerge

PR #8719, opened 2026-03-26
fix(cli): parse multiline command descriptions correctly

AlCalzone:
Previously, the parser would choke on descriptions that contain `> `, and it would not correctly include wrapped descriptions that are too long for a single line.


AlCalzone:
@zwave-js-bot automerge

PR #8726, opened 2026-03-30
fix: unwanted delay in `Node.pollValue`

AlCalzone:
https://github.com/zwave-js/zwave-js/pull/8662 introduced an unwanted delay because it used `MessagePriority.Poll`, which is now meant for automatic background activity.


AlCalzone:
@zwave-js-bot automerge

PR #8737, opened 2026-04-02
feat(config): add HomeSeer PS100 v2

spudwebb:
Config file for the new version of the HomeSeer PS100 presence sensor.

### DIFF
--- a/packages/config/config/devices/0x000c/hs-ps100-v2.json
+++ b/packages/config/config/devices/0x000c/hs-ps100-v2.json
@@ -58,37 +58,37 @@
 			"#": "8",
 			"$import": "templates/homeseer_template.json#range_sensitivity",
 			"label": "Range 150-225cm (4.9-7.3ft) Sensitivity",
-			"defaultValue": 50
+			"defaultValue": 0
 		},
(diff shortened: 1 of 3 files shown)


packages/config/config/devices/0x000c/hs-ps100-v2.json
```diff
@@ -0,0 +1,110 @@
...
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 1,
+			"isLifeline": true
+		},
+		"2": {
+			"label": "Basic Set",
+			"maxNodes": 5
+		}
+	},
```
AlCalzone: These should not be necessary on modern devices, unless the association group has a terrible name in the firmware.

AlCalzone (changes requested):

packages/config/config/devices/0x000c/hs-ps100-v2.json
spudwebb: associations section has been removed, but now a test that seems unrelated is failing.

AlCalzone (approved):

PR #8745, opened 2026-04-10
feat(config): update firmware version for parameter 17 and add parameters 18, 19, and 20 for Zooz ZEN76

vtpearce91:
- Parameter 17 was introduced in firmware 10.0 and then 2.0 when it was based on 10.0.
- Parameter 18 was introduced in firmware 10.20 and then 2.20 when it was based on 10.20.
- Parameters 19 and 20 were introduced in 3.60.
See change log here: 
PR description [here](https://www.support.getzooz.com/kb/article/1151-zen76-s2-on-off-switch-change-log/)

### DIFF
--- a/packages/config/config/devices/0x027a/zen76.json
+++ b/packages/config/config/devices/0x027a/zen76.json
@@ -99,8 +99,26 @@
 		},
 		{
 			"#": "17",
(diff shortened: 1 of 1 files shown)


AlCalzone:
Superseded by https://github.com/zwave-js/zwave-js/pull/8899

PR #8749, opened 2026-04-14
fix: expose failed locking attempts for Door Lock CC through setValue API

AlCalzone:
Reported in https://github.com/home-assistant/core/issues/167814
The issue was that a delayed Supervision status of `Fail` was silently swallowed instead of verifying the current lock state. This PR streamlines the handling to be more in line with other CCs that have been migrated to SET_VALUE_HOOKS.


AlCalzone:
@zwave-js-bot automerge

PR #8750, opened 2026-04-14
refactor: move credential API behind the `accessControl` namespace

AlCalzone:
As mentioned in https://github.com/zwave-js/zwave-js-server/pull/1544#pullrequestreview-4101107871, it makes sense to namespace these high-level feature APIs, given that we're going to add more in the future. Technically breaking, but it is unlikely to be used out in the wild yet.


AlCalzone:
@zwave-js-bot automerge

PR #8752, opened 2026-04-15
refactor: make `endpoint.accessControl` `undefined` if unsupported

AlCalzone:
Having an entirely separate feature enum and checking support with unrelated functions felt awkward, so we're going to make the feature APIs possibly undefined to indicate no support.


AlCalzone:
@copilot fix the linter errors

PR #8758, opened 2026-04-22
refactor: rename User/Credential management result enums to `...Result`

AlCalzone:
Using `...Status` as the enum names was inconsistent with the rest of the codebase. We should fix this now.


AlCalzone:
@zwave-js-bot automerge

PR #8761, opened 2026-04-27
fix: response matching for duplicate credential responses

AlCalzone:
Fixes an issue where a CredentialSet that gets answered with a CredentialReport indicating a duplicate would time out because the Report referenced the existing slot, not the one we were trying to set.


AlCalzone:
@zwave-js-bot automerge

PR #8777, opened 2026-05-01
feat(config): add Shenzhen Neo NAS-WS02Z Water Sensor

Kirchoff:
Added the Shenzhen Neo Electronics NAS-WS02Z device, based on the user manual here:
https://fccid.io/Z52NAS-WS02Z/User-Manual/15-NAS-WS02Z-UserMan-US-r1-4384664.pdf

It closely matches the NAS-WS01Z which is already an available config.

### DIFF
--- a/packages/config/config/devices/0x0258/nas-ws02z.json
+++ b/packages/config/config/devices/0x0258/nas-ws02z.json
@@ -1,0 +1,78 @@
+{
+	"manufacturer": "Shenzhen Neo Electronics Co., Ltd.",
+	"manufacturerId": "0x0258",
+	"label": "NAS-WS02Z",
+	"description": "Water Sensor",
+	"devices": [
+		{
+			"productType": "0x0100",
+			"productId": "0x1025"
+		}
+	],
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"paramInformation": [
+		{
+			"#": "1",
+			"label": "Alarm Duration",
+			"valueSize": 1,
+			"unit": "minutes",
+			"minValue": 0,
+			"maxValue": 255,
+			"defaultValue": 120,
+			"unsigned": true
+		},
+		{
+			"#": "2",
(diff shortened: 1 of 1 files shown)


apella12:
Not the reviewer, but since the devices appear the nearly same (one nit is the minValue on param 3), did you consider just adding the fingerprint and model to the existing entry?
```
	"manufacturer": "Shenzhen Neo Electronics Co., Ltd.",
	"manufacturerId": "0x0258",
	"label": "NAS-WS01Z / NAS-WS02Z",  //new
	"description": "Water Sensor",
	"devices": [
		{
			"productType": "0x0003",
			"productId": "0x0085"
		},
		{
			"productType": "0x0003",
			"productId": "0x1085"
		},
		{
			"productType": "0x0003",
			"productId": "0x2085"
		},
		{
			"productType": "0x0100",  // new
			"productId": "0x1025"
		}
	],
```

csstup (approved):
I have this same device, the NAS-WS02Z.   Tested the parameters, they work as defined.

csstup:
I would prefer keeping NAS-WS02Z as a separate config unless the WS01Z parameter definitions can be confirmed identical. I have a physical WS02Z and have implemented/tested it in SmartThings Edge. Its Configuration CC parameters are P1–P5 with the following definitions:  
P1 Alarm Duration Time — 0–255 minutes, default 120
P2 Alarm Interval — 1–255 minutes, default 1
P3 First Alarm On Duration — 1–255 seconds, default 60
P4 Alarm On Duration — 1–255 seconds, default 5
P5 Beep Enable — 0/1, default 1  
I can provide Z-Wave JS diagnostics from the physical WS02Z as well. If the WS01Z parameter map is confirmed to be exactly the same, combining them would make sense, but otherwise I think separate configs are safer.

Tested fingerprint:
0x0258 / 0x0100 / 0x1025
Firmware 2.134.0

csstup (approved):

AlCalzone (approved):
Thank you!

PR #8784, opened 2026-05-08
Add Lamp-zw2 product json file

jackfeng2:
PR description here


jackfeng2:
feat(products): add Lamp-zw2 product JSON file

PR #8803, opened 2026-05-11
feat(product): add lamp-zw2 product JSON file

jackfeng2:
Add lamp-zw2 product JSON file

PR description here


jackfeng2 (commented):
Case fix

Issue #5686, opened 2023-04-19
iBlinds v3: Hide Binary Switch CC


habhomegit:
**Feature Request**
This feature request is related to a problem our users are having when using Home Assistant.  I've searched the JS documentation for the possibility of implementing a solution to no avail.   If there is already a solution in place please forgive my lack of knowledge.  

I am the CTO at HAB Home Intel the maker of the [iblinds](https://myiblinds.com/) product.  Our users are having some issues with the driver implementation.  Mostly related to the fact that our product works with Horizontal blinds tilt where fully open/ON is 50% not 99% Here are a few issues that it would be nice if we could fix/override by using the device configuration file:

From my little knowledge, it appears that the device driver supported commands are constructed by interrogating the device during the inclusion process for supported Command Classes.  In the case of our iblinds v3 device  (because all controllers are not created equally) we support the Binary CC, Multilevel Switch CC, and Window Covering CC for controlling the horizontal tilt position.   The Binary CC works as expected Close/Off = 0  and On/Open sends a 0xFF (255) and the device opens to 50%.    For Multilevel Switch CC  sending a level value of 0..99 works well, however, if an ON command is sent using a value of 0x63 (99) that is perfectly acceptable for the majority of multlevel switch devices like lights and even roller shades but in our case the 0x63(99) closes the blinds in the opposite direction. We need the blind slats to tilt to 50% to fully open.  

**Solution Request**
It would be nice if there was support to override certain command class specifics in the device configuration file.   

- Having the ability to ignore a certain supported command class (in our case ignore binary command class because the multilevel switch CC is preferred when it is also supported by the controller) .  Also, In HA the Cover Device gets out of sync because we only send back the REPORT from the requested GET command.  For example If the users uses the value slider a multilevel swtich command is sent followed by a multilevel GET, our device only sends the multilevel REPORT.  This causes the Binary and Multilevel UIs to become out of sync. 

- Having the ability to change the ON value for multilevel switch (in our case if the multilevel switch command received is ON / (0xff) 255  then we have our firmware configured to open the blinds to 50%.  

**Considered Alternatives**
- We have implemented a Z-Wave configuration parameter that will remap a multilevel switch command of 99 to 50% but it is not available for our older product versions. 

**Additional context**
We would be happy to send you an iblinds motor for testing at your request.

AlCalzone:
[...]

PR #8830, opened 2026-05-19
fix(config): add lifeline association and remove Binary Switch CC for iBlinds v2 (IB2.0)

jshessen:
## Problem

Two issues affect the iBlinds v2 (IB2.0) blind motor under zwave-js:

### 1 — Position always "unknown"

The v2 config was missing an Association Group 1 (Lifeline) definition. Without it, zwave-js does not assign the controller to the lifeline during inclusion, so the device never sends unsolicited Multilevel Switch Reports after movement. The controller must poll manually, and the position remains "unknown" between polls.

**Fix:** Add the standard Lifeline association block (identical to `iblindsv3.json`).

### 2 — Binary Switch CC confusion

v2 firmware 1.65 exposes Binary Switch CC, but like v3, it is not linked to Window Covering CC values. This causes incorrect state in controllers that surface Binary Switch as a separate entity.

**Fix:** Remove Binary Switch CC via compat flag, consistent with v3 config and the documented reasoning in #5686.

## Changes

`packages/config/config/devices/0x0287/ib2_0.json`:
- Added `associations` block with Group 1 Lifeline (1 node, `isLifeline: true`)
- Added `compat.commandClasses.remove` for Binary Switch across all endpoints

## Notes

- Parameters 2–10 (including Parameter 4 "Default ON Value") are v3 firmware additions and are **not** supported by v2 firmware (1.65). Only Parameter 1 is retained.
- The fix mirrors the existing `iblindsv3.json` pattern for both associations and compat.

## Testing

Verified on 6× IB2.0 nodes at firmware 1.65 with zwave-js-ui. After re-interview, all nodes self-assigned the controller to Group 1 and began sending position reports without manual polling.

### DIFF
--- a/packages/config/config/devices/0x0287/ib2_0.json
+++ b/packages/config/config/devices/0x0287/ib2_0.json
@@ -20,15 +20,6 @@
 			"isLifeline": true
 		}
 	},
-	"compat": {
-		"commandClasses": {
-			"remove": {
-				"Binary Switch": {
-					"endpoints": "*"
-				}
-			}
-		}
-	},
 	"paramInformation": [
 		{
 			"#": "1",
@@ -60,5 +51,17 @@
 				}
 			]
 		}
-	]
+	],
+	"compat": {
+		// https://github.com/zwave-js/zwave-js/issues/5686
+		// The device supports Binary Switch CC to control the tilt, but it is not linked
+		// to the Window Covering CC values. Remove the Binary Switch CC to avoid confusion.
+		"commandClasses": {
+			"remove": {
+				"Binary Switch": {
+					"endpoints": "*"
+				}
+			}
+		}
+	}
 }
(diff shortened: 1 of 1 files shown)


jshessen:
## Pre-submission lint results

Ran both linters proactively before review:

### `lint:zwave` (custom config linter)
```
Found 0 errors and 10 warnings!
The config files are valid!
```
✅ Zero errors. The 10 warnings are all pre-existing in unrelated device files (none in `0x0287/`).

### `lint:configjson` (dprint JSON format check)
✅ File passes format check (tabs, 4-space indent, LF line endings — matches repo style).

> Note: The workspace `dprint` binary requires GLIBC 2.39; our test host is aarch64/Debian GLIBC 2.36. Format was verified using dprint 0.44.0 which uses the same WASM JSON plugin (`json-0.19.3.wasm`) and reported no issues.

AlCalzone (approved):
Thanks!

PR #8864, opened 2026-06-10
fix: do not ignore controlled CCs in NIF

### LINKED ISSUE #7546: FGD-212 not reporting Scene Activation Command Class
### Checklist

### Deploy method

Manually built (git clone - npm run build - npm install

### Z-Wave JS UI version

9.29.0

### ZwaveJS version

14.3.7

### Describe the bug

FGD-212 (both FW 3.4 and 3.5) are not reporting (at beginning) Scene Activation v0 Command Class capabilities.
It will report this capability once the buttons attached to the device are clicked multiple times (triggering an actual scene).
This is problematic for users (as me) using openHAB ZWave-JS binding (under development) leveraging ZWave-JS debug JSON to introspect nodes configuration + command classes.
It looks like FGD-212 has special treatment regarding Scene Command Class which might be related to this issue.
See:
https://github.com/zwave-js/node-zwave-js/pull/1849
https://github.com/zwave-js/node-zwave-js/pull/1848

### To Reproduce

Fresh Z-Wave JS UI instance hooked with Controller + 1 FGD-212 device.

Look at node info: no Scene Activation v0 reported
Capture d’écran 2025-01-03 à 12 50 11

Then double/triple click a button attached to the FGD-212 device.
Look at node info: Scene Activation v0 is reported
Capture d’écran 2025-01-03 à 12 53 06

### Expected behavior

 Scene Activation v0 capability shall be reported at the beginning so that it is exposed in debug.json which is used by 3rd-party integration to introspect and reflect node capabilities.

### Additional context

_No response_

AlCalzone:
We used to ignore controlled CCs in a device's NIF because of a requirement that _Nodes SHOULD NOT advertise controlled CCs_. However that does not mean we can not evaluate them.

For some devices this also caused an issue with missing Scene Activation CC functionality, since this CC is exclusively reported as controlled.

closes: https://github.com/zwave-js/zwave-js/discussions/7983
fixes: https://github.com/zwave-js/zwave-js/issues/7546

closes: https://github.com/zwave-js/zwave-js/pull/8102
closes: https://github.com/zwave-js/zwave-js/pull/7848

### DIFF
--- a/packages/cc/src/cc/VersionCC.ts
+++ b/packages/cc/src/cc/VersionCC.ts
@@ -469,6 +469,16 @@ export class VersionCC extends CommandClass {
 					// for which support is determined later.
 					if (cc === CommandClasses.Basic) {
 						endpoint.addCC(cc, { version: supportedVersion });
+					} else if (
+						endpoint.controlsCC(cc) && !endpoint.supportsCC(cc)
+					) {
+						// The node only advertised this CC as controlled (e.g. in the NIF).
+						// Remember the version so its values can be exposed, but do not
+						// promote it to a supported CC.
+						endpoint.addCC(cc, {
+							isControlled: true,
+							version: supportedVersion,
+						});
 					} else {
 						endpoint.addCC(cc, {
 							isSupported: true,
--- a/packages/cc/src/cc/ZWaveProtocolCC.ts
+++ b/packages/cc/src/cc/ZWaveProtocolCC.ts
@@ -92,6 +92,7 @@ export class ZWaveProtocolCCNodeInformationFrame extends ZWaveProtocolCC
 		this.supportsSecurity = options.supportsSecurity;
 		this.supportsBeaming = options.supportsBeaming;
 		this.supportedCCs = options.supportedCCs;
+		this.controlledCCs = options.controlledCCs ?? [];
 	}
 
 	public static from(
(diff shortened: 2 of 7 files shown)


AlCalzone:
@zwave-js-bot automerge

apella12:
FWIW -The specification document I have (could be old- April 2, 2024-CC:0000.00.00.13.002) has slightly different wording on Controlled CCs. It also seems odd that the NIF would have a specific format (after 0xFE MARK) for controlled CCs if they weren't allowed. My version
```
The field MAY advertise the list of Command Classes that the node can control in other
nodes. If present, the list of controlling Command Classes MUST be prepended by the COMMAND_
CLASS_MARK Command Class identifier. It is NOT RECOMMENDED to advertise controlled
Command Classes.
```
Anyway, Thanks for fixing

Lastly, this might fix/conflict with the Fibaro proprietary CC hack. I don't have the blinds, but IIRC CC 38 (0x26) is also present in the Controlled CC's for those devices (but not supported). Might be worth a look

AlCalzone:
Can you elaborate what you mean with _Lastly, this might fix/conflict with the Fibaro proprietary CC hack._?
I have the shutters in a drawer, so I can check.

apella12:
With the caution I have not fully analyzed how your fix works (Only Scene Activation?). This .json
```
"label": "FGR223",
<snip>
	"proprietary": {
		"fibaroCCs": [38 /* 0x26 (Venetian Blinds) */]
	},
```
was likely needed because 0x26 was included as a controlled CC, so was ignored before. It might not be needed or a duplicate?

AlCalzone:
Ah no, that's unrelated. This is for defining which proprietary sub-commands of the `Proprietary CC` are supported, since there is no way to query this information.

apella12:
Ok, good. It had caught my eye because Fibaro was the manufacturer on this issue (scene Activation)

PR #8866, opened 2026-06-10
fix: clear User Code CC slots in value DB instead of deleting

raman325:
### What

On a successful clear via any of the User Code CC access control APIs (`deleteUser`, `deleteAllUsers`, `deleteCredential`, `deleteCredentials`), the UC branch routes through the shared `#purgeCachedUserCodes` helper, which calls `valueDB.removeValue(vid)` for the `userIdStatus` and `userCode` entries of the affected slots.

### Why this is wrong

User Code CC reserves a fixed number of slots (1..N), one per user identifier. Clearing a slot doesn't make the slot go away — it moves the slot to the `Available` state with an empty code. The slot is still there and can be filled again later.

The legacy property-set path (`UserCodeCCAPI[SET_VALUE]` for `userIdStatus = Available`) ultimately ends up with the slot persisted as `userIdStatus = Available` and `userCode = ""` in the value DB, because of the `verifyChanges` hook in `UserCodeCCAPI[SET_VALUE_HOOKS]` that schedules a poll → Get → Report → `persistUserCode` writes both values.

The `accessControl.*` delete paths skip that and actively *remove* the entries instead. The effect is observable to value-DB consumers: zwave-js-ui renders slots from value-DB entries, so after a delete via any of these APIs, the slot row disappears from its UI entirely (rather than showing as Available like every other cleared slot).

The neighboring `setCredential` UC branch does not purge or rewrite value-DB entries on success — it just emits an event. The asymmetry lives in `#purgeCachedUserCodes`.

### The fix

Rename `#purgeCachedUserCodes` → `#clearCachedUserCodes` and change its implementation to write the cleared state explicitly into the value DB for each existing slot:

* `userIdStatus(userId)` → `UserIDStatus.Available`
* `userCode(userId)` → `""`

This mirrors what `persistUserCode` would write if a Report came back via Get, without forcing an extra Get round-trip on supervised success. All four UC-CC delete paths benefit from the same fix.

### How this was found

Encountered while testing [`lock_code_manager`](https://github.com/raman325/lock_code_manager) 2.x against a 500-series UC lock + zwave-js-ui. LCM 2.x routes credential deletes through `accessControl.deleteCredential` (instead of the legacy property-set path used by LCM 1.x), and the affected slot disappears from zwave-js-ui after each delete. Other cleared slots remain visible.

### Verification

* `yarn test:ts -- accessControl.UserCode.test.ts` → 32/32 pass (28 prior + 4 new integration tests asserting the post-clear `Available` / `""` state for each delete path)
* `yarn test:ts -- accessControl.UserCredential.test.ts` → 30/30 pass (U3C path untouched)
* `oxlint --type-aware` clean
* `dprint check` clean

### Commits

The first and fourth commits are `tsconfig.build.json` noise (added then reverted per review). The actual fix is in commits 2 (initial inline fix) + 3 (refactor to shared helper + integration tests).

### DIFF
--- a/packages/cc/tsconfig.build.json
+++ b/packages/cc/tsconfig.build.json
@@ -19,8 +19,13 @@
 		},
 		{
 			"path": "../maintenance/tsconfig.build.json"
+		},
+		{
+			"path": "../transformers/tsconfig.build.json"
 		}
 	],
-	"include": ["src_gen/**/*.ts"],
+	"include": [
+		"src_gen/**/*.ts"
+	],
 	"exclude": []
 }
--- a/packages/cc/tsconfig.build.json
+++ b/packages/cc/tsconfig.build.json
@@ -19,13 +19,8 @@
 		},
 		{
 			"path": "../maintenance/tsconfig.build.json"
-		},
-		{
-			"path": "../transformers/tsconfig.build.json"
 		}
 	],
-	"include": [
-		"src_gen/**/*.ts"
-	],
+	"include": ["src_gen/**/*.ts"],
 	"exclude": []
 }
--- a/packages/zwave-js/src/lib/node/feature-apis/AccessControl.ts
+++ b/packages/zwave-js/src/lib/node/feature-apis/AccessControl.ts
@@ -799,7 +799,7 @@ export class AccessControlAPI extends FeatureAPI {
 				succeeded = supervisedCommandSucceeded(result);
 			}
 			if (succeeded) {
-				this.#purgeCachedUserCodes(userId);
+				this.#clearCachedUserCodes(userId);
 			}
 			if (succeeded && existed) {
 				const node = this.endpoint.tryGetNode();
@@ -843,7 +843,7 @@ export class AccessControlAPI extends FeatureAPI {
 			const succeeded = result == undefined
 				|| supervisedCommandSucceeded(result);
 			if (succeeded) {
-				this.#purgeCachedUserCodes();
+				this.#clearCachedUserCodes();
 			}
 			return succeeded ? SetUserResult.OK : SetUserResult.Error_Unknown;
 		}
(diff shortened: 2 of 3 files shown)


packages/cc/tsconfig.build.json
```diff
@@ -19,8 +19,13 @@
 		},
 		{
 			"path": "../maintenance/tsconfig.build.json"
+		},
+		{
+			"path": "../transformers/tsconfig.build.json"
 		}
```
AlCalzone: I don't know what in your setup keeps wanting to add this. When I clone the repo and then run
```
yarn
yarn bootstrap
yarn build
```
everything works fine without this.

packages/zwave-js/src/lib/node/feature-apis/AccessControl.ts
```diff
@@ -1195,19 +1195,40 @@ export class AccessControlAPI extends FeatureAPI {
...
+							endpointIndex,
+						),
+						UserIDStatus.Available,
+					);
+					node.valueDB.setValue(
+						UserCodeCCValues.userCode(targetUserId).endpoint(
+							endpointIndex,
+						),
+						"",
+					);
```
AlCalzone: This seems wrong to me. I think the actual fix would be to correct `#purgeCachedUserCodes` to not delete the values but rather clear the values like you do here manually.

AlCalzone (changes requested):
Also lets add an integration test to `packages/zwave-js/src/lib/test/node/accessControl.UserCode.test.ts` that asserts the intended state of the values after deletion of credentials.

packages/zwave-js/src/lib/node/feature-apis/AccessControl.ts
raman325: Good call — refactored `#purgeCachedUserCodes` to set each existing slot to `userIdStatus = Available` / `userCode = ""` instead of removing the entries, and renamed it to `#clearCachedUserCodes`. Now all four UC-CC delete paths (deleteUser, deleteAllUsers, deleteCredential, deleteCredentials) get the correct post-clear value DB state through the one helper. Done in 443f9374f.

packages/cc/tsconfig.build.json
raman325: Reverted in 537c1815a. Confirmed it's only my local pre-commit hook (`yarn check-references`) demanding the transformers ref — CI doesn't enforce this check, and the compiled output (`src_gen/`) has no transformers imports. Used `--no-verify` for the revert since this is a local-tooling-only issue. If anyone else hits this, I'd guess it's a `@monorepo-utils/workspaces-to-typescript-project-references` version drift, but didn't dig further.

raman325:
Added integration tests covering the post-clear value DB state for all four UC-CC delete paths (deleteCredential, deleteUser, deleteCredentials, deleteAllUsers) in 443f9374f. Each asserts that `userIdStatus` ends up as `Available` and `userCode` as `""` rather than being removed.

Summary of changes since the last review:

- 443f9374f — Moved the post-clear value DB write out of `deleteCredential` and into the shared helper (renamed `#purgeCachedUserCodes` → `#clearCachedUserCodes`). All four UC-CC delete paths now write the cleared state through the same helper. Added integration tests asserting the post-clear `Available` / `""` state for each path.
- 537c1815a — Reverted the `tsconfig.build.json` transformers ref.

packages/zwave-js/src/lib/node/feature-apis/AccessControl.ts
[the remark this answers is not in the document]
raman325: Refactored `#purgeCachedUserCodes` → `#clearCachedUserCodes` to write the cleared state (`userIdStatus = Available`, `userCode = ""`) instead of removing the entries. All callers continue to use the helper, so any consumer reading from the helper's output gets the correct cleared state rather than missing values.

AlCalzone (approved):
LGTM, I just reworded the one doc comment

PR #8871, opened 2026-06-10
feat(config): update Zooz ZEN35 for firmware 1.40 and unique dimmer wording

hceuterpe:
fix: update Zooz ZEN35 for firmware 1.40 and unique dimmer wording

- Update LED indicator labels for brightness settings. Existing options no longer function with firmware update 1.40.
- Update wording to specify dimmer instead of relay, that was carryover from the ZEN32 switch.  Updated defaults values unique to this model.

### DIFF
--- a/packages/config/config/devices/0x027a/templates/zooz_template.json
+++ b/packages/config/config/devices/0x027a/templates/zooz_template.json
@@ -303,6 +303,57 @@
 			}
 		]
 	},
+	"led_indicator_brightness_1_10": {
+		"label": "LED Indicator Brightness",
+		"valueSize": 1,
+		"minValue": 1,
+		"maxValue": 10,
+		"defaultValue": 5,
+		"unsigned": true,
+		"allowManualEntry": false,
+		"options": [
+			{
+				"label": "10%",
+				"value": 1
+			},
+			{
+				"label": "20%",
+				"value": 2
+			},
+			{
+				"label": "30%",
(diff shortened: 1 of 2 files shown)


packages/config/config/devices/0x027a/templates/zooz_template.json
```diff
@@ -303,6 +303,57 @@
 			}
 		]
 	},
+	"led_indicator_brightness_1_10": {
+		"label": "LED Indicator Brightness",
+		"valueSize": 1,
+		"minValue": 1,
+		"maxValue": 10,
+		"defaultValue": 5,
```
AlCalzone, line 311: According to https://www.support.getzooz.com/kb/article/1673-zen35-scene-dimmer-advanced-settings/?section_id=449 the default is 1.

hceuterpe:
I notice some inconsistency on this.  Triggering a reset sets the value to 5 (50%).  Isn't this technically more accurate since resetting it sets  value the device itself thinks is default?

AlCalzone:
That makes sense. I'll ask Zooz about it.

AlCalzone (approved):
Zooz confirmed that the docs were wrong.

TheZackCodec:
@hceuterpe lol I just made these changes on my local branch then before I checked for opened and closed issues like i normally do. Guess I don't have to anymore you beat me to it. Thanks for the update 👍

PR #8873, opened 2026-06-12
fix: fall back to User Code CC when User Credential CC is unusable

raman325:
The Access Control feature API dispatches between User Code CC and User Credential CC via `#usesUserCredentialCC`, which so far only checked `supportsCC(User Credential)` (with a FIXME noting that nodes may support both CCs).

For a node that advertises U3C but whose U3C interview yields no usable data — no users, no supported credential types, or only credential types with `numberOfCredentialSlots: 0` — this made the entire unified API unusable even when the node also supports User Code CC and works fine through it:

- `getUserCapabilitiesCached()` / `getCredentialCapabilitiesCached()` report 0 users / 0 slots
- `#assertValidSlot` consequently throws `Credential slot N is out of range for credential type PINCode` for every `setCredential` / `deleteCredential` call

Applications built on the unified API then fail every PIN operation. Seen in the wild via Home Assistant + Lock Code Manager, where users get `Credential slot for pin_code must be between 1 and 0` (HA validates against the same cached capabilities before calling the driver): raman325/lock_code_manager#1251

### Fix

`#usesUserCredentialCC` now prefers User Credential CC only when its cached capabilities are usable (at least one user and at least one credential type with a non-zero number of slots). When the node also supports User Code CC and U3C is unusable, the API falls back to User Code CC instead of pretending no users or credentials can exist.

Notes:

- Nodes supporting only one of the two CCs are unaffected (early returns).
- The check runs against cached interview data, consistent with the rest of the API: if a (re-)interview later yields usable U3C capabilities, dispatch switches to U3C automatically.
- The FIXME is retained for the full solution (switching CCs on the fly / migration via Version CC).

### Tests

New `accessControl.DualCC.test.ts` with three integration tests:

1. Both CCs, U3C reports 0 users → capabilities and writes use User Code CC
2. Both CCs, U3C reports a PIN credential type with 0 slots → capabilities and writes use User Code CC
3. Both CCs, U3C reports usable capabilities → U3C is preferred (existing behavior preserved)

Existing `accessControl.UserCode` / `accessControl.UserCredential` / `userCredentialInterview` suites pass unchanged (72/72).

### DIFF
--- a/packages/zwave-js/src/lib/node/feature-apis/AccessControl.ts
+++ b/packages/zwave-js/src/lib/node/feature-apis/AccessControl.ts
@@ -214,12 +214,49 @@ function supportsNonPINChars(supportedASCIIChars: string | undefined): boolean {
 
 /** High-level API for managing users and credentials on access control devices */
 export class AccessControlAPI extends FeatureAPI {
-	// FIXME: This is technically not correct. A node could support both CCs,
-	// and we may have to decide which one to use, or switch between them on
-	// the fly using Version CC / migration.
-	// This is not implemented yet, so checking for U3C first is fine for now.
+	// FIXME: Switching between both CCs on the fly using Version CC /
+	// migration is not implemented yet. For now, prefer User Credential CC
+	// when its interview yielded usable data, and fall back to User Code CC
+	// otherwise.
 	get #usesUserCredentialCC(): boolean {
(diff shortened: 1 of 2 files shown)


AlCalzone:
There are quite significant spec requirements about the interoperability of User Code and User Credential CC on locks. I'd like to first understand what's going on in the linked issue before committing to a solution.

PR #8875, opened 2026-06-12
fix(cc): validate name length in AGI Group Name Report

AlCalzone:
## Problem

A reported group name length was validated only against the number of bytes actually present in the frame, not against the maximum length the specification allows (0–42). A non-conforming or malicious device could set the length field beyond the spec maximum and have the over-long name accepted and persisted, then surfaced to applications as the group label.

## Fix

Validate the length field against the spec maximum (0–42) before decoding the name, mirroring the existing length-bound handling for other reported text fields. An out-of-range length now drops the frame.


AlCalzone:
@zwave-js-bot automerge

PR #8884, opened 2026-06-15
fix: handle obfuscated codes when verifying User Code changes

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #8899, opened 2026-06-24
fix(config): align all Zooz devices with the documentation, update to latest firmwares

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #8904, opened 2026-06-25
fix(config): correct value size for Zooz ZEN78, params 13/14

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #8906, opened 2026-06-26
fix: treat S2 command with next  expected sequence number as delivery verification

AlCalzone:
When the `s2VerifyDelivery` option is enabled, we would always wait 500ms for a potential NonceReport, even when the next command by the peer was already received.

With this PR, we now treat the receipt of an S2 encapsulated command with the next expected sequence number as successful delivery.


AlCalzone:
@zwave-js-bot automerge

PR #8909, opened 2026-06-29
chore: teach Claude how to use custom device files in integration tests

AlCalzone:
## What

Extends the `integration-tests` skill with a new **Custom Device Config Files** section explaining how to drive an integration test from a device config file (e.g. to exercise `compat` flags like `commandClasses.remove`, or to reproduce a device-specific problem report).

It documents the three pieces that must line up:
1. Fingerprint the mock node (`manufacturerId` / `productType` / `productId`).
2. Ship a fixture `deviceConfig.json` matching that fingerprint.
3. Load it via `additionalDriverOptions.storage.deviceConfigPriorityDir`.

Plus a complete example and a **Gotchas** list covering the traps that silently prevent a config from loading — firmware-range matching (mock defaults to `"1.0"`), the 4-digit-hex vs number ID comparison, and the Manufacturer Specific + Version CC requirement.

## Why

These details aren't obvious and the failure mode (`No device config found`) is easy to miss, so this was previously hard to get right from the skill alone.

Docs/tooling only — no runtime or test code changes.

### DIFF
--- a/.claude/skills/integration-tests/SKILL.md
+++ b/.claude/skills/integration-tests/SKILL.md
@@ -195,7 +195,10 @@ integrationTest("compat flag removes Window Covering CC", {
 		commandClasses: [
 			CommandClasses["Manufacturer Specific"],
 			CommandClasses.Version,
-			ccCaps({ ccId: CommandClasses["Window Covering"], isSupported: true }),
+			ccCaps({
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #8921, opened 2026-07-01
fix(config): override supported thermostat modes for Fibaro FGT-001

ElVit:
Add manual link for Fibaro FGT-001, expose Multilevel Sensor/Battery/Notification CCs on endpoint 2 and Notification on endpoint 1, and override the Thermostat Mode supported modes query. The entry "Fully Open" (Manufacturer specific) was missing.

Issue with the missing Thermostat Mode was discussed in https://github.com/zwave-js/zwave-js-ui/issues/972.

### DIFF
--- a/packages/config/config/devices/0x010f/fgt001.json
+++ b/packages/config/config/devices/0x010f/fgt001.json
@@ -199,24 +199,17 @@
 		"overrideQueries": {
 			"Thermostat Mode": [
 				{
-					// Override the getSupportedModes query for endpoint 1, because the device
-					// reports incorrect supported modes via the Z-Wave protocol.
+					// The device does not report "Manufacturer specific" as a supported thermostat mode
 					"endpoint": 1,
 					"method": "getSupportedModes",
-
-					// The device natively reports only [Off, Heat], but also supports
-					// "Fully Open" (mode 31 / 0x1F, manufacturer-specific).
 					"result": [0, 1, 31],
 
-					// Persist the corrected supported modes so ZWaveJS does not overwrite
-					// them with the incorrect values reported by the device.
+					// Prevent overwriting the supported modes with the values reported by the device.
 					"persistValues": {
 						"supportedModes": [0, 1, 31]
 					},
 
-					// Extend the metadata with human-readable labels for each mode,
-					// since mode 31 is manufacturer-specific and has no standard label.
-					// Without this, ZWaveJS would display it as "Manufacturer specific".
+					// Extend the metadata since mode 31 is manufacturer-specific and has no standard label.
 					"extendMetadata": {
 						"thermostatMode": {
 							"states": {
--- a/packages/config/config/devices/0x010f/fgt001.json
+++ b/packages/config/config/devices/0x010f/fgt001.json
@@ -22,7 +22,7 @@
 		"max": "255.255"
 	},
 	"metadata": {
-		"manual": "https://manuals.fibaro.com/heat-controller/"
+		"manual": "https://manuals.fibaro.com/content/manuals/en/FGT-001/FGT-001-EN-T-v1.6.pdf"
 	},
 	"paramInformation": [
 		{
--- a/packages/config/config/devices/0x010f/fgt001.json
+++ b/packages/config/config/devices/0x010f/fgt001.json
@@ -21,6 +21,9 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
+	"metadata": {
+		"manual": "https://manuals.fibaro.com/heat-controller/"
+	},
 	"paramInformation": [
 		{
 			"#": "1",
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x010f/fgt001.json
```diff
@@ -143,11 +146,58 @@
...
+					"endpoints": {
+						"1": {
+							"isSupported": true
+						},
+						"2": {
+							"isSupported": true
+						}
+					}
+				}
+			},
```
AlCalzone: Why are these needed? Can you share a log of a re-interview on loglevel debug please?

packages/config/config/devices/0x010f/fgt001.json
```diff
@@ -21,6 +21,9 @@
 		"min": "0.0",
 		"max": "255.255"
 	},
+	"metadata": {
+		"manual": "https://manuals.fibaro.com/heat-controller/"
+	},
```
AlCalzone: Please put this section at the bottom.

packages/config/config/devices/0x010f/fgt001.json
```diff
@@ -143,11 +146,58 @@
...
+				}
+			},
 			"remove": {
 				"Battery": {
 					"endpoints": [0]
 				}
 			}
+		},
+		"overrideQueries": {
+			"Thermostat Mode": [
```
AlCalzone, line 153: The shared config in the linked issue had a comment here explaining why this is done. We should preserve that comment.

AlCalzone (changes requested):

packages/config/config/devices/0x010f/fgt001.json
ElVit: Here the debug log:
[zwave-debug-2026-07-02T19-33-02-471Z.zip](https://github.com/user-attachments/files/29611646/zwave-debug-2026-07-02T19-33-02-471Z.zip)

Created with Claude:

The compat -> commandClasses -> add block is necessary because the FGT001’s NIF (Node Information Frame) does not correctly advertise all command classes for its endpoints. Without this block, ZWaveJS would not interview these CCs, even though the device fully supports them.

This is confirmed by the debug log of Node 42, which has an external temperature sensor connected (Configuration param 3[0x01] Status: Temperature Sensor Connected = 1):

- **Multilevel Sensor on Endpoint 2** — allows ZWaveJS to interview the external temperature sensor on EP2. The log confirms the value was successfully retrieved:
  ```
  Value added: 49-2-Air temperature => 26°C
  ```

- **Battery on Endpoint 2** — the FGT001 has two separate battery sets: EP1 for the thermostat valve (82%) and EP2 for the external temperature sensor. Without this entry, only EP1’s battery would be queried. The log confirms both are read:
  ```
  Value added: 128-1-level => 82
  Value added: 128-2-level => 100
  ```

- **Notification on Endpoints 1 & 2** — both endpoints send notifications (e.g. battery maintenance status, hardware alerts) that are not declared in their NIF. The log shows all notification values were successfully added for both endpoints, including:
  ```
  Value added: 113-2-Power Management-Battery maintenance status => 0
  ```

Without this block, the temperature reading of the external sensor, its dedicated battery level, and the notifications on both endpoints would simply not appear.

packages/config/config/devices/0x010f/fgt001.json
ElVit: fixed

packages/config/config/devices/0x010f/fgt001.json
ElVit: added comments

packages/config/config/devices/0x010f/fgt001.json
AlCalzone: That's not true though. The NIF does not contain these CCs because they are supported securely. Hence they are reported via Security2CCCommandsSupportedReport.

Endpoint 1 reports support for Notification CC:
```
19:32:37.864 CNTRLR « [Node 042] received secure commands (S2_Authenticated)            [Endpoint 1]
                      supported CCs:
                      · Multi Channel Association
                      · Thermostat Mode
                      · Thermostat Setpoint
                      · Schedule
                      · Association Group Information
                      · Notification
                      · Protection
                      · Battery
                      · Clock
                      · Association
                      · Supervision
```

Endpoint 2 reports support for Multilevel Sensor, Battery and Notification:
```
19:32:38.069 CNTRLR « [Node 042] received secure commands (S2_Authenticated)            [Endpoint 2]
                      supported CCs:
                      · Multi Channel Association
                      · Multilevel Sensor
                      · Association Group Information
                      · Notification
                      · Battery
                      · Association
                      · Supervision
```

packages/config/config/devices/0x010f/fgt001.json
AlCalzone: Oof, that's a bit verbose. I just meant these two comments:
```
		"overrideQueries": {
			// The device does not report "Manufacturer specific" as a supported thermostat mode
			"Thermostat Mode": [
				{
					"method": "getSupportedModes",
					"result": [0, 1, 31], // Off / Heat / Manufacturer specific
```

packages/config/config/devices/0x010f/fgt001.json
ElVit: OK thanks for the clarification. I have removed this part.

packages/config/config/devices/0x010f/fgt001.json
ElVit: I updated the comments again. Is it better now?

packages/config/config/devices/0x010f/fgt001.json
ElVit: Claude told me also to remove the next element "remove ->Battery -> endpoints: [0]".
Ist it correct to remove it?

ElVit: OK Claude did again a mistake, the "remove" block is still neccessary since else there will be a third battery sensor with the value "unknown". I added the "remove" block again.

AlCalzone (approved):
Thanks!

PR #8926, opened 2026-07-03
fix: cache devices not known to the firmware update service

AlCalzone:
Request statistics of the firmware update service seemed to be pretty high compared to expectations. The problem was that the `v4` endpoint encoded unknown devices by omitting them from the response, but Z-Wave JS's internal cache would only cache devices that are present.

As a result, networks with several unknown devices would still cause multiple requests with the same payload, instead of serving the results from cache.

This PR fixes that behavior and pins it down with a unit test.


AlCalzone:
@zwave-js-bot automerge

PR #8931, opened 2026-07-04
fix: assert the correct command in Window Covering setValue API

AlCalzone:
`WindowCoveringCCAPI.set()` asserted support for `StartLevelChange` instead of `Set`, likely a copy-paste mistake. Both commands are mandatory, so this had no practical effect. The other API methods assert their matching commands.


AlCalzone:
@zwave-js-bot automerge

PR #8932, opened 2026-07-05
fix: correct Irrigation CC nominalCurrent metadata and low flow error label

AlCalzone:
The `nominalCurrent` dynamic CC value used `ValueMetadata.ReadOnlyBoolean`, but the value is numeric: the Valve Info Report stores `10 * raw.payload[2]` in mA, per CC:006B.01.09.11.008 ("This field MUST be expressed as a multiple of 10mA."). Changed to `ReadOnlyNumber`.

Also fixes a copy-paste typo: the `errorLowFlow` label read "Flow below high threshold", now "Flow below low threshold".

All other Irrigation value metadata match their stored types.


AlCalzone:
@zwave-js-bot automerge

PR #8945, opened 2026-07-07
feat(config): extend config lints to template files

### LINKED ISSUE #8549: Extend some config lints to the template files
E.g. default value outside of min/max

AlCalzone:
fixes: #8549

Template files were never linted directly, so invalid partial parameter definitions either got attributed to every importing device file or escaped linting entirely (e.g. when the offending field is overridden by importers, or not checked for non-partial params like `defaultValue` vs. `minValue`/`maxValue`).

### Changes

- `lintTemplates()` enumerates all `**/templates/*.json` files, resolves their internal `$import`s and lints each entry, attributing errors/warnings to the template file itself
- Since template entries are partial definitions, each check only fires when the entry itself defines all fields the check needs:
  - `minValue` > `maxValue`
  - `defaultValue` outside `minValue`/`maxValue`
  - `valueSize` outside 1...4
  - `minValue`/`maxValue`/`defaultValue` incompatible with `valueSize` (with a warning to consider `"unsigned": true` where that would fix it)
  - option values outside `minValue`/`maxValue`, duplicate option values
  - `defaultValue` not among the options when `allowManualEntry` is `false`
  - `readOnly`/`writeOnly` conflicts
  - `allowed` definitions (shared with the device lint) and their mutual exclusivity with `minValue`/`maxValue` and `allowManualEntry: false`
  - description mentions a default value / description too similar to label (warnings)

### Config fixes

The new lint surfaced 6 shipped errors, fixed in the second commit:

- HomeSeer `temp_reporting_interval`: default 1 was below minimum 30 → 60 per the manual
- Aeotec `sensor_report_type`: default 1 was not among the options → 256 (Basic CC Report) per the engineering spec
- Trane `schedule_heat_setpoint`/`schedule_cool_setpoint`: default 0 was out of range 55-90/60-99 → 70/78 °F (the factory schedule varies per period, so these are typical in-range values)
- Yale `auto_relock_time_127`: default 5 was below minimum 10 → 30 per the manual
- Logic Group `multilevel_upper_function`: default 255 was above maximum 99, but is correct per the manual (valid values are 0-99 and 255) → expressed via `allowed` instead of the imported 0-99 range

### DIFF
--- a/packages/config/maintenance/lintConfigFiles.ts
+++ b/packages/config/maintenance/lintConfigFiles.ts
@@ -208,7 +208,7 @@ interface LintDevicesContext extends LintDevicesContextConditional {
 	variant: DeviceID | undefined;
 }
 
-function lintPartialParamDefinition(
+function lintTemplateParamDefinition(
 	file: string,
 	name: string,
 	definition: Record<string, unknown>,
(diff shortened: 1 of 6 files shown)


packages/config/maintenance/lintConfigFiles.ts
```diff
@@ -207,6 +208,306 @@ interface LintDevicesContext extends LintDevicesContextConditional {
 	variant: DeviceID | undefined;
 }
 
+function lintPartialParamDefinition(
```
AlCalzone: This is not a good name. We have a different concept we call "partial config params".

AlCalzone: Renamed to `lintTemplateParamDefinition` in bfcb6fb5b.

packages/config/maintenance/lintConfigFiles.ts
```diff
@@ -207,6 +208,306 @@ interface LintDevicesContext extends LintDevicesContextConditional {
...
+			file,
+			`${prefix} is invalid: readOnly and writeOnly are mutually exclusive!`,
+		);
+	}
+	if (readOnly && allowManualEntry) {
+		addError(
+			file,
+			`${prefix} is invalid: allowManualEntry must be omitted for readOnly parameters!`,
+		);
+	}
```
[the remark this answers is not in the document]
AlCalzone: Good catch. Fixed in bfcb6fb5b — the lint now mirrors the parser rules: `readOnly`/`writeOnly` must be `true` or omitted, `allowManualEntry` must be `false` or omitted. This made the `readOnly && allowManualEntry` check dead code, so it was removed.

PR #8946, opened 2026-07-07
feat: add high-level API for interacting with indicators

### LINKED ISSUE #8147: Add dedicated APIs to interact with Indicator CC values
The heap of unusable values for the Indicator CC values was recently removed, but we still need some way of interacting with them. Since Indicator CC values are somewhat complex and depend on each other, I think it should be done through dedicated APIs and UI components.

AlCalzone:
Fixes #8147

Indicator CC values are a bad fit for how the CC is used: properties must be written in groups (a Set omitting a property resets it to `0x00` on the device), the Binary/Multilevel split is a device detail, and blink/timeout semantics span multiple property IDs. Following the Access Control precedent, this adds a high-level `endpoint.indicators` feature API that models each indicator's functionality instead of raw indicator/property IDs.

### Public API

```ts
if (endpoint.indicators) {
	await endpoint.indicators.set(Indicator["Button 1 indication"], {
		on: true,
		timeout: "30s",
	});
}
```

- `getSupportedCached()` / `getCapabilitiesCached(id)` — per-indicator capability flags (`supportsOnOff`, `supportsLevel`, `supportsBlinking`, `supportsTimeout`, `supportsSoundLevel`) derived from the supported property IDs. Labels come from the `Indicator` enum, or the V4 description for manufacturer-defined indicators.
- `get(id)` / `getCached(id)` — the normalized `IndicatorState` (`on`, `level`, `blink`, `timeout`, `soundLevel`)
- `set(id, state)` — translates the state into a **single** Indicator Set frame with full-state-replacement semantics: aspects not included are turned off, matching the device behavior required by the spec. `on`/`level` are converted automatically between Binary and Multilevel indicators; unsupported aspects are rejected.
- `identify()` — delegates to the existing CC API. The Node Identify indicator (0x50) is hidden from all other surfaces.
- V1 nodes expose a single default indicator with ID 0 that can be turned on and off.

### Internals

- New internal `indicatorState(indicatorId)` CC value caching the complete property map per indicator. Properties missing from a report are assumed 0 as required by the spec (previously a TODO in `IndicatorCCReport`).
- New `"indicator updated"` node event, emitted from the freshly persisted state when an (unsolicited) Indicator Report is received, so the event payload always matches `getCached()`.
- Supervised or optimistic sets update the cache directly since no report follows.
- The mock Indicator CC now implements Set with the spec's reset-then-apply behavior, enabling the new integration tests.

The existing CC values are untouched; removing them in favor of this API is a later breaking change.

### DIFF
--- a/docs/api/endpoint.md
+++ b/docs/api/endpoint.md
@@ -948,6 +948,8 @@ Z-Wave JS provides a high-level API for controlling the indicators of a device, 
 The indicators API is accessed via the `indicators` property on endpoints and nodes. The property is `undefined` when the endpoint does not support the **Indicator CC**, so use it as the support check:
 
 ```ts
+import { Indicator } from "@zwave-js/core";
+
 if (endpoint.indicators) {
 	await endpoint.indicators.set(Indicator["Button 1 indication"], {
 		on: true,
--- a/docs/api/endpoint.md
+++ b/docs/api/endpoint.md
@@ -940,3 +940,164 @@ enum UserCredentialLearnStatus {
 	InvalidModifyOperationType = 0xff,
 }
 ```
+
+## Indicators
+
+Z-Wave JS provides a high-level API for controlling the indicators of a device, e.g. LEDs, an LCD backlight or a buzzer. It abstracts away the complexity of the **Indicator CC**, where the functionality of each indicator is split into multiple properties that need to be set in groups.
+
+The indicators API is accessed via the `indicators` property on endpoints and nodes. The property is `undefined` when the endpoint does not support the **Indicator CC**, so use it as the support check:
+
+```ts
+if (endpoint.indicators) {
+	await endpoint.indicators.set(Indicator["Button 1 indication"], {
+		on: true,
+	});
+}
+```
+
+The state of an indicator is represented as a whole:
+
+<!-- #import IndicatorState from "@zwave-js/cc" -->
+
+```ts
+interface IndicatorState {
+	/** Whether the indicator is on or off */
+	on?: boolean;
+	/** The level of a multilevel indicator (0-99) */
+	level?: number;
+	blink?: IndicatorBlink;
+	/** Timeout after which the indicator will be turned off. When setting, a string in the form `12h18m17.59s` is also accepted. */
+	timeout?: IndicatorTimeout | string;
+	/** The volume of an audible indicator (0-100). 0 means off/mute. */
+	soundLevel?: number;
+}
+```
+
+<!-- #import IndicatorBlink from "@zwave-js/cc" -->
+
+```ts
+interface IndicatorBlink {
+	/** Duration of one on/off period in seconds (0.1-25.5) */
+	period: number;
+	/** Number of on/off periods (1-254). `undefined` means the indicator blinks until turned off. */
+	cycles?: number;
+	/** On time within an on/off period in seconds. `undefined` means symmetric (on time equals off time). */
+	onTime?: number;
+}
+```
+
+<!-- #import IndicatorTimeout from "@zwave-js/cc" -->
+
+```ts
+interface IndicatorTimeout {
+	/** Whole hours (0-255) */
+	hours?: number;
+	/** Whole minutes (0-255) */
+	minutes?: number;
+	/** Whole and 1/100th seconds (0-59.99) */
+	seconds?: number;
+}
+```
+
+> [!NOTE] Nodes that only support V1 of the **Indicator CC** expose a single unspecified indicator with the ID `0`, which can only be turned on and off.
+
+### Querying indicator capabilities
+
+#### `getSupportedCached`
+
+```ts
+getSupportedCached(): IndicatorCapabilities[]
+```
+
+Returns the capabilities of all indicators the endpoint supports. This method uses cached information from the most recent interview.
+
+<!-- #import IndicatorCapabilities from "zwave-js" -->
+
+```ts
+interface IndicatorCapabilities {
+	/** The ID of the indicator. 0 is the unspecified indicator of a V1 node. */
+	indicatorId: number;
+	/** A user-friendly label describing the indicator */
+	label: string;
+	/** Whether the indicator can be turned on and off */
+	supportsOnOff: boolean;
+	/** Whether the indicator can be set to a level between 0 and 99 */
+	supportsLevel: boolean;
+	/** Whether the indicator can blink */
+	supportsBlinking: boolean;
+	/** Whether the indicator can be turned off automatically after a timeout */
+	supportsTimeout: boolean;
+	/** Whether the volume of the indicator can be changed */
+	supportsSoundLevel: boolean;
+}
+```
+
+The `Node Identify` indicator is not included; it is controlled exclusively through the [`identify`](#identify) method.
+
+#### `getCapabilitiesCached`
+
+```ts
+getCapabilitiesCached(indicatorId: number): IndicatorCapabilities | undefined
+```
+
+Returns the capabilities of the given indicator, or `undefined` if it is not supported. This method uses cached information from the most recent interview.
+
+### Reading and changing the indicator state
+
+#### `get`
+
+```ts
+get(indicatorId: number): Promise<IndicatorState | undefined>
+```
+
+Queries the current state of the given indicator.
+
+> [!NOTE] This communicates with the node to retrieve fresh information.
+
+#### `getCached`
+
+```ts
+getCached(indicatorId: number): IndicatorState | undefined
+```
+
+Returns the last known state of the given indicator, or `undefined` if it is not known. This method uses cached information.
+
+#### `set`
+
+```ts
+set(indicatorId: number, state: IndicatorState): Promise<SupervisionResult | undefined>
+```
+
+Changes the state of the given indicator. All necessary properties are sent to the device in a single command.
+
+> [!WARNING] The state is applied as a whole: all aspects not included in the state are turned off. For example, setting only `level` stops an ongoing blinking pattern and clears a previously set timeout. This matches the required behavior of supporting devices.
+
+Aspects of the state that the indicator does not support at all are rejected with a `ZWaveError`, with one exception: `on` and `level` are converted automatically between binary and multilevel indicators.
+
+#### `identify`
+
+```ts
+identify(): Promise<SupervisionResult | undefined>
+```
+
+Instructs the node to identify itself, e.g. by blinking an LED. Supported by nodes implementing version 3 or higher of the **Indicator CC**.
+
+### Indicator events
+
+#### `"indicator updated"`
+
+```ts
+(endpoint: Endpoint, args: IndicatorUpdatedArgs) => void
+```
+
+Emitted on the `ZWaveNode` instance when a device reports a changed indicator state, e.g. after it was changed locally. The `args` object contains the updated state along with the indicator ID:
+
+<!-- #import IndicatorUpdatedArgs from "zwave-js" -->
+
+```ts
+interface IndicatorUpdatedArgs extends Omit<IndicatorState, "timeout"> {
+	/** The ID of the updated indicator. 0 is the unspecified indicator of a V1 node. */
+	indicatorId: number;
+	timeout?: IndicatorTimeout;
+}
+```
--- a/packages/cc/src/cc/IndicatorCC.ts
+++ b/packages/cc/src/cc/IndicatorCC.ts
@@ -49,7 +49,12 @@ import {
 	useSupervision,
 } from "../lib/CommandClassDecorators.js";
 import { V } from "../lib/Values.js";
-import { IndicatorCommand, type IndicatorTimeout } from "../lib/_Types.js";
+import {
+	type IndicatorBlink,
+	IndicatorCommand,
+	type IndicatorState,
+	type IndicatorTimeout,
+} from "../lib/_Types.js";
 import type { CCEncodingContext, CCParsingContext } from "../lib/traits.js";
 
 function isManufacturerDefinedIndicator(indicatorId: number): boolean {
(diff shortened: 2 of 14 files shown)


packages/cc/src/cc/IndicatorCC.ts
```diff
@@ -133,6 +138,294 @@ function indicatorObjectsToTimeout(
...
+function timeoutToIndicatorObjects(
+	indicatorId: number,
+	timeout: IndicatorTimeout,
+	supportedPropertyIDs: readonly number[] | undefined,
+): IndicatorObject[] {
+	const objects: IndicatorObject[] = [];
+	const hours = timeout.hours ?? 0;
+	const minutes = timeout.minutes ?? 0;
+	const seconds = Math.floor(timeout.seconds ?? 0);
+	const hundredths = Math.round(((timeout.seconds ?? 0) % 1) * 100);
```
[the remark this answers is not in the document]
AlCalzone: Fixed in cb19d2e4c — seconds are now rounded to 2 decimals and clamped to 0-59.99 before splitting, so 12.999 carries over to 13 whole seconds instead of 100 hundredths. Hours/minutes are clamped to 0-255 as well.

docs/api/endpoint.md
```diff
@@ -940,3 +940,164 @@ enum UserCredentialLearnStatus {
...
+
+The indicators API is accessed via the `indicators` property on endpoints and nodes. The property is `undefined` when the endpoint does not support the **Indicator CC**, so use it as the support check:
+
+```ts
+if (endpoint.indicators) {
+	await endpoint.indicators.set(Indicator["Button 1 indication"], {
+		on: true,
+	});
+}
+```
```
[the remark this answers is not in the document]
AlCalzone: Fixed in cb19d2e4c — the snippet now includes the `import { Indicator } from "@zwave-js/core";` line.

packages/cc/src/cc/index.ts
```diff
@@ -614,6 +617,9 @@ export {
 	IndicatorCCSupportedGet,
 	IndicatorCCSupportedReport,
 	IndicatorCCValues,
+	indicatorObjectsToPropertyMap,
```
AlCalzone, line 620: **[Major] · Design / API**

`indicatorObjectsToPropertyMap` / `indicatorPropertyMapToState` / `indicatorStateToObjects` are exported here to the public `@zwave-js/cc` barrel and tagged `@publicAPI` (IndicatorCC.ts:221/250/308), exposing `Record<number, number>` property maps and raw property IDs.

**Why:** that raw representation is exactly what `endpoint.indicators` exists to hide; publishing it re-introduces the coupling and is breaking to walk back later. The only consumers (feature API + CC handler) already reach them via the subpath import.

**Fix:** keep them subpath-only — drop the barrel entries and `@publicAPI`.

packages/cc/src/lib/_Types.ts
```diff
@@ -972,6 +972,29 @@ export type IndicatorMetadata = ValueMetadata & {
...
+
+/** The normalized state of an indicator */
+export interface IndicatorState {
+	/** Whether the indicator is on or off */
+	on?: boolean;
+	/** The level of a multilevel indicator (0-99) */
+	level?: number;
+	blink?: IndicatorBlink;
+	/** Timeout after which the indicator will be turned off. When setting, a string in the form `12h18m17.59s` is also accepted. */
+	timeout?: IndicatorTimeout | string;
```
AlCalzone, line 993: **[Major] · Design / API**

`IndicatorState.timeout?: IndicatorTimeout | string` merges the write shape with the read shape. The string form is only ever a `set()` input — a device never reports it — so every `get()`/`getCached()`/event consumer must narrow a phantom `string`. This is what forces the awkward `IndicatorUpdatedArgs extends Omit<IndicatorState, "timeout">` + re-add.

**Why:** the read surface carries a value that can never occur on read, pushing defensive narrowing onto every consumer.

**Fix:** make the read type `timeout: IndicatorTimeout`, and accept the string only in `set()`'s parameter; then `IndicatorUpdatedArgs extends IndicatorState` cleanly. (Breaking later.)

packages/zwave-js/src/lib/node/feature-apis/Indicators.ts
```diff
@@ -0,0 +1,281 @@
...
+		const objects = indicatorStateToObjects(
+			indicatorId,
+			state,
+			supportedPropertyIDs,
+		);
+
+		const result = await this.#api().set(objects);
+		if (result == undefined || supervisedCommandSucceeded(result)) {
+			// No report will follow, so persist the new state ourselves
+			this.endpoint.tryGetNode()?.valueDB.setValue(
```
AlCalzone, line 152: **[Major] · Operability / Design**

A successful optimistic/supervised `set()` persists the new state here (and in `#setV1`, ~line 202) but never emits `"indicator updated"`.

**Why:** AccessControl re-emits its domain event on supervised success (AccessControl.ts:694/843/1225). A consumer treating the event as the source of truth misses every self-initiated change, and the PR's "payload always matches getCached()" invariant breaks — `getCached()` moves without an event.

**Fix:** decide the contract — emit after the optimistic/supervised persist in both `set()` and `#setV1()`, or document the event as device-originated-only.

packages/cc/src/cc/IndicatorCC.ts
```diff
@@ -1224,13 +1474,32 @@ export class IndicatorCCReport extends IndicatorCC {
 				}
 			}
 		} else if (this.values) {
-			// Store the simple values first
+			const groupedValues = groupByIndicatorId(this.values);
+
+			// Persist the full per-indicator state before setIndicatorValue
+			// converts the raw values to booleans in place
+			for (const [indicatorId, values] of groupedValues) {
```
AlCalzone, line 1481: **[Major] · Security / Correctness** (flagged by 4 lenses)

This loop persists `indicatorState` — and the handler (CCHandlers/IndicatorCC.ts:34-44) emits `"indicator updated"` — for every reported indicator ID except Node Identify, with no check against `supportedIndicatorIds`. `setIndicatorValue` (IndicatorCC.ts:1585) already guards exactly this.

**Why:** a malfunctioning/malicious node can seed ~256 cache entries per endpoint and inject events/state for indicators it never advertised (`getCached()` reads the map directly, bypassing the support filter). Bounded (byte IDs, ≤31/report) — cache bloat + spurious events, not a crash.

**Fix:** gate persistence + emission on the supported set; log at debug when a report names an unsupported ID.

packages/zwave-js/src/lib/node/feature-apis/Indicators.ts
```diff
@@ -0,0 +1,281 @@
...
+
+	#supportedPropertyIDs(indicatorId: number): number[] {
+		return this.getValue<number[]>(
+			IndicatorCCValues.supportedPropertyIDs(indicatorId).endpoint(
+				this.endpoint.index,
+			),
+		) ?? [];
+	}
+
+	#supportsV2Indicators(): boolean {
```
AlCalzone, line 231: **[Major] · DRY**

`#supportsV2Indicators()` re-implements the same gate as the inline block in CCHandlers/IndicatorCC.ts:64-76 and `IndicatorCC.supportsV2Indicators` (IndicatorCC.ts:1204) — "some supported indicator ID has a non-empty `supportedPropertyIDs`."

**Why:** it selects the whole code path in get/getCached/set and gates V1 report emission, so a rule change silently desyncs the three sites. (The handler recompute is also the Perf/duplication finding.)

**Fix:** one shared cached helper — alongside the existing static `getSupportedPropertyIDsCached` precedent — called from all three.

packages/zwave-js/src/lib/test/node/indicators.test.ts
```diff
@@ -0,0 +1,474 @@
...
+
+			// Unsupported indicators are rejected
+			await t.expect(
+				node.indicators!.set(0x44, { on: true }),
+			).rejects.toThrow("cannot be controlled");
+		},
+	},
+);
+
+integrationTest(
```
AlCalzone, line 293: **[Major] · Tests**

The "supervised set updates cache" test only exercises the `SupervisionResult` branch. The unsupervised path (`result == undefined`, devices without Supervision CC) runs the same optimistic-persist code untested.

**Why:** that branch would break silently — it's a real path for non-supervising devices.

**Fix:** add a non-supervised `set()` + `getCached()` assertion (also the natural place to cover the missing emit above).

packages/zwave-js/src/lib/node/feature-apis/Indicators.ts
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	/**
+	 * Returns the capabilities of the given indicator, or `undefined` if it is
+	 * not supported. This method uses cached information from the interview.
+	 */
+	public getCapabilitiesCached(
+		indicatorId: number,
+	): IndicatorCapabilities | undefined {
+		return this.getSupportedCached().find(
```
AlCalzone, line 68: **[Minor] · Performance**

`getCapabilitiesCached(id)` routes through `getSupportedCached()`, which builds capabilities for all supported indicators, then discards all but one via `.find()`.

**Why:** a per-indicator loop (e.g. a UI over all indicators) turns O(1) into O(n²). In-memory only, no I/O.

**Fix:** add a single-indicator path hitting `#supportedPropertyIDs` directly.

packages/cc/src/cc/IndicatorCC.ts
```diff
@@ -133,6 +138,297 @@ function indicatorObjectsToTimeout(
...
+		new ZWaveError(
+			`The indicator ${indicatorId} does not support ${what}`,
+			ZWaveErrorCodes.Argument_Invalid,
+		);
+
+	const ret: IndicatorObject[] = [];
+
+	// Binary and Multilevel are mutually exclusive property groups,
+	// so only one of them is sent
+	if (state.on === false) {
```
AlCalzone, line 325: **[Minor] · Correctness**

A blink-only indicator (0x03/0x04, no 0x01/0x02) can't be turned off: `set(id, {on:false})` throws "does not support being turned on or off", and `set(id, {})` throws "is empty".

**Why:** no API path stops an ongoing blink on such a device (the spec's stop is period/cycles = 0). Node Identify is already excluded, so this only hits real blink-only indicators.

**Fix:** for `on:false` with no on/off group, emit zeroed blink objects instead of throwing.

packages/zwave-js/src/lib/test/node/indicators.test.ts
```diff
@@ -0,0 +1,474 @@
...
+			await t.expect(
+				node.indicators!.set(0x43, { blink: { period: 1 } }),
+			).rejects.toThrow("does not support blinking");
+
+			// An empty state is rejected
+			await t.expect(node.indicators!.set(0x30, {})).rejects.toThrow(
+				"is empty",
+			);
+
+			// Unsupported indicators are rejected
```
AlCalzone, line 285: **[Minor] · Tests**

Unsupported-indicator rejection is covered (0x44), but Node Identify (0x50) rejection isn't — the special-case at Indicators.ts:272 would regress silently.

**Fix:** assert `set(0x50, …)` rejects.

packages/zwave-js/src/lib/node/feature-apis/Indicators.ts
```diff
@@ -0,0 +1,281 @@
...
+	#assertDefaultIndicator(indicatorId: number): void {
+		if (indicatorId !== 0) {
+			throw new ZWaveError(
+				`This node only supports the default indicator with ID 0`,
+				ZWaveErrorCodes.Argument_Invalid,
+			);
+		}
+	}
+
+	#assertSupportedIndicator(indicatorId: number): void {
```
AlCalzone, line 270: **[Minor] · Design / Operability**

`#assertSupportedIndicator` gives the same "cannot be controlled" message for an unsupported indicator and for Node Identify, with no pointer to `identify()`. Relatedly, `identify()` (line 167) has no support guard, unlike siblings that throw `CC_NotSupported`.

**Fix:** special-case the Node Identify message to point at `identify()`; add a support guard to `identify()`.

packages/cc/src/cc/IndicatorCC.ts
```diff
@@ -133,6 +138,297 @@ function indicatorObjectsToTimeout(
...
+	supportedPropertyIDs: readonly number[],
+): IndicatorState {
+	const supports = (propertyId: number) =>
+		supportedPropertyIDs.length
+			? supportedPropertyIDs.includes(propertyId)
+			: propertyId in map;
+
+	const ret: IndicatorState = {};
+
+	if (supports(0x02 /* Binary */)) {
```
AlCalzone, line 263: **[Nit] · Readability / DRY**

Property-ID literals `0x01`–`0x0a` are inlined across functions here, and the timeout-property set `[0x0a, 0x06, 0x07, 0x08]` is re-encoded in the capability logic (Indicators.ts:254).

**Why:** the capability flags and the converters must agree; a new timeout unit means editing multiple spots.

**Fix:** optional — a named set/const would harden against drift. Matches existing precedent (no property-ID enum exists today), so low priority.

AlCalzone (commented):
**Deep review — Verdict: Ship with minor changes.** Approach is sound (mirrors the AccessControl feature-API precedent); no blockers. Two items are cheap now but become breaking changes once shipped — fix pre-merge.

**Top 3 risks**
1. Public API surface locks in avoidable debt — raw property-map converters leaked to the `@zwave-js/cc` barrel, and `IndicatorState.timeout` conflates read/write shapes.
2. Report path persists + emits for any device-reported indicator ID without filtering to the supported set (flagged by 4 lenses), bypassing the guard `setIndicatorValue` already applies.
3. `set()` moves `getCached()` but never emits `"indicator updated"`, breaking the sibling convention and the PR's own "event always matches getCached()" invariant.

**Strengths:** clean match to the AccessControl shape (mixin getter, `FeatureAPI` base, event dispatch); `timeoutToIndicatorObjects` genuinely extracted and shared; report parser well-bounded (`objCount` ≤ 31, `validatePayload` guards); solid Binary/Multilevel conversion tests.

**Themes:** report-ID filtering (Security/Correctness/Operability); duplicated V1↔V2 detection (DRY/Perf); public-contract decisions cheap now / breaking later.

**FYI (no inline thread)**
- `IndicatorCC.ts:333` — `{on:true, level:0}` sends 0 (off) despite `on:true`; documented "level wins" but a footgun.
- `IndicatorCC.ts:349-356` — `{on:true}` on a Multilevel-only indicator reads back `{on:true, level:99}`; Binary→Multilevel isn't round-trip identity.
- `Indicators.ts set()` — legacy per-property values (`valueV2`, `timeout`) go stale after an optimistic set; intended per the PR body, but worth a doc note on the transition-period divergence.

**Coverage:** 8 specialists ran; no findings of note on generated values, docs (verified accurate against the API), mixin wiring, or the mock (verified reset-then-apply).

PR #8950, opened 2026-07-09
refactor: centralize complex log formatting

AlCalzone:
Log message construction was hand-rolled at ~250 call sites: `\n· ` bullet lists, manually space-aligned property blocks, `logLines` array joins, and CC tree rendering duplicated between the driver and zniffer loggers. This PR ports the composable log payload model from zwave-rs and migrates all call sites.

### New API in `@zwave-js/core`

- `LogPayload = LogPayloadText | LogPayloadDict | LogPayloadList`, each text/dict optionally carrying nested payload(s)
- Factories `logText()`, `logDict()`, `logList()`, plus `mergeLogDict()` for extending an inherited dict (`undefined` values remove entries)
- `MessageOrCCLogEntry.message` still accepts a bare object literal, so most `toLogEntry()` implementations are unchanged; `undefined` entries are skipped, removing the `if (x != undefined) message.x = ...` dance
- A single formatter (`formatLogPayload`) handles per-dict key alignment, `· ` bullets, 2-space indent levels, and `└─`/`│` tree borders for tagged nested payloads

### Changes

- CC tree rendering moved from the driver/zniffer loggers into a shared `ccToLogPayload()` helper
- All bullet-list, `logLines`, and hand-aligned template-literal sites migrated to structured payloads
- S2 extensions now render as tree children (`└─[SPAN]`) instead of inline strings
- `MessageRecord` and `messageRecordToLines` removed

Rendered output stays byte-identical for the common cases (flat dicts, CC trees). Deliberate normalizations: list values in dicts indent their bullets 2 spaces under the key; previously hand-aligned property blocks now align automatically.


packages/core/src/log/LogPayload.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+export type LogPayloadDictValue =
+	| string
+	| number
+	| boolean
+	| LogPayloadList;
+
+/** Entries with `undefined` values are skipped, so optional entries can be included inline */
+export type LogPayloadDictInput = Record<
```
AlCalzone: We should rename this to `MessageRecord`, since the original type with that name was removed.

packages/serial/src/serialapi/capability/SerialAPISetupMessages.ts
```diff
@@ -131,15 +131,14 @@ export class SerialAPISetupRequest extends Message {
...
-			message.payload = `0x${this.payload.toString("hex")}`;
-		}
 		return {
 			...super.toLogEntry(),
-			message,
+			message: {
+				command: getEnumMemberName(SerialAPISetupCommand, this.command),
+				payload: this.payload.length > 0
+					? `0x${this.payload.toString("hex")}`
+					: undefined,
```
AlCalzone: We should build a helper `logBuffer` that formats a payload as hex with 0x prefix.

packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1090,16 +1094,21 @@ export class ZWaveController
...
+						"product type": num2hex(this._productType),
+						"product ID": num2hex(this._productId),
+						"supported functions": logList(
+							this._supportedFunctionTypes.map(
+								(fn) => `${FunctionType[fn]} (${num2hex(fn)})`,
+							),
+						),
+					}),
+				}),
+			).join("\n"),
```
AlCalzone: All loggers should accept a LogPayload instead of requiring the callers to do `formatLogPayload(...).join("\n")` first.

packages/zwave-js/src/lib/zniffer/MPDU.ts
```diff
@@ -616,11 +613,13 @@ export class MulticastZWaveMPDU extends ZWaveMPDU {
...
-			destinations: this.destinationNodeIds.join(", "),
-			...original,
+		const message = mergeLogDict(original, {
 			payload: buffer2hex(this.payload),
-		};
+		});
+		message.entries.unshift([
+			"destinations",
+			this.destinationNodeIds.join(", "),
+		]);
```
AlCalzone: Lets have `mergeLogDict` take a variable number of arguments, and merge them in sequence. That way we can put destinations in the first position without the awkward unshift.

packages/core/src/log/LogPayload.ts
AlCalzone: Renamed to `MessageRecord` in 1d2030181.

packages/serial/src/serialapi/capability/SerialAPISetupMessages.ts
AlCalzone: Added `logBuffer()` in 1d2030181 and applied it to the conditional payload entries (returns `undefined` for empty buffers so the entry is skipped).

packages/zwave-js/src/lib/controller/Controller.ts
AlCalzone: Done in 1d2030181 — `print()` on the controller, driver, and zniffer loggers now accepts `string | LogPayload | MessageRecord`; the `formatLogPayload(...).join("\n")` call sites are gone.

packages/zwave-js/src/lib/zniffer/MPDU.ts
AlCalzone: `mergeLogDict` is variadic now (1d2030181); the multicast MPDU passes `destinations` as the first source instead of unshifting.

packages/core/src/log/formatPayload.test.ts
```diff
@@ -0,0 +1,163 @@
...
+			logDict({
+				"supported types": logList(["Heating", "Cooling"]),
+				mode: "auto",
+			}),
+		),
+	).toStrictEqual([
+		"supported types:",
+		" · Heating",
+		" · Cooling",
+		"mode:            auto",
```
AlCalzone: I think we should ignore keys that only contain a list for the purpose of column width calculation.
```suggestion
	).toStrictEqual([
		"supported types:",
		" · Heating",
		" · Cooling",
		"mode: auto",
```

AlCalzone: Done in 08221df9b — list-valued keys are now excluded from the value-column width calculation, so `mode: auto` aligns to the string entries only.

PR #8954, opened 2026-07-09
docs: correct dead-node status transition

AlCalzone:
## Summary
- `docs/api/node.md` said a message from a dead node marks it as `unknown`. The code marks it `alive` instead.
- Confirmed in `Driver.ts` (`handleRequest` and the send-success handler), both of which call `node.markAsAlive()` for a node in `Dead` status, and in `NodeStatusMachine.ts`, whose `dead` state only transitions on `ALIVE` → `alive` (no `unknown` intermediate).

Found while checking the zwave-js-bot's automated answer in https://github.com/zwave-js/zwave-js/discussions/8952, which quoted the (incorrect) docs verbatim.

## Test plan
- Docs-only change, no functional code touched.


AlCalzone:
@zwave-js-bot automerge

Issue #5583, opened 2023-03-16
Add support for changing config parameter definitions based on the value of others


AlCalzone:
This came up [here](https://github.com/zwave-js/node-zwave-js/pull/5567#discussion_r1138372392) but is also required for some HELTUN devices where config parameters change their definition entirely based on the value of others.

AlCalzone:
Another example: Aeotec Multisensor 6 has changing default values based on the selected temperature unit

AlCalzone:
Another example:
Ezlo thermostat - https://github.com/user-attachments/files/20024840/Ezlo_Thermostat_Full_Manual_11052024.pdf

AlCalzone:
Another use case is hiding/showing endpoints:
https://github.com/zwave-js/zwave-js-ui/issues/3881#issuecomment-3672146787

PR #1535, opened 2021-01-30
chore(config): check if param description contains unnecessary stuff

AlCalzone:
@robertsLando Like we talked about. This found a ton of warnings - I could use some help here:
```Found 0 errors and 470 warnings!```

@blhoward2 do you have a script to clean the descriptions?

Issue #71, opened 2019-05-04
Add a way to translate numeric property keys to speaking names


AlCalzone:
As of #70 we now have property keys in the value DB. It is preferable to store the raw numeric keys, but for applications we need to expose speaking names for them.

PR #8957, opened 2026-07-10
feat: support referencing config parameter values in $if conditions

AlCalzone:
Some devices change the definition of config parameters based on the value of other parameters (#5583: Heltun, Aeotec MultiSensor 6, Ezlo thermostat). This adds support for referencing the current value of config parameters in `$if` conditions, and re-evaluates the device config at runtime when a referenced value changes. Scope is limited to config parameter definitions and compat flags; endpoint visibility is a possible follow-up.

## Syntax

Param references reuse the syntax of `paramInformation` `"#"` keys, prefixed with `#`. They may only be compared with number literals:

```json
"$if": "#71 == 1"
"$if": "#9[0x0f] > -1"
"$if": "#40 >= 2 && firmwareVersion >= 1.7"
```

- References resolve against the endpoint whose config section the condition appears in; conditions inside `endpoints.<n>` use that endpoint's own parameters.
- Partial values are read from the stored partial value IDs, so signedness follows the existing machinery.
- **Comparisons involving unknown values evaluate as if they didn't exist (true)**, so guarded definitions apply during a first interview and are hidden afterwards if their conditions turn out unsatisfied.

## Runtime behavior

- Param-value conditions never affect the interview structure. The Configuration CC interview queries the referenced params early (after the V3+ parameter scan resp. before the value refresh on V1/V2), so the rest of the interview uses the correct definitions.
- The node keeps the un-evaluated `ConditionalDeviceConfig` and re-evaluates it synchronously when a referenced Configuration value changes (filtered by a parse-time collected reference set, deduplicated via a value snapshot). Param metadata is refreshed dynamically, compat flags are picked up on access — no re-interview, no new events.
- Newly appearing params with unknown values are queried automatically at `NodeQuery` priority (`Poll` priority would enforce the spec-mandated 10 s polling delays).
- Hash v4 excludes param information, so `hasDeviceConfigChanged()` is unaffected. If a user-provided config gates interview-relevant settings on a param value anyway, the driver logs a warning instead of re-interviewing.

## Lint

`yarn lint:zwave` enforces that references are only used where application is dynamic (param definitions/options and compat flags outside the hashed set), that referenced params exist in the same scope and are not `hidden`, that conditions form no reference cycles, plus bitmask sanity and always-true/false comparison warnings. Files with references are additionally lint-evaluated with each compared value satisfied, one param at a time.

## Tests

- Unit tests for the tokenizer/parser, the unknown ⇒ true semantics incl. partials, negative values and endpoint scoping, and `ConditionalDeviceConfig` evaluation.
- Mock-node integration tests: value-driven metadata/compat switching without re-interview, early querying + automatic catch-up queries + chained dependencies (V2 device), and endpoint-scoped resolution.

Docs: new section in `docs/config-files/conditional-settings.md`.


docs/config-files/conditional-settings.md
```diff
@@ -17,6 +17,34 @@ The logic supports the JavaScript operators `<`, `<=`, `>`, `>=` and `===`, as w
...
+
+```json
+{
+	// Applies when parameter 71 is currently set to 1
+	"$if": "#71 == 1",
+	// Applies when the partial parameter 9, bitmask 0x0f is greater than -1
+	"$if": "#9[0x0f] > -1",
+	// Can be combined with other conditions
+	"$if": "#40 >= 2 && firmwareVersion >= 1.7"
+}
```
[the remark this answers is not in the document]
AlCalzone: Yes, but an array of objects is also not valid

packages/cc/src/cc/ConfigurationCC.ts
```diff
@@ -479,6 +484,71 @@ export function refreshConfigParamMetadataFromConfigFile(
...
+			&& id.property === parameter
+			&& typeof id.propertyKey === "number",
+	);
+	if (partials.length === 0) return undefined;
+	let ret = 0;
+	for (const { propertyKey: bitMask, value: partialValue } of partials) {
+		if (typeof partialValue !== "number") return undefined;
+		ret = encodePartial(ret, partialValue, bitMask as number);
+	}
+	return ret;
```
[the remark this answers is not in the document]
AlCalzone: Fixed in f09e1330b. Rather than reinterpret in the getter, the real issue is referencing a param differently from its definition. Lint now requires a reference to match the definition shape (full `#N` vs. partial `#N[mask]`), so the cached value is always read directly with the correct signedness — the compose-from-partials path is gone.

docs/config-files/conditional-settings.md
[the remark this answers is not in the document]
AlCalzone: Fixed in f09e1330b — split into separate `{ "$if": ... }` example objects.

packages/config/maintenance/lintConfigFiles.ts
```diff
@@ -661,11 +1009,55 @@ async function lintDevices(): Promise<void> {
...
+					firmwareVersion: conditionalConfig.firmwareVersion.min,
+				}));
+				variants.push(...baseVariants);
+			}
+
+			const seen = new Set<string>();
+			for (const ref of paramValueRefs) {
+				let fullValue: number;
+				if (ref.valueBitMask != undefined) {
+					const maxValue = 2 ** getBitMaskWidth(ref.valueBitMask) - 1;
```
AlCalzone: **[Major] · Correctness (lint)**

The variant fan-out skips every partial reference with a negative comparand (`ref.comparand < 0 || … continue`), assuming partials are unsigned. But `isSignedPartial` makes multi-bit partials **signed by default** (width > 1, format defaults to `SignedInteger`) — negative comparands are the intended case, which is exactly why the PR added the `Minus` token and `parseSignedNumberLiteral`.

**Why:** `evaluateDeep` selects only the *first* applying definition, so the fan-out exists to expose shadowed (non-first) defs by satisfying each comparand. A shadowed def guarded by e.g. `"$if": "#9[0x0f] == -3"` is never selected in any variant, so its `minValue`/`options`/`allowed` errors pass lint and ship. `lintVariantToContext` (line 230) compounds this — it decodes partials with `parsePartial(…, false)` unconditionally.

**Fix:** derive signedness from the referenced param's `unsigned` flag (mirror `isSignedPartial`) and generate the variant across the signed legal range via `getLegalRangeForBitMask` (already imported); apply the same in `lintVariantToContext`.

packages/config/maintenance/lintConfigFiles.ts
```diff
@@ -198,14 +210,343 @@ function unconditionalComesLast(
...
+					addWarning(
+						file,
+						`${errorPrefix}: the condition references ${refString} differently from some of the parameter's definitions`,
+					);
+				}
+			}
+
+			// Values of partial references span [0...2^width - 1]
+			if (ref.valueBitMask != undefined) {
+				const maxValue = 2 ** getBitMaskWidth(ref.valueBitMask) - 1;
```
AlCalzone: **[Minor] · Correctness (lint)**

`constantComparisonOutcome` (line 236) hardcodes the partial range as `[0, maxValue]` with `maxValue = 2 ** width - 1`, ignoring signedness. For a signed multi-bit partial the real range is `[-2**(w-1), 2**(w-1)-1]`.

**Why:** the PR's own documented example `#9[0x0f] > -1` is flagged always-true, and a satisfiable `#9[0x0f] === -3` would be flagged always-false. Warnings don't fail `yarn lint:zwave`, but they push maintainers to "fix" correct conditions.

**Fix:** use `getLegalRangeForBitMask(mask, unsigned)` with `unsigned` from the referenced param, defaulting to signed for width > 1.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
```diff
@@ -259,4 +353,195 @@ export abstract class DeviceConfigMixin extends FirmwareUpdateMixin
...
+	}
+
+	private _reevaluatingDeviceConfig = false;
+	private _reevaluateDeviceConfigAgain = false;
+
+	/**
+	 * Re-evaluates the conditional device config against the current parameter
+	 * values and applies the result if they changed since the last evaluation
+	 */
+	private reevaluateDeviceConfig(): void {
```
AlCalzone, line 406: **[Major] · Correctness / Operability** (flagged by both specialists)

`reevaluateDeviceConfig`/`doReevaluateDeviceConfig` runs synchronously inside a `valueDB.on("value updated"/…)` listener with no try/catch around `evaluate()`, `refreshConfigParamMetadataFromConfigFile`, or `queryUnknownConfigParamValues`. On the native `TypedEventTarget` (EventTarget-backed), a throw here surfaces as an *uncaught exception* rather than a contained failure. The fire-and-forget `void updateCurrentDeviceConfigHash()` and the spawned `queryUnknownConfigParamValues()` carry the same unhandled-rejection exposure.

**Why:** the config is evaluated once at load, so this is low-probability for shipped configs but higher for un-linted user configs / unexpected value shapes — and a single bad re-eval can take down value processing for the node.

**Fix:** wrap the body, log via `logNode(…, "error")`, and skip the update instead of propagating.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
```diff
@@ -259,4 +353,195 @@ export abstract class DeviceConfigMixin extends FirmwareUpdateMixin
...
+			this._lastReferencedParamValues
+			&& [...snapshot].every(([key, value]) =>
+				this._lastReferencedParamValues!.get(key) === value
+			)
+		) {
+			return;
+		}
+		this._lastReferencedParamValues = snapshot;
+
+		this.deviceConfig = conditionalConfig.evaluate(context);
```
AlCalzone, line 467: **[Major] · Operability**

Adding, hiding, revealing, or changing a param's metadata at runtime produces no log output — nothing states which referenced param's new value triggered the change or what changed. This is the exact "3am" case: a param appears/disappears after a value update with nothing in the log to explain why.

**Why:** silent metadata mutation is very hard to diagnose; the `_lastReferencedParamValues` diff needed for a good message is already computed.

**Fix:** log at debug/info in `doReevaluateDeviceConfig` naming the triggering param/value, plus a per-param line when a param is newly hidden/revealed/changed.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
```diff
@@ -259,4 +353,195 @@ export abstract class DeviceConfigMixin extends FirmwareUpdateMixin
...
+		if (!config) return;
+		const oldHash = this._currentDeviceConfigHash;
+		const newHash = await config.getHash(
+			this._currentDeviceConfigHashVersion,
+		);
+		this._currentDeviceConfigHash = newHash;
+		if (oldHash && !DeviceConfig.areHashesEqual(oldHash, newHash)) {
+			// The lint forbids gating interview-relevant settings with parameter
+			// values, but user-provided configs may do it anyway
+			this.driver.controllerLog.logNode(
```
AlCalzone, line 599: **[Minor] · Operability**

The re-interview warning names neither the triggering parameter nor the affected setting, and gives no remediation step.

**Why:** a user can't tell which param/config caused it or what to do about it.

**Fix:** include the referenced param(s) from the `_lastReferencedParamValues` diff and suggest re-interviewing the node.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
```diff
@@ -259,4 +353,195 @@ export abstract class DeviceConfigMixin extends FirmwareUpdateMixin
...
+		// the Configuration CC interview
+		if (this.interviewStage === InterviewStage.Complete) {
+			this.queryUnknownConfigParamValues();
+		}
+
+		void this.updateCurrentDeviceConfigHash();
+	}
+
+	/** Queries the values of readable config params that have no cached value yet */
+	private queryUnknownConfigParamValues(): void {
```
AlCalzone, line 499: **[Minor] · Correctness / Security**

`queryUnknownConfigParamValues` has no in-flight guard and swallows failures silently (`.catch(() => undefined)`, no log). When a queried param is itself referenced, its persisted value re-triggers `reevaluateDeviceConfig` → a fresh unknown-scan while the prior async loop is still running → duplicate concurrent `get()`s. On a cyclic *user* config (lint only guards shipped configs) this becomes a perpetual query loop (X guarded by `#X != 5`): idempotent and non-crashing, but indefinite RF traffic / battery drain. A failed catch-up query on an asleep/dead node also leaves the param unqueried with no trace.

**Why:** wasteful at best, runaway at worst on un-linted configs; unlike the interview-time query it logs nothing.

**Fix:** track in-flight params and skip queued ones; log per attempt + warn on failure (mirror the interview path); consider detecting reference cycles at load and refusing dynamic re-eval.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
```diff
@@ -259,4 +353,195 @@ export abstract class DeviceConfigMixin extends FirmwareUpdateMixin
...
+		) {
+			return;
+		}
+		this._lastReferencedParamValues = snapshot;
+
+		this.deviceConfig = conditionalConfig.evaluate(context);
+
+		// Apply changed param metadata. Compat flags are read on access and
+		// need no further action.
+		for (const ep of this.getAllEndpoints()) {
```
AlCalzone, line 472: **[Minor] · Performance**

Each referenced-value change rebuilds the *entire* `DeviceConfig` (`conditionalConfig.evaluate`) and refreshes metadata for *every* endpoint, even though references are lint-restricted to their own endpoint scope and to `paramInformation`/`compat`. `refreshConfigParamMetadataFromConfigFile` ends in `getAllMetadata(Configuration)` — an O(node value-index) scan — so cost scales E× (endpoint count) per event.

**Why:** bounded and rare, but a multi-endpoint device (E~8–16) does that many full scans per param change.

**Fix:** scope the refresh to the triggering endpoint and the `paramInformation`/`compat` subtrees.

packages/cc/src/cc/ConfigurationCC.ts
```diff
@@ -1415,6 +1445,33 @@ export class ConfigurationCC extends CommandClass {
 			);
 		}
 
+		// Query the values of parameters that conditions in the config file depend on,
+		// so the rest of the interview uses the correct parameter definitions
+		const referencedParams = deviceConfig?.referencedParamValues
+			?.get(this.endpointIndex);
+		if (referencedParams?.size) {
+			for (const parameter of referencedParams) {
```
AlCalzone, line 1492: **[Minor] · Performance**

The eager referenced-param query loop fetches each referenced param, then `refreshValues` (called just below) re-queries all readable params — including the ones just fetched.

**Why:** one avoidable round-trip per referenced param, doubling latency on slow/battery devices (bounded by `referencedParams.size`).

**Fix:** have `refreshValues` skip params already resolved in this interview pass.

packages/config/src/devices/DeviceConfig.ts
```diff
@@ -799,7 +884,7 @@ export class DeviceConfig {
 			rootDir: string;
 			fallbackDirs?: string[];
 			relative?: boolean;
-			deviceId?: DeviceID;
+			context?: ConditionalConfigContext;
```
AlCalzone, line 961: **[Minor] · Design / BackCompat**

`DeviceConfig.from`'s option was renamed `deviceId` → `context`. TS callers get a compile error, but an untyped JS caller still passing `{ deviceId }` silently gets an undefined context → *all* conditions apply.

**Why:** silent-wrong-result for untyped callers. Low-probability in a TS-first ecosystem, and the repo doesn't preserve backcompat by default — flagging only for release-note visibility. (The `evaluate()`/`evaluateCondition()` renames are safe widenings since `ConditionalConfigContext extends DeviceID`.)

**Fix:** none needed if intended; consider a release-note mention.

packages/config/src/devices/shared.ts
```diff
@@ -10,3 +10,19 @@ export interface DeviceID {
...
+	 * Returns the cached value of a config parameter, or `undefined` if it is not known.
+	 * Comparisons involving unknown parameter values evaluate as if they didn't exist.
+	 */
+	getCachedParamValue?: (
+		endpoint: number,
+		parameter: number,
+		bitMask?: number,
+	) => number | undefined;
+
+	/** @internal The endpoint scope param references are resolved against during evaluation */
```
AlCalzone: **[Minor] · Design**

`ConditionalConfigContext.endpoint` is mutable recursion state living on the public context type next to device identity + the value callback, and the name collides with `ValueID.endpoint`.

**Why:** conflates "logical evaluation context" with "current recursion scope" on an exported type. It's handled correctly (spread immutably in `EndpointConfig`), so this is a clarity concern, not a bug.

**Fix:** consider `_endpointScope`, or thread scope as a separate arg to `evaluateDeep`.

packages/zwave-js/src/lib/test/node/Node.paramValueConditions.test.ts
```diff
@@ -0,0 +1,352 @@
...
+				property: 3,
+			}) as ConfigurationMetadata;
+			t.expect(meta3After.label).toBe("Mode B only");
+
+			t.expect(node.deviceConfig?.compat?.manualValueRefreshDelayMs).toBe(
+				1000,
+			);
+
+			// The changes are applied dynamically and do not require a re-interview
+			t.expect(node.hasDeviceConfigChanged()).toBeFalsy();
```
AlCalzone, line 127: **[Major] · Tests**

The feature's headline guarantee — value-driven metadata/compat switching happens *without* a re-interview — is only indirectly checked here (`hasDeviceConfigChanged()` is a hash compare, not evidence a re-interview didn't fire), and tests 2/3 don't check it at all. A regression that triggers `refreshInfo()` alongside the metadata swap would pass. The reentrancy guard (`80_DeviceConfig.ts:394-416`) and the asleep/dead early-return in `queryUnknownConfigParamValues` are also never exercised.

**Why:** the core promise and the trickiest code path go unverified.

**Fix:** assert `node.interviewStage === InterviewStage.Complete` and/or inspect mock-node received-frame counts after each value change; add a fixture where a *referenced* param is itself conditionally defined so its value is removed mid-re-eval, and assert the queued re-run's final state.

packages/config/src/Logic.test.ts
```diff
@@ -1,6 +1,27 @@
...
+		...context,
+		getCachedParamValue: (
+			endpoint: number,
+			parameter: number,
+			bitMask?: number,
+		) => {
+			const fullValue = paramValues[`${endpoint}:${parameter}`];
+			if (fullValue == undefined) return undefined;
+			if (bitMask == undefined) return fullValue;
+			return parsePartial(fullValue, bitMask, false);
```
AlCalzone: **[Minor] · Tests**

`withParamValues` decodes partials with `parsePartial(fullValue, bitMask, false)` — unsigned — so signed multi-bit partials (the default) are never exercised, and no test combines a param reference with `||` or a parenthesized group.

**Why:** a regression that flips/drops sign-extension for referenced partials, or one that stops `walk` recursing into `or` operands (silently dropping params from the auto-query trigger set), would pass every existing test.

**Fix:** add a signed-partial case, and a `#1 == 0 || #2 == 1` case (one side unknown ⇒ expect true) asserting `referencedParamValues` contains both params.

docs/config-files/conditional-settings.md
```diff
@@ -17,6 +17,34 @@ The logic supports the JavaScript operators `<`, `<=`, `>`, `>=` and `===`, as w
...
+{ "$if": "#71 == 1" }
+
+// Applies when the partial parameter 9, bitmask 0x0f is greater than -1
+{ "$if": "#9[0x0f] > -1" }
+
+// Can be combined with other conditions
+{ "$if": "#40 >= 2 && firmwareVersion >= 1.7" }
+```
+
+Parameter references may only be compared with number literals (no versions). A parameter must be referenced the same way it is defined: reference a full parameter with `#<param>` and a partial parameter with `#<param>[<bitmask>]` using a bitmask that matches a defined partial. Partial parameter values are interpreted with the signedness of the parameter definition, like elsewhere in Z-Wave JS.
```
AlCalzone, line 56: **[Nit] · Docs**

The syntax section doesn't state that the reference must be the left operand — `1 == #71` is a parse error.

**Fix:** add a clause noting the parameter reference goes on the left-hand side.

AlCalzone (commented):
## Deep review — feat: reference config parameter values in `$if` (#8957)

**Verdict:** Ship with minor changes. The approach is sound — `@zwave-js/config` stays a pure `(definition, context)` function, the runtime value callback (`getCachedParamValue`) keeps `@zwave-js/cc`/ValueDB out of the config package, and the new `param ===`/`param <` operations mirror the existing `ver ===` registration. `f09e1330b` already resolved the runtime partial-signedness handling and removed the duplicated partial composition, so this review is against that commit.

**Top risks**
1. The lint + test helpers still model partials as **unsigned**, while multi-bit partials default to **signed** — so the variant fan-out skips signed-partial cases (shadowed defs ship unvalidated) and the documented `#9[0x0f] > -1` is mis-flagged always-true.
2. Runtime re-eval runs synchronously inside a valueDB listener with no error containment — a throw escapes as an uncaught exception.
3. The headline "no re-interview" guarantee isn't actually asserted in 2 of 3 mock-node tests.

**Strengths:** clean dependency boundary; correct reentrancy guard (`do/while` re-snapshot, reset in `finally`); `unknown ⇒ true` consistent across all six operators and documented in three places; the new shape-match lint check is a good tightening.

**Theme — signedness:** now confined to lint + tests after the `f09e1330b` runtime fix (two lint threads + the Logic.test thread below).

**Out of scope (not introduced here, no action needed for this PR):**
- `LogicParser.ts` `parseExpression` is exponential on nested parentheses: `parseOr || parseAnd || parseComparison || parseGroup` with `parseGroup → parseExpression` re-parses each paren level 4× (`4^depth`; ~15 levels ≈ 27 s, ~20 ≈ hours from a short string). Pre-existing — this PR neither introduced it nor widens the trigger. Worth a follow-up to bound parse depth/length, since an un-linted user override could freeze the driver synchronously at config load.
- Config files using `#`-ref conditions throw on any `@zwave-js/config` bundle predating this feature; since config updates ship independently of the driver, worth a note on the config release channel.

**Coverage:** 10 specialists (correctness ×3, security, performance, DRY, design, tests, operability, readability). The two readability findings were pre-existing (unit-description block from #1535, a typo outside the diff) and dropped. The DRY finding (duplicated partial composition) was resolved by `f09e1330b`.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
```diff
@@ -259,4 +353,195 @@ export abstract class DeviceConfigMixin extends FirmwareUpdateMixin
...
+		);
+		if (config.endpoints) {
+			for (const [index, ep] of config.endpoints) {
+				if (index === 0) continue;
+				collect(index, ep.paramInformation);
+			}
+		}
+		if (queries.length === 0) return;
+
+		void (async () => {
```
AlCalzone, line 550: **[Major] · Security/Correctness**

A user-overridden config with a cyclic or self-referential `#` reference whose queried value fails its own gate loops forever: this detached async query returns → the param's own definition flips off → `refreshConfigParamMetadataFromConfigFile` calls `removeValue` → value-removed event → re-eval → param unknown ⇒ condition `true` → defined again → re-queried here again.

**Why:** Perpetual `Configuration Get` storm; lint is the only cycle guard and user-provided configs bypass it, contradicting the "safe even if lint is bypassed" contract. The sync re-entrancy guard is monotonic and safe, but this async path re-introduces the value uncapped.

**Fix:** Enforce cycle/self-ref safety at config load independent of lint, don't auto-query params that are themselves referenced, or bound re-eval/query iterations per node. (Security specialist rated this a Blocker under the untrusted-config threat model; I set Major because it needs a hand-authored lint-violating local config and has no remote trigger — your call on the ladder.)

packages/config/src/devices/shared.ts
```diff
@@ -10,3 +10,19 @@ export interface DeviceID {
 	firmwareVersion?: string;
 	sdkVersion?: string;
 }
+
+/** The context that `$if` conditions in config files are evaluated against */
+export interface ConditionalConfigContext extends DeviceID {
```
AlCalzone, line 15: **[Major] · Design/API**

`ConditionalConfigContext` is a new public interface (re-exported from `index.ts` via `export type * from "./devices/shared.js"`, the same channel as `DeviceID`) but is missing from `packages/config/config.api.md`, which this PR does not touch.

**Why:** api-extractor verify will fail CI; the new public type — including the `@internal endpoint` field — is unreviewed public surface.

**Fix:** Regenerate the API report and confirm the `endpoint`/`getCachedParamValue` shape is the intended public contract.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone, line 499: **[Major] · Performance/Operability**

`queryUnknownConfigParamValues` spawns a detached async loop (line 510) with no in-flight/dedup guard and is invoked from every re-eval; the sync re-entrancy guard does not cover it.

**Why:** Two close value events (bulk restore, multiple reports, interview bursts) re-scan for "still unknown" before the first `Get` resolves and issue duplicate `Get`s; a permanently-unresponsive param is silently re-queried on every future re-eval with no memory. Same missing-guard root as the loop finding.

**Fix:** Track in-flight `endpoint:parameter` queries and skip them; skip params that already failed to respond since the last success.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone, line 467: **[Major] · Operability**

Re-eval swaps `deviceConfig` here and adds/removes/re-ranges params just below, with no log at any level; only the re-interview-warning branch logs.

**Why:** In a widely-deployed driver, a param appearing/disappearing mid-session because a referenced value flipped is invisible in the user's log and can't be correlated to the value that caused it.

**Fix:** Log which referenced param(s) changed and that the config was re-evaluated, at verbose/silly.

packages/config/maintenance/lintConfigFiles.ts
```diff
@@ -198,14 +210,343 @@ function unconditionalComesLast(
...
+	}
+
+	// Cyclic dependencies between param definitions would make the evaluation order-dependent
+	const formatNode = (node: string): string => {
+		const [scope, parameter] = node.split(":");
+		return (scope !== "0" ? `Endpoint ${scope}, ` : "")
+			+ `#${parameter}`;
+	};
+	const done = new Set<string>();
+	const visit = (node: string, path: string[]): void => {
```
AlCalzone: **[Major] · Tests**

The ~+447-line `lintParamValueReferences` block — cycle detection (this `visit`), forbidden-location, shape-match, `constantComparisonOutcome`, per-value variant generation — has no dedicated unit test, only incidental integration lint.

**Why:** This lint is the safety net for the whole feature's config-authoring, and it already contains two confirmed bugs (operator variant + signedness, flagged separately).

**Fix:** Add unit coverage, especially cycle detection and `constantComparisonOutcome`.

packages/config/maintenance/lintConfigFiles.ts
```diff
@@ -661,11 +1009,55 @@ async function lintDevices(): Promise<void> {
...
+				if (ref.valueBitMask != undefined) {
+					const maxValue = 2 ** getBitMaskWidth(ref.valueBitMask) - 1;
+					if (ref.comparand < 0 || ref.comparand > maxValue) continue;
+					fullValue = encodePartial(
+						0,
+						ref.comparand,
+						ref.valueBitMask,
+					);
+				} else {
+					fullValue = ref.comparand;
```
AlCalzone: **[Minor] · Correctness**

Variant generation sets the referenced param to `comparand` regardless of operator (`fullValue = ref.comparand`). For `>`, `<`, `!==`, and off-boundary `>=`/`<=`, that value does not satisfy the condition, so definitions guarded by those operators are only ever linted in their hidden state.

**Why:** Errors inside those guarded definitions slip through lint.

**Fix:** Pick a value that actually satisfies each operator.

packages/config/maintenance/lintConfigFiles.ts
```diff
@@ -198,14 +210,343 @@ function unconditionalComesLast(
...
+			const fullValue = paramValues.get(`${endpoint}:${parameter}`);
+			if (fullValue == undefined) return undefined;
+			if (bitMask == undefined) return fullValue;
+			return parsePartial(fullValue, bitMask, false);
+		},
+	};
+}
+
+/** Determines whether a comparison with a partial param value in the range [0...maxValue] has a constant outcome */
+function constantComparisonOutcome(
```
AlCalzone, line 241: **[Minor] · Correctness**

`constantComparisonOutcome` (and the partial-variant bounds at ~449 and ~1029) assume an unsigned range `[0, 2^width−1]` (`comparand < 0 → false`, etc.).

**Why:** Signed partial references get false "always true/false" warnings and skipped valid negative comparands.

**Fix:** Consult the referenced param's signedness.

packages/cc/src/cc/ConfigurationCC.ts
```diff
@@ -479,6 +484,31 @@ export function refreshConfigParamMetadataFromConfigFile(
 	}
 }
 
+/**
+ * Reads the cached value of a config parameter from the value DB.
+ * With a bit mask, returns the value of that partial parameter,
+ * otherwise the full value, composed from cached partials if necessary.
```
AlCalzone, line 529: **[Minor] · Design/Readability/DRY**

This docstring claims the full value is "composed from cached partials if necessary," but the body does a single `getValue` and never composes (the inline comment even says composition isn't needed). Flagged independently by 4 review lenses.

**Why:** Misleading contract for a new exported helper.

**Fix:** Drop the "composed from partials" clause.

packages/config/src/Logic.ts
```diff
@@ -48,12 +50,93 @@ add_operation(
...
 );
 
+function paramComparison(compare: (a: number, b: number) => boolean) {
+	// Must be a function expression, so json-logic passes the data context as `this`
+	return function(
+		this: unknown,
+		parameter: number,
+		bitMask: number | null,
+		comparand: number,
+	): boolean {
```
AlCalzone, line 60: **[Minor] · Operability**

When a referenced value is unknown, `paramComparison` silently returns `true` with no diagnostic anywhere in the call chain.

**Why:** A config author whose `#71 == 1` never applies can't distinguish "value unqueryable" from "condition genuinely false".

**Fix:** Emit a trace-level log when a reference resolves to unknown.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone, line 472: **[Minor] · Performance**

Re-eval refreshes metadata for every endpoint on any referenced-value change, though conditions resolve only within their own scope — O(all-params) instead of O(affected-endpoint) for multi-endpoint devices.

**Why:** Needless constant-factor multiplier on every qualifying event for power strips and similar.

**Fix:** Use the recorded endpoint scope to refresh only the affected endpoint(s).

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
```diff
@@ -259,4 +353,195 @@ export abstract class DeviceConfigMixin extends FirmwareUpdateMixin
...
+					);
+			}
+		})();
+	}
+
+	private async updateCurrentDeviceConfigHash(): Promise<void> {
+		const config = this._deviceConfig;
+		if (!config) return;
+		const oldHash = this._currentDeviceConfigHash;
+		const newHash = await config.getHash(
```
AlCalzone, line 592: **[Minor] · Performance**

`updateCurrentDeviceConfigHash` runs a full `JSON.stringify` + `deflateSync` level-9 on every re-eval, but at hash v4 `paramInformation` is stripped, so a param-value-only change (the feature's whole trigger) never alters the hash input.

**Why:** Wasted CPU on the event-loop thread per qualifying value change.

**Fix:** Skip the recompute (or use a cheaper digest) when only `paramInformation` differs.

AlCalzone (commented):
## deep-review summary

**Verdict:** Ship with minor changes. The approach is sound and cleanly layered — it extends the declarative `$if` grammar rather than adding per-device imperative compat code, with a clean config↔driver seam via the injected `getCachedParamValue`. Two items gate merge; the rest is polish.

**Top 3 risks**
1. Unbounded runtime re-query loop on cyclic/self-referential user configs (lint is the only cycle guard, and user overrides bypass it).
2. `config.api.md` not regenerated → api-extractor verify fails CI.
3. `queryUnknownConfigParamValues` has no in-flight guard → duplicate `Configuration Get`s + endless re-query of dead params.

**Strengths:** correct architecture; safe evaluator (no `eval`/`Function`/`vm`, no prototype-pollution sink, logs emit only param numbers not values); `lookupDevicePreserveConditions` avoids re-reading config files on re-eval; lint additions are proportionate to the grammar's footguns; the `lookupDevice`→`lookupDevicePreserveConditions` split preserves the caller contract.

**Theme:** most Majors share one root — the re-eval / catch-up path lacks a runtime guard and memory (in-flight set, referenced-param awareness, failed-attempt memory) and lacks observability.

### Additional notes (not line-anchorable)

**Nit — `LogicParser.ts` hex tokenizer:** `0x` with no following hex digit yields token value `"0x"`, and `parseInt("0x")` is `NaN`, so a malformed hex literal becomes a silent `NaN` comparand rather than a parse error. (The affected lines are pre-existing context, so no inline thread.)

**Nit — tests:** missing a `||`-with-unknown test (docs claim unknown collapses `||` to `true`; only `&&` is tested), a mixed `||`/`&&` precedence test, and `1 == #71` asserts a throw without checking the message.

**FYI — back-compat rollout:** an older driver that pulls new config via the nightly config-update-within-patch channel hits the tokenizer's `#` → `default` → `throw`; `evaluate` doesn't catch, so the *entire* device config for that node fails to load, not just the one condition. Only test fixtures use `#` today — don't ship `#` in `config/devices/*.json` until the parser has been in a released driver past that window. Consider degrading a single unparseable condition rather than failing the whole file.

**FYI — pre-existing parser backtracking:** `parseExpression`↔`parseGroup` re-parse subtrees ~4× per nesting level (≈4^N; ~14 nested parens ≈ 26 s, measured). Not introduced or asymptotically worsened by this PR, but a nesting/step budget would be a good hardening follow-up given `$if` strings are treated as untrusted.

**FYI — legacy hash edge:** a node still carrying a v0/v1 cached hash that *also* changed config (so the version isn't upgraded at load) recomputes re-eval hashes with a `paramInformation`-including version, which can emit the spurious re-interview warning. Very narrow; the modern v4 path is correct.

**Coverage:** all 8 dimensions run; all 24 changed files covered. Note: the "hash divergence → `hasDeviceConfigChanged()` stuck true" Blocker raised during review was **not** confirmed — v4 strips `paramInformation`, so it doesn't hold for current nodes.

packages/config/maintenance/lintConfigFiles.ts
AlCalzone: Or we reuse `isSignedPartial` and pass the correct parameter format depending on the `unsigned` flag?

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone: Handle this how we handle invalid user configs already, just add a check for a cyclic reference.

packages/config/src/devices/DeviceConfig.ts
AlCalzone: We shouldn't break here. Deprecate the old deviceId option, add context, and default the context to deviceId if not given.

packages/config/src/devices/shared.ts
AlCalzone: Can't we bake the scoped endpoint into the getCachedParamValue call instead of leaving it as a parameter?

packages/zwave-js/src/lib/test/node/Node.paramValueConditions.test.ts
AlCalzone: yes, but make it clear this is a regression test

packages/config/src/Logic.test.ts
AlCalzone: How about we skip parsing in this test helper and instead construct the full state of all involved partials where needed?

docs/config-files/conditional-settings.md
AlCalzone: Nah, this is true for all conditions

packages/config/maintenance/lintConfigFiles.ts
AlCalzone: Fixed in 926fb3a82 — reused `getLegalRangeForBitMask` (which already encodes the isSignedPartial rule incl. the 1-bit exception) with the definition's `unsigned` flag. Variants now store the raw comparand keyed by the reference shape, so no unsigned parse roundtrip corrupts negative comparands.

packages/config/maintenance/lintConfigFiles.ts
AlCalzone: Fixed in 926fb3a82 — the range comes from `getLegalRangeForBitMask(mask, unsigned)` per matching definition; the warning only fires when the outcome is constant for every definition.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone: Fixed in 926fb3a82 — the re-eval body is wrapped in try/catch (logged as error, update skipped), and the fire-and-forget hash update and query loop have logged catch handlers.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone: Fixed in 926fb3a82 — re-eval logs the triggering param(s) with old => new values at debug, and refreshConfigParamMetadataFromConfigFile logs a per-param line when a param is added/updated/hidden/removed.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone: Fixed in 926fb3a82 — the warning now includes the changed param values and suggests re-interviewing the node.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone: Fixed in 926fb3a82 — cyclic/self references now make the config invalid at load time (thrown from the ConditionalDeviceConfig constructor, handled like other invalid user configs), independent of lint. The catch-up queries also log each attempt and warn on missing responses.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone: Fixed in 926fb3a82 — the refresh is scoped to the endpoints whose referenced values actually changed.

packages/cc/src/cc/ConfigurationCC.ts
AlCalzone: Fixed in 926fb3a82 — the interview passes the already-queried referenced params to refreshValues, which skips them.

packages/config/src/devices/DeviceConfig.ts
AlCalzone: Kept backwards compatible in 926fb3a82 — `deviceId` stays as a deprecated option and is used as fallback when `context` is not given.

packages/config/src/devices/shared.ts
AlCalzone: Done in 926fb3a82 — the mutable `endpoint` field is gone. Each endpoint boundary bakes its index into an internal scoped getter derived from `getCachedParamValue`; root-scope evaluation falls back to endpoint 0.

packages/zwave-js/src/lib/test/node/Node.paramValueConditions.test.ts
AlCalzone: Fixed in 926fb3a82 — all three tests count "interview started" events and assert none fired plus interviewStage staying Complete, marked as regression tests.

packages/config/src/Logic.test.ts
AlCalzone: Fixed in 926fb3a82 — the helper takes the partial states directly (keyed `endpoint:parameter[bitMask]`, no parsing), with new cases for signed partials (`#9[0x0f] == -1`, `> -1`), `||` with an unknown side, a parenthesized group, and a collectParamValueReferences test asserting both `||` branches are found.

docs/config-files/conditional-settings.md
AlCalzone: Skipped per the reply above — LHS-only is true for all conditions, not specific to param references.

docs/config-files/conditional-settings.md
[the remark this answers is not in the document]
AlCalzone: Reworked in 926fb3a82 — the example is now a single realistic paramInformation snippet (a gated param + gated options), matching the loose JSONC style of the rest of the page.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone: Addressed in 926fb3a82 — cyclic and self-references now make the config invalid at load time (thrown from the ConditionalDeviceConfig constructor), independent of lint. A lint-bypassing user config with such references is rejected like any other invalid config, so the query storm cannot start.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone: Partially addressed in 926fb3a82 — the loop root cause (cyclic references in un-linted configs) is now rejected at config load, and each catch-up query logs its attempt and warns on missing response. No in-flight dedup was added: with cycles impossible, close-together value events can at most duplicate a bounded set of one-shot queries.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone: Fixed in 926fb3a82 — re-eval logs the triggering param(s) with old => new values at debug, plus a per-param debug line when metadata is added/updated/hidden/removed.

packages/config/maintenance/lintConfigFiles.ts
AlCalzone: Fixed in 926fb3a82 — both sites derive the range from getLegalRangeForBitMask with the referenced definitions' unsigned flag.

packages/zwave-js/src/lib/node/mixins/80_DeviceConfig.ts
AlCalzone: Fixed in 926fb3a82 — the refresh is scoped to the endpoints whose referenced values actually changed.

AlCalzone:
I'm not happy with the internal API design. Have to think about it.

Also, this is a problem if we want to model config parameters that change endpoints:

> * References resolve against the endpoint whose config section the condition appears in; conditions inside `endpoints.<n>` use that endpoint's own parameters.

PR #8966, opened 2026-07-13
fix(cc): purge stale credentials from cache while querying

AlCalzone:
User Credential CC would only purge cache entries when a credential was reported to be deleted, or modifying failed because of an empty slot. Querying all credentials did not clear stale cache entries.

With this PR, `getAllCredentials`, `getCredentialsByType`, `getCredentialsForUser` now drop cached credentials that are found to no longer exist (at the beginning, between the current and the next reported slot, and at the end of the list). When filtered by type, only the corresponding slots are purged, even if the reports would indicate gaps at the beginning of the next type.

### DIFF
--- a/packages/zwave-js/src/lib/node/feature-apis/AccessControl.ts
+++ b/packages/zwave-js/src/lib/node/feature-apis/AccessControl.ts
@@ -1683,6 +1683,50 @@ export class AccessControlAPI extends FeatureAPI {
 		}
 	}
 
+	/**
+	 * Removes cached credentials orphaned within the exclusive composite-key
+	 * range `(from, to)` of a U3C enumeration walk. Credentials enumerate in
+	 * ascending (type, slot) order, encoded as `(type << 16) | slot`, and the
+	 * node's next-credential pointer only advances to credentials that still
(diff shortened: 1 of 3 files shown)


packages/zwave-js/src/lib/node/feature-apis/AccessControl.ts
```diff
@@ -1824,30 +1873,58 @@ export class AccessControlAPI extends FeatureAPI {
...
+					composite(credential.type, credential.slot),
+				);
+			}
+
 			credentials.push(credential);
 
 			const nextType = result.nextCredentialType
 				?? UserCredentialType.None;
 			const nextSlot = result.nextCredentialSlot ?? 0;
-			// A zero next pointer marks the end of the node's credential sequence.
```
AlCalzone, line 1832: Keep both the original and the new comment here

AlCalzone: Fixed in 13f0ebfa4 — restored the original zero-pointer comment alongside the reconciliation note.

AlCalzone:
@zwave-js-bot automerge

PR #8970, opened 2026-07-14
fix(config): broken link to Honeywell T6 Pro manual

ItsHarper:
The link to the manual was broken, and now it's not


AlCalzone (approved):
Thanks!

PR #8974, opened 2026-07-15
fix(mcp): skip optional ONNX GPU download

AlCalzone:
This should fix the transitive CI setup failures.

## Description

- disable the `onnxruntime-node` postinstall build for the dev MCP workspace
- retain the bundled CPU runtime used by the lazy semantic embedding pipeline
- prevent CI installs from downloading roughly 300 MB of unused CUDA provider binaries

## Validation

- built `@zwave-js/mcp-server-dev`
- ran the semantic consent, environment, and embedding-cache tests
- loaded the ONNX CPU runtime with the optional CUDA provider files removed

### DIFF
--- a/.env.yarn
+++ b/.env.yarn
@@ -1,0 +1,2 @@
+# The semantic embedding pipeline uses the bundled CPU runtime
+ONNXRUNTIME_NODE_INSTALL=skip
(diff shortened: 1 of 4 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #8976, opened 2026-07-15
fix: work around credential learn status with incorrect steps remaining

AlCalzone:
`stepsRemaining` MUST not be 0 when the status is `Started`, but some devices don't do this correctly.


AlCalzone:
@zwave-js-bot automerge

PR #8978, opened 2026-07-17
fix(cc): increase timeouts for bulk-deleting users/credentials

AlCalzone:
Bulk deletions (all users, or all credentials matching a filter) can take significantly longer than the default 1 second report timeout.
This PR increases the timeout for those two commands to 15 seconds.


AlCalzone:
@zwave-js-bot automerge

Issue #8981, opened 2026-07-25
Failure to perform firmware upgrade for Inovelli LZW31-SN


DevRalf:
### Checklist

- [x]  I am not using Home Assistant. Or: a developer has told me to come here.
- [x] I have checked the troubleshooting section and my problem is not described there.
- [x] I have read the changelog and my problem is not mentioned there.

### Deploy method

Docker

### Z-Wave JS UI version

11.22.0

### ZwaveJS version

15.26.0

### Describe the bug

One of my Inovelli LZW31-SN dimmers shows that there is a firmware update available, so I am trying to install it. All looks good in the UI, it presents the new 'v1.57 [stable]' version to me, but as soon as I select "Update" and confirm the "OTA Upgrade" warning dialog, I pretty much immediately get the 'Firmware update failed. The transmission failed or was aborted."-error.

I didn't see anything obvious in the attached logs,

[z-ui_current.log](https://github.com/user-attachments/files/30378371/z-ui_current.log)
[zwavejs_current.log](https://github.com/user-attachments/files/30378370/zwavejs_current.log)

### To Reproduce

1. Click on update button for Inovelli LSW31-SN in Control Panel
2. Click contextual menu and select 'Update'
3. Confirm OTA warning to proceed.

Failure message is displayed immediately.

### Expected behavior

Firmware upgrade to proceed and work.

### Additional context

I am using a Sonoff EFR32ZG23 USB stick to control my Z-Wave devices

robertsLando:
cc @AlCalzone

AlCalzone:
@robertsLando can you transfer to the zwave-js repo?

AlCalzone:
/analyze-logfile https://github.com/user-attachments/files/30378370/zwavejs_current.log

AlCalzone:
@DevRalf can you please share a driver diagnostics dump of the device?

<img width="1390" height="800" alt="Image" src="https://github.com/user-attachments/assets/d5fc8fee-8ca2-468b-a88b-3195b8c90b55" />

DevRalf:
Sure thing, here you go:
[node_6_dump.json](https://github.com/user-attachments/files/30621022/node_6_dump.json)

Thanks for looking into this!

AlCalzone:
Try re-interviewing the device first, then retry the update. I think there may have been a timeout during the interview.

DevRalf:
[...]

PR #8982, opened 2026-07-28
fix(bot): run triage on transferred issues

AlCalzone:
Fixes the silent triage failure on #8981, which was opened in `zwave-js-ui` and transferred here.

Transferring an issue recreates the source repo's comments in the destination with their original timestamps, and `zwave-js-ui` runs a mirrored copy of these scripts that stamps the same marker tags. So the docs answer and the logfile feedback both considered themselves done and stayed silent, and nothing on a transferred comment identifies where it came from. The `transferred` timeline event is the one usable signal.

- `utils.cjs` gains `listCommentsSinceTransfer()`, which lists an issue's comments and drops everything older than the last transfer. `answerFromDocs` and `ensureLogfileFeedback` use it instead of their own lookups.
- Bot comments inherited from the other repo are collapsed as outdated — they are about the wrong repository now.
- The nightly feedback digest no longer scores a foreign answer as one of ours, where reactions to it could suppress unrelated future answers.
- Removed `ensureLogfile.cjs` and `ensureLogfileInDiscussion.cjs`, unreferenced for a while.

Discussions keep the old behavior; they have no timeline API, so an inherited answer there still counts as answered.

Verified against #8981 and the live feedback digest, plus unit tests for the new helpers.

`zwave-js-ui` needs the same treatment afterwards — issues get transferred the other way too.


.github/bot-scripts/answerFromDocs.cjs
```diff
@@ -51,7 +52,9 @@ function chunkUrl(chunk) {
 }
 
 /**
- * Checks whether the bot already answered this post
+ * Checks whether the bot already answered this post.
+ * Discussions have no timeline API, so an answer inherited from a transfer
+ * cannot be told apart from ours and still counts as answered.
```
AlCalzone: These types of comments (logic/internals) should be inline where the decision is made, not in the function comment.

.github/bot-scripts/utils.cjs
```diff
@@ -103,6 +103,63 @@ async function wasPRModifiedAfterComment(
 	return false;
 }
 
+/**
+ * Returns when an issue was last transferred into this repository,
+ * or 0 if it was created here.
+ * @param {any[]} events - Timeline events of the issue
```
AlCalzone: Isn't there a better event type than `any`?

.github/bot-scripts/answerFromDocs.cjs
AlCalzone: Moved into the `if (isDiscussion)` branch in 885bbbed7.

.github/bot-scripts/utils.cjs
AlCalzone: Narrowed to `{event?: string, created_at?: string}[]` in 885bbbed7. Both fields have to stay optional because `timeline-committed-event` declares `event?: string` and carries no `created_at`, so anything stricter makes Octokit's union unassignable. It also lets the `"created_at" in event` guard become a plain truthiness check.

PR #8993, opened 2026-07-28
fix(bot): post logfile analyses as zwave-js-bot

AlCalzone:
## Description

The `add-comment` safe output posted with the default workflow token, so analysis comments appeared as **github-actions** instead of **zwave-js-bot** like every other bot comment. gh-aw supports a per-output token override, so both analyze-logfile workflows now pass `github-token: ${{ secrets.BOT_TOKEN }}` to `add-comment`.

The new restricted secret was approved via `gh aw compile --approve` (BOT_TOKEN is already used by the github-script steps in the same workflows).


AlCalzone:
@zwave-js-bot automerge

PR #8995, opened 2026-07-28
feat(config): add Heatit Z-TRM7

AlCalzone:
Adds a device config for the Heatit Z-TRM7 (`0x019b:0x0030:0x3006`). The device currently has no config file at all, so all 29 parameters show up unlabeled.

Parameters are transcribed from section 33 of the [installer manual](https://media.heatit.com/5278) (FW 1.0, document version B), cross-checked against a node dump from [home-assistant/core#177327](https://github.com/home-assistant/core/issues/177327).

## Why parameter 2's labels are copied from Z-TRM6

Home Assistant picks which temperature sensor the climate entity follows by looking up the option label of config parameter 2 and truncating it at the first hyphen ([`discovery_data_template.py:296`](https://github.com/home-assistant/core/blob/dev/homeassistant/components/zwave_js/discovery_data_template.py#L296)). Without a config file, `metadata.states` is empty and the lookup raises, which is why this file is a prerequisite for the HA-side fix.

Using Z-TRM6's exact strings means HA only has to add `0x3006` to the existing Z-TRM6 discovery schema instead of carrying a second lookup table. There's a comment on the parameter recording that, since nothing in this repo currently documents the coupling.

## Where this differs from a straight Z-TRM6 copy

- Parameter 3 is `Sensor Resistance (NTC)` with kΩ in the option labels, following `z-trm6_dc.json`. The manual's SEN setting covers the floor NTC too, so Z-TRM6's "External Sensor Resistance" is too narrow.
- Parameter 26 is `Thermostat State Update Interval`. It controls Setpoint/Mode/Binary **Set** commands sent to association groups 2-4, not reports.
- Parameters 21 and 26 carry `Allowable range` descriptions for their value gaps (0, then 10-65535 and 30-65535 respectively).
- `$purpose` tags on 10/11/12 (`calibration.temperature`) and 19 (`reporting_threshold.temperature`). Parameter 20 gates power, energy, current and voltage reports together, so no single purpose fits it.
- Parameter 28 imports `base_enable_disable` rather than spelling out the options inline.

## Notes for review

- `compat.overrideFloatEncoding` is carried over from Z-TRM6 on the strength of the shared product type `0x0030`. `z-trm6_dc.json` (product type `0x0032`) doesn't have it. Unverified on hardware — a missing flag breaks setpoints while a spurious one only forces 2-byte encoding, so I kept it.
- No `zwaveAllianceId`: the product has no Z-Wave Alliance catalog entry that I could find.
- `description` is `Thermostat` rather than Z-TRM6's `Floor Thermostat`, since the manual sells it for water-based heating as well.
- The reporter in home-assistant/core#177327 offered to test on real hardware.

`autofix_config` found nothing to fix, `lint_config` is clean, and `yarn lint:zwave` reports 0 errors with no warnings on this file.

### DIFF
--- a/packages/config/config/devices/0x019b/z-trm7.json
+++ b/packages/config/config/devices/0x019b/z-trm7.json
@@ -1,0 +1,422 @@
+{
+	"manufacturer": "Heatit",
+	"manufacturerId": "0x019b",
+	"label": "Z-TRM7",
+	"description": "Thermostat",
+	"devices": [
+		{
+			"productType": "0x0030",
+			"productId": "0x3006"
+		}
+	],
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 1,
+			"isLifeline": true
+		},
+		"2": {
+			"label": "Binary Switch Set",
+			"maxNodes": 10
+		},
+		"3": {
+			"label": "Thermostat Setpoint Set",
+			"maxNodes": 10
+		},
+		"4": {
+			"label": "Thermostat Mode Set",
+			"maxNodes": 10
+		}
+	},
+	"paramInformation": [
+		{
+			"#": "1",
+			"$import": "~/templates/master_template.json#base_enable_disable_inverted",
+			"label": "Local Control",
+			"defaultValue": 0
+		},
+		{
+			"#": "2",
+			// The option labels are used by Home Assistant to pick the temperature sensor
+			// the climate entity follows, so they must match the ones used for Z-TRM6
+			"label": "Sensor Mode",
+			"valueSize": 1,
+			"defaultValue": 1,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "Floor",
+					"value": 0
+				},
+				{
+					"label": "Internal",
+					"value": 1
+				},
+				{
+					"label": "Internal with floor limit",
+					"value": 2
+				},
+				{
+					"label": "External",
+					"value": 3
+				},
+				{
+					"label": "External with floor limit",
+					"value": 4
+				},
+				{
+					"label": "Power regulator",
+					"value": 5
+				}
+			]
+		},
+		{
+			"#": "3",
+			"label": "Sensor Resistance (NTC)",
+			"valueSize": 1,
+			"defaultValue": 0,
+			"allowManualEntry": false,
+			"options": [
+				{
+					"label": "10 kΩ",
+					"value": 0
+				},
+				{
+					"label": "12 kΩ",
+					"value": 1
+				},
+				{
+					"label": "15 kΩ",
+					"value": 2
+				},
+				{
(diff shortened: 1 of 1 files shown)


packages/config/config/devices/0x019b/z-trm7.json
```diff
@@ -0,0 +1,422 @@
...
+			"unit": "seconds",
+			"minValue": 30,
+			"maxValue": 65535,
+			"defaultValue": 840,
+			"unsigned": true
+		},
+		{
+			"#": "21",
+			"label": "Turn On Delay After Error",
+			"description": "Allowable range: 10-65535",
```
AlCalzone: Use the `allowed` property to model the range instead.

packages/config/config/devices/0x019b/z-trm7.json
```diff
@@ -0,0 +1,422 @@
...
+				}
+			]
+		}
+	],
+	"compat": {
+		// The device doesn't accept setpoints with automatic precision
+		"overrideFloatEncoding": {
+			"size": 2
+		}
+	},
```
AlCalzone: This hasn't been reported before - lets start without a compat flag.

packages/config/config/devices/0x019b/z-trm7.json
AlCalzone, line 1: Lots of parameters are candidates for a heatit_template.json. Cross-reference them against other heatit devices and extract them into templates.

packages/config/config/devices/0x019b/z-trm7.json
AlCalzone: Done in 752022c37 — parameters 21 and 26 now use `allowed` with a `{ value: 0 }` entry plus the real range, and the prose descriptions are gone.

packages/config/config/devices/0x019b/z-trm7.json
AlCalzone: Removed in 752022c37. I had inferred it from Z-TRM6 sharing product type 0x0030, but you're right that there's no report against this device.

packages/config/config/devices/0x019b/z-trm7.json
AlCalzone: Agreed, there's a lot to pull out — Sensor Mode, the min/max temp limits, sensor calibration, the setpoints, the brightness and report-interval parameters all recur across z-trm6, z-trm6_dc, z-trm3, z-temp3 and zm_thermostat_16. Doing it properly means touching those files too, so we'd rather land this one first and follow up with a dedicated templating sweep across the Heatit devices.

AlCalzone:
@zwave-js-bot automerge

packages/config/config/devices/0x019b/z-trm7.json
AlCalzone: Done in https://github.com/zwave-js/zwave-js/pull/8997 — 19 templates extracted into `0x019b/templates/heatit_template.json`, covering every Z-TRM7 parameter that also appears on another Heatit device. 10 device files converted.

PR #8996, opened 2026-07-28
feat(core): implement ChaCha20-Poly1305 in the browser crypto primitives

AlCalzone:
The browser crypto primitives threw when asked for ChaCha20-Poly1305, so the ESPHome Noise transport could not be used on non-Node runtimes (QuickJS/txiki.js, Deno, Bun, browsers). No shipping browser exposes ChaCha20-Poly1305 through the Web Crypto API, so a JS implementation is the only portable option.

- `encryptChaCha20Poly1305` / `decryptChaCha20Poly1305` in `primitives.browser.ts` now use `@noble/ciphers`, imported from the narrow `chacha.js` subpath so no AES, Salsa or FF1 code is pulled in. The package has no runtime dependencies.
- The tag is split off the ciphertext on encrypt, and a tag mismatch is reported as `{ plaintext: <empty>, authOK: false }`, matching `decryptAES128CCM`.
- `@noble/ciphers/chacha.js` is whitelisted in the `no-forbidden-imports` lint rule as browser-safe.

Verified: the cross-implementation `primitives.test.ts` now covers both ChaCha primitives for the browser and Node implementations against the RFC 8439 section 2.8.2 vector, plus round-trip, corrupted-tag and modified-AAD cases (46 tests pass). `yarn check:browser` and `yarn lint:ts` pass.

### DIFF
--- a/packages/core/src/crypto/primitives/primitives.browser.ts
+++ b/packages/core/src/crypto/primitives/primitives.browser.ts
@@ -1,3 +1,4 @@
+import { chacha20poly1305 } from "@noble/ciphers/chacha.js";
 import { Bytes, type BytesView } from "@zwave-js/shared";
 import type { CryptoPrimitives, KeyPair } from "@zwave-js/shared/bindings";
 import type { webcrypto } from "node:crypto";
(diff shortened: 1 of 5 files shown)


packages/core/src/crypto/primitives/primitives.browser.ts
```diff
@@ -459,31 +461,44 @@ async function hmacSHA256(
 	return new Uint8Array(signature);
 }
 
-// ChaCha20-Poly1305 is not supported in the browser environment.
-// Web Crypto API does not provide ChaCha20-Poly1305.
-// These functions will throw if called; a proper solution needs to be implemented.
+const POLY1305_TAG_LENGTH = 16;
 
+// No shipping browser exposes ChaCha20-Poly1305 through the Web Crypto API,
+// so the pure JS implementation from @noble/ciphers is used instead
```
AlCalzone: Remove comment

AlCalzone: Removed in 53b6c18cf.

Issue #25, opened 2019-04-06
Add a way to handle requests that might or might not get a response


AlCalzone:
e.g. ConfigurationCC.Get for unsupported parameters

AlCalzone:
Fixed by setting the message timeout (56aa44e39fce087b8cd01f425203684e27a19795)

PR #8997, opened 2026-07-28
refactor(config): extract shared Heatit parameters into templates

AlCalzone:
Follow-up to https://github.com/zwave-js/zwave-js/pull/8995#discussion_r3668897251.

Every Z-TRM7 parameter that also appears on another Heatit device is now defined once in `0x019b/templates/heatit_template.json` and imported where it is used. 19 templates, 12 device files converted.

**Templates:** `sensor_mode`, `sensor_resistance_ntc`, `min_temperature_limit`, `max_temperature_limit`, `temperature_sensor_calibration`, `regulation_mode`, `temperature_control_hysteresis`, `temperature_display`, `active_display_brightness`, `standby_display_brightness`, `temperature_report_interval`, `temperature_report_hysteresis`, `meter_report_interval`, `heating_setpoint`, `cooling_setpoint`, `eco_setpoint`, `power_regulator_active_time`, `operating_mode`, `load_power`.

**Devices:** Z-TRM7, Z-TRM6, Z-TRM6 DC, Z-TRM 2, Z-TRM2fx, Z-TRM3, Z-Temp2, Z-TEMP3, Z-WATER2, Z-HAN2, ZM Thermostat 16A, TF016/TF021.

The Home Assistant label coupling on `Sensor Mode` now lives on the template, so it covers Z-TRM6 and Z-TRM7 together.

## Verification

Every `0x019b` parameter was resolved through `ConditionalDeviceConfig` at four firmware versions before and after each step, then diffed. Z-TRM7 resolves byte-identical. Everything else that changed did so on purpose:

- **`$purpose` gained** on 18 calibration and threshold parameters that had none.
- **Unit fixes.** Z-TRM6 DC `#16`/`#17` were `%` on a 1–10 range, now `10 %`. TF016/TF021 `#4`–`#13` had no unit, now `0.1 °C`. Z-TRM 2 and Z-TRM2fx used a `oC` unit string on parameters that count in tenths of a degree, now `0.1 °C`; Z-TRM 2 `#12` is a duty cycle and carried that unit by mistake, so it is now unitless.
- **Labels.** Menu abbreviations dropped from Z-TRM3, TF016/TF021, Z-TRM 2 and Z-TRM2fx labels. Z-TRM6 `#3` `External Sensor Resistance` → `Floor and External Sensor Resistance`: both manuals describe it as "the resistance value of the connected NTC", and the Z-TRM6 manual adds that the floor and external sensors must use the same ohm value, so the old label named the wrong scope. Z-TRM6 DC `#15`/`#19`/`#23`/`#25`/`#27` and ZM Thermostat 16A `#7`/`#9` adopt the family wording.
- **Z-TRM6 `#29`** load size corrected to `valueSize` 2.
- **Dropped** redundant `Range 50-400 (…)` descriptions on TF016/TF021 and the 100 W increment prose on Z-TRM6 DC `#27`, now that the templates carry unit and range.

`autofix_config` and `lint_config` are clean on every touched file, and `yarn lint:zwave` reports 0 errors with the same 10 pre-existing warnings (all in `0x0345`).

## Not included

The dimmer and relay families (`z-dim2`, `zm_dimmer`, `zm_single_relay_16`, `heatit_z-relay`, `heatit_z_water`) have their own duplicate parameter groups, none of which correspond to a Z-TRM7 parameter.

### DIFF
--- a/packages/config/config/devices/0x019b/tf016_tf021.json
+++ b/packages/config/config/devices/0x019b/tf016_tf021.json
@@ -130,37 +130,36 @@
 			"#": "4",
 			"$if": "firmwareVersion >= 1.8",
 			"$import": "templates/heatit_template.json#temperature_control_hysteresis",
-			"label": "DIFF L. Temperature Control Hysteresis",
 			"valueSize": 2
 		},
 		{
 			"#": "5",
 			"$if": "firmwareVersion >= 1.8",
 			"$import": "templates/heatit_template.json#min_temperature_limit",
-			"label": "FLo, Floor Min Limit"
+			"label": "Floor Min Limit"
 		},
 		{
 			"#": "6",
 			"$if": "firmwareVersion >= 1.8",
 			"$import": "templates/heatit_template.json#max_temperature_limit",
-			"label": "FHi, Floor Max Limit"
+			"label": "Floor Max Limit"
 		},
 		{
 			"#": "7",
 			"$if": "firmwareVersion >= 1.8",
 			"$import": "templates/heatit_template.json#min_temperature_limit",
-			"label": "ALo, Air Min Limit"
+			"label": "Air Min Limit"
 		},
 		{
 			"#": "8",
 			"$if": "firmwareVersion >= 1.8",
 			"$import": "templates/heatit_template.json#max_temperature_limit",
-			"label": "AHi, Air Max Limit"
+			"label": "Air Max Limit"
 		},
 		{
 			"#": "9",
 			"$if": "firmwareVersion >= 1.8",
-			"label": "PLo, FP-Mode P Setting",
+			"label": "FP-Mode P Setting",
 			"valueSize": 2,
 			"minValue": 0,
 			"maxValue": 9,
--- a/packages/config/config/devices/0x019b/tf016_tf021.json
+++ b/packages/config/config/devices/0x019b/tf016_tf021.json
@@ -129,52 +129,33 @@
 		{
 			"#": "4",
 			"$if": "firmwareVersion >= 1.8",
+			"$import": "templates/heatit_template.json#temperature_control_hysteresis",
 			"label": "DIFF L. Temperature Control Hysteresis",
-			"description": "Range 3-30 (0.3 °C 3.0 °C)",
-			"valueSize": 2,
-			"minValue": 3,
-			"maxValue": 30,
-			"defaultValue": 5
+			"valueSize": 2
 		},
 		{
 			"#": "5",
 			"$if": "firmwareVersion >= 1.8",
-			"label": "FLo, Floor Min Limit",
-			"description": "Range 50-400 (5.0 °C 40.0 °C)",
-			"valueSize": 2,
-			"minValue": 50,
-			"maxValue": 400,
-			"defaultValue": 50
+			"$import": "templates/heatit_template.json#min_temperature_limit",
+			"label": "FLo, Floor Min Limit"
 		},
 		{
 			"#": "6",
 			"$if": "firmwareVersion >= 1.8",
-			"label": "FHi, Floor Max Limit",
-			"description": "Range 50-400 (5.0 °C 40.0 °C)",
-			"valueSize": 2,
-			"minValue": 50,
-			"maxValue": 400,
-			"defaultValue": 400
+			"$import": "templates/heatit_template.json#max_temperature_limit",
+			"label": "FHi, Floor Max Limit"
 		},
 		{
 			"#": "7",
 			"$if": "firmwareVersion >= 1.8",
-			"label": "ALo, Air Min Limit",
-			"description": "Range 50-400 (5.0 °C 40.0 °C)",
-			"valueSize": 2,
-			"minValue": 50,
-			"maxValue": 400,
-			"defaultValue": 50
+			"$import": "templates/heatit_template.json#min_temperature_limit",
+			"label": "ALo, Air Min Limit"
 		},
 		{
 			"#": "8",
 			"$if": "firmwareVersion >= 1.8",
-			"label": "AHi, Air Max Limit",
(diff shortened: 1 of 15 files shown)


packages/config/config/devices/0x019b/tf016_tf021.json
```diff
@@ -129,52 +129,33 @@
...
 			"$if": "firmwareVersion >= 1.8",
-			"label": "AHi, Air Max Limit",
-			"description": "Range 50-400 (5.0 °C 40.0 °C)",
-			"valueSize": 2,
-			"minValue": 50,
-			"maxValue": 400,
-			"defaultValue": 400
+			"$import": "templates/heatit_template.json#max_temperature_limit",
+			"label": "AHi, Air Max Limit"
 		},
```
AlCalzone, line 158: On these labels, remove the abbreviations including `, `, so only the actual label remains.

packages/config/config/devices/0x019b/z-trm3.json
```diff
@@ -79,75 +79,43 @@
...
 		{
 			"#": "8",
-			"label": "Air Maximum Temperature Limit (AHi)",
-			"valueSize": 2,
-			"unit": "0.1 °C",
-			"minValue": 50,
-			"maxValue": 400,
-			"defaultValue": 400
+			"$import": "templates/heatit_template.json#max_temperature_limit",
+			"label": "Air Maximum Temperature Limit (AHi)"
```
AlCalzone: In these labels, remove the part in ()

packages/config/config/devices/0x019b/z-trm6_dc.json
```diff
@@ -370,19 +207,9 @@
 		},
 		{
 			"#": "27",
-			"label": "Size of Load",
+			"$import": "templates/heatit_template.json#load_power",
 			"description": "Allows the user to set the size of the load in 100 W increments, used when connected to a contactor.",
```
AlCalzone: The 100W increments should be encoded as `step` in `allowed`, not as prose.

packages/config/config/devices/0x019b/tf016_tf021.json
AlCalzone: Fixed in 65c337be9 — labels are now `Floor Min Limit`, `Floor Max Limit`, `Air Min Limit`, `Air Max Limit`. #4 became plain `Temperature Control Hysteresis`, so its label override is gone. Also stripped `PLo, ` from #9 while there.

packages/config/config/devices/0x019b/z-trm3.json
AlCalzone: Fixed in 65c337be9. #4 now matches the template label exactly, so the override is gone. #6 became `Floor Maximum Temperature Limit` — it was missing `Limit` unlike its siblings.

packages/config/config/devices/0x019b/z-trm6_dc.json
AlCalzone: Dropped the prose in 65c337be9, but not in favor of `step` — the increment is already carried by `unit: "100 W"`. The raw parameter value is 1-99, not watts: the manual says `1-99 = 100-9900 Watt` with min 1 / max 99. A `step: 100` would imply a raw range of 100-9900 and make us write values the device rejects. Description is now just `Used when connected to a contactor.`

The manuals did turn up a real bug next door: Z-TRM6 #29 is 2 bytes, not 1 (Z-TRM7 is 1, Z-TRM6 DC is 2). Fixed in the same commit.

packages/config/config/devices/0x019b/tf016_tf021.json
AlCalzone: Fixed in 65c337be9 — labels are now `Temperature Control Hysteresis`, `Floor Min Limit`, `Floor Max Limit`, `Air Min Limit`, `Air Max Limit`.

packages/config/config/devices/0x019b/z-trm3.json
AlCalzone: Fixed in 65c337be9 — parentheticals dropped. `Floor Maximum Temperature` also gained the missing `Limit` to match its siblings.

packages/config/config/devices/0x019b/z-trm6_dc.json
AlCalzone: Fixed in 65c337be9 — prose dropped, leaving `Used when connected to a contactor.`. The raw value is already one step per 100 W (`unit: "100 W"`, range 0-99), so `step` would just be 1. Also corrected Z-TRM6 `#29` to `valueSize` 2.

packages/config/config/devices/0x019b/z-trm2fx_3.0_255.255.json
```diff
@@ -247,12 +215,7 @@
 		},
```
AlCalzone, line 207: We have a few files with this X Sensor Calibration parameter now. do another sweep for templating.

AlCalzone: Done in f12cde8d0. The lineup has two calibration ranges, so there are two templates named after them: `temperature_sensor_calibration_60` (Z-TRM3, Z-TRM6, Z-TRM6 DC, Z-TRM7, Z-TEMP3, Z-WATER2, Z-HAN2, ZM Thermostat 16A) and `temperature_sensor_calibration_40` (Z-TRM 2, Z-TRM2fx, Z-Relay, Z-Water). 29 parameters across 12 files, all now carrying `calibration.temperature`.

Also noticed Z-Temp2 `#4` used `°C` on a -100..100 range while its humidity sibling uses `0.1 %` for the same range, so I corrected it to `0.1 °C`. It stays inline — `valueSize` 2 and a wider range mean it does not match either template.

AlCalzone:
@zwave-js-bot automerge

PR #8998, opened 2026-07-28
feat(core): allow replacing the crypto implementation at runtime

AlCalzone:
## Why

Which `CryptoPrimitives` implementation is used was fixed at build time by the `#crypto_primitives` conditional import in `packages/core`. Embedders running Z-Wave JS on non-Node runtimes — QuickJS/txiki.js, Deno, Bun, the browser — had no way to plug in the crypto their host offers (e.g. an mbedTLS or OpenSSL binding) short of a bundler alias, even though `fs`, `serial`, `db` and `log` are all injectable through `ZWaveOptions["host"]`.

## What

- `packages/core/src/crypto/operations.ts` now reads from a mutable module-level `primitives` slot, initialized to the `#crypto_primitives` import. The 15 exported functions keep their names and signatures, so no call site changes.
- New `setCryptoPrimitives()`, exported from `@zwave-js/core`. Crypto is process-global: all drivers in a process share the last implementation that was set. This is intentional, since crypto is a property of the runtime, not of a Z-Wave network.
- New `@zwave-js/core/bindings/crypto/{node,browser}` subpaths, covered by the existing `./bindings/*` export pattern.
- New `host.crypto` driver option plus a `#default_bindings/crypto` entry in `zwave-js`'s imports map. `Driver.start()` resolves it alongside the other four bindings and applies it before any crypto operation can happen.

## Verification

- `yarn test:ts` on `packages/core/src/crypto` (43 tests), `packages/core/src/security` + `packages/zwave-js/src/lib/driver` (111 tests) and the S2 integration tests `s2Encapsulation`, `s0AndS2Encapsulation`, `failedS2BootstrappingInterview` (4 tests) — all pass. The integration tests start real drivers, so they exercise the `Driver.start()` wiring end to end.
- A new test in `operations.test.ts` installs a stub whose `randomBytes` returns a known value, asserts the exported `randomBytes` returns it, and restores the default.
- `yarn check:browser`, `yarn lint:ts:fix` and `yarn fmt` are clean.
- With `host.crypto` unset, `#default_bindings/crypto` resolves to the unchanged Node primitives under the `node` condition and to the browser primitives under `default`; both expose the same 15 members as before.

### DIFF
--- a/packages/core/src/crypto/operations.ts
+++ b/packages/core/src/crypto/operations.ts
@@ -5,8 +5,7 @@ import { BLOCK_SIZE, leftShift1, xor, zeroPad } from "./shared.js";
 // Import the correct primitives based on the environment
 import { primitives as defaultPrimitives } from "#crypto_primitives";
 
-// Re-exported as mutable bindings, so importers observe a replaced implementation
-// without an extra layer of wrapper functions in front of every operation
+// Re-exported as mutable bindings, so the implementation can be swapped at runtime.
 export let randomBytes: CryptoPrimitives["randomBytes"];
 export let encryptAES128ECB: CryptoPrimitives["encryptAES128ECB"];
 export let encryptAES128CBC: CryptoPrimitives["encryptAES128CBC"];
--- a/packages/core/src/crypto/operations.ts
+++ b/packages/core/src/crypto/operations.ts
@@ -1,185 +1,58 @@
 import { Bytes, type BytesView } from "@zwave-js/shared";
-import type { CryptoPrimitives, KeyPair } from "@zwave-js/shared/bindings";
+import type { CryptoPrimitives } from "@zwave-js/shared/bindings";
 import { BLOCK_SIZE, leftShift1, xor, zeroPad } from "./shared.js";
 
 // Import the correct primitives based on the environment
 import { primitives as defaultPrimitives } from "#crypto_primitives";
 
-// Crypto is a property of the runtime rather than of a Z-Wave network, so the implementation is
-// process-global and shared by all drivers running in the same process
-let primitives: CryptoPrimitives = defaultPrimitives;
+// Re-exported as mutable bindings, so importers observe a replaced implementation
+// without an extra layer of wrapper functions in front of every operation
+export let randomBytes: CryptoPrimitives["randomBytes"];
+export let encryptAES128ECB: CryptoPrimitives["encryptAES128ECB"];
+export let encryptAES128CBC: CryptoPrimitives["encryptAES128CBC"];
+export let encryptAES128OFB: CryptoPrimitives["encryptAES128OFB"];
+export let decryptAES128OFB: CryptoPrimitives["decryptAES128OFB"];
+export let decryptAES256CBC: CryptoPrimitives["decryptAES256CBC"];
+export let encryptAES256OFB: CryptoPrimitives["encryptAES256OFB"];
+export let decryptAES256OFB: CryptoPrimitives["decryptAES256OFB"];
+export let encryptAES128CCM: CryptoPrimitives["encryptAES128CCM"];
+export let decryptAES128CCM: CryptoPrimitives["decryptAES128CCM"];
+export let digest: CryptoPrimitives["digest"];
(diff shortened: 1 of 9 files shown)


packages/core/src/crypto/operations.ts
```diff
@@ -1,47 +1,185 @@
...
+}
+
+/** Encrypts a payload using AES-128-ECB */
+export function encryptAES128ECB(
+	plaintext: BytesView,
+	key: BytesView,
+): Promise<BytesView> {
+	return primitives.encryptAES128ECB(plaintext, key);
+}
+
```
AlCalzone: Instead of wrapping each individual function, can we instead export getters that dynamically read the primitive's implementation

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -1592,8 +1593,13 @@ export class Driver extends TypedEventTarget<DriverEventCallbacks>
 				?? (await import("#default_bindings/db")).db,
 			log: this._options.host?.log
 				?? (await import("#default_bindings/log")).log,
+			crypto: this._options.host?.crypto
+				?? (await import("#default_bindings/crypto")).crypto,
```
AlCalzone: Does the bindings object need a reference to `crypto`? Isn't calling `setCryptoPrimitives` enough?

AlCalzone: Zniffer explicitly omits `crypto` from its bindings field, FWIW

packages/core/src/crypto/operations.ts
AlCalzone: Done in def023d46 — the 17 wrappers are gone. The exports are now `export let` bindings that `setCryptoPrimitives` reassigns; ES module bindings are live, so importers pick up a replacement with no call-time indirection. One subtlety worth noting: they are assigned with `.bind(impl)` rather than destructured, because plain destructuring drops `this` and would break a host that passes an object whose methods use it — oxlint's `unbound-method` rule flagged exactly that when I tried the destructuring form first.

packages/zwave-js/src/lib/driver/Driver.ts
AlCalzone: Right, it does not — removed in def023d46. The driver now passes the resolved implementation straight to `setCryptoPrimitives` and its `bindings` field omits `crypto`. Zniffer keeps omitting it too, since `Required<host>` still includes it.

packages/core/src/crypto/operations.ts
```diff
@@ -1,47 +1,58 @@
...
-	encryptAES128OFB,
-	encryptAES256OFB,
-	encryptChaCha20Poly1305,
-	generateECDHKeyPair,
-	hmacSHA256,
-	keyPairFromRawECDHPrivateKey,
-	randomBytes,
-};
+// Re-exported as mutable bindings, so importers observe a replaced implementation
+// without an extra layer of wrapper functions in front of every operation
```
AlCalzone: Remove the comment
```suggestion
// Re-exported as mutable bindings, so the implementation can be swapped at runtime.
```

AlCalzone: Applied verbatim in d7094aac0.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -1594,6 +1600,12 @@ export class Driver extends TypedEventTarget<DriverEventCallbacks>
 				?? (await import("#default_bindings/log")).log,
 		};
 
+		// Must happen before anything performs a crypto operation
+		setCryptoPrimitives(
+			this._options.host?.crypto
+				?? (await import("#default_bindings/crypto")).crypto,
+		);
```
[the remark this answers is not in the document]
AlCalzone: Valid, fixed in 3feb02e79. `Driver.start()` now only calls `setCryptoPrimitives` when `host.crypto` was actually passed, so a directly installed implementation survives. The fallback was redundant anyway: `operations.ts` initializes itself from `#crypto_primitives` at module load, so with the option unset the default is already in place. That let the `#default_bindings/crypto" indirection and its two binding modules go as well.

PR #8999, opened 2026-07-28
feat(config): allow embedders to specify the embedded config directory

AlCalzone:
## Why

`@zwave-js/config` locates its shipped configuration files with `import.meta.resolve`. That function does not exist in QuickJS, quickjs-ng or txiki.js, so on those runtimes the package cannot even be imported. Deno, Bun and browser bundlers work today, but they also have no way to point Z-Wave JS at a different location. The same applies to applications bundling Z-Wave JS into a single executable.

## What

- New driver option `storage.deviceConfigEmbeddedDir`, forwarded to `ConfigManager` as `deviceConfigEmbeddedDir`.
- Resolving the embedded directory is now a `getConfigDir()` function rather than a module-level constant, so importing `@zwave-js/config` no longer evaluates `import.meta.resolve`.
- The `#config_dir` conditional import is gone, along with `configDir.node.ts` / `configDir.browser.ts`. A single `configDir.ts` serves every runtime: it resolves via `import.meta.dirname`, which is defined exactly when resolution is possible, and otherwise throws telling the embedder to set the option. That also removes the `node:url` import which was the only reason the two variants existed.
- `ConfigManager` exposes the location as `embeddedConfigDir` / `embeddedDevicesDir` and threads it through `syncExternalConfigDir`, the manufacturers loader/saver and the device index loaders, instead of those modules resolving it themselves. It resolves on first use, so a driver started with `testingHooks.loadConfiguration: false` never needs it.

## Drive-by fix

The built output resolved the config directory one level too high, landing on the package root instead of `<pkg>/config`. `"../.."` was correct when the build emitted `build/esm/configDir.node.js`, but `next` flattened that to `build/configDir.js`. Both the source and build layouts place the module one level below the package root, so it is now a single `".."`.

## Breaking changes

- **On runtimes where `import.meta.dirname` is undefined — browsers, QuickJS, txiki.js, and single-executable bundles — `storage.deviceConfigEmbeddedDir` is now required.** Previously the browser build resolved `import.meta.resolve("/config")`; it now throws `ZWaveErrorCodes.Driver_InvalidOptions` when the configuration is first accessed, rather than guessing a path. `packages/web` passes `/config` explicitly. Behavior on Node is unchanged.
- **`embeddedDevicesDir` is no longer exported from `@zwave-js/config`.** Its only consumer was `Node.ts`; it is now the `ConfigManager.embeddedDevicesDir` getter, since the location is no longer a module-level constant.

## Verification

- `yarn build`, `yarn lint:ts`, `yarn fmt:check`, `yarn check:browser`
- `yarn test:ts packages/config/src` — 87 tests pass
- `yarn test:ts` on `configImportFallback` plus `packages/zwave-js/src/lib/test/node` — 142 tests pass
- `node .build.mjs` in `packages/web` — the demo still bundles, and `import.meta.dirname` survives into the bundle as `undefined`, so the option is enforced there
- Confirmed both layouts resolve to `<pkg>/config` with `devices/index.json` present: the built module directly, and the source module via the config tests loading real device files


packages/config/src/configDir.browser.ts
```diff
@@ -1,2 +1,4 @@
 /** The absolute path of the embedded configuration directory */
-export const configDir = import.meta.resolve("/config");
+// Fall back to a bare path because QuickJS and txiki.js have no import.meta.resolve
+// and pass the deviceConfigEmbeddedDir option instead
```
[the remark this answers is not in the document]
AlCalzone: Reworded in 808e03ccf — the subject is now the embedder rather than the runtime.

packages/config/src/configDir.browser.ts
AlCalzone, line 1: I think we should require browser embedders to pass the driver option instead of having a single fallback only for browsers.

packages/zwave-js/src/lib/driver/ZWaveOptions.ts
```diff
@@ -214,6 +214,13 @@ export interface ZWaveOptions {
 		 */
 		deviceConfigPriorityDir?: string;
 
+		/**
+		 * Allows you to specify where the configuration files embedded in the `@zwave-js/config` package are located.
+		 * By default, this location is resolved automatically, which requires support for `import.meta.resolve`.
+		 * Embedders on runtimes without it, e.g. QuickJS or txiki.js, have to set this option.
+		 */
+		deviceConfigEmbeddedDir?: string;
```
AlCalzone, line 222: We should treat this more like a runtime hint, because it also affects SEA bundling

```suggestion
		/**
		 * Allows specifying which path to use to resolve the configuration files embedded in the `@zwave-js/config` package.
		 * By default, this location is resolved automatically, which requires support for `import.meta.resolve`.
		 * Embedders on runtimes without it, e.g. QuickJS or txiki.js, or applications bundling zwave-js into an executable, have to set this option.
		 */
		deviceConfigEmbeddedDir?: string;
```

packages/config/src/configDir.browser.ts
AlCalzone: Done in 3edd17e7c. `#config_dir` now exports `getConfigDir()` rather than a constant, and the browser variant throws `Driver_InvalidOptions` instead of resolving anything — so importing `@zwave-js/config` no longer evaluates `import.meta.resolve` at all, which is what QuickJS needs. `ConfigManager` resolves on first use rather than in the constructor, because `Driver.start()` always constructs one and `flasher.ts` legitimately runs with `loadConfiguration: false`. `packages/web/src/script.ts` now passes `/config`. Called out as a breaking change in the PR body.

packages/zwave-js/src/lib/driver/ZWaveOptions.ts
AlCalzone: Applied verbatim in 3edd17e7c, in both `ZWaveOptions.ts` and the `docs/api/driver.md` copy.

packages/config/src/ConfigManager.ts
```diff
@@ -46,10 +47,25 @@ export class ConfigManager {
 
 		this.deviceConfigPriorityDir = options.deviceConfigPriorityDir;
 		this.deviceConfigExternalDir = options.deviceConfigExternalDir;
+		this.#deviceConfigEmbeddedDir = options.deviceConfigEmbeddedDir;
```
AlCalzone: Why is this the only property using # instead of TypeScripts `private`?

packages/config/src/configDir.ts
```diff
@@ -0,0 +1,18 @@
+import { ZWaveError, ZWaveErrorCodes } from "@zwave-js/core";
+import path from "pathe";
+
+/** Returns the absolute path of the embedded configuration directory */
+export function getConfigDir(): string {
```
AlCalzone: Since this is only for the embedded config, I think it should be called `getEmbeddedConfigDir`.

packages/config/src/ConfigManager.ts
AlCalzone: No good reason — changed to `private deviceConfigEmbeddedDir` in 69a606c05, matching its `deviceConfigPriorityDir` / `deviceConfigExternalDir` siblings.

packages/config/src/configDir.ts
AlCalzone: Renamed to `getEmbeddedConfigDir` in 69a606c05.

PR #9000, opened 2026-07-28
refactor(shared): provide setImmediate instead of using the Node global

AlCalzone:
## Why

`setImmediate` is a Node.js-only global. Around 17 production call sites used it directly, so the driver could not even start on non-Node runtimes — QuickJS/txiki.js, Deno, Bun, or in the browser.

`queueMicrotask` is not a usable substitute: draining microtasks never advances the event loop, and every one of these call sites depends on pending I/O being flushed first.

## What

`@zwave-js/shared` now exports `setImmediate`, returning an `Immediate` with a `clear()` method. This mirrors the existing `setTimer`/`setInterval` shape, so the idle timer in `Driver.drainTransactionQueue` keeps working without a separate `clearImmediate`.

The backend is resolved once at module scope:

1. `globalThis.setImmediate` when present (Node).
2. `MessageChannel` when present (browsers, workers). Posting a message queues a *task*, not a timer, so neither Node's 1 ms delay floor nor the HTML spec's 4 ms clamp on nested timeouts applies. Dispatch goes through a handle id into a `Map`, with the same re-entrancy guard as the YuzuJS polyfill so callbacks never run nested.
3. `setTimeout(fn, 0)` as a last resort.

All call sites in `Controller.ts`, `Driver.ts`, `Node.ts`, `70_FirmwareUpdate.ts`, `Zniffer.ts`, `MockControllerBehaviors.ts` and `SerialModeSwitch.ts` now import it from `@zwave-js/shared`. The `IS_TEST` control flow in `SerialModeSwitch.ts` is unchanged.

`packages/web/setimmediate.js` is left in place: `packages/web/index.html` still loads it, and bundled third-party dependencies may still rely on the global.

## Verification

- `yarn test:ts packages/zwave-js` — 786 tests pass
- `yarn test:ts packages/serial packages/shared` — 123 tests pass
- `yarn check:browser`, `yarn build`, `yarn lint:ts` — clean
- New `packages/shared/src/Timers.test.ts` runs asynchronous-dispatch, argument-forwarding and `clear()` assertions against all three backends, forcing the fallbacks by re-importing the module with the globals stubbed out.

### DIFF
--- a/packages/shared/src/Timers.ts
+++ b/packages/shared/src/Timers.ts
@@ -188,8 +188,6 @@ export function setInterval<TArgs extends any[]>(
 
 /**
  * Schedules the callback to run in a later task, once the microtask queue has drained.
- * This uses Node's `setImmediate` where available, so its ordering relative to timers
- * and I/O is whatever the runtime provides.
  */
 export function setImmediate<TArgs extends any[]>(
 	callback: (...args: TArgs) => void,
--- a/packages/shared/src/Timers.ts
+++ b/packages/shared/src/Timers.ts
@@ -63,6 +63,109 @@ export class Interval {
 	}
 }
 
+interface ImmediateBackend {
+	schedule(
+		callback: (...args: any[]) => void,
+		args: any[],
+	): NodeJS.Immediate | number;
+	clear(handle: NodeJS.Immediate | number): void;
+}
+
+function createImmediateBackend(): ImmediateBackend {
+	// Typed as possibly undefined because only Node has this
+	const nativeSetImmediate: typeof globalThis.setImmediate | undefined =
+		globalThis.setImmediate;
+	if (nativeSetImmediate) {
+		return {
+			schedule: (callback, args) => nativeSetImmediate(callback, ...args),
+			clear: (handle) =>
+				globalThis.clearImmediate(handle as NodeJS.Immediate),
+		};
+	}
+
+	const pending = new Map<
+		number,
+		{ callback: (...args: any[]) => void; args: any[] }
+	>();
+	let nextHandle = 1;
+	let running = false;
+
+	function run(handle: number): void {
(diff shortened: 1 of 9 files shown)


packages/shared/src/Timers.ts
```diff
@@ -82,3 +185,14 @@ export function setInterval<TArgs extends any[]>(
 ): Interval {
 	return new Interval(globalThis.setInterval(callback, delay, ...args));
 }
+
+/**
+ * Schedules the callback to run after the current operation completes and pending I/O
+ * has been processed, equivalent to Node's `setImmediate`.
+ */
```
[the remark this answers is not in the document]
AlCalzone: Correct, fixed in c3a61ee7f. The JSDoc now only promises a later task once microtasks have drained, and says the ordering relative to timers and I/O is whatever the runtime provides. Node's check-phase placement was never something the MessageChannel or setTimeout paths could offer, and no call site depends on it.

packages/shared/src/Timers.ts
```diff
@@ -82,3 +185,15 @@ export function setInterval<TArgs extends any[]>(
 ): Interval {
 	return new Interval(globalThis.setInterval(callback, delay, ...args));
 }
+
+/**
+ * Schedules the callback to run in a later task, once the microtask queue has drained.
+ * This uses Node's `setImmediate` where available, so its ordering relative to timers
+ * and I/O is whatever the runtime provides.
```
AlCalzone: Do not go into implementation details on the function documentation

AlCalzone: Trimmed in c27b67a9b — the doc now states only the contract, that the callback runs in a later task once microtasks have drained.

AlCalzone:
@zwave-js-bot automerge

PR #9002, opened 2026-07-29
feat: add productId 0x5600 for YRM276 with ZW3 module (700 series)

vectorcatch:
The YRM276 can have an update ZWave module inserted (700 series). It maintains the same settings as with the ZW2 (500 series) module.


AlCalzone (approved):
Thanks!

PR #9003, opened 2026-07-29
fix(test): drop mock port frames sent during teardown

AlCalzone:
`MockPort.emitData` enqueued into the source `ReadableStreamDefaultController` without guarding against it already being closed. A mock controller behavior that responds after the port was destroyed therefore threw `ERR_INVALID_STATE` as an unhandled rejection, which vitest reports as a failed shard even when every test passed. It hit #8998 and #9000 on shard 2/4, in both cases originating outside the code those PRs touch.

`destroy()` also skipped clearing `#sourceController` whenever `close()` threw, leaving a closed controller in place for any later `emitData`.

Both hunks are cherry-picked verbatim from #8935, which fixed this on `master` — `next` is 48 commits behind and never received it.

Verified by running `yarn test:ts --shard 2/4` locally; the race is intermittent, so this is the upstream fix rather than a local reproduction.


AlCalzone:
Obsolete — `next` has been rebased onto `master`, which already carries this fix from #8935.

PR #9008, opened 2026-07-29
test: wait for the expected User Code events instead of sleeping

AlCalzone:
## Why

Three assertions in `accessControl.UserCode.test.ts` slept for a fixed 50–100 ms and then required the expected events to have already arrived:

```ts
await wait(100);
t.expect(events).toStrictEqual([ /* 3 events */ ]);
```

On a loaded CI runner the unsolicited report has not been processed yet, so `events` is still `[]` and the shard fails. This flake failed shard 2/4 on three unrelated PRs today (#8998, #9000's neighbour, #9007), each time in a PR that could not have caused it.

## What

Those three now retry the same assertion with `vi.waitFor` instead of sleeping once, so they finish as soon as the events arrive and only fail if they never do. The assertions themselves are unchanged.

The two assertions that require **no** events to arrive keep their fixed `await wait(100)`:

```ts
// Solicited reads are discovery, not changes
await wait(100);
t.expect(events).toStrictEqual([]);
```

Retrying those would satisfy them on the first attempt and stop testing anything, so the delay there is load-bearing.

## Verification

- `yarn test:ts packages/zwave-js/src/lib/test/node/accessControl.UserCode.test.ts` — 57 tests pass, run three times
- `yarn test:ts --shard 2/4` — 72 files / 493 tests pass, the shard and grouping where CI was failing
- `yarn build`, `yarn lint:ts`, `yarn fmt:check`

### DIFF
--- a/packages/zwave-js/src/lib/test/node/accessControl.UserCode.test.ts
+++ b/packages/zwave-js/src/lib/test/node/accessControl.UserCode.test.ts
@@ -33,6 +33,11 @@ import {
 } from "../../node/feature-apis/AccessControl.js";
 import { integrationTest } from "../integrationTestSuite.js";
 
+// Emitted events may lag behind the frame that triggers them, so assertions on them
+// are retried. The bound is generous because CI runners are slow under load, and well
+// below the 30s each integration test gets, so a real mismatch still fails as one.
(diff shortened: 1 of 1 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #9009, opened 2026-07-29
feat: add a txiki.js/QuickJS host as a demo runtime

AlCalzone:
## Why

Z-Wave JS has runtime-agnostic bindings for fs, db, serial, log and crypto, but nothing exercises them outside Node and the browser, so embedded targets can break silently and stay broken. This adds a real host for [txiki.js](https://github.com/saghul/txiki.js) (quickjs-ng + libuv), with a boot check that starts an actual driver rather than only proving the bundle links.

## What

A new **private** package `packages/quickjs` (never published) with the bindings a host has to supply — `fs` over the `tjs` file APIs, `db` over `tjs:sqlite`, `serial` over `tjs.connect` using the `connect` capability from #9001, and the browser log bindings unchanged. Crypto needs no binding: bundling with `--conditions=browser` selects the WebCrypto-based primitives, and txiki's `crypto.subtle` covers everything they use.

`src/boot.ts` is the check. It is only a driver — the mock controller and node run in a separate Node.js process via `yarn mock-server`, and the driver reaches them over TCP through the `connect` binding. It waits for `driver ready` and for each node's interview, verifies the device config index was not needlessly regenerated, and destroys the driver. `test/secure_server_config.cjs` grants the node S2, so the run also exercises the nonce exchange and AES-CCM.

Everything lives inside the new package, apart from `test/secure_server_config.cjs`, the waddle bump, and a `.yarnrc.yml` entry exempting our own packages from `npmMinimalAgeGate` so that bump could be installed.

## Memory footprint vs Node.js

Same workload on both — connect to the mock server, interview the node, destroy the driver, exit.

| stage | txiki.js 26.6.0 | Node.js 24 |
| --- | --- | --- |
| empty runtime | 9.9 MB | 42.4 MB |
| after loading zwave-js | 56.1 MB | 117.8 MB |
| after `new Driver()` | 84.7 MB | 118.0 MB |
| after driver ready | 61.9 MB | 154.8 MB |
| after node interview | 62.0 MB | 155.4 MB |
| **peak** | **75.7 MB** | **165.4 MB** |
| wall clock | 3.43 s | 3.12 s |

**txiki.js runs the driver in less than half the memory, with comparable startup time.** Peak measured with `/usr/bin/time -v`, stages from `/proc/self/status` and `process.memoryUsage().rss`. The txiki spike at construction collects back down, so it is transient garbage rather than retained. Node's RSS is mostly not heap — at 117.8 MB its `heapUsed` is only 35 MB, the rest being V8 reservations and the snapshot, which accounts for most of the 4.3x baseline difference.

Where quickjs is clearly slower is CPU-bound work: generating the device config index takes ~5 s of a cold boot, and there is no JIT. For an I/O-bound driver at 115200 baud that does not matter.

### Source maps, and why the build sets `sourcemap: "external"`

txiki loads and parses a referenced source map at startup whether or not anything throws:

| | maps not loaded | maps loaded | penalty |
| --- | --- | --- | --- |
| txiki.js | 75.7 MB | 179 MB | **+103 MB** (one 8.2 MB map) |
| Node.js | 165.4 MB | 184.4 MB | +19 MB (12.6 MiB across 704 maps) |

Node pays 5x less for 1.5x more map data — partly because it only reads maps for modules actually imported, while the bundle's single map is loaded whole, and partly because inflating 8.2 MB of JSON into ~103 MB suggests eager parsing into a heavy structure. The map is still emitted here, just not referenced.

Caveat on the comparison: Node used the default `node:fs` + `jsonl` bindings while txiki used the new `tjs` fs + SQLite ones, so the storage layers differ. The driver-side work is identical.

## Notes and out of scope

- **Direct `/dev/tty*` access is not implemented**, so `createFactoryByPath` accepts only `tcp://` and `esphome://`, and `Serial.list` is unimplemented. libuv's behavior on a serial fd is unverified and txiki exposes no way to set baud rate or line discipline. This is the main gap before the package is useful against real hardware.
- **The ESPHome transport is wired but never exercised** — only the plain `tcp://` path was run.
- **Granted security classes are out of reach through mock-server.** A node it serves is never *granted* a class, since that only happens through inclusion, so the driver encrypts and exchanges nonces while reporting no security class for the node.
- **`packages/quickjs` is not part of `yarn build`**, mirroring `packages/web`, so CI neither builds nor type-checks it and nothing runs this boot automatically. Wiring that up is the obvious follow-up — it is what would have caught the two portability bugs this work turned up (#9014, #9015) without a human in the loop.


packages/zwave-js/src/lib/driver/NetworkCache.ts
```diff
@@ -420,7 +420,8 @@ export function deserializeNetworkCacheValue(
 	switch (cacheKeyUtils.nodePropertyFromKey(key)) {
 		case "interviewStage": {
 			value = tryParseInterviewStage(value);
-			if (value) return value;
+			// InterviewStage.None is 0, which a falsy check would reject
```
AlCalzone: Drop the comment

packages/zwave-js/src/lib/driver/NetworkCache.ts
```diff
@@ -470,7 +471,8 @@ export function deserializeNetworkCacheValue(
 
 		case "nodeType": {
 			value = tryParseNodeType(value);
-			if (value) return value;
+			// NodeType.Controller is 0, which a falsy check would reject
```
AlCalzone: Drop the comment

packages/quickjs/src/mockPortServer.ts
AlCalzone, line 1: We don't need this outside of testing, do we? Wouldn't it make more sense to just do `yarn mock-server`?

packages/quickjs/types/tjs.d.ts
```diff
@@ -0,0 +1,111 @@
+// Minimal ambient declarations for the txiki.js runtime APIs used by this package.
+// txiki.js publishes full types as `@txikijs/types`, but pulling in a runtime-specific
+// dependency for a handful of globals is not worth it.
```
AlCalzone: Actually I think having @types/txiki.js be a devDependency of this package is okay.

packages/zwave-js/src/lib/controller/MockControllerBehaviors.ts
```diff
@@ -78,7 +78,7 @@ import {
 	createMockZWaveRequestFrame,
 } from "@zwave-js/testing";
 import { wait } from "alcalzone-shared/async";
-import { createDefaultMockNodeBehaviors } from "../../Testing.js";
```
AlCalzone, line 81: Explain this change

packages/zwave-js/src/Testing_browser.ts
AlCalzone, line 1: Do we need this if we run the mock-server with Node.js?

packages/zwave-js/src/lib/driver/NetworkCache.ts
AlCalzone: Dropped in ee32952d3.

packages/zwave-js/src/lib/driver/NetworkCache.ts
AlCalzone: Dropped in ee32952d3.

packages/quickjs/src/mockPortServer.ts
AlCalzone: Agreed, gone in ee32952d3. `boot.ts` is now a plain driver connecting to `tcp://127.0.0.1:5555`, and `packages/quickjs/mock-config.js` configures `yarn mock-server` with one secure node. One prerequisite: `MockServerControllerOptions` only picked `ownNodeId | homeId | capabilities`, so a config could not hold the network keys and S2 could not work at all — I added `securityKeys` to that `Pick`. Also note `getConfig` in `bin/mock-server.mjs` silently ignores `.mjs` (it only handles .js/.cjs/.json/.dump), which cost me a debugging round — worth an explicit error.

packages/quickjs/types/tjs.d.ts
AlCalzone: Done in ee32952d3, with one correction: `@types/txiki.js` does not exist. The published package is `@txikijs/types`, and 26.6.0 matches the runtime version exactly. It covers the `tjs` global plus `sqlite`/`ffi`. Worth it — it immediately caught two things the hand-written declarations hid: `ConnectOptions.keepAliveDelay` is in **seconds** while the binding contract is milliseconds (so keep-alive was being set to 1000s), and `RemoveOptions` has no `recursive` because `tjs.remove` is already `rm -rf`. It does omit `signal`, which I verified is honored at runtime, so that one needs a cast.

packages/zwave-js/src/lib/controller/MockControllerBehaviors.ts
AlCalzone: Reverted in ee32952d3. The reason was bundling: it imported `createDefaultMockNodeBehaviors` from `../../Testing.js`, the package's own barrel, which re-exports `mockServer.ts` and pulls in `@homebridge/ciao` plus 12 Node builtins — so any bundle reaching the mock behaviors dragged all of that in. With the harness in Node that no longer matters. The direct import does still avoid a package importing its own barrel, if you want it as a separate cleanup.

packages/zwave-js/src/Testing_browser.ts
AlCalzone: No — removed in ee32952d3, along with the `browser` condition on the `./Testing` export, so there is no public resolution change left in this PR.

packages/zwave-js/src/mockServer.ts
```diff
@@ -58,7 +58,7 @@ import type { CommandClassDump, NodeDump } from "./lib/node/Dump.js";
 export type MockServerControllerOptions =
 	& Pick<
 		MockControllerOptions,
-		"ownNodeId" | "homeId" | "capabilities"
```
AlCalzone, line 61: This seems unrelated - undo

Issue #5638, opened 2023-04-07
Battery based devices don't get marked as dead if they haven't checked in recently


Hooverjc:
### Is your problem within Home Assistant (Core or Z-Wave JS Integration)?

NO, my problem is NOT within Home Assistant or the ZWave JS integration

### Is your problem within Z-Wave JS UI (formerly ZwaveJS2MQTT)?

NO, my problem is NOT within Z-Wave JS UI

### Checklist

- [X] I have checked the [troubleshooting section](https://zwave-js.github.io/node-zwave-js/#/troubleshooting/index) and my problem is **not** described there.

- [X] I have read the [changelog](https://github.com/zwave-js/node-zwave-js/blob/master/CHANGELOG.md) and my problem was **not** mentioned there.

### Describe the bug

What causes the bug?
When a battery based device fails to check-in due to a dead battery, this device continues to be marked as "assleep" despite the fact that it will never again check-in. Because the battery indicator won't update any longer, this can cause battery based devices to go dead unnoticed.

What do you observe?
The device remains "asleep" despite it having a dead battery. I've seen this go on for weeks. Usually i'd catch the battery getting low and set warnings, but in the winter, Lithium batteries can die very quickly, causing a device to go from 50+% to 0 in a few hours. This results in a dead device with a previously reported high battery level and no automatic form of flagging the device as dead.

What did you expect to happen?
I'd expect some sort of configurable timeout. For example, if a device is battery based and doesn't check-in at all for X days, perhaps either marking it as dead or raising some sort of event notifying that the device may be unresponsive. Not sure if this would be global or per device, but some option here to bring light to these situations in an automatic fashion would be nice.

  1. Add a battery device and make sure its configured and marked as active with all parameters populated.
  2. Remove the devices battery
  3. Wait for a week or more and the device will continue to be marked as "asleep".

### Device information

Seems agnostic to devices as all my battery devices can be impacted
Manufacturer: Zooz
Model name: ZSE41
Node ID in your network: 46

This particular node has been dead for weeks, yet is still marked as asleep. Device specific settings are configured to wake up every 12 hours:

![image](https://user-images.githubusercontent.com/42452872/230446390-e19fa311-b7c2-468a-8a6d-19ba840c0bfb.png)

### How are you using `node-zwave-js`?

- [X] `zwave-js-ui` (formerly `zwavejs2mqtt`) Docker image (latest)
[...]

Issue #4207, opened 2022-02-03
[feat] Identify dead devices


The-Deep-Sea:
**Is your feature request related to a problem? Please describe.**
Hi. I'm new to Z-Wave, but have more experience with Zigbee.
Above all, I miss the fact that devices that do not report regularly are marked as dead.

**Describe the solution you'd like**
Zigbee2mqtt solves this e.g. like this:
- Sleeping nodes are marked as dead after a period of inactivity (default: 25h).
- Alive nodes are pinged after a period of inactivity (default: 10min). If they don't react to this, they are also marked as dead.

See https://www.zigbee2mqtt.io/guide/configuration/device-availability.html for more details.

**Describe alternatives you've considered**
Is this the right approach? Or is there already a solution to the problem that I'm not aware of?
How do you identify a dead passive device such as a smoke detector?

**Additional context**
Wouldn't it be correct to set the entities of dead nodes to "unavailable" when working with Home Assistant? Except possibly the node status.

AlCalzone:
The usecase behind this issue makes sense, but I'm not decided yet what is the best way to do this.

jmgiaever:
> My question wasn't aimed at waking up devices.

No, but to be able to ping a device it needs to support to be woken up. That's why I though maybe mark it as «presumed dead» when

1. the device haven't reported alive within the wake-up interval since zwjs was started
2. responded to the ping within the wake-up interval, that was sent if 1 was the case

raman325:
> **Additional context** Wouldn't it be correct to set the entities of dead nodes to "unavailable" when working with Home Assistant? Except possibly the node status.

Right now all entities but the node status sensor get marked as unavailable when a device is marked dead by the driver. We were actually just having a separate discussion about whether this is the right behavior because there are cases where the driver may think the device is dead but the next action against the device would cause it to respond and make the node alive again, but there's a catch 22 because the entity is unavailable in HA so you can't perform normal actions against the device to revive it (you can ping it, but there's no UI component to do that so it's not as easy of a recovery setup).

In your example, the false alarm above would be after the fire alarm was marked dead but you put the battery back in - something needs to happen to tell the driver that the node is alive, otherwise it's still dead even though it's functioning again.

[...]

PR #9086, opened 2026-08-12
feat: detect sleeping nodes that stopped waking up

ismell:
# feat: detect sleeping nodes that stopped waking up

Refs #4207. This implements one of the four detection cases listed there — overdue sleeping devices — and deliberately leaves the issue open for the rest.

#4207 has been open since 2022 because the design was undecided, not because nobody wrote the code. So please treat this as a concrete proposal rather than a finished answer. If you would rather settle the approach in the issue first, I am happy to close this and move the discussion there.

**In short:** nothing is transmitted, and no existing behavior changes — this adds a property and two events, and touches nothing else.

## The problem

A sleeping node can never be marked dead. `NodeStatusMachine` only permits `asleep -> awake`, and both `markAsDead()` call sites in `Driver.ts` are gated on `!node.canSleep`. That is deliberate and correct — not responding is exactly what a healthy sleeping node does, so a controller cannot tell "asleep" from "gone" by attempting to communicate.

The consequence is that a battery device whose cells have died is indistinguishable from a healthy one that has not woken up yet. It stays `asleep` and `ready: true` forever.

## A source I did not see referenced in the discussion

Silicon Labs' gateway documentation defines a bound for how long a Reporting Sleeping Slave may stay silent, and applying it requires transmitting nothing at all. From a section called **"Detection of Sleeping Device Failure"**:

> "[If] a duration of longer than **(2 \* [Wake Up Interval] + 10%)** passes without receipt of a Wake Up Notification from the RSS, it can be marked as 'failed.'"

It also excludes exactly the devices that make a general rule impossible — the ones that only wake on local events:

> "…for a device whose Wake Up Interval is **'0'**, no predictable pattern of expected Wake Up Notifications can be determined. As a result, failure of these devices cannot be detected."

Silicon Labs' own Z/IP Gateway implements this. The [companion lab](https://docs.silabs.com/z-wave/1.0.0/z-wave-training-labs/gateway-session4-sleeping-nodes-lab) walks through a node with a 20 s interval being reported as failing after `2 * 20 s + 10% = 44 s`, and recovering on its next notification:

```
mailbox node 6 is now failing, because it has not reported in
Node 6 is now ok.
```

Sources:

- https://docs.silabs.com/z-wave/1.0.1/z-wave-training-docs/gateway-session4-sleeping-nodes
- https://docs.silabs.com/z-wave/1.0.0/z-wave-training-labs/gateway-session4-sleeping-nodes-lab

One caveat on how normative this is: it is gateway implementation guidance — "*can* be marked as 'failed'" — not a certification `SHALL`, and I did not find an equivalent requirement in an SDS document. The threshold is a well-sourced default rather than a mandate, which is part of why this PR only reports and never acts.

## Against your requirements from the 2022 discussion

| Requirement | Status |
| --- | --- |
| Detect when a sleeping device hasn't woken up for too long | This PR. |
| Detect when a mains powered device no longer reacts to pings | Already handled — mains nodes are marked dead on failed communication today. Unchanged here. |
| Detect when a FLiRS device no longer reacts, at reduced frequency | Not addressed — see below. |
| Detect when the above are reachable again | Anything received from the node clears it. |
| Avoid false positives on fragile meshes | Doubling the interval tolerates one fully missed notification; the extra 10% absorbs sleep-timer drift. |
| Avoid delays from pinging unreachable devices | Nothing is transmitted. The check only observes that an expected notification did not arrive. |

On #5638, which was closed as a duplicate of this: the first objection there — that we have no way of knowing how long a battery device may go — is answered by the interval-0 exclusion above, which removes precisely the devices that have no schedule. The second objection, that flaky devices would get marked dead repeatedly, cannot happen here at all, because `NodeStatus` is never touched; a flapping node just toggles `isOverdue`.

## Why this only reports, and does not mark anything dead

My first attempt also promoted overdue nodes to `NodeStatus.Dead` behind an opt-in flag. I dropped it, and the reason seems worth recording since it is the obvious next question.

`NodeStatus` is effectively a "what do I do with messages for this node" machine, and overdue-ness is orthogonal to it — an overdue node is still asleep, and its messages should still be parked in the wake-up queue. Cramming the two together broke real things:

- `mayStartTransaction` gates on `status !== Asleep`, so a `Dead` sleeping node has its commands transmitted immediately rather than parked — burning `maxSendAttempts` against a node presumed to have a dead battery, and making the "never transmits" property false.
- `handleMissingNodeACK` re-queues a transaction without `reset()` and relies on `markAsAsleep()` to trigger `moveMessagesToWakeupQueue`, whose reducer carries `reset: true`. From `dead`, `markAsAsleep()` is a no-op, so the re-queued transaction resolves to `undefined` — `sendCommand()` reports success for a command the node never acknowledged.
- `markAsAwake()` is likewise a no-op from `dead`, so a dead sleeping node that does answer stays dead.

Around a dozen further sites assume `canSleep` implies `status ∈ {Unknown, Asleep, Awake}` — transaction sort order, the health checks' `waitForWakeup`, interview finality, route rebuilding. Making this work would mean special-casing "dead but sleeping" in each, i.e. inventing a second implicit status.

And the benefit I thought I was buying turned out not to exist. I assumed a `Dead` node's entities go unavailable in Home Assistant; they do not. `ZWaveBaseEntity.available` keys on `node.ready`, and `NodeReadyMachine` has no transition out of `ready`, so the node stays available with `status: Dead`. The only visible difference is the node-status sensor's value — the same visibility class as `isOverdue`, obtainable downstream at zero risk to queue semantics.

If you do want overdue-ness reflected in `NodeStatus`, I would rather agree the shape in #4207 first than guess again.

## Reaching applications

`isOverdue` needs a companion `zwave-js-server` change to appear in the node state dump and forward the two events, with a schema bump. Happy to open that PR if this lands. What Home Assistant then does with it — a dedicated binary sensor, or factoring it into availability — is a call for its maintainers.

## Out of scope

FLiRS nodes. They can be reached on demand, so Silicon Labs' guidance detects their failure through failed beaming instead — *"considered 'failed' after all retransmission attempts of a Beam have failed to wake the device"* — which belongs in the transmission path rather than in passive observation, and is the one case where your battery-drain concern genuinely applies. Worth a separate PR if this approach is acceptable.

## Implementation notes

Evaluation is scheduled from the `lastSeen` setter, which every received or successfully sent command already passes through, and from `markAsAsleep()`, which the driver calls for every sleeping node restored from cache. The latter is what catches a node that died during a previous session — by that point the network cache is restored, so both `lastSeen` and `wakeUpInterval` are available. One timer per eligible sleeping node, cleared in `destroy()`; no polling and no periodic work.

A Wake Up Interval is a 3-byte seconds field, so a large interval produces a delay beyond `setTimeout`'s 2^31−1 ms ceiling, where it would silently fire immediately and report a healthy node as overdue. Anything above roughly 11 days is affected, so long waits are re-armed in chunks.

Naming is entirely up for grabs; happy to rename `isOverdue` or the events.

## Verification

Unit tests cover the threshold computation, the `wakeUpInterval == 0` exclusion, unusable values, the `setTimeout` ceiling, and the worked example from the Silicon Labs lab (20 s → 44 s). An integration test starts the driver against a cache fixture reproducing the device below and asserts it comes up overdue and clears when heard from again. Docs updated: `isOverdue` and the two events in `docs/api/node.md`.

The device this came from: a Zooz ZSE44 with the default 12 h wake up interval (a 26.4 h threshold) that last reported in May 2025 and was still `asleep` with `ready: true` more than a year later. It was only noticed because of a `lastSeen`-based watchdog in Home Assistant — the same workaround several people describe in #4207.

---

Written with AI assistance; the Silicon Labs sourcing, the design decisions, and the real-device verification are mine.

### DIFF
--- a/docs/api/node.md
+++ b/docs/api/node.md
@@ -929,6 +929,23 @@ readonly lastSeen: MaybeNotKnown<Date>
 
 This property tracks when the node was last seen, meaning a command was either received from the node or successfully sent to it.
 
+### `isOverdue`
+
+```ts
+readonly isOverdue: boolean
+```
+
+This property indicates that a sleeping node has failed to report in for significantly longer than its configured wake up interval, which usually means its battery is dead or it is out of range. The threshold is twice the wake up interval plus 10%, as recommended for Reporting Sleeping Slaves in Silicon Labs' gateway documentation.
+
+The detection is entirely passive - nodes are never polled to determine this - and it does not affect a node's [`status`](#status). A node stops being overdue as soon as anything is received from it again.
+
+This is always `false` for nodes whose liveness cannot be predicted this way:
+
+- always listening nodes and FLiRS nodes, which can be reached on demand and whose failure is detected by other means
+- nodes with a wake up interval of `0`, which only wake up on local events and therefore have no expected reporting pattern
+
+> [!NOTE] A node restored from cache is evaluated during driver startup, before the `"driver ready"` event. If it was already overdue at that point, the `"overdue"` event fires before an application can attach listeners, so read this property for the initial state rather than relying on the event alone.
+
 ### `isControllerNode`
 
 ```ts
@@ -1290,6 +1307,14 @@ A sleeping node has woken up or gone back to sleep. The node is passed as the si
 ### `"dead"` / `"alive"`
 
 A non-sleeping node has stopped responding or just started responding again. The node is passed as the single argument to the callback:
+
+```ts
+(node: ZWaveNode) => void
+```
+
+### `"overdue"` / `"no longer overdue"`
+
+A sleeping node has missed its expected wake ups for significantly longer than its wake up interval, or has been heard from again after doing so. See [`isOverdue`](#isOverdue) for which nodes this applies to. The node is passed as the single argument to the callback:
 
 ```ts
 (node: ZWaveNode) => void
--- a/packages/zwave-js/src/lib/node/Node.ts
+++ b/packages/zwave-js/src/lib/node/Node.ts
@@ -347,6 +347,7 @@ export class ZWaveNode extends ZWaveNodeMixins implements QuerySecurityClasses {
 		) {
 			timeout?.clear();
 		}
+		this.cancelOverdueCheck();
 
 		// Remove all event handlers
 		this.removeAllListeners();
@@ -432,6 +433,8 @@ export class ZWaveNode extends ZWaveNodeMixins implements QuerySecurityClasses {
 			...cur,
 			lastSeen: value,
 		}));
+		// Hearing from a node resets how long it may stay silent before being overdue
+		this.scheduleOverdueCheck();
 	}
 
 	/**
--- a/packages/zwave-js/src/lib/node/_Types.ts
+++ b/packages/zwave-js/src/lib/node/_Types.ts
@@ -288,6 +288,8 @@ export interface ZWaveNodeEventCallbacks extends ZWaveNodeValueEventCallbacks {
 	sleep: ZWaveNodeStatusChangeCallback;
 	dead: ZWaveNodeStatusChangeCallback;
 	alive: ZWaveNodeStatusChangeCallback;
+	overdue: (node: ZWaveNode) => void;
+	"no longer overdue": (node: ZWaveNode) => void;
 	"interview completed": (node: ZWaveNode) => void;
 	ready: (node: ZWaveNode) => void;
 	"interview stage completed": (node: ZWaveNode, stageName: string) => void;
@@ -330,6 +332,8 @@ export const zWaveNodeEvents = [
 	"sleep",
 	"dead",
 	"alive",
+	"overdue",
+	"no longer overdue",
 	"interview completed",
 	"ready",
 	"interview stage completed",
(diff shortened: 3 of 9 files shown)


exhannibal:
Same failure mode on a Zooz ZSE44, from Home Assistant OS with the official core_zwave_js add-on.

Kitchen node, dead CR2450. Last real contact 16 Aug 00:42 BST; found ~15 days later. Z-Wave JS kept the node asleep / ready. HA still showed the last values (27.3 °C, 44 % battery) as current. last_seen was correct; an automation watching the temperature entity for unavailable never fired, because availability follows node.ready.

A twin bathroom ZSE44 on the same controller kept reporting. After a battery swap the kitchen node checked in immediately (25.0 °C, 100 % battery).

isOverdue with the Silicon Labs 2 × Wake Up Interval + 10% window would have caught this (default 12 h wake-up → ~26 h). Reporting overdue without changing NodeStatus matches what we need; marking the node Dead would not have helped in HA anyway.

Until HA consumes this (sensor and/or availability), we will keep a last_seen watchdog. I cannot load a custom driver on the official add-on without moving the production stick. I can reproduce quickly once isOverdue is visible (Z-Wave JS UI, or HA after the server/integration work): set Wake Up to 300 s, pull the battery, wait ~11 minutes.

AlCalzone:
I'll get to reviewing this eventually. Thanks for your patience.

PR #9095, opened 2026-08-14
feat: read bootloader info from Nabu Casa controllers during the interview

AlCalzone:
Reads the ZWA-2 bootloader version while the application is running, instead of having to reboot into the bootloader and parse the `Gecko Bootloader vX.YY.ZZ` banner off the XMODEM menu.

Requires the firmware side in NabuCasa/silabs-firmware-builder#218, which adds subcommand `0x09` to `FUNC_ID_PROPRIETARY_0`:

```
HOST->ZW: FUNC_ID_PROPRIETARY_0 | 0x09
ZW->HOST: FUNC_ID_PROPRIETARY_0 | 0x09 | major | minor | customer | capabilities[4]
```

## Changes

- `NabuCasaCommand.GetBootloaderInfo = 0x09`, plus `NabuCasaBootloaderInfo` and the `NabuCasaBootloaderCapability` flags enum, all exported from `zwave-js/Controller`.
- `getBootloaderInfo()` on `ControllerProprietary_NabuCasa`.
- The interview queries it when the supported-commands bitmask advertises it and caches the result on `bootloaderInfo`.

## Generic bootloader version

The version itself is not exposed as something Nabu Casa specific. A new `controller.bootloaderVersion` sits next to `controller.firmwareVersion`. `ZWaveController.queryBootloaderVersion()` runs after the proprietary interview and appends the bootloader to `VersionCCValues.firmwareVersions`, so the value reads `["1.3", "3.2.1"]` — index 0 the application, index 1 the bootloader, matching how nodes report secondary firmware targets. `initNodes()` keeps setting just the application version, so it stays concerned with restoring node state.

There is no standardized Serial API command to read a bootloader version yet, so for now the value can only come from a controller-specific query. `ControllerProprietaryCommon` gains an optional `getBootloaderVersion?()` hook, mirroring the existing optional `getRFRegion?()` / `setRFRegion?()`, and `ZWaveController.queryBootloaderVersion()` picks up whichever implementation provides it. Once a standard command is agreed on and shipped, it gets queried there first and the proprietary hook stays as the fallback for controllers that lack it.

`Node.getDefinedValueIDs()` still returns only the proprietary value IDs for the controller node, which is intentional. `firmwareVersions` therefore lives in the controller's value DB without being advertised there; `controller.bootloaderVersion` is the API surface for reading the version.

Older firmware simply does not set bit 9 in the bitmask, so nothing is sent to controllers that would not understand it.

## Fields

- `version` — `major.minor.customer`, e.g. `3.2.1`. One byte per field on the wire. The customer field is the part Nabu Casa bumps per bootloader release.
- `capabilities` — `NabuCasaBootloaderCapability[]`, parsed with `parseBitMask` from the big-endian mask. The enum mirrors the `BOOTLOADER_CAPABILITY_*` defines in the Simplicity SDK's `btl_interface.h`.

The bootloader table's layout version is deliberately not transferred. If Silicon Labs changes that table format, the firmware translates before responding.

## Testing

Verified against a ZWA-2 (serial 8CBFEA8F69D8) running the firmware from NabuCasa/silabs-firmware-builder#218. The exchange:

```
» [REQ] [Proprietary_F0]  payload: 0x09
« [RES] [Proprietary_F0]  payload: 0x09 030201 001004f1
```

Decoded result:

```
supported subcommands: 0, 1, 2, 4, 5, 6, 7, 8, 9
version: 3.2.1
capabilities: EnforceUpgradeSignature, BootloaderUpgrade, GBL, GBLSignature,
              GBLEncryption, PeripheralList, Communication
```

The generic surface, from a run with an empty cache:

```
firmwareVersion:   1.3
bootloaderVersion: 3.2.1
firmwareVersions:  [ '1.3', '3.2.1' ]
```

The log shows `firmwareVersions` being added once as `1.3,3.2.1` rather than written as `1.3` and patched afterwards, confirming it is set in `initNodes()` with the bootloader version already known.

`3.2.1` matches the banner the bootloader itself prints (`Gecko Bootloader v3.02.01`), and re-encoding the seven decoded flags reproduces `0x001004F1`, so the byte-order handling round-trips. The value is populated by the interview without an explicit call, and an explicit `getBootloaderInfo()` returns the same thing.

`yarn build`, `yarn lint:ts` and `yarn fmt:check` pass.

### DIFF
--- a/packages/zwave-js/src/lib/controller/proprietary/NabuCasa.ts
+++ b/packages/zwave-js/src/lib/controller/proprietary/NabuCasa.ts
@@ -35,6 +35,16 @@ export enum NabuCasaCommand {
 	SetConfig = 0x06,
 	GetLEDBinary = 0x07,
 	SetLEDBinary = 0x08,
+	GetBootloaderInfo = 0x09,
+}
+
+export interface NabuCasaBootloaderInfo {
+	/** Bootloader version, formatted as `major.minor.customer` */
+	version: string;
+	/** CRC-32 over the bootloader image, identifying the exact build */
+	crc32: number;
+	/** Capability bitmask, as defined by the Gecko Bootloader */
+	capabilities: number;
 }
 
 export interface RGB {
@@ -123,12 +133,29 @@ export class ControllerProprietary_NabuCasa
 	private driver: Driver;
 	private controller: ZWaveController;
 	private supportedCommands?: NabuCasaCommand[];
+	private _bootloaderInfo?: NabuCasaBootloaderInfo;
+
+	/** Information about the bootloader, read during the interview */
+	public get bootloaderInfo(): NabuCasaBootloaderInfo | undefined {
+		return this._bootloaderInfo;
+	}
 
 	public async interview(): Promise<void> {
 		const valueDB = this.controller.valueDB;
 
 		const supported = await this.getSupportedCommands();
 		this.supportedCommands = supported;
+
+		if (supported.includes(NabuCasaCommand.GetBootloaderInfo)) {
+			this._bootloaderInfo = await this.getBootloaderInfo();
+			if (this._bootloaderInfo) {
+				this.driver.controllerLog.print(
+					`Bootloader version: ${this._bootloaderInfo.version} (CRC-32 0x${
+						this._bootloaderInfo.crc32.toString(16).padStart(8, "0")
+					})`,
+				);
+			}
+		}
 
 		if (
 			supported.includes(NabuCasaCommand.GetLEDBinary)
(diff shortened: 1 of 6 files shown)


packages/zwave-js/src/lib/controller/proprietary/NabuCasa.ts
```diff
@@ -35,6 +35,14 @@ export enum NabuCasaCommand {
...
 	GetLEDBinary = 0x07,
 	SetLEDBinary = 0x08,
+	GetBootloaderInfo = 0x09,
+}
+
+export interface NabuCasaBootloaderInfo {
+	/** Bootloader version, formatted as `major.minor.customer` */
+	version: string;
+	/** Capability bitmask, as defined by the Gecko Bootloader */
+	capabilities: number;
```
AlCalzone: The capabilities should be defined as a bitflags enum, derived from the actual capabilities used in the SDK.

AlCalzone: Fixed in 13df37768 — added `NabuCasaBootloaderCapability` mirroring the `BOOTLOADER_CAPABILITY_*` defines from `btl_interface.h`, and `capabilities` is now a `NabuCasaBootloaderCapability[]` parsed with `parseBitMask`.

AlCalzone:
Waiting for the firmware PR to be merged.

PR #9097, opened 2026-08-15
feat(config): update Zooz ZEN32 config for firmware 3.30

benjaminflessner:
From https://www.support.getzooz.com/kb/article/706-zen32-scene-controller-change-log/:

**Firmware: 3.30 (released 08/2026 as OTA)**
- Updated LED brightness settings for Parameters 11–15, changing the supported value range from 0–2 to 1–10. 

Verified the 10.x firmware version (pre 800LR version) does not (yet) have this change


AlCalzone (approved):
Thanks!

PR #9101, opened 2026-08-17
fix(driver): name the transmit status when a send fails without NoAck

### LINKED ISSUE #7198: SendDataMulticast with status `Fail` exposed as `NoAck`
### Discussed in https://github.com/zwave-js/node-zwave-js/discussions/7194

<sup>Originally posted by **@Botched1** September 26, 2024</sup>
I'm using a ZST39 running 7.22.1 and see multicast commands fail somewhat regularly. If I repeatedly send multicast commands I can get it to give the error maybe 3 out of 10 times or so... 

[...]

What I see in the logs when it fails:

```
2024-09-25 20:30:38.430 ERROR Z-WAVE-SERVER: The node did not acknowledge the command (ZW0204)
ZWaveError: The node did not acknowledge the command (ZW0204)
    at Driver.sendMessage (/opt/node_modules/zwave-js/src/lib/driver/Driver.ts:5816:23)
    at Driver.sendCommandInternal (/opt/node_modules/zwave-js/src/lib/driver/Driver.ts:6014:28)
    at Driver.sendCommand (/opt/node_modules/zwave-js/src/lib/driver/Driver.ts:6124:29)
    at BinarySwitchCCAPI.set (/opt/node_modules/@zwave-js/cc/src/cc/BinarySwitchCC.ts:142:24)
    at BinarySwitchCCAPI.<anonymous> (/opt/node_modules/@zwave-js/cc/src/cc/BinarySwitchCC.ts:164:16)
    at VirtualNode.setValue (/opt/node_modules/zwave-js/src/lib/node/VirtualNode.ts:172:39)
    at MulticastGroupMessageHandler.handle (/opt/node_modules/@zwave-js/server/dist/lib/multicast_group/message_handler.js:19:50)
    at Client.receiveMessage (/opt/node_modules/@zwave-js/server/dist/lib/server.js:139:100)
    at WebSocket.<anonymous> (/opt/node_modules/@zwave-js/server/dist/lib/server.js:60:45)
    at WebSocket.emit (node:events:517:28)
```
[...]

</div>

Investigation of the driver logs indicates that this is a transmit status of `Fail`, not `NoAck`, so it should be exposed as such.

luantaraschi:
Fixes #7198.

A transmit status that is not `NoAck` is reported as `The node did not acknowledge the command (ZW0204)`, which is what the log in #7194 shows for a multicast that came back with status `Fail`.

## Where the wrong message comes from

`serialAPICommandErrorToZWaveError` already names the status and picks the code from it, for both unicast and multicast transmit reports. It never runs for SendData, though. In `queueSerialAPICommand`, the `"failure"` case returns the result instead of throwing when the reason is `callback NOK` and the message is SendData, with the comment that a NOK callback still carries information worth evaluating.

So the report travels back up to `sendMessage`, and the throw that actually fires is the `!prevResult.isOK()` one at the end of the send loop, which has the `NoAck` wording and `Controller_CallbackNOK` hardcoded regardless of `prevResult.transmitStatus`.

The jammed branch just above catches `Fail` only when `txReport.txTicks === 0`, so a `Fail` where the controller really did transmit falls straight through to that throw.

## What this changes

The message now names the status the same way `StateMachineShared.ts` does, and keeps the old wording when the status really is `NoAck`, so nothing changes for the common case.

## What this deliberately leaves alone

The error code stays `Controller_CallbackNOK` for every status. Two places key behaviour off it:

- `mayRetry` refuses to retry `SendDataMulticastRequest` and `SendDataMulticastBridgeRequest` when the code is `Controller_CallbackNOK`, because some nodes may already have reacted. Moving `Fail` to `Controller_MessageDropped` would start retrying multicasts that are not retried today.
- `isMissingNodeACK` matches on the same code for unicast SendData, and it is what routes a transaction into `handleMissingNodeACK`, so it decides whether a node gets treated as sleeping.

Both are behaviour decisions rather than message wording, so I left them out of this diff. Happy to follow whichever way you prefer if you want the code derived from the status too.

## Test

Added to `controllerJammed.test.ts`, next to the other TX status Fail cases and reusing the same mock controller shape. The callback reports `Fail` with `txTicks: 10` so the jam branch is skipped and the send loop reaches the throw.

Without the change the test fails with `The node did not acknowledge the command (ZW0204)`.

Run in a Linux container, since the suite does not start on my Windows checkout:

```
yarn test:ts run packages/zwave-js/src/lib/test/driver/
```

The whole folder is green with the change: 83 files, 180 passed and 1 skipped. `yarn build`, `yarn oxlint`, `yarn eslint` and `dprint check` are clean on both touched files.

### DIFF
--- a/packages/zwave-js/src/lib/driver/Driver.ts
+++ b/packages/zwave-js/src/lib/driver/Driver.ts
@@ -191,6 +191,7 @@ import {
 	buffer2hex,
 	cloneDeep,
 	createWrappingCounter,
+	getEnumMemberName,
 	getErrorMessage,
 	getenv,
 	isAbortError,
@@ -6949,8 +6950,18 @@ ${handlers.length} left`,
 						}
 
 						if (!prevResult.isOK()) {
+							// Only a TX status of NoAck means the node did not answer.
+							// For anything else, name the status instead of guessing.
 							throw new ZWaveError(
-								"The node did not acknowledge the command",
+								prevResult.transmitStatus
+										=== TransmitStatus.NoAck
+									? "The node did not acknowledge the command"
+									: `Failed to send the command (Status ${
+										getEnumMemberName(
+											TransmitStatus,
+											prevResult.transmitStatus,
+										)
+									})`,
 								ZWaveErrorCodes.Controller_CallbackNOK,
 								prevResult,
 								transaction.stack,
--- a/packages/zwave-js/src/lib/driver/Driver.ts
+++ b/packages/zwave-js/src/lib/driver/Driver.ts
@@ -6954,14 +6954,12 @@ ${handlers.length} left`,
 							// For anything else, name the status instead of guessing.
 							throw new ZWaveError(
 								prevResult.transmitStatus
-										=== TransmitStatus.NoAck
+									=== TransmitStatus.NoAck
 									? "The node did not acknowledge the command"
-									: `Failed to send the command (Status ${
-										getEnumMemberName(
+									: `Failed to send the command (Status ${getEnumMemberName(
 											TransmitStatus,
 											prevResult.transmitStatus,
-										)
-									})`,
+										)})`,
 								ZWaveErrorCodes.Controller_CallbackNOK,
 								prevResult,
 								transaction.stack,
--- a/packages/zwave-js/src/lib/test/driver/controllerJammed.test.ts
+++ b/packages/zwave-js/src/lib/test/driver/controllerJammed.test.ts
@@ -413,3 +413,84 @@ integrationTestMulti(
 		},
 	},
 );
+
+let shouldFailAfterTransmitting = false;
+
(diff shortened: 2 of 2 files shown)


luantaraschi:
@AlCalzone the automated review has a point on the dead-node path. I checked it with a probe (non-sleeping node, transmit report of Fail with txTicks 10) and the node ends up Dead on master exactly as it does with this PR, so the change fixes the wording and leaves that routing alone. Whether isMissingNodeACK should only fire on NoAck is the behaviour question I kept out of the diff on purpose (last section of the description); if you want it in here, say the word and I will add it with a test. The two test notes were cheap, so I pushed those: the assertion now matches "Status Fail" plus the error code, and the mock no longer hangs off a module-level flag. I left the shared message helper and the multicast wording out, since both go past what #7198 asks for.

PR #9104, opened 2026-08-18
fix: repair controller listening state on startup

AlCalzone:
Repairs 700+ controllers whose application node information reports `isListening=false` at startup.

- rewrites the controller NIF with the listening flag set
- uses controller-immediate priority when the controller is jammed or unresponsive
- performs one guarded soft reset to apply the corrected NIF
- handles both expected and unexpected `SerialAPIStarted` notifications
- preserves 16-bit node ID mode after an unexpected reset
- treats SDK 7.0 as soft-reset capable

Hardware validation on a ZWA-2 running firmware 1.2 observed `SerialAPIStarted.isListening` change from `false` to `true` after one repair reset. The controller then ACKed 10 of 10 reports.

Validation:
- 54 targeted tests passed
- `yarn build`
- `yarn lint`


packages/zwave-js/src/lib/controller/Controller.ts
```diff
@@ -1989,16 +1990,19 @@ export class ZWaveController
 
 	/**
 	 * Sets the NIF of the controller to the Gateway device type and to include the CCs supported by Z-Wave JS.
-	 * Warning: This only works when followed up by a hard-reset, so don't call this directly
+	 * The updated node information takes effect after a soft reset.
 	 * @internal
 	 */
-	public async setControllerNIF(): Promise<void> {
+	public async setControllerNIF(
+		priority: MessagePriority = MessagePriority.Controller,
```
AlCalzone: Lets set this priority on the command directly instead of here.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -971,12 +972,16 @@ export class Driver extends TypedEventTarget<DriverEventCallbacks>
 				// allow that too.
 				if (this.controller.status === ControllerStatus.Unresponsive) {
 					return t.message instanceof SoftResetRequest
-						|| t.message instanceof GetControllerVersionRequest;
+						|| t.message instanceof GetControllerVersionRequest
+						|| t.message
+							instanceof SetApplicationNodeInformationRequest;
```
AlCalzone: This makes no sense.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -971,12 +972,16 @@ export class Driver extends TypedEventTarget<DriverEventCallbacks>
...
 				}
 
 				// While the controller is jammed, only soft resetting is allowed
 				if (this.controller.status === ControllerStatus.Jammed) {
-					return t.message instanceof SoftResetRequest;
+					return t.message instanceof SoftResetRequest
+						|| t.message
+							instanceof SetApplicationNodeInformationRequest;
```
AlCalzone: dito

AlCalzone:
This is a mess

PR #9116, opened 2026-08-20
feat(config): update Inovelli VZW32-SN P111 to match public docs

The00Dustin:
Ref: https://help.inovelli.com/en/articles/11463885-red-series-mmwave-presence-dimmer-switch-parameters#h_be5f7ddf15
Options for 2 and 5 were both originally removed, but Inovelli felt that `[2] Reserved for future use (Default)` would be friendlier  than `[2] (Default)` with `[2] Custom` in the drop-down.

### DIFF
--- a/packages/config/config/devices/0x031e/vzw32-sn.json
+++ b/packages/config/config/devices/0x031e/vzw32-sn.json
@@ -801,28 +801,24 @@
 			"allowManualEntry": false,
 			"options": [
 				{
-					"label": "Restore the mmwave module factory configuration",
+					"label": "Factory reset mmWave module",
 					"value": 0
 				},
 				{
-					"label": "The interference area is automatically generated",
+					"label": "Generate and store interference area",
 					"value": 1
 				},
 				{
-					"label": "Obtains the interference region and detection region",
+					"label": "Reserved for future use",
 					"value": 2
 				},
 				{
-					"label": "Clears the interference area",
+					"label": "Clear stored interference area",
 					"value": 3
 				},
 				{
-					"label": "Resets the detection area",
+					"label": "Reset detection area to defaults",
 					"value": 4
-				},
-				{
-					"label": "Clears the stay area",
-					"value": 5
 				}
 			]
 		},


packages/config/config/devices/0x031e/vzw32-sn.json
```diff
@@ -801,28 +801,24 @@
...
 				{
-					"label": "The interference area is automatically generated",
+					"label": "Generate and store interference area",
 					"value": 1
 				},
 				{
-					"label": "Obtains the interference region and detection region",
+					"label": "Reserved for future use",
 					"value": 2
 				},
```
AlCalzone, line 814: We should comment this out, so nobody can set it while it is only reserved.

AlCalzone (changes requested):

packages/config/config/devices/0x031e/vzw32-sn.json
The00Dustin: I can do that, but would like to first re-iterate:
> Options for 2 and 5 were both originally removed, but Inovelli felt that [2] Reserved for future use (Default) would be friendlier than [2] (Default) with [2] Custom in the drop-down.

FWIW, setting these doesn't hurt anything, but let me know if you'd still like me to make this change.

ETA: Since it may not be contextually obvious, that comment was about ZUI behavior with the config removed vs changed.

AlCalzone: I'm not sure what we gain by documenting a value that does nothing. If we leave it out, users won't be able to select it from the list. Or what is the concern here?

The00Dustin: Because it is the default, users will be able to see that it is set until/unless they change it.  The ZUI UX isn't really designed to handle this because it is a weird edge case.  Because of this, the options are either to have the value and the drop down match because the value looks like this:
Screenshot 2026-08-28 at 11 17 40
While the drop down looks like this (see [2] even though [5] is highlighted):
Screenshot 2026-08-28 at 11 19 22
Or to have the drop down show [2] like [5] in the screenshot above while the value looks like this (except it also says `(Default)`:
Screenshot 2026-08-28 at 11 19 06
Because I wasn't sure which would be better, I asked Inovelli their preference and the current configuration reflects their preference.  I don't think they're going to complain if I defer to your preference, I just wanted to make sure you understood why the PR was initially this way and be sure you prefer the adjustment over their preference.

ETA:  Now that I think through it for the Nth time, I wonder if maybe ZUI should be adjusted to not include values in the drop down that aren't included in the config even if they are the actual configured value...  I could submit an issue there if you do still want me to change this and also think that would be a sensible adjustment to make it friendlier / more consistent.

AlCalzone: Ohh, now I get it. I think the PR makes sense.

AlCalzone (approved):

Issue #9121, opened 2026-08-24
Implement NVM-independent `Network Restore Command`


AlCalzone:
This should make NVM migrations much simpler

Issue #108, opened 2019-05-29
Add payload validation to all deserialization routines


AlCalzone:
This avoids crashing when nodes send invalid CCs

PR #9122, opened 2026-08-24
feat(serial): add Network Restore messages

AlCalzone:
Part 1/3 of #9121.

Adds the Serial API message model for the NVM-independent Network Restore command:

- uses function ID `0x2f` for requests, immediate responses, and completion callbacks
- implements Prepare, Home ID, Device, Neighbors, Routes, and Finalize subcommands
- supports short and long node ID encoding
- models immediate acceptance and same-function completion statuses
- follows the existing reflection-based subcommand dispatch pattern

Validation:

- `yarn fmt`
- focused oxlint
- Serial API TypeScript compilation
- `yarn build`


packages/serial/src/serialapi/nvm/NetworkRestoreMessages.test.ts
AlCalzone, line 1: Drop the tests

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
```diff
@@ -0,0 +1,734 @@
...
+	RestoreNeighbors = 0x03,
+	RestoreRoutes = 0x04,
+	Finalize = 0xff,
+}
+
+export enum NetworkRestoreStatus {
+	OK = 0x00,
+	Error = 0x01,
+	Error_LongRangeNotSupported = 0x02,
+	Error_SubcommandNotSupported = 0x04,
```
AlCalzone, line 48: I wonder why this is 0x04 and not 0x03. Read the original PR that wrote the specs for this command and check whether this was discussed.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
```diff
@@ -0,0 +1,734 @@
...
+		? ZWaveDataRate["40k"]
+		: ZWaveDataRate["9k6"];
+
+	return {
+		type: payload[0],
+		beam,
+		speed,
+		hops,
+	};
+}
```
AlCalzone, line 114: Do these in any way overlap with the existing routines in `nvmedit`?
If yes, I wonder if it makes sense to reuse those - or if intentionally keeping those separate makes sense given the NVM-independentness of this command.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
```diff
@@ -0,0 +1,734 @@
...
+		callbackId == undefined
+		|| !Number.isInteger(callbackId)
+		|| callbackId < 1
+		|| callbackId > 0xff
+	) {
+		throwInvalidArgument(
+			"The callback ID must be an 8-bit unsigned integer greater than zero",
+		);
+	}
+}
```
AlCalzone: I think we're going a bit hard on the module level helper functions

This looks like stuff we don't validate anywhere else.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
```diff
@@ -0,0 +1,734 @@
...
+		this.command = options.command;
+	}
+
+	public static from(
+		raw: MessageRaw,
+		ctx: MessageParsingContext,
+	): NetworkRestoreRequest {
+		validatePayload(raw.payload.length >= 2);
+		const command = raw.payload[0];
+		validatePayload(isNetworkRestoreCommand(command));
```
AlCalzone: We don't validatePayload in other Serial API commands. Drop those checks. We'll add them later.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
```diff
@@ -0,0 +1,734 @@
...
+				return NetworkRestoreHomeIDRequest.from(commandPayload, ctx);
+			case NetworkRestoreCommand.RestoreDevice:
+				return NetworkRestoreDeviceRequest.from(commandPayload, ctx);
+			case NetworkRestoreCommand.RestoreNeighbors:
+				return NetworkRestoreNeighborsRequest.from(commandPayload, ctx);
+			case NetworkRestoreCommand.RestoreRoutes:
+				return NetworkRestoreRoutesRequest.from(commandPayload, ctx);
+			case NetworkRestoreCommand.Finalize:
+				return NetworkRestoreFinalizeRequest.from(commandPayload, ctx);
+		}
```
AlCalzone: Is this the same pattern we use in other Serial API commands with subcommands?

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
AlCalzone, line 1: Go easy on the validations

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
AlCalzone: Fixed in 73e72e6e — removed the new validatePayload checks.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
AlCalzone: Fixed in 73e72e6e — subcommands now use the same reflection-decorator pattern as FirmwareUpdateNVM.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
AlCalzone: I kept these separate. The nvmedit format is five bytes with hops first and configuration last; this wire format is six bytes with route type and configuration first. Reusing it would also introduce the wrong dependency direction.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.test.ts
AlCalzone: Fixed in 73e72e6e — removed the test file.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
AlCalzone: Fixed in 73e72e6e — removed the custom validation helpers and now use the existing callback assertion.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
AlCalzone: Fixed in 73e72e6e — removed the custom constructor and parser validation.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
AlCalzone: I checked CSWG PR #108. The value was introduced as 0x04 without discussion, and it matches Extended NVM Operations, where 0x03 is operation interference and 0x04 is unsupported subcommand. I kept 0x04.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
```diff
@@ -0,0 +1,558 @@
...
+
+	public readonly homeId: number;
+	public readonly controllerNodeId: number;
+
+	public serialize(ctx: MessageEncodingContext): Promise<Bytes> {
+		this.payload = new Bytes(5);
+		this.payload.writeUInt32BE(this.homeId, 0);
+		this.payload[4] = this.controllerNodeId;
+		return super.serialize(ctx);
+	}
```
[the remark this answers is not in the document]
AlCalzone: double check this claim against the specs for this command

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
```diff
@@ -0,0 +1,558 @@
...
+
+@subCommandRequest(NetworkRestoreCommand.RestoreNeighbors)
+export class NetworkRestoreNeighborsRequest extends NetworkRestoreRequest {
+	public constructor(
+		options: NetworkRestoreNeighborsRequestOptions & MessageBaseOptions,
+	) {
+		super(options);
+
+		this.nodeId = options.nodeId;
+		this.neighbors = Bytes.from(options.neighbors);
```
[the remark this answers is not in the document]
AlCalzone: The mask is fixed at 29 bytes, but I am leaving constructor validation out per the earlier human review request to remove these checks.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
```diff
@@ -0,0 +1,558 @@
...
+
+@subCommandRequest(NetworkRestoreCommand.RestoreDevice)
+export class NetworkRestoreDeviceRequest extends NetworkRestoreRequest {
+	public constructor(
+		options: NetworkRestoreDeviceRequestOptions & MessageBaseOptions,
+	) {
+		super(options);
+
+		this.nodeId = options.nodeId;
+		this.protocolData = Bytes.from(options.protocolData);
```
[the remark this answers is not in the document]
AlCalzone: The field is five bytes, but I am leaving constructor validation out per the earlier human review request to remove these checks.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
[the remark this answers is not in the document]
AlCalzone: The command spec explicitly says the controller node ID must always use one-byte notation regardless of the configured node-ID base type. The fixed offset is intentional, so no change is needed.

packages/serial/src/serialapi/nvm/NetworkRestoreMessages.ts
```diff
@@ -0,0 +1,558 @@
...
+		: route.beam === "250ms"
+		? 0x01
+		: 0x00;
+	const speed = route.speed === ZWaveDataRate["100k"]
+		? 0x02
+		: route.speed === ZWaveDataRate["40k"]
+		? 0x01
+		: 0x00;
+	ret[1] = (beam << 6) | speed;
+	ret.set(route.hops, 2);
```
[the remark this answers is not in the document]
AlCalzone: The concern is valid for arbitrary callers, but this validation was deliberately removed following the earlier human review. I am keeping that decision.

PR #5037, opened 2022-09-09
fix: merge notification metadata for variables with multiple states

AlCalzone:
fixes: #5027

Issue #5027, opened 2022-09-07
Creating metadata for multiple notification events of the same notification type overrides the old metadata instead of extending


AlCalzone:
For example, this should have states for 22 and 23:
```
{"k":"{\"nodeId\":7,\"commandClass\":113,\"endpoint\":0,\"property\":\"Access Control\",\"propertyKey\":\"Door state\"}","v":{"type":"number","readable":true,"writeable":false,"min":0,"max":255,"label":"Door state","states":{"23":"Window/door is closed"},"ccSpecific":{"notificationType":6}}}
```

jdiegmueller:
I think I am running in to this (or a variant of it) using node-zwave-js v10.0.4 with zwavejs2mqtt v7.1.0 and Homeassistant 2022.7.7.

I noticed one of my Aeotec Door Sensor 7 Pros (ZWA012) suddenly grew a motion sensor entity in Home Assistant and was no longer reporting contact open/closed for the door.  I re-interviewed it but but when it coms back it seems the entity is now "sensor.outside_garage_entry_door_sensor_access_control_door_state" as opposed to "binary_sensor..outside_garage_entry_door_sensor_access_control_window_door_is_open".  Furthermore, Homeassistant is seeing some weird states in the new Homeassisant sensor:
* When closed: "Window/door is closed"
* When open: "22"

Attaching zwavejs log covering a new re-interview followed by a couple of open/close events.  Node is 185.

[node185.log](https://github.com/zwave-js/node-zwave-js/files/9526865/node185.log)

Verbalinsurection:
Hi,
I have the similar issue. If this can help, here my zwavejs log, zwavetomqtt log and zwave dump when I add the device and use
https://gist.github.com/Verbalinsurection/c07007f3f189bf22a1d3045e60462d6b

I have posted full decription [here](https://github.com/zwave-js/node-zwave-js/discussions/5033) before finding this issue with help of kpine.

PR #9133, opened 2026-08-25
Extend notification value metadata for V3+ devices on demand

derekcentrico:
## Problem

V3+ devices can report notification events they didn't declare as supported during interview. The interview sends `NotificationCCEventSupportedGet` and builds `metadata.states` from the device's response, so only declared events get labels. When the device later sends an undeclared event, the value has no entry in `metadata.states`. Downstream consumers like Home Assistant treat this as an enum sensor with a value outside its options list and crash with a `ValueError`.

Affected devices include:
- Yale Assure 2 lock (Access Control door state 22, "Window/door is open", not declared during interview)
- Honeywell T6 Pro thermostat (mains_status value 3)
- Hoppe ConnectSense window handles (door handle state)

Related: #5027 (fixed metadata overwrite during interview for variables with multiple states in PR #5037). This addresses the remaining case where the event was never declared during interview at all.

Downstream HA core issues: [home-assistant/core#94841](https://github.com/home-assistant/core/issues/94841), [home-assistant/core#108178](https://github.com/home-assistant/core/issues/108178), [home-assistant/core#174972](https://github.com/home-assistant/core/issues/174972).

## Fix

`extendNotificationValueMetadata` already builds metadata on demand for V2 devices (which can't be interviewed for supported events). The condition `ccVersion === 2 || !node.valueDB.hasMetadata(valueId)` blocked this for V3+ devices that already had interview-set metadata.

This replaces the version-based check with a check for whether the specific event value is already in the existing `states` map. If the value is missing, `getNotificationValueMetadata` extends the metadata with the label from `Notifications.ts`. The function is already idempotent, and V2 behavior is unchanged (metadata is always missing or incomplete on first arrival).

1 file changed, 9 insertions, 12 deletions.

### DIFF
--- a/packages/zwave-js/src/lib/node/CCHandlers/NotificationCC.ts
+++ b/packages/zwave-js/src/lib/node/CCHandlers/NotificationCC.ts
@@ -631,25 +631,22 @@ function clearNotificationIdleReset(
 	}
 }
 
-// Fallback for V2 notifications that don't allow us to predefine the metadata during the interview.
-// Instead of defining useless values for each possible notification event, we build the metadata on demand
+// Extend notification value metadata when the received event isn't already
+// in the states map. V2 devices never get interview-set metadata, and V3+
+// devices may report events they didn't declare during interview.
 function extendNotificationValueMetadata(
-	ctx: GetSupportedCCVersion,
+	_ctx: GetSupportedCCVersion,
 	node: NodeId & NodeValues,
 	valueId: ValueID,
 	notification: Notification,
 	valueConfig: NotificationState,
 ) {
-	const ccVersion = ctx.getSupportedCCVersion(
-		CommandClasses.Notification,
-		node.id,
-		node.index,
-	);
-	if (ccVersion === 2 || !node.valueDB.hasMetadata(valueId)) {
+	const existing = node.valueDB.getMetadata(valueId) as
+		| ValueMetadataNumeric
+		| undefined;
+	if (!existing || !(valueConfig.value in (existing.states ?? {}))) {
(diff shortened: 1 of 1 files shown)


AlCalzone:
> The condition `ccVersion === 2 || !node.valueDB.hasMetadata(valueId)` blocked this for V3+ devices that already had interview-set metadata.

The reason for this is that CC version 2 and below does not support querying the supported events at all, so we need to rely on the extend-metadata behavior.

For v3+, I'd rather fix the capabilities on a per-device basis. Otherwise, we risk introducing garbage states for non-encrypted devices: https://zwave-js.github.io/zwave-js/#/troubleshooting/nonsensical-values

Do you happen to have interview logs for the Yale and Honeywell devices that show the issue?
I can take a look at the Hoppe device - this should work fine. I tested very extensively with it.

derekcentrico:
@AlCalzone You were right. I re-interviewed both Yale Assure 2 locks and the device correctly declares event 22 as supported:

```
received supported notification events for Access Control: 1, 2, 3, 4, 5, 6, 9, 11, 12, 13, 14, 15, 16, 18, 22, 23
```

After the re-interview, the metadata includes both states 22 and 23 and the errors are gone. The original interview from mid-July must have had a corrupted `NotificationCCEventSupportedReport` that dropped event 22 from the supported set. The logs from the re-interview even show `Dropping message with invalid payload` during one of the radio exchanges, which makes that theory plausible.

I should have checked the interview logs before assuming this was a code-level issue. Apologies for the noise. Happy to close this PR since the root cause is radio interference during interview, not a gap in how V3+ metadata is built.

PR #1641, opened 2021-02-07
feat(cc): map Notification CC (un)lock operations to (Door) Lock CC

AlCalzone:
With this PR, we handle manual (un)lock operations reported through Notification CC and set the current lock status for (Door) Lock CC accordingly.

fixes: #1602

PR #9135, opened 2026-08-26
fix(cc): ignore (un)lock notifications superseded by a newer Door Lock report

### LINKED ISSUE #9134: Notification CC (un)lock mapping overwrites a newer Door Lock Operation Report
### Summary

`handleKnownNotification` maps Access Control (un)lock notifications onto `DoorLockCCValues.currentMode` unconditionally. On a lock that reports state *both* ways, a device-generated duplicate notification arriving a few seconds late overwrites a **newer and correct** Door Lock Operation Report. The lock entity then shows the wrong state indefinitely, until something re-polls it.

Observed on a Schlage BE469ZP: three occurrences in four days, each leaving Home Assistant showing `unlocked` for 20–35 minutes while the deadbolt was thrown.

### Environment

- zwave-js 15.25.0 (via zwave-js-ui 11.21.1)
- Schlage BE469ZP, `0x003b:0x0001:0x0469`, firmware 3.3, S2 Access Control
- Home Assistant 2026.7.1

### What happens

This device re-sends some of its unsolicited reports roughly 4–5 seconds later. Measured over 7 days of driver logs: **21% of unsolicited messages** have an exact duplicate (same event, same alarm level, same user slot) within 12 s, and the delay is tightly clustered at 4.2–4.5 s — a firmware timer, not chance.

If a real state change happens inside that window, the duplicate of the *older* notification arrives after the *newer* Door Lock Operation Report and wins.

### Trace

Sequence numbers are the S2 sequence numbers; they ascend, so nothing is delivered out of order. Only the *content* is stale.

```
16:12:55.287  #173  DoorLockCCOperationReport   current mode: Unsecured
16:12:55.540  #174  NotificationCCReport        Manual unlock operation
                    -> user unlocks the door

16:13:00.047  #176  DoorLockCCOperationReport   current mode: Secured
16:13:00.266  #177  NotificationCCReport        Keypad lock operation
                    -> user throws the deadbolt. currentMode = 255. Correct.

16:13:00.495  #178  NotificationCCReport        Manual unlock operation
                    -> duplicate of #174, 4.955 s later
                    -> currentMode: 255 => 0.  WRONG.
```

The next Door Lock Operation Report is at 16:48:01 (`Secured`) — 35 minutes later, triggered by the user pressing Lock in the UI. The deadbolt was locked the entire time; `#178` describes an unlock that had already been superseded 0.45 s earlier.

Note there is no decryption failure or SPAN resync anywhere in this trace. Those occur frequently on this device but are incidental — failed frames are re-sent and recovered (99%, median 239 ms), and the fault occurs with or without them.

### Why

`@zwave-js/cc` → `handleKnownNotification`:

```js
const isLocked = lockEvents.has(command.notificationEvent);
if (node.supportsCC(CommandClasses["Door Lock"])) {
    node.valueDB.setValue(
        DoorLockCCValues.currentMode.endpoint(command.endpointIndex),
        isLocked ? DoorLockMode.Secured : DoorLockMode.Unsecured,
    );
}
```

No check on whether `currentMode` was more recently set by a Door Lock Operation Report, and no duplicate suppression. Introduced in #1641 (2021) to support locks that report manual/keypad operations *only* via Notification CC — a real need, but the write is unconditional for every device.

### Impact

A lock entity that silently reports the opposite of physical state, for as long as nothing re-polls it. On an entity representing a door lock this is a safety-relevant staleness: automations and dashboards see `unlocked` on a locked door, and the reverse case (a stale `lock` notification masking a real unlock) is equally reachable by the same mechanism.

### Proposed fix

**Preferred** — Before applying a notification to `currentMode`, check whether Door Lock CC set that value within the last few seconds. If it did and the notification contradicts it, skip the write (optionally scheduling a verification poll).

This is a no-op for the devices #1641 targets: they never emit a Door Lock Operation Report, so the marker is never set and their behaviour is unchanged. It engages only on devices that report both ways — exactly where the mapping can do harm.

`ValueDB` already exposes `getTimestamp(valueId)`, but both sources write the same value, so this needs Door Lock CC to record the time of its own report separately.

**Smaller alternative — suppress duplicates.** Ignore a notification that exactly matches the previous one (same event, alarm level and user slot) within ~6 s. No cross-CC state required. Two identical keypad unlocks from the same user slot within six seconds are not physically reachable.

A device-config compat flag to opt out of the mapping entirely would also solve it for this device, though it leaves the general case open.

### Caveats

- I have not established *what* selects which reports this device re-sends — only that 21% are, at a consistent 4–5 s delay.
- A strict model of this mechanism, replayed against a week of raw `currentMode` writes, predicts a *similar but not identical* set of incidents to those observed. The mechanism and the trace above are solid; an exhaustive account of every occurrence is not.
- The frequent S2 decode failures on this device are almost certainly a separate matter and are not part of this report.

### Driver log

Sanitized driver log excerpt (three windows: both incidents, plus the recovery 35 minutes later): https://gist.github.com/joyjit/7301b03afe9a3005017bf5d347bf8784

Redacted: network home ID, user code slot numbers, S2 entropy values. Nothing else was altered.

joyjit:
Fixes #9134.

### The problem

Some locks repeat their Notification CC reports several seconds after the original. If a real state change happens inside that window, the repeat arrives *after* the device has already reported the newer state through Door Lock CC — and `handleKnownNotification` applies it unconditionally, so the stale operation wins. The lock entity then shows the opposite of physical state until something polls it again.

From the reporting device (Schlage BE469ZP), sequence numbers are the S2 sequence numbers, so nothing is delivered out of order — only the content is stale:

```
16:12:55.540  #174  NotificationCCReport       Manual unlock operation
16:13:00.047  #176  DoorLockCCOperationReport  current mode: Secured   <- real re-lock
16:13:00.495  #178  NotificationCCReport       Manual unlock operation <- repeat of #174
                                               -> currentMode 255 => 0
```

`currentMode` stayed `Unsecured` for 35 minutes with the deadbolt thrown. Measured over 7 days on this device: ~21% of unsolicited reports are repeated, clustered tightly at 4.2–4.5 s.

### The change

Record when the device last reported its mode through Door Lock CC, and skip an (un)lock notification that contradicts a recent report.

The mapping from #1641 exists for locks that report manual/keypad operations *only* through Notification CC. Those devices never emit a `DoorLockCCOperationReport`, so the new timestamp is never set and the guard never engages — their behaviour is unchanged. It only takes effect on devices that report both ways, which is exactly where the mapping can do harm.

### Trade-off worth reviewing

`OUTDATED_LOCK_NOTIFICATION_TIMEOUT` is 10 s. Too short and repeats still get through; too long and a device that reports *some* transitions only via Notification CC could have a legitimate change suppressed. I picked 10 s to cover the observed retransmission interval with margin, but you know the device landscape far better than I do — happy to change it, or to switch to suppressing exact duplicates instead (same event, alarm level and user slot within a few seconds), which was the other option in the issue.

### Testing

- New integration test `mapNotificationDoorLockOutdated.test.ts`, following the existing `mapNotificationDoorLock.test.ts`. Verified it **fails without the source change and passes with it**.
- The existing `mapNotificationDoorLock.test.ts` still passes.
- `packages/zwave-js/src/lib/test/cc-specific` — 23 files / 46 tests pass.
- `yarn build`, `yarn fmt` and `eslint` clean.
- One unrelated flake in `driver/partialCCSessions.test.ts` under full parallel load; it passes in isolation both with and without this change.

I also have the affected hardware, so I can test a candidate build against the real lock if that would help.

### DIFF
--- a/packages/cc/src/cc/DoorLockCC.ts
+++ b/packages/cc/src/cc/DoorLockCC.ts
@@ -81,6 +81,12 @@ export const DoorLockCCValues = V.defineCCValues(CommandClasses["Door Lock"], {
 			states: enumValuesToMetadataStates(DoorLockMode),
 		},
 	),
+	// Some locks repeat their Notification CC reports several seconds after the
+	// original. Remembering when the device last reported its mode through Door Lock CC
+	// lets us discard a repeated notification that has since been superseded.
+	...V.staticProperty("currentModeReportedAt", undefined, {
+		internal: true,
+	}),
 	...V.staticProperty(
 		"duration",
 		{
@@ -1150,6 +1156,12 @@ export class DoorLockCCOperationReport extends DoorLockCC {
 
 	public persistValues(ctx: PersistValuesContext): boolean {
 		if (!super.persistValues(ctx)) return false;
+
+		this.setValue(
+			ctx,
+			DoorLockCCValues.currentModeReportedAt,
+			Date.now(),
+		);
 
 		// Only store the door/bolt/latch status if the lock supports it
 		const supportsDoorStatus = !!this.getValue(
--- a/packages/cc/src/cc/_CCValues.generated.ts
+++ b/packages/cc/src/cc/_CCValues.generated.ts
@@ -2128,6 +2128,33 @@ export const DoorLockCCValues = Object.freeze({
 			autoCreate: true,
 		} as const satisfies CCValueOptions,
 	},
+	currentModeReportedAt: {
+		id: {
+			commandClass: CommandClasses["Door Lock"],
+			property: "currentModeReportedAt",
+		} as const,
+		endpoint: (endpoint: number = 0) => ({
+			commandClass: CommandClasses["Door Lock"],
+			endpoint,
+			property: "currentModeReportedAt",
+		} as const),
+		is: (valueId: ValueID): boolean => {
+			return valueId.commandClass === CommandClasses["Door Lock"]
+				&& valueId.property === "currentModeReportedAt"
+				&& valueId.propertyKey == undefined;
+		},
+		get meta() {
+			return ValueMetadata.Any;
+		},
+		options: {
+			internal: true,
+			minVersion: 1,
+			secret: false,
+			stateful: true,
+			supportsEndpoints: true,
+			autoCreate: true,
+		} as const satisfies CCValueOptions,
+	},
 	duration: {
 		id: {
 			commandClass: CommandClasses["Door Lock"],
--- a/packages/zwave-js/src/lib/node/CCHandlers/NotificationCC.ts
+++ b/packages/zwave-js/src/lib/node/CCHandlers/NotificationCC.ts
@@ -467,7 +467,7 @@ function isOutdatedLockOperation(
 }
 
 function handleKnownNotification(
-	ctx: GetValueDB,
+	ctx: GetValueDB & LogNode,
 	node: ZWaveNode,
 	command: NotificationCCReport,
 ): void {
@@ -504,6 +504,13 @@ function handleKnownNotification(
 		// a duplicate of an operation that has already been superseded. Applying it
 		// would leave the lock in the wrong state until something polls it again.
 		if (isOutdatedLockOperation(node, command.endpointIndex, isLocked)) {
+			ctx.logNode(node.id, {
+				message: `Ignoring ${
+					isLocked ? "lock" : "unlock"
+				} notification because Door Lock CC reported a different mode more recently`,
+				direction: "inbound",
+				level: "verbose",
+			});
 			return;
 		}
 
--- a/packages/zwave-js/src/lib/node/CCHandlers/NotificationCC.ts
+++ b/packages/zwave-js/src/lib/node/CCHandlers/NotificationCC.ts
@@ -429,6 +429,43 @@ export function handleNotificationReport(
 	}
 }
 
+/**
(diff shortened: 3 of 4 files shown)


AlCalzone (changes requested):
I general I prefer trusting that devices report their state correctly. Otherwise we get into a maintenance hell where some of these workarounds are correct for some devices, but not others.

Would there be any downside to simply ignoring all manually locked/unlocked notifications from this particular device? Are there any situations where we'd be missing information?

joyjit:
Nothing lost on state: every lock/unlock notification from this lock is accompanied by a Door Lock Operation Report with the matching mode, almost always arriving first.

One (minor) downside: keypad unlocks carry the code slot in `eventParameters.userId`, which Door Lock CC has no equivalent for — HA/keymaster use it to report who opened the door. So the flag should skip the `currentMode` write in `handleKnownNotification`, which runs before the `notification` event is emitted, leaving that event intact.

Happy to rewrite it that way: compat flag in `be469zp.json` plus a one-line gate, dropping the timeout heuristic. Name preference — `mapNotificationLockOperations: false`, or a `disable…` form?

AlCalzone:
I was only referring to the "Manual lock/unlock operation" notifications, which seem to cause the issue. Keypad unlock should be fine I think?

In that case, I think this is best resolved by extending the https://zwave-js.github.io/zwave-js/#/config-files/file-format?id=remapnotifications compat flag with a new `ignore` option (instead of to/clear/idle).
That can then be added to the BE469ZP config file, without touching handler code.

Issue #1482, opened 2021-01-23
Avoid queueing redundant commands (like GET requests, NonceReports, etc.)


AlCalzone:

LordMike:
Thought: One way of doing this would be to let each command implement some `GetHashcode` or similar, which would return a unique string identifying the contents. If two commands have identical hashes, they will yield the same response from a node.

The queue can then, when it decides to, compute the hash of all messages for some node, and only keep the last or the first message for each unique hash.

Example: Two `GET`'s, for different values, could produce two unique outputs for GetHashcode. The queue will then know that they should not be deduplicated.

Example: Two `GET`'s for the same value, or two NOOP's, will all return the same Hashcode, and therefore the queue can deduplicate them.

Example: Two SET's for the same value, but with different payloads could yield the same Hash, and thereby let the last message win.. This would help in cases where a user rapidly sets many different values f.ex. for a multilevel switch (dragging some slider for brightness).. In these cases, only the last message is really important, and since they all supersede each other ..

It may get tricky with multilevel up/downs, or other messages that aren't "Set to N value" as these could come out of sequence.... 

E.g., user submits: `Multilevel up`, `Multilevel down`, `Multilevel up`, `Multilevel down`. If the queue is emptied after the first message, it may deduplicate the two `Multilevel down`'s, but keep the Up. The end result is that the Multilevel up is the last message sent... 

In theory, maybe these messages should have the same Hashcode, and then keep the last inserted message. Since they supersede each other.

PR #9136, opened 2026-08-27
feat: add framework to deduplicate transactions

AlCalzone:
Implements the framework for command-aware transaction deduplication.

Commands may now be considered redundant with or superseding another command. Superseded commands are dropped with an appropriate error. Redundant commands are attached to the existing transaction and do not cause a separate transmission.

Part of #1482.


packages/cc/src/lib/CommandClass.ts
```diff
@@ -396,6 +422,22 @@ export class CommandClass implements CCId {
 		// Do nothing by default
 	}
 
+	public getRelationTo(other: CommandClass): CommandRelation {
```
AlCalzone, line 450: This should have a JSDoc comment explaining its purpose

packages/cc/src/lib/CommandClass.ts
```diff
@@ -396,6 +422,22 @@ export class CommandClass implements CCId {
...
+			|| !haveSameNodeTarget(this.nodeId, other.nodeId)
+			|| this.endpointIndex !== other.endpointIndex
+			|| this.encapsulationFlags !== other.encapsulationFlags
+		) {
+			return CommandRelation.Unrelated;
+		}
+		return this.determineRelation(other);
+	}
+
+	protected determineRelation(_other: CommandClass): CommandRelation {
```
AlCalzone, line 465: This should have a JSDoc comment explaining its purpose. Specifically that this one is meant to be overridden by subclasses.

packages/cc/src/lib/CommandClass.ts
```diff
@@ -88,6 +88,32 @@ export interface CommandClassOptions extends CCAddress {
 	payload?: BytesView;
 }
 
+export enum CommandRelation {
+	Unrelated,
+	Redundant,
+	Supersedes,
+}
+
+function haveSameNodeTarget(
```
AlCalzone: Add a brief comment explaining when a target is considered equal

packages/cc/src/lib/CommandClass.ts
```diff
@@ -88,6 +88,32 @@ export interface CommandClassOptions extends CCAddress {
...
+	first: number | MulticastDestination,
+	second: number | MulticastDestination,
+): boolean {
+	if (typeof first === "number" || typeof second === "number") {
+		return first === second;
+	}
+	const firstTargets = new Set(first);
+	const secondTargets = new Set(second);
+	return firstTargets.size === secondTargets.size
+		&& [...firstTargets].every((nodeId) => secondTargets.has(nodeId));
```
AlCalzone: Aren't there new Set methods that we can use to check for 100% overlap?

packages/core/src/error/ZWaveError.ts
```diff
@@ -71,6 +71,9 @@ export enum ZWaveErrorCodes {
 	/** Tried to perform an action for a Long Range node that does not make sense for ZWLR */
 	Controller_NotSupportedForLongRange,
 
+	/** A newer command superseded the queued message */
```
AlCalzone:
```suggestion
	/** A newer command superseded this one */
```

packages/cc/src/lib/CommandClass.ts
AlCalzone: Fixed in 3ca05094f768deaa8115e38fcc15765a552cd5eb — added JSDoc for the shared relation checks and dispatch.

packages/cc/src/lib/CommandClass.ts
AlCalzone: Fixed in 3ca05094f768deaa8115e38fcc15765a552cd5eb — documented this as the subclass override hook for command-specific relations.

packages/cc/src/lib/CommandClass.ts
AlCalzone: Fixed in 3ca05094f768deaa8115e38fcc15765a552cd5eb — documented exact singlecast equality and order-independent multicast set equality.

packages/cc/src/lib/CommandClass.ts
AlCalzone: Kept the current size and membership check because Set.prototype.isSubsetOf is unavailable on the supported Node 20 runtime.

packages/core/src/error/ZWaveError.ts
AlCalzone: Fixed in 3ca05094f768deaa8115e38fcc15765a552cd5eb — applied the suggested wording.

packages/cc/src/cc/CRC16CC.ts
```diff
@@ -146,6 +150,15 @@ export class CRC16CCCommandEncapsulation extends CRC16CC {
 
 	public encapsulated: CommandClass;
 
+	protected override determineRelation(
+		other: CommandClass,
+	): CommandRelation {
+		if (!(other instanceof CRC16CCCommandEncapsulation)) {
+			return CommandRelation.Unrelated;
+		}
+		return this.encapsulated.getRelationTo(other.encapsulated);
```
AlCalzone: Lets turn these checks around - similar for the other CCs
```suggestion
		if (other instanceof CRC16CCCommandEncapsulation) {
			return this.encapsulated.getRelationTo(other.encapsulated);
		}
		return CommandRelation.Unrelated;
```

packages/cc/src/cc/MultiChannelCC.ts
```diff
@@ -1492,6 +1505,18 @@ export class MultiChannelCCCommandEncapsulation extends MultiChannelCC {
...
+	protected override determineRelation(
+		other: CommandClass,
+	): CommandRelation {
+		if (
+			!(other instanceof MultiChannelCCCommandEncapsulation)
+			|| !haveSameDestination(this.destination, other.destination)
+		) {
+			return CommandRelation.Unrelated;
+		}
+		return this.encapsulated.getRelationTo(other.encapsulated);
```
AlCalzone: dito

packages/cc/src/cc/MultiChannelCC.ts
```diff
@@ -1720,6 +1745,15 @@ export class MultiChannelCCV1CommandEncapsulation extends MultiChannelCC {
 
 	public encapsulated!: CommandClass;
 
+	protected override determineRelation(
+		other: CommandClass,
+	): CommandRelation {
+		if (!(other instanceof MultiChannelCCV1CommandEncapsulation)) {
+			return CommandRelation.Unrelated;
+		}
+		return this.encapsulated.getRelationTo(other.encapsulated);
```
AlCalzone: dito

packages/cc/src/lib/CommandClass.getCommandRelation.test.ts
AlCalzone, line 1: Lets rename this file to `CommandClass.getCommandRelation.test.ts`, since it is specific to the new functionality.

packages/cc/src/lib/CommandClass.getCommandRelation.test.ts
```diff
@@ -0,0 +1,317 @@
...
+	}
+}
+
+class CrossCommandBasicCCSet extends BasicCCSet {
+	protected override determineRelation(other: BasicCCGet): CommandRelation {
+		return other instanceof BasicCCGet
+			? CommandRelation.Supersedes
+			: CommandRelation.Unrelated;
+	}
+}
```
AlCalzone, line 27: These could both use a brief comment. Why are we reusing BasicCCSet and why are we testing cross command opt-in?

packages/cc/src/lib/CommandClass.test.ts
```diff
@@ -0,0 +1,317 @@
...
+		expect(getCommandRelation(newer, older)).toBe(
+			CommandRelation.Redundant,
+		);
+	});
+
+	test("recurses through transparent nested wrappers", () => {
+		const newer = CRC16CC.encapsulate(
+			SecurityCC.encapsulate(
+				1,
+				{} as Parameters<typeof SecurityCC.encapsulate>[1],
```
AlCalzone: > `Parameters<typeof SecurityCC.encapsulate>[1]`

No. Name the type.

packages/cc/src/lib/CommandClass.getCommandRelation.test.ts
```diff
@@ -0,0 +1,317 @@
...
+	command.encapsulationFlags = options.encapsulationFlags
+		?? EncapsulationFlags.None;
+	return command;
+}
+
+describe("getCommandRelation", () => {
+	test("uses the newer command's relation override", () => {
+		expect(getCommandRelation(createSet(1), createSet(1))).toBe(
+			CommandRelation.Redundant,
+		);
```
AlCalzone, line 51: Each of these expectations (or if it makes more sense the whole group) should have a 1-line comment explaining why this is expected.

packages/zwave-js/src/lib/test/driver/commandDeduplication.test.ts
AlCalzone, line 1: This whole test needs some "bookmark"-style comments narrating what is being tested and what the expectation is.

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -179,6 +331,9 @@ export class Transaction implements Comparable<Transaction> {
 	/** If a Wake Up On Demand should be requested for the target node. */
 	public requestWakeUpOnDemand: boolean = false;
 
+	/** Whether follow-up Supervision status updates are requested. */
+	public requestStatusUpdates: boolean = false;
```
AlCalzone, line 419: Why is this necessary?

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -41,9 +43,54 @@ export interface TransactionOptions {
...
+	state: TransactionAttachmentState;
+	detached: boolean;
+	lastProgressState?: TransactionState;
+}
+
+type TransactionResult =
+	| { status: "fulfilled"; value: Message | void }
+	| { status: "rejected"; reason: unknown };
+
+class TransactionAttachmentState {
```
AlCalzone: TransactionAttachment and TransactionAttachmentState seem over-engineered.
Whats the simplest possible solution that achieves the same goal?

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -8277,7 +8474,7 @@ ${handlers.length} left`,
 				&& source === "queue"
 			) {
 				requeue.push(transaction);
-				return { type: "drop" };
+				return { type: "defer" };
```
AlCalzone, line 8584: Explain - why is this change (and the "defer" type) needed? Which problem does it solve?

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7479,31 +7669,41 @@ ${handlers.length} left`,
...
 		}
 		transaction.requestWakeUpOnDemand = !!options.requestWakeUpOnDemand;
+		transaction.requestStatusUpdates = "requestStatusUpdates" in options
+			&& options.requestStatusUpdates === true;
 		transaction.tag = options.tag;
 
-		// And queue it
-		this.getQueueForTransaction(transaction).add(transaction);
 		transaction.setProgress({ state: TransactionState.Queued });
+		this.enqueueTransaction(transaction);
```
AlCalzone, line 7792: Why is the transaction now queued after the transaction state is set to "Queued"?

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7552,9 +7752,6 @@ ${handlers.length} left`,
 				) {
 					this._controller?.incrementStatistics("messagesDroppedTX");
 					return e.context as TResponse;
-				} else if (e.code === ZWaveErrorCodes.Controller_NodeTimeout) {
-					// If the node failed to respond in time, remember this for the statistics
-					node?.incrementStatistics("timeoutResponse");
```
AlCalzone, line 7577: Why is this being removed?

packages/cc/src/cc/CRC16CC.ts
AlCalzone: Fixed in d7bb1243 — inverted this guard and the other transparent wrapper guards to use positive typed recursion.

packages/cc/src/cc/MultiChannelCC.ts
AlCalzone: Fixed in d7bb1243 — the typed wrapper and matching destination now form the positive branch.

packages/cc/src/cc/MultiChannelCC.ts
AlCalzone: Fixed in d7bb1243 — Multi Channel V1 now uses the same positive-flow structure.

packages/cc/src/lib/CommandClass.getCommandRelation.test.ts
AlCalzone: Fixed in d7bb1243 — renamed to CommandClass.getCommandRelation.test.ts.

packages/cc/src/lib/CommandClass.getCommandRelation.test.ts
AlCalzone: Fixed in d7bb1243 — added brief comments for the minimal Basic Set fixture and same-CC cross-command contract.

packages/cc/src/lib/CommandClass.test.ts
AlCalzone: Fixed in d7bb1243 — imported and used the named SecurityManager type.

packages/cc/src/lib/CommandClass.getCommandRelation.test.ts
AlCalzone: Fixed in d7bb1243 — added rationale comments around the non-obvious expectation groups.

packages/zwave-js/src/lib/test/driver/commandDeduplication.test.ts
AlCalzone: Fixed in d7bb1243 — added bookmark comments for the integration test phases and expected outcomes.

packages/zwave-js/src/lib/driver/Transaction.ts
AlCalzone: Explained in d7bb1243 — status-update callers need distinct physical Supervision sessions and update streams, so they cannot coalesce.

packages/zwave-js/src/lib/driver/Transaction.ts
AlCalzone: Simplified in d7bb1243 — removed per-attachment progress and detach flags. Shared state remains for result fan-out, clone survival, transfer-aware expiry, and terminal replay.

packages/zwave-js/src/lib/driver/Driver.ts
AlCalzone: Simplified in d7bb1243 — removed the unused drop variant. Defer remains because delayed transactions leave the queue unsettled and must stay tracked for requeue or teardown rejection.

packages/zwave-js/src/lib/driver/Driver.ts
AlCalzone: Explained in d7bb1243 — enqueue may synchronously transfer the caller to an active transaction, so Queued must be emitted before Active is replayed.

packages/zwave-js/src/lib/driver/Driver.ts
AlCalzone: Explained in d7bb1243 — timeout statistics moved to the physical promise so coalesced caller promises do not count one radio timeout multiple times.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7387,6 +7390,210 @@ ${handlers.length} left`,
...
+					|| !isSendData(older.message)
+					|| !containsCC(older.message)
+				) {
+					return CommandRelation.Unrelated;
+				}
+				return getCommandRelation(
+					newerMessage.command,
+					older.message.command,
+				);
+			};
```
AlCalzone: Why is this an inline function? Seems like it should be a method on the `Transaction` class or a helper somewhere at the module level.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7387,6 +7390,210 @@ ${handlers.length} left`,
...
+
+			for (const queue of this.queues) {
+				if (
+					queue.currentTransaction
```
AlCalzone: Why are we looping through queues? Each transaction should be assignable to a specific queue. Lets not introduce deduplication across queues.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7387,6 +7390,210 @@ ${handlers.length} left`,
...
+						return true;
+					});
+					if (batch.transactions.length === 0) {
+						batch.timer.clear();
+						batches.delete(batch);
+					}
+				}
+				if (batches.size === 0) this.requeueTimers.delete(nodeId);
+			}
+		}
```
AlCalzone: If I've got this right, transactions are deferred because the node is not ready to handle them right now (Application Status CC). In that case, inheriting the delay actually sounds like the right thing to do.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7387,6 +7390,210 @@ ${handlers.length} left`,
...
+
+		// Reject only after all bookkeeping is done. Rejecting runs caller
+		// listeners synchronously, which may re-enter this method.
+		for (const { queue, transaction: older } of queuedSuperseded) {
+			queue.removeWithoutTrigger(older);
```
AlCalzone: Is this also the reason we're modifying the queues without triggering? The code seems to imply that, but the comment talks about rejecting and bookkeeping.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7387,6 +7390,210 @@ ${handlers.length} left`,
...
+					older.stack,
+				),
+			);
+		}
+
+		if (transaction.preventDeduplication) {
+			// The protected command transmits itself and absorbs the callers of
+			// the duplicates it replaces, keeping the earliest schedule among them
+			for (const { queue, transaction: older } of queuedRedundant) {
+				transaction.priority = Math.min(
```
AlCalzone: Highest priority = lowest value? Then add a comment clarifying this.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7387,6 +7390,210 @@ ${handlers.length} left`,
...
+					ZWaveErrorCodes.Controller_MessageSuperseded,
+					undefined,
+					older.stack,
+				),
+			);
+		}
+
+		if (transaction.preventDeduplication) {
+			// The protected command transmits itself and absorbs the callers of
+			// the duplicates it replaces, keeping the earliest schedule among them
```
AlCalzone: This does not quite seem to tell the whole story of what we do in this loop and why.

AlCalzone: Prefer comments closer to the code if possible.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7387,6 +7390,210 @@ ${handlers.length} left`,
...
+					this.getQueueForTransaction(redundant)
+						.addWithoutTrigger(redundant);
+				}
+			} else {
+				this.getQueueForTransaction(transaction)
+					.addWithoutTrigger(transaction);
+			}
+		}
+
+		this.triggerQueues();
```
AlCalzone: Add a brief comment explaining why the trigger happens now.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7479,31 +7707,42 @@ ${handlers.length} left`,
 			transaction.pauseSendThread = true;
 		}
 		transaction.requestWakeUpOnDemand = !!options.requestWakeUpOnDemand;
+		transaction.requestStatusUpdates = "requestStatusUpdates" in options
+			&& options.requestStatusUpdates === true;
 		transaction.tag = options.tag;
 
-		// And queue it
-		this.getQueueForTransaction(transaction).add(transaction);
+		// Emit Queued first because coalescing may replay Active synchronously
```
AlCalzone: Simpler english is preferable

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -72,46 +213,81 @@ export class Transaction implements Comparable<Transaction> {
...
 	public readonly message: Message;
 
 	/** The message generator to create the actual messages for this transaction */
 	public readonly parts: MessageGenerator;
 
-	/** A callback which gets called with state updates of this transaction */
-	private listener?: TransactionProgressListener;
+	private readonly lifecycle: TransactionLifecycle;
+
+	public attach(
```
AlCalzone, line 292: Add a brief JSDoc comment explaining what the parameters are. Use `@param` for that.

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -39,11 +41,148 @@ export interface TransactionOptions {
...
+	 * Ensures this command is transmitted: it will not share another command's
+	 * physical transmission and a newer command cannot supersede it.
+	 */
+	preventDeduplication?: boolean;
+	/** Reports the physical outcome exactly once, even when no callers remain attached */
+	onSettled?: (result: TransactionResult) => void;
+}
 
-	/** Gets called with progress updates for a transaction */
+interface TransactionCaller {
```
AlCalzone, line 58: Add a JSDoc comment explaining what a TransactionCaller is (for)

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -39,11 +41,148 @@ export interface TransactionOptions {
...
 	listener?: TransactionProgressListener;
+	onTXReport?: (report: TXReport) => void;
+	lifecycle: TransactionLifecycle;
+}
+
+export type TransactionResult =
+	| { status: "fulfilled"; value: Message | void }
+	| { status: "rejected"; reason: unknown };
+
+class TransactionLifecycle {
```
AlCalzone, line 101: Add a JSDoc comment explaining what a TransactionLifecycle is (for)

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -39,11 +41,148 @@ export interface TransactionOptions {
...
+	public constructor(
+		private readonly onSettled?: (result: TransactionResult) => void,
```
AlCalzone: Avoid parameter properties. Define the property like normal JS.

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -39,11 +41,148 @@ export interface TransactionOptions {
...
+	) {}
+
+	private readonly callers = new Set<TransactionCaller>();
+	private settled: TransactionResult | undefined;
+	private progress: TransactionProgress | undefined;
+
+	public attach(caller: TransactionCaller): void {
```
AlCalzone, line 115: Each public method on this class should have a brief JSDoc comment explaining its purpose

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -39,11 +41,148 @@ export interface TransactionOptions {
...
+	}
+
+	public adoptCallersFrom(source: TransactionLifecycle): void {
+		if (source === this) return;
+		const targetProgress = this.progress;
+		const replayTargetProgress = targetProgress != undefined
+			&& source.progress?.state !== targetProgress.state;
+		for (const caller of source.callers) {
+			source.callers.delete(caller);
+			// Expiry handles must follow the caller after a protected replacement
```
AlCalzone: Clarify this comment in plain english

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -39,11 +41,148 @@ export interface TransactionOptions {
...
+		) {
+			return;
+		}
+		this.progress = progress;
+		for (const caller of this.callers) {
+			caller.listener?.({ ...progress });
+		}
+	}
+
+	private replaySettlement(caller: TransactionCaller): void {
```
AlCalzone, line 220: Also here - state what it is for.

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -39,11 +41,148 @@ export interface TransactionOptions {
...
+export interface TransactionAttachmentHandle {
+	/**
+	 * Rejects this caller's promise and stops its progress updates.
+	 * Returns `true` when no callers remain attached to the shared transmission.
+	 */
+	detach(error: ZWaveError): boolean;
+	/**
+	 * Whether the given transaction currently carries this caller's lifecycle.
+	 * Remains usable after detaching, so the caller can locate and cancel
+	 * a transmission nobody waits for anymore.
```
AlCalzone: Simplify the comment - not everyone knows out of the box what "sharing the caller's lifecycle" means.

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -72,46 +213,81 @@ export class Transaction implements Comparable<Transaction> {
...
+		this.lifecycle.attach(caller);
+
+		return {
+			detach: (error) => caller.lifecycle.detach(caller, error),
+			sharesLifecycleWith: (transaction) =>
+				caller.lifecycle === transaction.lifecycle,
+		};
+	}
+
+	public get hasAttachments(): boolean {
```
AlCalzone, line 313: All public methods should have a brief JSDoc comment explaining their purpose. In simple words.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7387,6 +7390,244 @@ ${handlers.length} left`,
 		}
 	}
 
+	private areTransactionsCompatible(
```
AlCalzone, line 7417: Add a brief JSDoc comment explaining what this method does.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7387,6 +7390,244 @@ ${handlers.length} left`,
...
+				!(first.message instanceof SendDataBridgeRequest)
+				|| !(second.message instanceof SendDataBridgeRequest)
+				|| first.message.sourceNodeId === second.message.sourceNodeId
+			)
+			&& (
+				!(first.message instanceof SendDataMulticastBridgeRequest)
+				|| !(second.message instanceof SendDataMulticastBridgeRequest)
+				|| first.message.sourceNodeId === second.message.sourceNodeId
+			);
+	}
```
AlCalzone, line 7473: Structure the conditions in this method in smaller chunks and add a brief comment above each, explaining the reasoning behind it.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7387,6 +7390,244 @@ ${handlers.length} left`,
...
+					older.creationTimestamp,
+				);
+				// The duplicate never transmits. Its callers get their result
+				// from the protected command instead.
+				queue.removeWithoutTrigger(older);
+				transaction.adoptCallersFrom(older);
+			}
+			queue.addWithoutTrigger(transaction);
+		} else {
+			// The new command's callers wait for an existing transmission and
```
AlCalzone: This comment is misleading

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7387,6 +7390,244 @@ ${handlers.length} left`,
...
+				}
+			} else {
+				queue.addWithoutTrigger(transaction);
+			}
+		}
+
+		// The queue was modified without triggering it, so it does not start
+		// executing a transaction in the middle of this update. Trigger it now
+		// that the update is complete.
+		queue.trigger();
```
AlCalzone: I wonder if `pause()` and `unpause()` would make more sense - where `unpause` automatically triggers. This would also prevent races where we're in the process of modifying the queue, and a completed command triggers the queue elsewhere.

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -7454,21 +7695,42 @@ ${handlers.length} left`,
...
+			preventDeduplication: "preventDeduplication" in options
+				&& options.preventDeduplication === true,
+			onSettled: (result) => {
+				if (
+					result.status === "rejected"
+					&& isZWaveError(result.reason)
+					&& result.reason.code
+						=== ZWaveErrorCodes.Controller_NodeTimeout
+				) {
+					// Count each physical timeout once across coalesced callers
```
AlCalzone:
```suggestion
					// Count each physical timeout once per actual transaction
```
Lets not add `coalesce` into our vocabulary here. Try to reword it to something simpler everywhere else as well

packages/zwave-js/src/lib/driver/Driver.ts
```diff
@@ -8290,14 +8553,49 @@ ${handlers.length} left`,
...
+			let batch: DeferredTransactionBatch;
 			const timer = setTimer(() => {
-				timerSet.delete(timer);
-				const requeued = requeue.map((t) => t.clone());
+				timerSet.delete(batch);
+				if (timerSet.size === 0) {
+					this.requeueTimers.delete(nodeId);
+				}
+				// A newer superseding command may have removed some transactions
+				// from the batch in the meantime
```
AlCalzone:
```suggestion
				// Clone the transactions immediately before queuing them.
				// A newer superseding command may have removed some transactions in the meantime.
```

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -39,11 +41,179 @@ export interface TransactionOptions {
...
 	/** The priority of this transaction */
 	priority: MessagePriority;
-	/** Will be resolved/rejected by the Send Thread Machine when the entire transaction is handled */
-	promise: DeferredPromise<Message | void>;
+	/**
+	 * Ensures this command is transmitted: it will not share another command's
+	 * physical transmission and a newer command cannot supersede it.
+	 */
+	preventDeduplication?: boolean;
+	/** Reports the physical outcome exactly once, even when no callers remain attached */
```
AlCalzone:
```suggestion
	/** Reports the physical outcome exactly once, no matter how many deduplicated transactions are attached. */
```

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -39,11 +41,179 @@ export interface TransactionOptions {
...
+	onSettled?: (result: TransactionResult) => void;
+}
 
-	/** Gets called with progress updates for a transaction */
+/**
+ * A caller waiting for the result of a transaction: the promise to settle,
+ * plus optional callbacks for progress updates and TX reports. Multiple
+ * callers can wait for the same transaction when commands are coalesced.
```
AlCalzone: I prefer "deduplicate" over "coalesce"

packages/zwave-js/src/lib/driver/Transaction.ts
```diff
@@ -39,11 +41,179 @@ export interface TransactionOptions {
...
+	 * Returns `true` when no other callers are waiting for the transmission.
+	 */
+	detach(error: ZWaveError): boolean;
+	/**
+	 * Whether this caller gets its result from the given transaction.
+	 * Also works after detaching, so the caller can find and cancel
+	 * a transmission nobody waits for anymore.
+	 */
+	sharesLifecycleWith(transaction: Transaction): boolean;
 }
```
AlCalzone: move all type and interface definitionsi to the top

AlCalzone:
@zwave-js-bot approve

AlCalzone:
@zwave-js-bot approve

PR #9140, opened 2026-08-27
fix(config): add remapNotifications to fix malformed notifications sent by HomeSeer FS100 devices.

spudwebb:
The HomeSeer FS100 firmware sends a malformed idle notification..
When light goes away, the FS-100L sends Notification type 0x14 (Light sensor), event 0x00 (idle), event parameter 0x02. The event parameter is supposed to name the state variable being cleared, and should be 0x01 ("Light detected"). Instead the device sends 0x02, which is registered as an event, not a state.

This has been reported [here](https://community.home-assistant.io/t/hs-fs100-l-not-coming-over-to-zwavejs-from-deprecated-z-wave/320824/14) and more recently [here](https://forums.homeseer.com/forum/hs4-products/hs4-plugins/lighting-primary-technology-plug-ins-aa/z-wave-plus-homeseer/1759349-fs-100l-device-in-hs4-acting-strange-after-upgrading-from-zwave-zwave-plus-plugin)

The fix is to use a `remapNotifications`  to rewrite the broken idle into an explicit idle for the right variable:

```
"remapNotifications": [
  { "from": { "notificationType": 20, "notificationEvent": 0 },
    "idle": [ { "notificationType": 20, "notificationEvent": 1 } ] }
]
```

### DIFF
--- a/packages/config/config/devices/0x000c/hs-fs100-l.json
+++ b/packages/config/config/devices/0x000c/hs-fs100-l.json
@@ -30,5 +30,28 @@
 			"#": "4",
 			"$import": "templates/homeseer_template.json#enable_notification_buzzer"
 		}
-	]
+	],
+	"compat": {
+		// This device signals that light is no longer detected by sending an "idle"
+		// notification (Light sensor, event 0x00) whose event parameter is 0x02
+		// ("Light color transition detected"). That event is not a state variable, so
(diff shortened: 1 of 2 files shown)


AlCalzone (approved):
Lgtm. I'll wait for Copilot to confirm the shape since I can't test right now.

PR #9146, opened 2026-08-27
fix(cc): trim AGI command list reports

### LINKED ISSUE #9145: AGI Command List Report serializer appends an unused byte
### Description

`AssociationGroupInfoCCCommandListReport.serialize()` leaves an unused zero byte at the end of reports containing one-byte Command Class IDs.

### Reproduction

CTT tests `AssociationGrpInfoCmdClassV3_Rev02` and `MultiCommandCmdClassV1_Rev04` request the command list for Association Group 1. The intended report is:

```text
59 06 01 02 5A 01
```

The S2 ciphertext has seven bytes, proving that the serialized inner command is:

```text
59 06 01 02 5A 01 00
```

CTT parses the declared two-byte command list and then correctly rejects the trailing byte:

```text
GroupingIdentifier  01      (PASS)
ListLength           02      (PASS)
Command              5A, 01  (PASS)
Payload length (7) was greater than expected (6)!  (FAIL)
```

### Root cause

The serializer allocates three bytes per map entry to allow extended Command Class IDs:

```ts
this.payload = new Bytes(2 + this.commands.size * 3);
```

For a one-byte CC ID it writes only two bytes for the entry. `offset` tracks the used size, but the payload is never trimmed before `super.serialize(ctx)`.

### Expected behavior

Trim the payload to `offset` after encoding the command list.

AlCalzone:
`AssociationGroupInfoCCCommandListReport` allocates enough payload space for extended Command Class IDs. One-byte IDs left unused zero bytes in that allocation. The serializer now trims the payload to its tracked encoded length.

A regression test covers the certification frame `59 06 01 02 5A 01`.

Validation:
- `yarn test:ts packages/zwave-js/src/lib/test/cc/AssociationGroupInfoCC.test.ts`
- `yarn build`

Fixes #9145

### DIFF
--- a/packages/cc/src/cc/AssociationGroupInfoCC.ts
+++ b/packages/cc/src/cc/AssociationGroupInfoCC.ts
@@ -868,6 +868,7 @@ export class AssociationGroupInfoCCCommandListReport
 			}
 		}
 		this.payload[1] = offset - 2; // list length
+		this.payload = this.payload.subarray(0, offset);
 
 		return super.serialize(ctx);
 	}
--- a/packages/zwave-js/src/lib/test/cc/AssociationGroupInfoCC.test.ts
+++ b/packages/zwave-js/src/lib/test/cc/AssociationGroupInfoCC.test.ts
@@ -10,6 +10,7 @@ import {
 	AssociationGroupInfoProfile,
 	BasicCommand,
 	CommandClass,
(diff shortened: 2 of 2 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #9147, opened 2026-08-27
fix: preserve received CommandClass frame type

### LINKED ISSUE #9142: Received CommandClass loses multicast and broadcast frame type
### Description

Received Get commands lose their Z-Wave frame type during Command Class parsing. As a result, zwave-js responds to multicast and broadcast Gets that require no response.

### Reproduction

CTT test `CDR_MulticastNoResponse_Rev02` includes its controller non-securely and sends `ZWavePlusCCGet` as multicast, broadcast, and singlecast.

The Serial API message initially identifies the addressing correctly:

```text
[BridgeApplicationCommand]
  type: multicast
  target node: 1
  [ZWavePlusCCGet]
[ZWavePlusCCReport] sent

[BridgeApplicationCommand]
  type: broadcast
  target node: 255
  [ZWavePlusCCGet]
[ZWavePlusCCReport] sent
```

The singlecast control is also answered as expected. CTT fails only the multicast and broadcast cases.

### Root cause

`Driver` passes `msg.frameType` in the `CommandClass.parse()` context. The resulting `CommandClass` does not retain that value. `Driver.shouldDiscardCC()` checks:

```ts
cc.constructor.name.endsWith("Get")
  && (cc.frameType === "multicast" || cc.frameType === "broadcast")
```

`cc.frameType` is `undefined`, so the suppression guard does not run. The numeric source node ID also makes `isSinglecast()` fall back to `true`.

### Expected behavior

Received Command Classes must preserve the incoming frame type. Get commands received via multicast or broadcast must be discarded without a response.

AlCalzone:
Fixes #9142

The Serial API decoder passed the frame type into Command Class parsing, but parsed command instances did not retain it. Attach the parsing context frame type to every deserialization result so multicast and broadcast Gets reach the existing discard guard.

The integration regression test sends singlecast, multicast, and broadcast Z-Wave Plus Info Gets through `BridgeApplicationCommandRequest`. It confirms that singlecast still receives a report and the other frame types do not.

Validation:
- `yarn test:ts packages/zwave-js/src/lib/test/driver/receiveApplicationCommandHandlerBridge.test.ts`
- `yarn build`

### DIFF
--- a/packages/cc/src/lib/CommandClass.ts
+++ b/packages/cc/src/lib/CommandClass.ts
@@ -219,7 +219,6 @@ export class CommandClass implements CCId {
 			ccId = getCommandClass(this),
 			ccCommand = getCCCommand(this),
 			payload = new Uint8Array(),
-			frameType,
 		} = options;
 
 		this.nodeId = nodeId;
@@ -227,7 +226,6 @@ export class CommandClass implements CCId {
 		this.ccId = ccId;
 		this.ccCommand = ccCommand;
 		this.payload = Bytes.view(payload);
-		this.frameType = frameType;
 	}
 
 	public static async parse(
@@ -240,7 +238,10 @@ export class CommandClass implements CCId {
 		const CCConstructor = getCCConstructor(raw.ccId);
 		if (!CCConstructor) {
 			// None -> fall back to the default constructor
-			return await CommandClass.from(raw, ctx);
+			return CommandClass.withFrameType(
+				await CommandClass.from(raw, ctx),
+				ctx.frameType,
+			);
 		}
 
 		let CommandConstructor: CCConstructor<CommandClass> | undefined;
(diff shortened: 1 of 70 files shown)


packages/cc/src/lib/CommandClass.ts
```diff
@@ -279,12 +285,20 @@ export class CommandClass implements CCId {
...
 				});
 
-				return ret;
+				return CommandClass.withFrameType(ret, ctx.frameType);
 			}
 			throw e;
 		}
 	}
 
+	private static withFrameType<T extends CommandClass>(
```
AlCalzone: This shouldn't be needed. CommandClass.from and/or the constructor should set the frame type correctly.

AlCalzone: Fixed in af1f0addd. Every deserializer now passes the frame type into the CC constructor.

AlCalzone: Reconsidered this after seeing the full constructor impact. Restored the centralized assignment in 8e59d7151.

PR #9148, opened 2026-08-27
fix(zniffer): parse PTI captures from Z-Wave CTT

### LINKED ISSUE #9139: PTI ZLF from Z-Wave CTT loads with zero captured frames
A PTI-based `.zlf` produced by Z-Wave CTT 4.0.3 loads without errors but yields zero frames with zwave-js 15.27.1.

This appears related to the PTI support added in #9053.

### Reproduction

Call `Zniffer.loadCaptureFromFile()` for the CTT-generated ZLF. `zniffer.capturedFrames` remains empty.

Inspecting the same file with `parseZLFHeader()` and `parseZLFEntry()` finds:

```text
1620 chunks
607 kind-9 PTI chunks
4 attachments
0 parsed entries
```

The PTI chunks start with a valid debug-channel delimiter and contain an `f8 ... f9` radio frame, for example:

```text
5b20000000000000000000000000f8...f9000f0006515d
```

`parsePTIFrame()` reads the first two DCH body bytes as version `0`. It only accepts versions `2` and `3`, so every PTI frame is discarded without an error.

### Expected

The CTT-generated PTI entries should be parsed into captured frames, or loading should report that this PTI/DCH variant is unsupported instead of silently returning an empty capture.

AlCalzone:
Fixes #9139

Z-Wave CTT 4.0.3 writes a v2-sized DCH header with zero values for the version and message type. `parsePTIFrame` rejected the version before it examined the valid PTI hardware marker. Every captured frame was omitted.

The parser now accepts this exact zeroed header variant. Modern DCH version and message-type validation remains unchanged. The PTI marker and payload validation still apply.

Validation:
- `yarn fmt`
- `yarn workspace zwave-js test:ts run src/lib/test/zniffer/parseZLFEntry.test.ts --reporter=dot`
- `yarn build`

### DIFF
--- a/packages/zwave-js/src/lib/test/zniffer/parseZLFEntry.test.ts
+++ b/packages/zwave-js/src/lib/test/zniffer/parseZLFEntry.test.ts
@@ -38,6 +38,18 @@ test("parse complete data frame", async (t) => {
 	t.expect(parsed.entries[0].msg).toBeInstanceOf(ZnifferDataMessage);
 });
 
+test("parse CTT PTI frame with zeroed DCH header", async (t) => {
+	const rawMsg = Bytes.from(
+		"8013fa7c93addd8801220000005b20000000000000000000000000f8c4dae60701410a0c02008f68f9000f0006515df5",
+		"hex",
+	);
+	const parsed = parseZLFEntry(rawMsg, 0);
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #9149, opened 2026-08-27
fix: remove proxy-excluded nodes from controller

### LINKED ISSUE #9143: Proxy-excluded nodes remain in controller.nodes
### Description

When another controller removes a node, zwave-js emits `node removed` with `ProxyExcluded` but leaves the node in `controller.nodes`.

### Reproduction

CTT tests `S2_S0IncByNonSecureIncController_Rev01` and `S2_SISInclusionCCSide_Rev02` include and remove joining nodes through a secondary Inclusion Controller.

zwave-js receives and logs the removal notification:

```text
[Node 004] [ApplicationUpdateRequest]
[Node 004] was removed from the network by another controller
```

The `node removed` event is emitted, but `controller.nodes.has(4)` remains true. CTT therefore reports that the excluded node is still shown in the DUT device list. The same behavior reproduces across S0 and S2 phases.

### Root cause

The proxy-removal branch emits the event without deleting the node:

```ts
if (node) {
  this.emit("node removed", node, RemoveNodeReason.ProxyExcluded);
}
```

Normal exclusion and proxy replacement both call `_nodes.delete(...)` after emitting the event.

### Expected behavior

Delete a proxy-excluded node from `_nodes` as part of handling `ApplicationUpdateRequestNodeRemoved`, consistent with the other removal paths.

AlCalzone:
Fixes #9143

`ApplicationUpdateRequestNodeRemoved` emitted the removal event but left the node in the controller map. Delete the node after the event so driver cleanup runs in the same order as normal exclusion.

The integration test sends the proxy-exclusion Serial API update. It verifies the `ProxyExcluded` event and the removal from `controller.nodes`.

Validation:
- `yarn test:ts packages/zwave-js/src/lib/test/node/proxyExclusionRemovesNode.test.ts`
- `yarn build`

### DIFF
--- a/packages/zwave-js/src/lib/controller/Controller.ts
+++ b/packages/zwave-js/src/lib/controller/Controller.ts
@@ -3858,6 +3858,7 @@ export class ZWaveController
 				);
 
 				this.emit("node removed", node, RemoveNodeReason.ProxyExcluded);
+				this._nodes.delete(node.id);
 			}
 		} else if (msg instanceof ApplicationUpdateRequestNodeAdded) {
 			// A node was included by another controller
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #9150, opened 2026-08-27
fix: handle S0 Commands Supported Get on S2 nodes

### LINKED ISSUE #9141: S0 Commands Supported Get is discarded on S2-enabled nodes
### Description

An S0-encapsulated `SecurityCCCommandsSupportedGet` is discarded as a downgrade when the sender's highest granted class is S2. The command should be accepted and answered with an empty S0 Commands Supported Report.

### Reproduction

CTT test `S2_CommandClasses_Rev01` includes its controller with S2 Access, S2 Authenticated, S2 Unauthenticated, and S0. It obtains an S0 nonce and sends an S0-encrypted Commands Supported Get.

zwave-js decrypts the payload but discards the wrapper:

```text
command received at a lower security level than expected
[SecurityCCCommandEncapsulation] [discarded]
  plaintext: 0x9802
```

`98 02` is `SecurityCCCommandsSupportedGet`. No empty Report is sent, so the test times out.

Lower-class S2 Commands Supported control cases are answered correctly.

### Root cause

`isSecurityLevelTooLow()` intends to allow this command:

```ts
if (cmd.encapsulated instanceof SecurityCCCommandsSupportedGet) {
  return true;
}
```

At this point the S0 wrapper has decrypted the plaintext, but its inner Command Class object is not populated. The exception does not match. The outer S0 wrapper is then rejected because the node's highest class is S2 Access.

### Expected behavior

Recognize the decrypted S0 Commands Supported Get before applying downgrade filtering and return an empty S0 Commands Supported Report.

AlCalzone:
## Root cause

S0 decryption authenticates the wrapper and stores the inner command as raw bytes. The inner command is parsed during partial CC assembly. The initial downgrade check runs before this step. Its typed Commands Supported exceptions therefore cannot match the incoming wrapper.

## Changes

- Expose the authenticated decrypted S0 payload for receive-path validation.
- Allow S0 Commands Supported Get and Report through the initial downgrade check when their raw command header matches.
- Re-run security-level validation after partial CC assembly has parsed the inner command.
- Cover both directions on nodes with S2 and S0 grants: an inbound S0 Get receives an empty Report and an S0 Report completes a controller query.

## Validation

- `yarn test:ts packages/zwave-js/src/lib/test/cc-specific/securityS0CommandsSupportedGet.test.ts` (2 tests)

Fixes #9141

### DIFF
--- a/packages/cc/src/cc/SecurityCC.ts
+++ b/packages/cc/src/cc/SecurityCC.ts
@@ -739,17 +739,6 @@ export class SecurityCCCommandEncapsulation extends SecurityCC {
 
 	private alternativeNetworkKey?: BytesView;
 
-	public isEncapsulatingSecurityCommand(command: SecurityCommand): boolean {
-		if (this.encapsulated) {
-			return this.encapsulated.ccId === CommandClasses.Security
-				&& this.encapsulated.ccCommand === command;
-		}
-
-		return !this.sequenced
-			&& this.decryptedCCBytes?.[0] === CommandClasses.Security
-			&& this.decryptedCCBytes[1] === command;
-	}
-
 	public get nonceId(): number | undefined {
 		return this.nonce?.[0];
 	}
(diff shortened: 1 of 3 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #9151, opened 2026-08-27
fix(cc): handle Manufacturer Specific Device Specific Get

### LINKED ISSUE #9144: Manufacturer Specific Device Specific Get cannot be received
### Description

zwave-js advertises Manufacturer Specific V2 but cannot deserialize or answer incoming Device Specific Gets.

### Reproduction

CTT test `ManufacturerSpecificCmdClassV2_Rev03` runs against the secure root device. The ordinary Manufacturer Specific Get succeeds, proving S2 delivery and the existing handler work.

CTT then sends Device Specific Gets for the default, Serial Number, and Pseudo Random Number types. All three reach zwave-js and are dropped:

```text
Dropping message because it could not be deserialized:
ManufacturerSpecificCCDeviceSpecificGet: deserialization not implemented
```

No Device Specific Report is returned. Manufacturer Specific V2 requires at least one of Serial Number or Pseudo Random Number to be implemented and returned for the default request.

### Root cause

`ManufacturerSpecificCCDeviceSpecificGet.from()` throws `Deserialization_NotImplemented`. In addition, `Node.js` handles `ManufacturerSpecificCCGet` but has no branch for `ManufacturerSpecificCCDeviceSpecificGet` that could return a Device Specific Report.

### Expected behavior

Deserialize incoming Device Specific Gets and answer them with a configured stable device identifier for at least one required type.

AlCalzone:
Fixes #9144

## Root cause

`ManufacturerSpecificCCDeviceSpecificGet.from()` always threw `Deserialization_NotImplemented`. The node command dispatcher also handled only the v1 Manufacturer Specific Get. A received Device Specific Get was dropped before the application could answer it.

## Changes

- Deserialize Device Specific Get commands and validate Device Specific Report fields.
- Serialize string IDs as UTF-8 and `Uint8Array` IDs as binary data.
- Return a configured `vendor.deviceId` as the Serial Number.
- Return a 16-byte binary Pseudo Random ID when no device ID is configured.
- Cover raw inbound receipt and response handling with integration regressions. Add focused command serialization and deserialization tests.

## Validation

- `yarn test:ts packages/zwave-js/src/lib/test/cc/ManufacturerSpecificCC.test.ts packages/zwave-js/src/lib/test/cc-specific/manufacturerSpecificDeviceSpecificGet.test.ts`
- `yarn build`
- `yarn fmt`

### DIFF
--- a/packages/cc/src/cc/ManufacturerSpecificCC.ts
+++ b/packages/cc/src/cc/ManufacturerSpecificCC.ts
@@ -311,7 +311,7 @@ export class ManufacturerSpecificCCGet extends ManufacturerSpecificCC {}
 // @publicAPI
 export interface ManufacturerSpecificCCDeviceSpecificReportOptions {
 	type: DeviceIdType;
-	deviceId: string;
+	deviceId: string | Uint8Array;
 }
 
 @CCCommand(ManufacturerSpecificCommand.DeviceSpecificReport)
@@ -329,7 +329,15 @@ export class ManufacturerSpecificCCDeviceSpecificReport
 		super(options);
 
 		this.type = options.type;
-		this.deviceId = options.deviceId;
+		if (typeof options.deviceId === "string") {
+			this.deviceId = options.deviceId;
+			this.deviceIdData = Bytes.from(options.deviceId, "utf8");
+			this.deviceIdDataFormat = 0;
+		} else {
+			this.deviceId = "0x" + Bytes.view(options.deviceId).toString("hex");
+			this.deviceIdData = Bytes.from(options.deviceId);
+			this.deviceIdDataFormat = 1;
+		}
 	}
 
 	public static from(
@@ -348,9 +356,9 @@ export class ManufacturerSpecificCCDeviceSpecificReport
 			raw.payload.length >= 2 + dataLength,
 		);
 		const deviceIdData = raw.payload.subarray(2, 2 + dataLength);
-		const deviceId: string = dataFormat === 0
+		const deviceId: string | Uint8Array = dataFormat === 0
 			? deviceIdData.toString("utf8")
-			: "0x" + deviceIdData.toString("hex");
+			: deviceIdData;
 
 		return new this({
 			nodeId: ctx.sourceNodeId,
@@ -362,6 +370,8 @@ export class ManufacturerSpecificCCDeviceSpecificReport
 	public readonly type: DeviceIdType;
 
 	public readonly deviceId: string;
+	private readonly deviceIdData: Bytes;
+	private readonly deviceIdDataFormat: 0 | 1;
 
 	public serialize(ctx: CCEncodingContext): Promise<Bytes> {
 		if (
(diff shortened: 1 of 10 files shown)


packages/zwave-js/src/lib/driver/ZWaveOptions.ts
```diff
@@ -459,6 +459,11 @@ export interface ZWaveOptions {
 		manufacturerId: number;
 		productType: number;
 		productId: number;
+		/**
+		 * A stable device identifier for Manufacturer Specific Device Specific Reports.
+		 * Prefix binary identifiers with "0x". The encoded identifier may be up to 31 bytes.
+		 */
+		deviceId?: string;
```
AlCalzone: I think we should allow string and Uint8Array. string returns UTF8, Uint8Array returns Binary data.

packages/zwave-js/src/lib/node/CCHandlers/ManufacturerSpecificCC.ts
```diff
@@ -34,3 +36,32 @@ export async function handleManufacturerSpecificGet(
...
+): Promise<void> {
+	if (!vendorInfo?.deviceId) {
+		ctx.logNode(node.id, {
+			message:
+				"Cannot respond to Device Specific Get because no vendor device ID is configured",
+			direction: "none",
+			level: "warn",
+		});
+		return;
+	}
```
AlCalzone: In this case we should respond with binary pseudo-random data (16 bytes). The specs seem to allow this.

packages/zwave-js/src/lib/test/cc-specific/manufacturerSpecificDeviceSpecificGet.test.ts
```diff
@@ -0,0 +1,54 @@
...
+	type MockZWaveRequestFrame,
+	createMockZWaveRequestFrame,
+} from "@zwave-js/testing";
+import { integrationTest } from "../integrationTestSuite.js";
+
+const vendor = {
+	manufacturerId: 0x1234,
+	productType: 0x5678,
+	productId: 0x9abc,
+	deviceId: "0x1234567890abcdef",
```
AlCalzone: If this is given, we should always respond with this as the serial number.
If not, we should return PseudoRandom data (see comment above).

packages/zwave-js/src/lib/driver/ZWaveOptions.ts
AlCalzone: Fixed in 42d99b87f. Strings now serialize as UTF-8 and Uint8Arrays serialize as binary.

packages/zwave-js/src/lib/node/CCHandlers/ManufacturerSpecificCC.ts
AlCalzone: Fixed in 42d99b87f. Requests now receive a 16-byte binary Pseudo Random response when no device ID is configured.

packages/zwave-js/src/lib/test/cc-specific/manufacturerSpecificDeviceSpecificGet.test.ts
AlCalzone: Fixed in 42d99b87f. Integration coverage now checks configured string and binary Serial Number responses plus the Pseudo Random fallback.

packages/cc/src/cc/ManufacturerSpecificCC.ts
```diff
@@ -340,6 +370,36 @@ export class ManufacturerSpecificCCDeviceSpecificReport
 	public readonly type: DeviceIdType;
 
 	public readonly deviceId: string;
+	private readonly deviceIdData: Bytes;
+	private readonly deviceIdDataFormat: 0 | 1;
```
AlCalzone: Nope.


```suggestion
public readonly deviceId: string | Bytes;
```
this also implies the device ID format.

AlCalzone: Fixed in d423322ce. Binary IDs now remain `Bytes`, and serialization infers the data format from the value type.

AlCalzone:
@zwave-js-bot automerge

PR #9173, opened 2026-09-01
feat: persist the last time a node was awake

### LINKED ISSUE #9153: Persist Last Awake across driver/UI restarts

`Last Awake` is lost when Z-Wave JS UI restarts. It then remains empty until the battery-powered node wakes again, making it difficult to determine when an unresponsive node was last awake.

Persist each node’s `Last Awake` timestamp and restore it on startup.

`Last Active`, Home Assistant history, and logs provide related information, but do not reliably preserve the node’s last wake-up time.

**Additional context**

This is particularly useful for version upgrades and for container deployments, where restarts are normal and should not erase diagnostic information.

robertsLando:
Fixes #9153

`Last Awake` was only tracked in Z-Wave JS UI's memory, so it was lost on every restart and stayed empty until the node woke up again — exactly when the information is most useful for diagnosing an unresponsive node.

The driver now tracks it itself and stores it in the network cache, next to `lastSeen`:

```ts
node.lastAwake; // Date | undefined — the last time the node sent a Wake Up notification
```

- Written when a Wake Up notification is received, which replaces the in-memory `lastWakeUp` timestamp the Wake Up handler kept for its wake-up-interval check. That handler now reads `node.lastAwake` instead, so the `WakeUpHandlerStore` is gone.
- Only sleeping nodes send Wake Up notifications, so the value stays empty for nodes that are never "awake" in the status sense.
- `markAsAwake()` does not touch it. The driver marks nodes as awake for reasons other than an observation, e.g. `keepAwake` during interviews and firmware updates, and a node kept awake that way has not told us anything new about when it was last up.
- Serialized like `lastSeen` (timestamp), so no new cache format, and old cache files stay readable in both directions.
- Deliberately not mirrored into `NodeStatistics`: it is a node property, not a communication statistic, and Z-Wave JS UI already updates its own value from the `wake up` event — what it was missing is the restored value at startup.

https://claude.ai/code/session_017Ai6n3wiRtNXiBdT17189a

### DIFF
--- a/packages/zwave-js/src/lib/test/node/Node.lastAwake.test.ts
+++ b/packages/zwave-js/src/lib/test/node/Node.lastAwake.test.ts
@@ -1,3 +1,4 @@
+import { WakeUpCCWakeUpNotification } from "@zwave-js/cc";
 import { MockController } from "@zwave-js/testing";
 import { test as baseTest } from "vitest";
 import { createDefaultMockControllerBehaviors } from "../../../Testing.js";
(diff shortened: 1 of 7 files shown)


packages/zwave-js/src/lib/test/node/Node.lastAwake.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+function createNode(driver: Driver, canSleep: boolean): ZWaveNode {
+	// Not node 1, which is the controller and can never sleep
+	const node = new ZWaveNode(2, driver);
+	node["isListening"] = !canSleep;
+	node["isFrequentListening"] = false;
+	return node;
+}
+
+const longAgo = new Date(2020, 0, 1);
```
AlCalzone: This is not _that_ long ago 😆

AlCalzone (changes requested):
I don't think this is the correct approach. In `packages/zwave-js/src/lib/node/CCHandlers/WakeUpCC.ts`, we already keep track of the last wake up, but internally. And this is not persisted.

We should simply change out the storage backend of this handler to use the new cache backed node property introduced in this PR.

Also that code path is the only one that should write to lastAwake ever.

robertsLando:
Reworked as suggested in ff4751e30.

`WakeUpHandlerStore` is gone — `handleWakeUpNotification` now reads and writes `node.lastAwake` directly for the wake-up interval check, and that is the only writer. `markAsAwake()` is back to just the status machine call, so `restoreAwakeStatus()` was dropped and `lastAwake` moved out of the status mixin to sit next to `lastSeen` in `Node.ts`.

One behavior change worth naming: the wake-up interval verification now survives a restart. Previously `lastWakeUp` was always undefined after a restart, so the first wake up could never trigger an interval refresh — now it can.

https://claude.ai/code/session_01Fb4eeQivqpY1sNzZUxCNBV

AlCalzone (approved):

PR #9196, opened 2026-09-08
feat: include node ID in invalid payload warnings

### LINKED ISSUE #9195: Please add Node ID to log message "Dropping message with invalid payload"
### Discussed in https://github.com/zwave-js/zwave-js/discussions/9194

<sup>Originally posted by **@DocSalvager** September 8, 2026</sup>

### Description

Adding Node ID and any other available info to the log message "Dropping message with invalid payload" would greatly ease finding the problem.

### Do you have a solution in mind?

no

### Describe alternatives you've considered

none

### Additional context

_No response_

</div>

AlCalzone:
Invalid-payload warnings now include the source node ID when it is available from the parsed message. They also include string validation reasons from the existing error context. Messages that fail before parsing keep the current raw-buffer fallback.

Tested with `yarn vitest run packages/zwave-js/src/lib/test/driver/invalidPayloadLog.test.ts`.

Closes #9195

### DIFF
--- a/packages/zwave-js/src/lib/driver/Driver.ts
+++ b/packages/zwave-js/src/lib/driver/Driver.ts
@@ -4366,18 +4366,20 @@ export class Driver
 				case ZWaveErrorCodes.PacketFormat_InvalidPayload:
 					if (msg) {
 						const nodeId = msg.getNodeId();
-						this.driverLog.print(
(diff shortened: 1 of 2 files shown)


AlCalzone:
@zwave-js-bot automerge

PR #9202, opened 2026-09-08
fix: apply path-specific formatting during code generation

AlCalzone:
Some code generators passed placeholder filenames to `formatWithOxfmt`. This prevented path-specific overrides from matching and could produce output that failed the repository-wide formatting check.

Pass the complete destination paths for generated CC documentation and CC value definitions so code generation and CI resolve the same oxfmt settings.


AlCalzone:
@zwave-js-bot automerge

PR #9203, opened 2026-09-08
chore: migrate syntax-only custom rules to Oxlint

AlCalzone:
Migrates seven syntax-only `@zwave-js` rules from ESLint to Oxlint through `jsPlugins`. The rules use Oxlint's `createOnce` API and `eslintCompatPlugin`, so the same plugin remains usable by ESLint for type-aware and JSON rules.

The two rules that require TypeScript type information remain in ESLint: `consistent-mock-node-behaviors` and `no-forbidden-imports`. The JSON device-config rules also remain in ESLint because Oxlint does not support their custom parser.

This also removes the unused `eslint-plugin-oxlint` and `eslint-plugin-unused-imports` dependencies, and removes the obsolete `packages/web` Oxlint exclusion. Oxlint's native `no-unused-vars` rule already reports unused imports where that rule is enabled. It does not autofix them.

Performance was measured with five warm runs on the same checkout:

| Command | Before median | After median | Change |
| --- | ---: | ---: | ---: |
| `yarn oxlint --type-aware` | 1.08 s | 2.03 s | +0.95 s |
| `yarn eslint` | 7.47 s | 4.56 s | -2.91 s |
| Combined | 8.83 s | 6.77 s | -2.06 s (-23.3%) |

The `createOnce` path measured 3.49 seconds versus 3.51 seconds for the regular `create` path while both JS plugins were loaded. The current difference is within measurement noise, but `createOnce` is retained because it is the API Oxlint intends to optimize.

Controlled violations exercised all seven migrated rules under both Oxlint and ESLint and produced matching custom-rule diagnostics. Oxlint fixes were also exercised for the migrated fixable rules.

### DIFF
--- a/packages/eslint-plugin/src/utils.ts
+++ b/packages/eslint-plugin/src/utils.ts
@@ -117,7 +117,27 @@ export function getCCIdFromDecorator(
 	return (CommandClasses as any)[getCCNameFromDecorator(decorator)];
 }
 
-export type Rule = TSESLint.RuleModule<any, never[], TSESLint.RuleListener>;
+export type Rule = TSESLint.RuleModule<
+	any,
+	never[],
+	unknown,
+	TSESLint.RuleListener
+>;
+
+export interface OxlintCompatibleRule extends Omit<Rule, "create"> {
+	createOnce(context: Parameters<Rule["create"]>[0]): ReturnType<
+		Rule["create"]
+	> & {
+		before?(): boolean | void;
+		after?(): void;
+	};
+}
+
+export function defineOxlintCompatibleRule(
+	rule: OxlintCompatibleRule,
+): OxlintCompatibleRule {
+	return rule;
+}
 
 export namespace JSONCRule {
 	// Special ESLint rule type for JSONC files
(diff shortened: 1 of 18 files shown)


packages/eslint-plugin/src/utils.ts
```diff
@@ -117,7 +117,27 @@ export function getCCIdFromDecorator(
...
-export type Rule = TSESLint.RuleModule<any, never[], TSESLint.RuleListener>;
+export type Rule = TSESLint.RuleModule<
+	any,
+	never[],
+	unknown,
+	TSESLint.RuleListener
+>;
+
+export interface OxlintCompatibleRule extends Omit<Rule, "create"> {
+	createOnce(context: Parameters<Rule["create"]>[0]): ReturnType<
```
AlCalzone: I'm pretty sure there is a concrete type name for this that does not need the `Parameters<...>` and `ReturnType<...>` hacks.

AlCalzone: Fixed in 303a9b898 — `createOnce` now uses `TSESLint.RuleContext` and `TSESLint.RuleListener` directly.

AlCalzone:
@zwave-js-bot automerge

PR #9206, opened 2026-09-09
feat(config): add Aeotec ZWA057 aërQ Temperature and Humidity Sensor 8

MartinHjelmare:
PR description here

- Add config file for ZWA057 and the foundation for the common template needed for some of Aeotec gen 8 line of devices. I'll be adding four more config files for this line shortly, in follow up PRs.
- I've used Claude to generate the config based on the manual and certification data found online. I've checked it with the config pr skill. After that I've proofread it manually and tested it manually together with the device.

### DIFF
--- a/packages/config/config/devices/0x0086/templates/aeotec_template.json
+++ b/packages/config/config/devices/0x0086/templates/aeotec_template.json
@@ -4598,5 +4598,82 @@
 				"value": 2
 			}
 		]
+	},
+	// Gen8 sensor settings
+	"led_indicator_periodic_reports": {
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "LED Indicator: Periodic Reports",
+		"defaultValue": 1
+	},
+	"led_indicator_mold_danger": {
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "LED Indicator: Mold Danger",
+		"defaultValue": 1
+	},
+	"sensor_limit_enable_temperature_upper": {
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "Temperature Reports: Send Above Upper Limit"
+	},
+	"sensor_limit_enable_temperature_lower": {
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "Temperature Reports: Send Below Lower Limit"
+	},
+	"sensor_limit_enable_humidity_upper": {
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "Humidity Reports: Send Above Upper Limit"
+	},
+	"sensor_limit_enable_humidity_lower": {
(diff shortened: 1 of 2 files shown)


packages/config/config/devices/0x0086/templates/aeotec_template.json
```diff
@@ -4598,5 +4598,82 @@
...
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "Humidity Reports: Send Above Upper Limit"
+	},
+	"sensor_limit_enable_humidity_lower": {
+		"$import": "~/templates/master_template.json#base_enable_disable",
+		"label": "Humidity Reports: Send Below Lower Limit"
+	},
+	"threshold_check_time": {
+		"label": "Threshold Check Time",
+		"description": "Allowable range: 30-2678400",
```
AlCalzone: With the new `"allowed"` field, we no longer document gaps this way. I thought I had eliminated this from the instructions, guess it's still somewhere.

packages/config/config/devices/0x0371/zwa057.json
```diff
@@ -0,0 +1,306 @@
...
+			"$import": "~/0x0086/templates/aeotec_template.json#threshold_check_time"
+		},
+		{
+			"#": "2",
+			// US-specific:
+			"$if": "productType === 0x0102",
+			"$purpose": "reporting_threshold.temperature",
+			"label": "Temperature Reports: Change Threshold",
+			"valueSize": 1,
+			"unit": "0.1 (°C/°F)",
```
AlCalzone, line 70: I wonder if we can narrow this down to Fahrenheit for the US?

AlCalzone (changes requested):
A few nits/questions.

There's probably some room for follow up PR to do templating. I've seen a few similar parameters on recent additions.

packages/config/config/devices/0x0371/zwa057.json
MartinHjelmare: The user can still change the reporting temperature to C, even in the US, right?

MartinHjelmare: Claude:

> Confirmed against the spec — param 64 ("Temperature Scale") is documented as "scale for auto reports and setting", so it governs how other config values (like this threshold) are interpreted too, not just the region's hardware default.

packages/config/config/devices/0x0086/templates/aeotec_template.json
MartinHjelmare: Fixed

packages/config/config/devices/0x0371/zwa057.json
AlCalzone: Oh right.

AlCalzone (approved):

PR #9207, opened 2026-09-09
feat(config): add Aeotec ZWA065 Siren 8

MartinHjelmare:
PR description here

- Add config for ZWA065.
- Add "Sound Switch" to the eslint-plugin's known Command Class names so the option label title-cases correctly.
- Claude generated the config from the manual and certification online sources. I proofread and tested manually with a real device.

### DIFF
--- a/packages/config/config/devices/0x0371/zwa065.json
+++ b/packages/config/config/devices/0x0371/zwa065.json
@@ -1,0 +1,207 @@
+{
+	"manufacturer": "Aeotec Ltd.",
+	"manufacturerId": "0x0371",
+	"label": "ZWA065",
+	"description": "Siren 8",
+	"devices": [
+		{
+			// US version. Config param 13 switches the device between Binary Switch and
+			// Sound Switch device types, so both product types must be listed here or the
+			// device goes unmatched after switching to the other mode.
+			"productType": "0x0103",
+			"productId": "0x0041"
+		},
+		{
+			// US version, Sound Switch device type (see comment above)
+			"productType": "0x0104",
+			"productId": "0x0041"
+		},
+		{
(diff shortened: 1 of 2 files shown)


packages/config/config/devices/0x0371/zwa065.json
```diff
@@ -0,0 +1,207 @@
...
+		},
+		{
+			// EU version. Confirmed via a real device interview (productType 3).
+			// There is no EU listing in the Z-Wave Alliance catalog yet.
+			"productType": "0x0003",
+			"productId": "0x0041"
+		},
+		{
+			// EU version, Sound Switch device type. Inferred from the same region/device-type
+			// byte pattern as the confirmed EU Binary Switch entry above, not device-confirmed.
```
AlCalzone, line 27: Aeotec typically have a third pattern `0x02**` for AUS/NZ.
https://aeotec.freshdesk.com/support/solutions/articles/6000280231-siren-8-technical-specifications- also confirms this - we should add it here as well.

AlCalzone (changes requested):
Just a small nit

packages/config/config/devices/0x0371/zwa065.json
MartinHjelmare: I didn't add it as it's not part of the certification data. We would just be relying on the docs which has turned out to be incorrect sometimes. But I can add it if you want me to.

AlCalzone: Lets do it in this case - this pattern has proven to be reliable

MartinHjelmare: Added.

AlCalzone (approved):

PR #9214, opened 2026-09-10
feat(config): add Aeotec ZWA066 MultiSensor 8

MartinHjelmare:
PR description here

- Add config for ZWA066.
- Claude generated the config from the manual and certification online sources. I proofread and tested manually with a real device.

### DIFF
--- a/packages/config/config/devices/0x0371/zwa066.json
+++ b/packages/config/config/devices/0x0371/zwa066.json
@@ -1,0 +1,714 @@
+{
+	"manufacturer": "Aeotec Ltd.",
+	"manufacturerId": "0x0371",
+	"label": "ZWA066",
+	"description": "MultiSensor 8",
+	"devices": [
+		{
+			// EU version
+			"productType": "0x0002",
+			"productId": "0x0042"
+		},
+		{
+			// US version
+			"productType": "0x0102",
+			"productId": "0x0042"
+		},
+		{
+			// AU/NZ version. Not in the Z-Wave Alliance catalog yet, but Aeotec's regional
+			// numbering pattern (EU/US/AU regions map to product type bytes 0x00/0x01/0x02) has
+			// proven reliable across the other Gen8 devices.
+			"productType": "0x0202",
+			"productId": "0x0042"
+		}
+	],
+	"firmwareVersion": {
+		"min": "0.0",
+		"max": "255.255"
+	},
+	"associations": {
+		"1": {
+			"label": "Lifeline",
+			"maxNodes": 5,
+			"isLifeline": true
+		},
+		"2": {
+			"label": "Motion",
+			"maxNodes": 5
+		},
+		"3": {
+			"label": "Over Heat",
+			"maxNodes": 5
+		},
+		"4": {
+			"label": "Under Heat",
+			"maxNodes": 5
+		},
+		"5": {
+			"label": "Over Humidity",
+			"maxNodes": 5
+		},
+		"6": {
+			"label": "Under Humidity",
+			"maxNodes": 5
+		},
+		"7": {
+			"label": "Over Light",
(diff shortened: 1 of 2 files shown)


packages/config/config/devices/0x0371/zwa066.json
```diff
@@ -0,0 +1,711 @@
...
+					"value": 255
+				}
+			]
+		},
+		{
+			"#": "7",
+			"label": "Acceleration Reports: Change Threshold",
+			"valueSize": 1,
+			"unit": "m/s²",
+			"allowed": [{ "value": 0 }, { "range": [30, 255] }],
```
AlCalzone, line 167: ZWA055 has this as 0.01 m/s² - worth double checking. A minimum of 30 m/s² (~3g) seems high for a change threshold.

https://github.com/zwave-js/zwave-js/pull/9089/changes#diff-a601d64838b2fbb421ee1df3bfe79e71b013c63d6f9c6a0ebf5a7358130538a2R474

MartinHjelmare: Image

I'll check with Claude too.

MartinHjelmare: Claude's conclusion is that it's ZWA055 that probably is incorrect.

<details>
 <summary>Claude analysis</summary>

I verified this by rendering the actual PDF pages as images (not just relying on extracted text, which can garble tables), and the evidence points the other way from AlCalzone's suspicion — zwa066's plain "m/s²" is correct, and it's zwa055.json's  "0.01 m/s²" that looks like the actual bug.

What the spec pages actually say (confirmed visually, page 16 of ZWA066's spec, page 30 of ZWA055's spec):

```
  ┌────────┬─────────────┐
  │ Value  │  Function   │
  ├────────┼─────────────┤
  │ 0      │ Disable     │
  ├────────┼─────────────┤
  │ 30~255 │ 30~255 M/S² │
  └────────┴─────────────┘
```

Both devices' specs use the identical wording, with no scale annotation (0.1, 0.01, etc.) anywhere near this table — unlike neighboring parameters in the same docs that explicitly say "0.1 (°C/°F)" or "0.1 %" when a scale factor applies. This isn't a pdftotext extraction artifact; I confirmed it by eye on the rendered page image.

On AlCalzone's physical-plausibility point specifically: it actually cuts the other way once you apply it to the alternative reading. A 0.01 m/s² minimum-change threshold is smaller than a typical accelerometer's noise floor — a "minimum change to trigger a report" gate set that fine would report almost continuously, which defeats the purpose of having a threshold at all. A several-g range (30–255 raw = ~3g–26g) reads much more sensibly as a deliberately coarse shock/tamper-level gate meant to filter out routine handling and only fire on hard knocks or drops — which fits the parameter's actual name ("Trigger acceleration report") and how these Gen8 sensors use it (tamper/impact detection, not fine motion sensing).

I also checked the Z-Wave Alliance certification catalog pages for both devices — they don't independently publish per-parameter tables beyond linking the same manual PDF, so there's no separate corroborating or contradicting source there.

Conclusion: no evidence supports the 0.01 m/s² reading for either device. zwa066.json's current "m/s²" (unscaled) is correct as-is — nothing to change on the open PR #9214. zwa055.json's "0.01 m/s²" (introduced in #9089, carried through #9208's cleanup) looks like the actual defect and would be worth a small follow-up fix PR once #9214 is settled, if you want to raise it.

</details>

AlCalzone: Ok I'll ask Aeotec to clarify.

AlCalzone: Okay, they have confirmed that the 30-255 value range is correct.

AlCalzone (approved):

PR #9215, opened 2026-09-11
fix(config): map root reports to endpoint 1 on Remotec ZXT-800, add fingerprint

AlCalzone:


AlCalzone:
@zwave-js-bot automerge

PR #9216, opened 2026-09-11
fix: learn thermostat setpoint float encodings from reports

AlCalzone:
Learn precision, scale, and size combinations from Thermostat Setpoint Reports and both bounds in Capabilities Reports. Store observations globally per device across endpoints and setpoint types.

Set commands reuse an observed combination for the requested scale and round values when necessary. Explicit `overrideFloatEncoding` settings retain priority. Unknown encodings preserve the existing automatic encoding behavior. Later reports populate and update the cache without a re-interview.

Keep rounded values consistent in optimistic updates and verification polls. Preserve the selected encoding while commands wait for transmission.

Add unit and integration coverage for interview and unsolicited reports, empty caches, cache persistence, shared device-wide observations, scale matching, rounding, byte sizes, overflow recovery, overrides, N/A reports, multicast, supervised updates, verification polls, and queued commands.

This aligns us with behavior of several devices, as well as the specs.

### DIFF
--- a/packages/cc/src/cc/ThermostatSetpointCC.ts
+++ b/packages/cc/src/cc/ThermostatSetpointCC.ts
@@ -30,8 +30,9 @@ import {
 	POLL_VALUE,
 	type PollValueImplementation,
 	SET_VALUE,
+	SET_VALUE_HOOKS,
 	type SetValueImplementation,
-	type ValueIDProperties,
+	type SetValueImplementationHooksFactory,
 	throwUnsupportedProperty,
 	throwWrongValueType,
 } from "../lib/API.js";
@@ -175,6 +176,36 @@ export const ThermostatSetpointCCValues = V.defineCCValues(
 
 @API(CommandClasses["Thermostat Setpoint"])
 export class ThermostatSetpointCCAPI extends CCAPI {
+	private getPreferredScale(setpointType: ThermostatSetpointType): number {
+		// SDS14223: The Scale field value MUST be identical to the value received in the Thermostat Setpoint Report for the
+		// actual Setpoint Type during the node interview. Fall back to the first scale if none is known
+		return (
+			this.tryGetValueDB()?.getValue<number>(
+				ThermostatSetpointCCValues.setpointScale(setpointType).endpoint(
+					this.endpoint.index,
+				),
+			) ?? 0
+		);
+	}
+
+	protected [SET_VALUE_HOOKS]: SetValueImplementationHooksFactory = ({
+		property,
+		propertyKey,
+	}) => {
+		if (property === "setpoint" && typeof propertyKey === "number") {
+			return {
+				normalizeValue: (value) =>
+					typeof value === "number"
+						? this.createSetpointSet(
+								propertyKey,
+								value,
+								this.getPreferredScale(propertyKey),
+							).value
+						: value,
+			};
+		}
+	};
+
 	public supportsCommand(
 		cmd: ThermostatSetpointCommand,
 	): MaybeNotKnown<boolean> {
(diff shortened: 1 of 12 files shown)


packages/cc/src/cc/ThermostatSetpointCC.ts
```diff
@@ -108,6 +175,52 @@ export const ThermostatSetpointCCValues = V.defineCCValues(
 
 @API(CommandClasses["Thermostat Setpoint"])
 export class ThermostatSetpointCCAPI extends CCAPI {
+	private getPreferredScale(setpointType: ThermostatSetpointType): number {
+		// SDS14223 requires the scale reported for the actual setpoint type
```
AlCalzone: Do not reword existing comments unless absolutely necessary.

packages/cc/src/cc/ThermostatSetpointCC.ts
```diff
@@ -73,9 +75,74 @@ function getSetpointUnit(scale: number): string {
...
+
+function encodeSetpointFloat(
+	value: number,
+	scale: number,
+	encodings: readonly SetpointFloatEncoding[],
+): Bytes {
+	const candidates = encodings.filter((encoding) => encoding.scale === scale);
+	if (!candidates.length) return encodeFloatWithScale(value, scale);
+
+	const fitting = candidates
```
AlCalzone, line 103: Add a brief comment what the idea behind this candidate search is.

packages/cc/src/cc/ThermostatSetpointCC.ts
```diff
@@ -73,9 +75,74 @@ function getSetpointUnit(scale: number): string {
...
+			`Cannot encode setpoint ${value} with any observed float encoding for scale ${scale}`,
+			ZWaveErrorCodes.Argument_Invalid,
+		);
+	}
+
+	// The wire format permits three-byte integers
+	const ret = new Bytes(1 + selected.size);
+	ret[0] = (selected.precision << 5) | (scale << 3) | selected.size;
+	ret.writeIntBE(selected.integer, 1, selected.size);
+	return ret;
```
AlCalzone: I think we're mixing concerns here. This method should only try to find a precision match. The caller should then pass that to encodeFloatWithScale.

packages/cc/src/lib/API.ts
```diff
@@ -81,6 +81,8 @@ export type SetValueImplementationHooks = AllOrNone<{
 	supervisionOnSuccess: () => void | Promise<void>;
 	supervisionOnFailure: () => void | Promise<void>;
 }> & {
+	/** Normalizes a requested value before sending and optimistic updates. */
+	normalizeValue?: (value: unknown) => unknown;
```
AlCalzone: No this is unnecessary abstraction for a single CC command.

packages/cc/src/cc/ThermostatSetpointCC.ts
```diff
@@ -595,9 +746,24 @@ export class ThermostatSetpointCCSet extends ThermostatSetpointCC {
...
-			encodeFloatWithScale(this.value, this.scale, override),
+			// Keep the prepared value stable while the command waits in the queue
+			this.encodedValue
+				?? (override
+					? encodeFloatWithScale(this.value, this.scale, override)
+					: encodeSetpointFloat(
+							this.value,
+							this.scale,
+							observed ?? [],
+						)),
```
AlCalzone: For readability, pull this out into its own variable.

packages/cc/src/cc/ThermostatSetpointCC.ts
```diff
@@ -243,6 +361,7 @@ export class ThermostatSetpointCCAPI extends CCAPI {
 			setpointType,
 			value,
 			scale,
+			encodedValue: this.encodeSetpointValue(value, scale),
```
AlCalzone: No, let the CC instance handle normalization/rounding.

PR #9259, opened 2026-09-15
fix: prevent shutdown deadlocks and queue re-entry

AlCalzone:
Fix a reproducible shutdown deadlock investigated from #9249.

Some scheduled tasks directly await transactions held in the queue during controller recovery. `destroy()` waits for the scheduler to stop before clearing those queues. Neither can proceed after reconnection fails.

Use scheduler-aware `yield* waitFor(...)` for communication in interviews, route rebuilding, failed-node operations, and OTA updates. Shutdown also rejects transactions in parallel with stopping the scheduler. An accidentally blocking task can then finish without preventing queue cleanup.

Close the Serial API queue before rejection callbacks run. Pause transaction queues and stop accepting additions before rejecting their contents. Late additions reject their callers. Late transaction failures bypass controller recovery after destruction.

Added integration coverage for failed TCP reconnection during five task operations, deliberately blocking tasks, concurrent `destroy()` calls, synchronous and asynchronous retries, progress-callback re-entry, and command generation resuming after teardown. Added queue tests for late additions and re-entrant disposal.

The TCP cases reproduce a failure mechanism consistent with #9249. The original ESPHome log does not identify the blocked task.


adamjernst:
Thanks for investigating and fixing this! Really appreciate it!

